PR #25, opened 2015-06-11
Adding font files and references

juliaelman:
This adds all of the font files for Source Sans Pro and creates a reference to them in the head.


maya:
Looks good, @juliaelman. This adds regular and bold font (400 and 700 weight). This currently does not include italics. Can we verify with design if we need italics or other weights? https://www.google.com/fonts#UsePlace:use/Collection:Source+Sans+Pro cc @angel @mollieru

angel:
I think we'll need at least 400 italics. We'll also need 600. That's what I've been using for the buttons. Thanks!

maya:
Do we need 700?

angel:
Yes, headers are in 700. If you think the page load will be too much, I would rather take out the 600 weight.

maya:
We can try leaving it for now and if we feel it's too heavy, take it out. Although it would simplify things with only two weights, then we only have to declare normal and bold, rather than 400, 600, 700. Curious to hear other people's thoughts.

maya:
On second thought, we might as well be explicit with our weights (i.e. 400 and 700, not normal and bold) in case other people want to extend this and add additional weights.

mjovel:
@maya and @angel I think limiting to two weights might be prudent from a performance perspective. Maybe discussions of what a performance budget for the framework that comes with the pattern library should be? I would think this would be less than the average sites budget as content imagery and media by the frameworks user. That might make the decision easier to quantify?

maya:
True, @mjovel. Also the last commit removes Source Code Pro, the monospace version of Source Sans Pro for this reason. We are using prism.css which uses `Consolas, Monaco, 'Andale Mono', monospace` as its monospace font stack which relies on browser fonts.

PR #26, opened 2015-06-16
Adding basic pages and Jekyll collections

### LINKED ISSUE #16: Add pages folder for the sites page infrastructure
We currently do not have a `pages` folder in our setup. We'll want to make sure to add that in as we build out the page infrastructure for the library.

juliaelman:
This pull request includes the following, based on the initial IA presented https://github.com/18F/govt-wide-patternlibrary/issues/21#issuecomment-112095912:

- Adding `includes/scripts.html` and moving jQuery to load at the bottom of the page.
- Adding Jekyll collections for visual style, layout system and patterns to `_config.yml`
- Adding pages for each section of the v3 IA outlined in the image below:

- Adding some basic logic to display collection information on each page.

Fixes #16


_patterns/header_footer.md
```diff
@@ -0,0 +1,5 @@
+---
+layout: default
+type: component
+title: Headers and Footer
```
maya: I would separate this out to two different files.

_patterns/charts_graphs.md
```diff
@@ -0,0 +1,5 @@
+---
+layout: default
+type: component
+title: Charts and Graphs
```
maya: This has been moved out of our Sprint priorities, so I think this can be removed for now.

carodew:
I can say that on my previous team we used patternlab.io as a framework and initially it didn't provide different URLs for each pattern, so when I would tell a dev "go find the project tile and use that" I couldn't actually send them a URL straight to the pattern and it was a pain Also, we might have had it configured wrong, because at some point URLs did start working for us and it was much easier. 

I was thinking that each of the top three sections would have one long page with all their subsections on it, but you could use anchor links to jump to the relevant spot on the page (and also be able to use that in the URL). In the new wireframe I started getting at that by including the top of the next component in line (Icons).

juliaelman:
@carodew right! And that is the methodology and the structure of the way the files/folders are currently structured. Glad we're on the same page! (no pun intended)

colinpmacarthur:
I'm in agreement with @carodew: I think what she describes is a good way to go.

Also, so glad each component will have a URL! You all are the best.

juliaelman:
@colinpmacarthur thanks! YOU are the best!!!!! :rainbow:

maya:
Thanks for jumping in and clarifying folks!

PR #51, opened 2015-06-19
Adding basic main navigation

juliaelman:
This pull request includes the main navigation for the pattern library, as outlined in the desktop wireframes #21. A few other things to note in this pull request:

- Added a skipnav for 508 accessibility
- Changed /pages files from `.md` to `.html` and replaced markup, fixes #50 
- Moved references to pattern library sections in index.html and created a more DRY approach

NOTE: This pull request does not include the slide out nav, as we will most likely do this once the side navigation has been laid out.

PR #52, opened 2015-06-19
Add basic side navigation

juliaelman:
@maya this is the initial layout we discussed for the side navigation to help get you started. It is fairly bare bones at the moment and ready for your styling magic! :sparkles: Here are some notes to think about:

- The side navigation (and main navigation) is not responsive or uses a slide out. Here is a link for the wireframe for how that interaction should work: https://github.com/18F/govt-wide-patternlibrary/issues/21#issuecomment-113263565
- The links under each main section are currently only dummy links. I was not sure how you wanted to best handle that for this iteration.

**NOTE:** This pull request is downstream from #51, so that will need to be merged in prior to this.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -6,7 +6,7 @@ url: https://pages.18f.gov/govt-wide-patternlibrary
 # Markdown config
 markdown: redcarpet
 highlighter: rouge
-name: Government-wide Pattern Library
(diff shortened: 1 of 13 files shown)


juliaelman:
+1!

PR #53, opened 2015-06-19
Removed 'fix' for font smoothing from _typography.scss

andrewhughey:
Please stop fixing font smoothing:
http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/


juliaelman:
+1 agreed. Thanks for the PR!

PR #55, opened 2015-06-19
Style sidebar navigation

maya:
This styles the sidebar navigation.

This is what it looks like:
sidebar-nav


maya:
Do we want Elements and Components to link to their location on the page? Right now they do not and are just text. The current styling makes them look like they're links.

mollieru:
I recommend adjusting the _font-size_ of these words to _1.2em_ so that they look less like links. Meanwhile, I recommend that 'Patterns' link to a title 'Patterns' (just as 'Layout' does). 
screen shot 2015-06-22 at 1 35 20 pm

angel:
I think if we are going to separate the "patterns" to "elements" and "components" those sub sections should be linked to sub headers that say "elements" and "components", so then it would make sense that those are clickable as well.

mollieru:
Resolution: what @angel said :) 

+ removing 'Patterns' for now and elevating these to top level nav.

maya:
This has been resolved and fixed in my last commits. This is what it looks like:

screen shot 2015-06-22 at 12 05 35 pm

maya:
This PR also separates Elements and Components into separate pages to match the new sidebar navigation structure. Previously they were within "Patterns" page, which has been removed.

maya:
@juliaelman: I was incorrect about removing the aside. I previously read here: http://html5doctor.com/understanding-aside/ "Navigation [...] are not directly related to the article and therefore do not justify the use of an <aside>” That it should be used for content tangentially related to it. 

However, I’m seeing in a newer article: http://html5doctor.com/aside-revisited/ “aside is now acceptable for secondary content when not nested within an article element." "When used outside of an article element, the contents should be related to the site (e.g., a blogroll, groups of additional navigation, and even advertising if that content is related to the page)." So it should be fine for navigation. I can go ahead and add it back in. Apologies!

juliaelman:
Looks good! Since we are no longer using "Patterns" as a top level nav item for "Elements" and "Components", let's do the same for our collections infrastructure too. We'll need to create a collection for "Elements" and another for "Components" and folders for each of these too.

carodew:
Hey guys – there is a lot going on with edits to the nav today and I am concerned that we are making a lot of smaller decisions without looking at the bigger picture. I was hoping we could wait until Colin and I had a chance to test the navigation so we can make smart, informed changes, rather than making changes based on our gut reactions. I'm a little concerned that the team is doing a lot of work to modify the nav now that will just have to be undone when we do get user feedback.

juliaelman:
@carodew @colinpmacarthur I whole heartedly agree and would much rather make informed decisins based off of testing.

juliaelman:
Looks good!

PR #56, opened 2015-06-22
Moving skip nav link to fix link/breakage.

juliaelman:
There is currently an conflict when trying to click on the "Government Wide Pattern Library" and the skip nav. Due to the changes done in 6bec201079d96ef14c05c533cbbafa2ae4562b13, the positioning on of the skip nav overlays when trying to click on that link. This PR solves the issue by moving the skip nav link directly underneath the `<body>` tag.

### DIFF
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -1,10 +1,10 @@
+<a id="skipnav" href="#main">Skip to main content</a>
 <header role="banner">
   <div id="logo">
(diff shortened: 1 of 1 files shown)


maya:
Looks good to me @juliaelman. Adding a screenshot so others can see what it looks like.
screen shot 2015-06-22 at 12 39 10 pm

Screen recording after hitting tab button: http://cl.ly/bjss

Issue #43, opened 2015-06-17
[Icons] design draft


mollieru:
For iconography, we have been considering forking the way that USPTO has organized their reuse of Font-Awesome, which we're also suggesting we use. Would be helpful to hear how others feel about this idea. Screenshot below and here's the live link: https://uspto.github.io/designpatterns/1.x/docs/foundation/icons.html ]

Leaving this here for feedback:
![icons uspto ui design library](https://cloud.githubusercontent.com/assets/3680494/8219678/b2a58616-1519-11e5-8da9-1c48a8b107d6.png)

arvalarva:
So far, Font Awesome has met the vast majority of the U.S. Patent and Trademark Office's needs. The community can also request new icons, or simply add icons within GitHub to the core Font Awesome library. Also, Font Awesome is becoming so widespread, that the goal of easily recognizable icons (instructions) for our customers is easily met.

logantpowell:
:+1:  "God is in the details" - Mies Van der Rohe

carodew:
We've heard some concerns in user interviews that Font Awesome can be a bit cartoony, but it's not been widespread feedback. I do think that it would be worthwhile at some point to remove some of the cheesier icon options (beer? Checkered flag?) and/or recommend specific icons for specific contexts.

logantpowell:
:-1: "Beer is the proof that god loves us and wants us to be happy" - Benjamin Franklin

carodew:
@logantpowell touché.

andrewmaier:
Icon fonts are _almost_ an anti-pattern for most applications: they're difficult to update/maintain, not semantic, and poorly rendered/supported. Instead, I'd recommend we use SVGs with a PNG fallback. There's a really good presentation on the benefits of this approach, here: https://docs.google.com/presentation/d/1CNQLbqC0krocy_fZrM5fZ-YmQ2JgEADRh3qR6RbOOGk/edit#slide=id.p

logantpowell:
I'm not an expert, but I'd say that - if the intention is to create an icon set for everyone to share, this is a good way of doing it. I.e., in terms of maintenance/updates/etc... However, you have a good point about the support. 

I would say it depends on the market you're thinking of providing a solution for... Here's a [take](https://clicky.com/marketshare/us/web-browsers/mobile/) on the US mobile browser market:

[...]

Issue #21, opened 2015-06-10
Create basic site architecture for the pattern library


juliaelman:
For Sprint 2, our focus will be to create a basic site architecture based on the initial wireframes created by the core UX team //cc @carodew @colinpmacarthur 

- [x] Main navigation include
- [x] Side navigation include
- [x] Basic main content structure

juliaelman:
Here are the v1 wireframes for us to get started on the front end work. 

**NOTE** There is another round of changes that are coming up with some feedback from today's overview session. Great work @carodew and @colinpmacarthur!

![gwpl-desktop-wireframes_page_2](https://cloud.githubusercontent.com/assets/955558/8095266/c5e6ae28-0f9a-11e5-98d4-72d427bb0494.png)

mjovel:
Well done!  I really like the use and accessibility content. I was wondering what the use case was for the css and scss tabs within the source tabs? Is it for people to extract those styles and use them within their own set-up? customize them? Most other pattern libraries I have seen have tended to focus on the html. Does listing the css and scss encourage presentational changes leading to increased abandonment of patterns through customization?

juliaelman:
@mjovel thanks for your feedback! I think it will be a matter of testing, iteration and feedback to determine whether or not this approach will cause increased abandonment. //cc @carodew @colinpmacarthur @maya

carodew:
Ditto what @juliaelman said. That's good to know that there seems to be a bigger focus on HTML. We do  plan to do further research and testing into how people want to be able to use/access code – in order to put out something quickly that we can test we are making a few assumptions (and code use is one of those areas). 

On that note, @maya and @juliaelman – since that area is more in question we can talk about things we can do to make it easier to build while still testable. For example, we probably don't need a 'copy' button right off the bat, since people will be able to manually copy/paste if they want to.

maya:
Good point @mjovel. One possibility is we just offer HTML snippets for V1 and allow our users to download the SCSS/CSS files. Then we can continue research the other options. How does this sound? This certainly makes implementation simpler.

juliaelman:
*UPDATE* Here are the v2 wireframes with updates and notes from the team.

![1434055570864](https://cloud.githubusercontent.com/assets/955558/8162628/85e6ce10-134c-11e5-8c99-c152f2611d65.png)

maya:
Also we need to add in the main content area as part of our list.

juliaelman:
[...]

PR #57, opened 2015-06-22
Adding slide out side navigation for smaller screen sizes

juliaelman:
This pull request includes only the functionality for the slide out menu for the pattern library website. No styling has been added to the menu itself and am leaving that for @maya to add her styling :sparkles: See screenshots for current state below the bullet points.

A few things about this pull request:

- I've used the unicode formatting for the menu icon. I've tested on several devices and browsers and it works ok, but we might want to move over to Font Awesome if we decide to move to that route, see #43 for reference.
- The side nav is still needing the accordion functionality added, per the wireframe spec found here: https://cloud.githubusercontent.com/assets/10144074/8239942/1571d372-15cd-11e5-864d-44edc555eb37.png
- The anchor links on the homepage are removed, as the menu itself covered up the content when linking on the homepage.
- Adding the text "Menu" next to the slide out button for accessibility purposes, article reference from @vz3: http://exisweb.net/menu-eats-hamburger

**NOTE:** This is downstream from #55 and #56 and should not be merged in prior to those PR.

screen shot 2015-06-22 at 1 44 20 pm

screen shot 2015-06-22 at 1 44 33 pm


maya:
@juliaelman I made a couple of fixes, looks good to me!

I realized that you removed the anchor links from the index.html file bc we switched to absolute links instead of relative links so in page linking does not work on that "show all" page. One possible fix is a different sidebar include for "all" page, but realize it’s not super high priority, but want it to be documented here.

maya:
With our updated nav structure, this is what the slideout nav looks like: 
screen shot 2015-06-22 at 3 43 56 pm

juliaelman:
Refs #21

PR #61, opened 2015-06-23
Hide sidebar navigation items

maya:
This hides sidebar navigation items _not_ in the current page. It will only show the child list items for the page it's on.

This resolves one of the bullet points in https://github.com/18F/govt-wide-patternlibrary/issues/21

This is what it looks like when you're on the Components page:

screen shot 2015-06-22 at 7 09 34 pm

### DIFF
--- a/_layouts/styleguide.html
+++ b/_layouts/styleguide.html
@@ -5,7 +5,7 @@
     {% include head.html %}
   </head>
 
-  <body>
+  <body class="{{ site.title }}">
 
     	{% include navbar.html %}
   	
--- a/_layouts/styleguide.html
+++ b/_layouts/styleguide.html
@@ -5,7 +5,7 @@
     {% include head.html %}
   </head>
 
-  <body class="{{ site.title }}">
+  <body class="{{ page.slug }}">
 
     	{% include navbar.html %}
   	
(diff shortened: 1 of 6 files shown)


maya:
Also the last commit adds in a new frontmatter called "slug".

juliaelman:
Looks good! The only thing I found was some weirdness with the hyperlinking. It looks like the the very last hyperlinked component on the page does not hit right above the title, as the others do. See preview below:

bhputhvwin

maya:
@juliaelman This is bc the page is too short, so there's nothing to scroll to at the moment. Once content is added to those last ones it will work as expected.

juliaelman:
@maya hrmm... not sure that adding more content to the last list items will fix this? It may be more ambiguous on the spacing and therefore still cause the same interaction.

Also, I saw that you changed the `title` to `slug` in that commit and sorry I missed that!

maya:
@juliaelman I'm pretty sure it will work, it's a common issue with one-page scroll sites. I've seen this issue many times before and as soon as content is added, the page height increases and it has something to jump to and works as expected. You can test this by adding an image and text to the last component. If you still experience the issue let me know and I'll take a look.

juliaelman:
814dfa9 fixes the issue with a temporary solve until we get the final content. Thanks @maya!

PR #68, opened 2015-06-23
Update _config.yml

ErieMeyer:
Great catch @juliaelman -- updating the file here to go along with https://github.com/18F/govt-wide-patternlibrary/pull/66 and https://github.com/18F/govt-wide-patternlibrary/pull/65

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -35,6 +35,9 @@ collections:
   visual:
     output: true
     permalink: /visual-style/:path/
+  content:
+    output: true
+    permalink: /content-style/:path/
   layout-system:
     output: true
     permalink: /layout-system/:path/


maya:
@ErieMeyer thanks for your contribution! Quick question: If this is just a page of text, do we need to create a jekyll collection? Basically, are you planning on breaking this out into partials? At the current state it looks like a single page, which is fine and can live on the site as is.

ErieMeyer:
> Single page is fine! Thank you!

I can see a world where there are multiple languages that would need more pages, but for now, one is great.

maya:
@ErieMeyer Also, @juliaelman just brought to my attention that all of our current pages are written in HTML, not markdown, while the components feeding in to the pages are in markdown. Would you rather this content be in markdown or HTML?

maya:
@ErieMeyer actually leaning towards converting this to HTML. Would you like to do it or should we?

ErieMeyer:
To make sure it's standard with the rest of the stuff you're doing, I would love if you took a run at it. Let me know if it's a pain, and I'll do it!

juliaelman:
@ErieMeyer happy to help and done in #106! :boom: There are some comments and notes to address in there too.

ErieMeyer:
Excellent -- thank you!!

🇺🇸🌷💯

PR #70, opened 2015-06-23
Fix h2 placement

maya:
This fixes the issue where the placement of the h2 heading was not falling in the correct place. This resolves: https://github.com/18F/govt-wide-patternlibrary/issues/63

This is what it looks like:

heading-fix


juliaelman:
+1

PR #71, opened 2015-06-23
Add smooth scrolling to sidebar navigation

maya:
This adds smooth scrolling to the sidebar nav. This will need to be tweaked and added after: https://github.com/18F/govt-wide-patternlibrary/pull/70

This is what it looks like: 
http://cl.ly/0C2x1n3I2f2u


juliaelman:
+1

PR #105, opened 2015-06-30
Add typography content

maya:
This adds the typography live content to the library: headings, body text, links, and lists.

Note: 

- Some of these can be considered "elements" (HTML elements) and may be able to be moved into their own file. We'll need further research on this. Currently, everything is in located in `typography.md` and `_typography.scss`.
- I'm using ems for the margins on headings so it scales with the font being used. You'll see larger margins for larger fonts, smaller margins for smaller fonts.
- Body text is not using the 500px max-width as I will consider the best approach to do this. We want users not to be restricted when adding in body text if they don't need the constraint. For example, they may need longer space for text for a hero image.
- This does not add in Example

This is what it looks like:
typography

### DIFF
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -20,12 +20,26 @@ title: Typography
 
   <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
 
-
   <!-- Links title -->
   <h3>Links</h3>
 
   <p>This is <a href="#">linked text</a></p>
   <p>This is <a href="#">linked text</a></p>
+
+  <!-- Lists title -->
+  <h3>Lists</h3>
+
+  <ul>
+    <li>Unordered list item</li>
+    <li>Unordered list item</li>
+    <li>Unordered list item</li>
+  </ul>
+
+  <ol>
+    <li>Ordered list item</li>
+    <li>Ordered list item</li>
+    <li>Ordered list item</li>
+  </ol>
 
   <img src="{{ site.baseurl }}/assets/img/static/Typography_UI_v1.png">
 </div>
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -25,7 +25,7 @@ title: Typography
   <h3>Links</h3>
 
   <p>This is <a href="#">linked text</a></p>
-  <p>This is <a href="#">linked text</a></p>
+  <p>This is <a class="usa-link" href="#">linked text</a></p>
 
   <!-- Lists title -->
   <h3>Lists</h3>
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -25,7 +25,7 @@ title: Typography
   <h3>Links</h3>
 
   <p>This is <a href="#">linked text</a> for body copy</p>
-  <p class="sans">This is <a href="#">linked text</a> for everything else</p>
+  <p class="usa-sans">This is <a href="#">linked text</a> for everything else</p>
 
   <!-- Lists title -->
   <h3>Lists</h3>
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -5,11 +5,11 @@ title: Typography
 
 <div class="preview">
   <!-- Add HTML markup for example here -->
-  <h1>Heading Level 1</h1>
-  <h2>Heading Level 2</h2>
-  <h3>Heading Level 3</h3>
-  <h4>Heading Level 4</h4>
-  <h5>Heading Level 5</h5>
+  <h1>Heading 1 in 36px in Source Sans Pro 700</h1>
+  <h2>Heading 2 in 24px in Source Sans Pro 700</h2>
+  <h3>Heading 3 in 19px in Source Sans Pro 700</h3>
+  <h4>Heading 4 in 16px in Source Sans Pro 700</h4>
+  <h5>Heading 5 in 14px in Source Sans Pro 700</h5>
 
   <img src="{{ site.baseurl }}/assets/img/static/Typography_UI_v1.png">
 </div>
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -5,6 +5,12 @@ title: Typography
 
 <div class="preview">
   <!-- Add HTML markup for example here -->
+  <h1>Heading Level 1</h1>
+  <h2>Heading Level 2</h2>
+  <h3>Heading Level 3</h3>
+  <h4>Heading Level 4</h4>
+  <h5>Heading Level 5</h5>
+
   <img src="{{ site.baseurl }}/assets/img/static/Typography_UI_v1.png">
 </div>
 
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -5,11 +5,20 @@ title: Typography
 
 <div class="preview">
   <!-- Add HTML markup for example here -->
+
+  <!-- Heading title -->
+  <h3>Headings</h3>
+
   <h1>Heading 1 in 36px in Source Sans Pro 700</h1>
   <h2>Heading 2 in 24px in Source Sans Pro 700</h2>
   <h3>Heading 3 in 19px in Source Sans Pro 700</h3>
   <h4>Heading 4 in 16px in Source Sans Pro 700</h4>
   <h5>Heading 5 in 14px in Source Sans Pro 700</h5>
+
+  <!-- Body title -->
+  <h3>Body</h3>
+
+  <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
 
   <img src="{{ site.baseurl }}/assets/img/static/Typography_UI_v1.png">
 </div>
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -9,6 +9,7 @@ title: Typography
   <!-- Heading title -->
   <h3>Headings</h3>
 
+  <h3 class="usa-display">Display 48px in Source Sans Pro 700</h3>
   <h1>Heading 1 in 36px in Source Sans Pro 700</h1>
   <h2>Heading 2 in 24px in Source Sans Pro 700</h2>
   <h3>Heading 3 in 19px in Source Sans Pro 700</h3>
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -25,7 +25,7 @@ title: Typography
   <h3>Links</h3>
 
   <p>This is <a href="#">linked text</a> for body copy</p>
-  <p>This is <a class="usa-link" href="#">linked text</a> for everything else</p>
+  <p class="sans">This is <a href="#">linked text</a> for everything else</p>
 
   <!-- Lists title -->
   <h3>Lists</h3>
--- a/_visual/typography.md
+++ b/_visual/typography.md
@@ -24,8 +24,8 @@ title: Typography
   <!-- Links title -->
   <h3>Links</h3>
 
-  <p>This is <a href="#">linked text</a></p>
-  <p>This is <a class="usa-link" href="#">linked text</a></p>
+  <p>This is <a href="#">linked text</a> for body copy</p>
+  <p>This is <a class="usa-link" href="#">linked text</a> for everything else</p>
 
   <!-- Lists title -->
   <h3>Lists</h3>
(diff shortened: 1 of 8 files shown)


maya:
Also note: We do not have inline styles in the typography section (italics, em, bold, strong, preformatted code, highlighted text). Do we want to add this in?

assets/_scss/elements/_list.scss
```diff
@@ -21,7 +21,7 @@ article {
     }
```
msecret: Why is this under an article rule? What happens if I want this list style when I'm not using an article? What about users that aren't really using article tags because they don't know much about html5 elements? How it is going to be documented that you need to nest it in an article tag?

assets/_scss/elements/_typography.scss
```diff
@@ -17,12 +17,22 @@ p {
...
-  color: $color-base-link;
-  &:hover, &:active, &:visited {
-    color: $color-base-link;
+  border-bottom: 1px dotted;
+  &:hover {
+    border-bottom: 1px solid;
   }
 }
 
+.usa-link {
```
msecret: Classes should eventually be moved out of elements. This is partly due to organizational reasons: you would only expected to see bare html elements in an elements folder. This will also help with a trending upward specificity curve when you import the files. So elements imported first (lowest specificity), then components (medium specificity) then patterns (high specificity) etc.

assets/_scss/elements/_typography.scss
```diff
@@ -17,12 +17,22 @@ p {
...
+  border-bottom: none;
+  font-family: $font-sans;
```
msecret: If this is already defined on `body`, isn't it not needed here?

assets/_scss/elements/_typography.scss
```diff
@@ -17,12 +17,22 @@ p {
...
+  border-bottom: 1px dotted;
+  &:hover {
+    border-bottom: 1px solid;
   }
 }
 
+.usa-link {
+  border-bottom: none;
+  font-family: $font-sans; 
+  font-weight: 700;
```
msecret: You might want to make font weights variables. If a user changes to a different fonts, they may need to reset what the various weights are according to their font.

assets/_scss/elements/_typography.scss
```diff
@@ -32,6 +42,10 @@ h1, h2, h3, h4, h5, h6 {
   line-height: $heading-line-height;
 }
 
+h5 {
+  text-transform: uppercase;
+}
+
 // Create heading mixins 
 
 @mixin title {
```
msecret: The indentation for these lines below seems off by 1 tab.

assets/_scss/elements/_list.scss
maya: Updated here so it's easier to use: https://github.com/18F/govt-wide-patternlibrary/commit/e05f7d179e5caa657775c09e5782249cdf8eec07

assets/_scss/elements/_typography.scss
maya: Where would we add this instead? In core?

assets/_scss/elements/_typography.scss
maya: Like `$bold`?

assets/_scss/elements/_typography.scss
maya: Can't see this issue, looks good to me.

assets/_scss/themes/_usa.scss
```diff
@@ -16,13 +17,13 @@ li {
 }
 
 a {
-  color: $color-main;
+  color: $color-primary;
```
msecret: So I've been thinking about this:

How will people use the theme file or create new themes?

Whats to be gained by putting the color rules in the themes file? Like if I want to change a color for my site, I'd prefer to just set a variable, I won't need all this code here because it's just setting the color for various things. 

One worry I have here is that this file is already pretty large, and it's the begging of the other code, meaning this file could potentially get huge. It migiht be good to really think about what goes in the theme file, (I'm still thinking about it right now). I think variables themselves are the best way to control theme.

assets/_scss/elements/_typography.scss
maya: the normal `<a>` tag has has a border-bottom and serif style, which needs to get removed for the alternate sans-serif link.

assets/_scss/themes/_usa.scss
maya: Good point and this was previously in variables, but just removed to put it in theme instead. I don't have an answer yet about what style should be considered a theme or variable, but we should think more about it.

assets/_scss/elements/_typography.scss
msecret: Technically this is a component, its a link component. I'm not sure how you were going to do your component files but it could just be `components/link.scss`

assets/_scss/themes/_usa.scss
maya: Someone would just add another theme file to the themes folder to change their custom theme. Also variables does not contain any live code or styling, just variables. The theme files contain the implementation of variables or other customizations.

assets/_scss/elements/_typography.scss
maya: We're thinking of components as a combination of elements. They are more complex than elements.

assets/_scss/elements/_typography.scss
msecret: Yea it could follow typography conventions: $light, $book, $medium, $bold, $exbold, whatever. In reality, most web fonts will only use 2 or 3. Maybe just $light, $medium, $bold will do, but you might want to talk to the designers.

assets/_scss/elements/_typography.scss
msecret: Isn't this technically more complex then an element because it uses a class? It kinda means that it takes the base element (all the rules of the a tag) and builds on top of them.

assets/_scss/elements/_list.scss
maya: I reverted this in 3412d7b as it was messing up the sidebar navigation. I think the original intention of lists being unstyled by default is sound, as sometimes you'll want a list to have semantic structure but not styling. 

Maybe we'll have to document using a bit of HTML5. Can you think of another way to do this? Use something besides an article? Use a class for styled or unstyled?

assets/_scss/elements/_typography.scss
maya: Structurally it's still an HTML element (a link), just with a variant styling. So I think it's fine to include with elements. I think this will add more confusion by separating it out from the default links. 

We also have `components` for more complex things made up of more than one elements.

assets/_scss/elements/_typography.scss
maya: Added in 79dc573.

assets/_scss/elements/_typography.scss
maya: What if we utilize the cascade here so `p a` gets the serif link style and `a` gets sans-serif style? AND/OR we provide targeted classes like `sans-link` or `serif-link`.

assets/_scss/elements/_list.scss
mjovel: I would opt for having an oocss class of "unstyled-list" over changing the default behavior of a list.  the same would go with inline lists as well.

msecret: I think if you want to have unstyled lists as the default you should move these rules under a "styled" list class. I think my issue here is that it doesn't necessarily make sense that the styled list would be under an article tag.

assets/_scss/elements/_typography.scss
msecret: You should nest the nest the style rules if that's how you want the links to be used. People will probably want both, default styles to work for most situations and ways to override them with classes.

I still stand by what I said before though. As a developer, when I see an `elements` folder, I would suspect only rules for html elements, a, h1, etc, would be in there, no classes. So I would have trouble finding where the `usa-link` class would be if I was working on the code base. It would also help for specificity organization that classes don't come until later in the import order in the final file to ensure you have low specificity first so don't override unwanted things.

I think you have to make the best decision on the research you guys did. Just be careful with specificity problems, that's like one of the only things in CSS architecture that really matters.

_visual/typography.md
```diff
@@ -5,6 +5,43 @@ title: Typography
...
+    <li>Unordered list item</li>
+  </ul>
+
+  <ol>
+    <li>Ordered list item</li>
+    <li>Ordered list item</li>
+    <li>Ordered list item</li>
+  </ol>
+
   <img src="{{ site.baseurl }}/assets/img/static/Typography_UI_v1.png">
```
juliaelman: Will there be another pull request that adds the example?

assets/_scss/themes/_usa.scss
maya: This is good discussion around themes but we should move this to its own issue, outside of the typography PR.

_visual/typography.md
maya: That's what I was thinking unless you feel that it must be included here.

juliaelman: :+1: works for me! Just wanted to clarify.

assets/_scss/elements/_list.scss
mjovel: If you make unstyled list the default and styled-list what content creators need to use to create what is otherwise an html default expected behavior it will cause greater confusion.

msecret: Yea I agree with @mjovel, I think styled lists should be the default, and use a class for the unstyled lists. I think I remember it being what I've seen in other libraries.

maya: @mjovel @msecret Added `unstyled-list` class here: 322ba49. Lists are now styled by default.

assets/_scss/elements/_typography.scss
maya: Updated links in 66574b2. Links will be serif style by default. When parent is a sans-serif font, it will have the sans styling (`.sans a`).

assets/_scss/elements/_list.scss
```diff
@@ -1,29 +1,33 @@
...
-}
-
-article {
-  ul, ol {
-    padding-left: 2em;
+  margin: {
+    top: 2em;
+    bottom: 2em;
+  padding-left: 2em;
+  }
```
msecret: Space between closing bracket and `li`

assets/_scss/elements/_typography.scss
```diff
@@ -17,9 +17,9 @@ p {
 }
 
 a {
-  color: $color-base-link;
-  &:hover, &:active, &:visited {
-    color: $color-base-link;
+  border-bottom: 1px dotted;
```
msecret: newline after this.

assets/_scss/themes/_usa.scss
```diff
@@ -16,13 +17,13 @@ li {
...
   text-decoration: none;
```
msecret: There should be a space after this line

assets/_scss/themes/_usa.scss
```diff
@@ -16,13 +17,13 @@ li {
...
   &:hover, &:active {
-    color: $color-main-dark;
+    color: $color-primary-darker;
   }
```
msecret: There should be a space after this line.

assets/_scss/themes/_usa.scss
```diff
@@ -88,7 +89,7 @@ aside {
   background-color: #fff;
   border-bottom: 1px solid $color-grey-300;
   h1 {
-    color: $color-main;
+    color: $color-primary;
   }
```
msecret: There should be a space after this line.

assets/_scss/themes/_usa.scss
```diff
@@ -88,7 +89,7 @@ aside {
   background-color: #fff;
   border-bottom: 1px solid $color-grey-300;
```
msecret: There should be a space after this line.

assets/_scss/elements/_list.scss
```diff
@@ -1,29 +1,35 @@
-// Make lists unstyled by default
 ul, ol {
```
msecret: I added a rule in our styleguide that selectors separated by commas only have to be on new lines if they are more then 5 chars, so this would be fine. Let me know if you want them to always have newlines, so like this:

```
ul,
ol {
...

```

maya: I don't feel strongly either way about this. I wonder if it's easier to have the same rule applied to all selectors.

msecret: I like it this way, I'll keep it for now and we can change later if needed.

juliaelman:
@maya thanks for getting this work completed! are we planning on incorporating the download button outlined in the wireframes in the next iteration of this page? 

See the following screenshot FYR: https://cloud.githubusercontent.com/assets/10144074/8410509/af388820-1e4c-11e5-8f12-44180f86a764.png

maya:
@juliaelman we are adding the download buttons on the two components we are testing with, accordions and footers. This is not one of them. Thanks for pointing this out!

PR #109, opened 2015-07-01
Add accordion component with basic code snippets layout

juliaelman:
This pull request is the start of work for the accordion menu component. A few things to note:

- The arrow is currently an image based on @angel original mock-up [here](https://github.com/18F/govt-wide-patternlibrary/blob/18f-pages-staging/assets/img/static/Accordions_UI_v1.png). The images are based off of Font Awesome icons, of which we have not determined is something we want to use in the pattern library, see: https://github.com/18F/govt-wide-patternlibrary/issues/43
- The pull request **does not** include the copy code or the "Show Code/Hide Code" functionality based in the following wireframe: https://cloud.githubusercontent.com/assets/10144074/8410482/8724b944-1e4c-11e5-9e63-4e156d0ec269.png. It will be added in at later pull requests which will be downstream to this code base.


_components/accordions.md
```diff
@@ -4,9 +4,153 @@ type: component
 title: Accordions
 ---
 
-<div class="preview">
-  <!-- Add HTML markup for example here -->
-  <img src="{{ site.baseurl }}/assets/img/static/Accordions_UI_v1.png">
+<h3>Bordered Accordion</h3>
+
+<div class="code-sample">
+  <ul id="accordion-menu">
```
msecret: Why are we using ids here? Using a class like `usa-accordion` would allow people to put multiple on the page, would lessen potential issues with specificity in the CSS code.

Let me know if you're OK with changing to IDs and I can update the code.

assets/_scss/components/_accordions.scss
```diff
@@ -0,0 +1,46 @@
+#accordion-menu {
+  border-color: $color-grey-600;
+  border-style: solid;
+  border-width: 1px;
+  border-bottom-width: 0;
+  color: $color-grey-900;
+  margin: 0;
+  padding: 0;
+
+  li {
```
msecret: These selectors are fragile, in that if somebody wants to make an accordion out of different elements then ps for headers, the code wouldn't work. I suggest using classes instead to ensure the code breaks in less situations.

assets/_scss/components/_accordions.scss
```diff
@@ -0,0 +1,46 @@
...
+        display: none;
+      }
+
+      p {
+        background: url('../img/arrow-right.png') no-repeat;
+        background-position: 15px;
+      }
+    }
+
+    p {
```
msecret: Same comment as above for here. This using a p tag here has the added problem of hurting accessibility because it's not going to be shown as a link or button.

_components/accordions.md
juliaelman: While I can understand the assumption that users would want to use an accordion more than once on a page, I'd rather leave that up to our UX team to determine. Let's go ahead and leave this as an ID for now and point out to the team that this is a good thing to test for. //cc @carodew @colinpmacarthur

assets/_scss/components/_accordions.scss
juliaelman: This is a good point in moving these to classes, but again I think that would be something we'd want to test for.

Also, these pieces will be part of code snippets in which users are able to change/modify these snippets as they see fit. It will be interested to see how folks use, change and interact with them!

juliaelman:
@msecret thanks for taking a look! I placed some answers to your questions and let me know if there are any other pieces you'd like changed before we merge in.

juliaelman:
@msecret I found an issue with the borders on the accordion. Looks like commit 889893f571f761f038f6ff476bf3694d0040b29e removed the border at the bottom of the `<li>` in the accordion. See screenshot below:

screen shot 2015-07-02 at 9 33 01 am

_components/accordions.md
colinpmacarthur: Accordions are tricky. They increase the number of clicks necessary to traverse a page's content. They make it more difficult to compare content on two pages. They're generally a bit of a usability minefield. See http://www.nngroup.com/articles/accordions-complex-content/.

That said, if a developer makes a conscious decision to use one accordion, I can't think of a reason why it would be bad to use multiple ones.

I'd love to hear @carodew's thoughts.

colinpmacarthur: @carodew and I just chatted about this.

We think there are several legitimate uses of multiple accordions per page. (e.g. Accordion nav and accordion in the content.) We also think that the library should lean towards accommodating lots of different uses for a given component (think that's been the strategy so far). So the UX team is all for making it possible to have multiple accordions per page. 

cc: @juliaelman and @mollieru

juliaelman: Right on. Thanks to the UX team for your feedback and letting us know the best way to move forward :+1: 

@msecret feel free to change this to a class and update the JavaScript to reflect this change.

msecret:
@juliaelman fixed the border problem in  7fbcd54, sorry about that

assets/_scss/components/_accordions.scss
msecret: OK thats fine, we can keep things like this until we see there being an actual problem in testing.

It would be good to make sure all devs are on the same page in how we are dealing with nested parts of components like this.

juliaelman:
@msecret +1 thanks for the changes and feel free to merge in!

PR #111, opened 2015-07-01
Add buttons

maya:
This adds buttons to the library.

This is what it looks like:
screen shot 2015-07-01 at 4 33 39 pm

Here are the normal, active, and hover states, which are live:
active and hover

Thanks @noahmanger for helping with the focus state for rounded corners!

### DIFF
--- a/assets/_scss/elements/_buttons.scss
+++ b/assets/_scss/elements/_buttons.scss
@@ -12,15 +12,15 @@ button[type="reset"] {
   border: 0;
   border-radius: $border-radius;
   font-size: $base-font-size;
+  line-height: 1;
   text-align: center;
-  width: 8em;
   margin: {
     top: .5em;
     bottom: .5em;
     right: .5em;
   }
   outline: none;
-  padding: .5em .7em;
+  padding: 1rem 2rem;
   &:hover {
     background-color: $color-grey-400;
   }
--- a/assets/_scss/elements/_buttons.scss
+++ b/assets/_scss/elements/_buttons.scss
@@ -31,8 +31,41 @@ button[type="reset"] {
 }
 
 input[type="submit"]:disabled {
+  background-color: $color-grey-300;
   color: $color-grey-700;
+}
+
+input[type="button"].button-secondary,
+input[type="submit"].button-secondary,
+input[type="reset"].button-secondary,
+input[type="image"].button-secondary,
+button[type="button"].button-secondary,
+button[type="submit"].button-secondary,
+button[type="reset"].button-secondary {
+  background-color: $color-grey-600;
   &:hover {
-    background-color: $color-grey-300;
-  }
+    background-color: $color-grey-800;
+  }  
+}
+
+input[type="button"].button-outlined,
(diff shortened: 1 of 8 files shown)


assets/_scss/elements/_buttons.scss
```diff
@@ -0,0 +1,77 @@
...
+
+input[type="button"].button-big,
+input[type="submit"].button-big,
+input[type="reset"].button-big,
+input[type="image"].button-big,
+button[type="button"].button-big,
+button[type="submit"].button-big,
+button[type="reset"].button-big {
+  font-size: 1.9rem;
+  padding: 1.5rem 3rem;
```
msecret: From your research, do you think people will want to control button size? We could potentially variablize this sizing stuff.

assets/_scss/elements/_buttons.scss
```diff
@@ -0,0 +1,77 @@
...
+button[type="reset"] {
+  -webkit-appearance: none;
+  background-color: $color-grey-900;
+  border: 0;
+  border-radius: $usa-border-radius;
+  color: $color-white;
+  font-size: $base-font-size;
+  font-weight: $font-bold;  
+  line-height: 1;
+  text-align: center;
```
msecret: Please keep properties alphabetically ordered according to style guide.

assets/_scss/elements/_buttons.scss
```diff
@@ -0,0 +1,77 @@
...
+  margin: {
+    top: .5em;
+    bottom: .5em;
+    right: .5em;
+  }
+  outline: none;
+  padding: 1rem 2rem;
+  -webkit-font-smoothing: antialiased;
```
msecret: Please keep - properties together either at top or bottom.

assets/_scss/elements/_buttons.scss
maya: Fixed in d2f6ac5

msecret:
Future considerations:

- Allowing a `button` tag to be styled like a text link. This is helpful for when you're using JS for links and don't want to put an `href` attribute. So this would probably be `.unstyled` class on `button`.
- Being able to style a `<a>` tag like a button.

assets/_scss/elements/_buttons.scss
```diff
@@ -0,0 +1,77 @@
+// Buttons
+
+input[type="button"],
+input[type="submit"],
+input[type="reset"],
+input[type="image"],
+button[type="button"],
+button[type="submit"],
+button[type="reset"] {
+  -webkit-appearance: none;
```
msecret: Remove all prefixes and replace with what would be in auto prefixer:

https://github.com/postcss/autoprefixer/wiki/support-list

assets/_scss/elements/_buttons.scss
maya: Done in  d2f6ac5

Scotchester:
:+1: to being able to stile `a`, `button`, and `input` elements using these button styles.

PR #113, opened 2015-07-02
Add namespaced class to buttons

maya:
This adds a namespace class to our button classes. I forgot to do it on the other PR and fixing here.


juliaelman:
+1

PR #114, opened 2015-07-02
Accordion  adding accessibility

msecret:
@juliaelman this is me working off your accordion pull request. This shows what an accessible accordion would be like, but as mentioned in commit message, puts a little more work on the user. 

We could also try to build out some code that adds some of the aria attributes for the user, so it's simplified for them. We should discuss.

**(Julia) NOTE:** This pull request is downstream from #109.


juliaelman:
Nice @msecret! A few notes:

- Be sure to update the code snippet boxes with the updated HTML, CSS and JavaScript (starting here: https://github.com/18F/govt-wide-patternlibrary/pull/114/files#diff-14cc5ce9ad55a36abdbd1b1254840ae3R58)
- See the following comment for an issue with the borders on the accordion: https://github.com/18F/govt-wide-patternlibrary/pull/109#issuecomment-118034227
- There is an issue with the last accordion not opening properly. See screen recording below:
wukvq5qp5k

Also, in the future it might be good to either submit the pull request against the one you are enhancing or be explicit that it is downstream from another request.

juliaelman:
@msecret thanks for making those fixes! A few notes:

- It looks like you are anticipating that the user will be downloading the `components.js` file, therefore making the `accordion` function callable. AFAIK for testing purposes we are wanting each of the code snippets to be able to stand alone and that the user will merely have the ability to copy/paste the code in order to create the accordion box. While I agree with the method you've incorporated, it is making a few assumptions that our research has concluded to. Open to other ideas, but would like to stay as close to the testing plan as possible.
- Closing PR #138, as it looks like you made the same changes in this PR instead.

msecret:
@juliaelman I was thinking that the user would copy the whole code sample I wrote, but I think I forgot the update the JavaScript code snippet. I can do that now.

msecret:
I just merged this in. It was a difficult merge, so I may have caused some issues.

Some questions:
- I put in all the JS in the JS code section. Should I make an effort to make it more concise? @juliaelman,  I wasn't sure what you mean't by your comment.
- This code uses the `.button-unstyled` class defined in `buttons.scss`. This was a similar issue to @maya list thing. I want to discuss solutions to this, but can just copy it over to the CSS for now.

_components/accordions.md
```diff
@@ -154,39 +184,35 @@ title: Accordions
...
-  background: url("https://cdn.rawgit.com/18F/govt-wide-patternlibrary/18f-pages-staging/assets/img/arrow-right.png") no-repeat;
+.usa-accordion ul button[aria-expanded=false] {
+  background: url("../img/arrow-right.png") no-repeat;
   background-position: 15px;
-  border-bottom: 0; 
+  border-bottom: 0;
 }
 
-#accordion-menu ul li p {
-  background: url("https://cdn.rawgit.com/18F/govt-wide-patternlibrary/18f-pages-staging/assets/img/arrow-down.png") no-repeat;
```
juliaelman: Please put the correct reference back in so that users are able to use the image resource without having to download it.

_components/accordions.md
```diff
@@ -154,39 +184,35 @@ title: Accordions
...
 }
 
-#accordion-menu ul li.hidden .content {
-  display: none; 
+.usa-accordion ul li:last-child {
+  border-bottom: 0;
 }
 
-#accordion-menu ul li.hidden p {
-  background: url("https://cdn.rawgit.com/18F/govt-wide-patternlibrary/18f-pages-staging/assets/img/arrow-right.png") no-repeat;
```
juliaelman: Please put the correct reference back in so that users are able to use the image resource without having to download it.

juliaelman:
@msecret nice! There do seem to be a few issues in records to the `.button-unstyled` class that are effecting the look/feel of the accordion:

- There is a border radius on the bottom of each `button` that should not be there, per design spec.
government_wide_pattern_library___components

- On `focus`, each button is highlighted yellow. This is also not outlined in the design spec and should be addressed.
screen shot 2015-07-10 at 11 30 29 am

My suggestion would be that our `.unstyled` button should be exactly that... unstyled. :) I'll open a new issue for us to come up with a solution as a team. It would also be good to have this resolved prior to merging this PR in.

assets/_scss/elements/_buttons.scss
```diff
@@ -96,4 +96,12 @@ input[type="reset"].usa-button-big,
 input[type="image"].usa-button-big {
   font-size: 1.9rem;
   padding: 1.5rem 3rem;
-}
\ No newline at end of file
+}
+
+.button-unstyled {
```
juliaelman: Should this be called `.usa-unstyled-button`?

juliaelman:
@msecret perfection! I pushed up a few typo fixes (all were legacy) and am merging in.

PR #137, opened 2015-07-02
Add button update

maya:
This updates the buttons by adding classes for hover and active states and showing those states visually rather than interactively. It removes the different "button types" (button, submit, reset) from the markup but preserves it in the CSS. 

This is what it looks like:
screen shot 2015-07-02 at 1 29 17 pm

### DIFF
--- a/_elements/buttons.md
+++ b/_elements/buttons.md
@@ -8,23 +8,23 @@ title: Buttons
   <!-- Add HTML markup for example here -->
 
   <h3>Primary Buttons</h3>
-  <button type="button">Button</button>
-  <button type="reset">Reset</button>
-  <button type="submit">Submit</button>
+  <button type="button">Static</button>
+  <button type="reset">Active</button>
+  <button type="submit">Hover</button>
 
   <h3>Secondary Buttons</h3>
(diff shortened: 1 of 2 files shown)


_elements/buttons.md
```diff
@@ -8,24 +8,24 @@ title: Buttons
...
+  <button>Static</button>
+  <button class="usa-button-active">Active</button>
+  <button class="usa-button-hover">Hover</button>
 
   <h3>Secondary Buttons</h3>
-  <button class="usa-button-secondary" type="button">Button</button>
-  <button class="usa-button-secondary" type="reset">Reset</button>
-  <button class="usa-button-secondary" type="submit">Submit</button>
+  <button class="usa-button-secondary">Static</button>
+  <button class="usa-button-secondary usa-button-secondary-active">Active</button>
```
msecret: Do you need a separate class for secondary? Will `usa-button-active` not be enough?

maya: We don't right now, but I was planning for the future, in case the primary active and secondary active states would look different. Like if someone chose different colors from what we have, so they may want more control.

PR #138, opened 2015-07-02
Changed accordion to use class

msecret:
Using a class for the accordion allows for multiple to be put on one
page and keeps CSS specificity low to ensure styles are easily
extendible.


juliaelman:
Closing this PR in favor for the same changes made in #114.

Issue #103, opened 2015-06-29
Create dynamic layout for pattern library pages


juliaelman:
In Sprint 2, we moved forward with using flat designs for each of the pages. For Sprint 3, we are starting to flesh out these interior pages with their assigned layouts.

carodew:
Adding the wireframes. 

Since we agreed that we'll need to do testing before we know for sure what works best, the team will build out the easier version (to reduce development time before testing) and build out each component page with a single download button. That button will download the files (HTML, CSS, JS if needed) for that component only. Since the download section layout is single column and pretty simple anyway, it will look the same in a mobile view.

![download button desktop](https://cloud.githubusercontent.com/assets/10144074/8410509/af388820-1e4c-11e5-8f12-44180f86a764.png)


Component page with code snippets (use for Accordion only for testing):

![gwpl-desktop-wireframes v2](https://cloud.githubusercontent.com/assets/10144074/8410482/8724b944-1e4c-11e5-9e63-4e156d0ec269.png)
![gwpl-mobile-wireframe v2](https://cloud.githubusercontent.com/assets/10144074/8410490/8f3bc564-1e4c-11e5-8f72-90a00ce5de2f.png)

juliaelman:
Completed, per merge from #112.

PR #158, opened 2015-07-06
Add show/hide functionality for accordion page

juliaelman:
This pull request includes a very basic show/hide functionality for testing purposes on the accordion page. It is in reference to the following wireframe https://cloud.githubusercontent.com/assets/10144074/8410482/8724b944-1e4c-11e5-9e63-4e156d0ec269.png and helps complete the functionality outlined here in Issue #103.

lzjrqsdpfh

### DIFF
--- a/assets/_scss/components/_accordions.scss
+++ b/assets/_scss/components/_accordions.scss
@@ -1,51 +1,53 @@
 #accordion-menu {
-  border: 1px solid $color-grey-600;
-  border-radius: 5px;
-  color: $color-grey-900;
-  margin: 0;
-  overflow: hidden;
-  padding: 0;
-
-  li {
-    background-color: $color-grey-200;
-    border-bottom: 1px solid $color-grey-600;
-    font-family: $font-sans;
-    list-style: none;
+  ul {
+    border: 1px solid $color-grey-600;
+    border-radius: 5px;
+    color: $color-grey-900;
     margin: 0;
+    overflow: hidden;
+    padding: 0;
 
-    &:last-child {
-      border-bottom: 0;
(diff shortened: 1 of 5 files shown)


assets/_scss/components/_accordions.scss
```diff
@@ -1,51 +1,53 @@
...
-  margin: 0;
-  overflow: hidden;
-  padding: 0;
-
-  li {
-    background-color: $color-grey-200;
-    border-bottom: 1px solid $color-grey-600;
-    font-family: $font-sans;
-    list-style: none;
+  ul {
```
msecret: I thought this PR was just for adding functionality to the style guide itself, not the actual pattern library code. Why was this selector added?

assets/js/code-snippets.js
```diff
@@ -0,0 +1,5 @@
+$(function() {
```
msecret: Is there a chance there will be any more JS code needed for the style guide itself like this? In that case, I'd call this file styleguide.js, otherwise this is fine as is.

juliaelman: Very possibly, but for now I'd like to leave as `code-snippets.js`. I'll be adding in the ability to copy the code in an upcoming pull request and would rather make this it's own file specifically for the code snippets portion of the pattern library website.

assets/_scss/components/_accordions.scss
juliaelman: @msecret good question! I moved the `code-sample` class to be more specific to the area in which the code snippets live, rather than wrapping the entire area. It made more sense to separate those out.

PR #159, opened 2015-07-06
[WIP] Add visual tweaks

maya:
This implements various fixes:

- fixes main nav links by adding unstyled list class (was dropped below navbar)
- makes disclaimer bar shorter and uses rem
- moves skipnav up 2px so it doesn’t show
- adds max width 50rem (500px) for p text used within `usa-content` class. Utilizing the cascade here bc doesn’t make sense to apply a class for such a commonly used element. Idea for nesting this from CFPB.
- reduces size of outline button by .2rem (2px) all around bc border was making it larger. This makes outline button the same size as the rest.

### DIFF
--- a/assets/_scss/elements/_list.scss
+++ b/assets/_scss/elements/_list.scss
@@ -8,8 +8,8 @@ ul, ol {
   li {
     line-height: $base-line-height;
     margin: {
-      top: .3em;
-      bottom: .3em;
+      top: .75em;
+      bottom: .75em;
     }
   }
 }
--- a/assets/_scss/elements/_list.scss
+++ b/assets/_scss/elements/_list.scss
@@ -17,6 +17,7 @@ ul, ol {
   list-style: none;
   padding-left: 0;
   display: table;
+
   li:after {
     content: "";
     display: block;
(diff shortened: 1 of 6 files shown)


assets/_scss/elements/_list.scss
```diff
@@ -2,24 +2,53 @@ ul, ol {
...
+    content: "";
+    display: block;
+    margin-bottom: 0.5em;
+  }
 }
 
-ol {
-  list-style-type: decimal;
+ul li {
+  display: table-row;
```
msecret: Whats the reason for building the list like this, as a table with the :before rule with the bullet?

maya: @msecret in order to get multi-digit list alignment to flush left. See this issue: https://github.com/18F/govt-wide-patternlibrary/issues/42#issuecomment-118968237

juliaelman:
@msecret I believe that this pull request is marked off as a [WIP] by @maya and should not have been merged in yet. Opened an issue (#173) for a visual bug that now appears due to some conflicts with the accordion menu and will be submitting a pull request shortly with a fix.

PR #161, opened 2015-07-06
Add typography example

maya:
This adds the typography example to the library. 

Note: I opted not to alter our header top margins used in the image bc you never know what size of image someone will use. I'm going with our default margins (1.5em) instead, so this will scale with the font size.

This is what it looks like:

**Large screen**

screen shot 2015-07-08 at 5 39 37 pm copy

**Medium**

screen shot 2015-07-08 at 5 41 19 pm

**Small**

screen shot 2015-07-08 at 5 41 49 pm


juliaelman:
+1

PR #175, opened 2015-07-07
Add unstyled list to accordion

maya:
This adds the unstyled list class to the accordion component and makes some changes to the unstyled list classes.

This is how it looks:

screen shot 2015-07-07 at 8 44 59 am

### DIFF
--- a/_components/accordions.md
+++ b/_components/accordions.md
@@ -7,7 +7,7 @@ title: Accordions
 <h3>Bordered Accordion</h3>
 
 <div id="accordion-menu">
-  <ul>
+  <ul class="usa-unstyled-list">
     <li>
       <p>Apple Pie h4</p>
       <div class="content">
--- a/assets/_scss/elements/_list.scss
+++ b/assets/_scss/elements/_list.scss
@@ -59,11 +59,15 @@ li {
   list-style-type: none;
   
   li {
+    display: list-item;
     margin: 0;
 
     &:before {
-      content: "";
-      padding-right: 0;
+      display: none;
     }
+
+    &:after {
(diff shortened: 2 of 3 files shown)


juliaelman:
Added a note about the accordion component and how this change will effect the purpose of this test component. It's probably best to hold off on these visual tweaks until after this sprint and wait till we have more user feedback to determine the best solutions.

maya:
Great points @juliaelman! I think we gotta chew on this issue some more. Some thoughts: If a user does not include our list styles then we won't have the issue you reported. It will only occur when they include it, when that happens the unstyled list class will prevent the styling issue. 

I'm trying to think if there's another smart way of including the fancy flush left lists. For example, with the `max-width: 50opx" for body text, I'm utilizing the cascade so when there is paragraph text inside a `usa-content` class, it takes that max-width. Could something similar work? I'd hate relying on the cascade too much, but maybe it's necessary for highly used elements like this.

juliaelman:
@maya I think this is a separate issue outside of this particular fix. Good point that the user would not have the issue if they did not include the pattern libraries code base! I am okay with merging this in, but ask that WIP pull requests not get merged in the future please :)

As for your second point, let's move this discussion into a new issue! Feel free to open and add your thoughts/ideas there - thanks!

msecret:
I agree with Julia that if some visual tweaks are causing problems in other components we should hold off on them.

I'm also concerned that we're going to have some major headaches dealing with dependencies if we continue like this where adding some code in the base CSS could break any number of components both when they're alone or included with the library. We should consider solutions for mitigating that problem.2

PR #179, opened 2015-07-07
Fixes overlay issues between the header and the main content area.

### LINKED ISSUE #178: Scrolling overlay issue on smaller screen view
When viewing the pattern library website < 850px screen width, there is a overlay issue between the header and the main content. See movie below for reference:

t3z6jwdz7p

juliaelman:
Fixes #178 and hides a few header elements that we failed to hide before.

### DIFF
--- a/_components/accordions.md
+++ b/_components/accordions.md
@@ -9,7 +9,7 @@ title: Accordions
 <div id="accordion-menu">
   <ul class="usa-unstyled-list">
     <li>
-      <p>Apple Pie h4</p>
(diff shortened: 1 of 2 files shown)


juliaelman:
Referencing https://github.com/18F/govt-wide-patternlibrary/issues/178#issuecomment-119341943 here:

@maya thanks for the review! I added z-index ordering of the sidenav to be the equivalent of the `.main-content` should solve the issue.

@msecret want to take a quick review?

PR #177, opened 2015-07-07
Updating styles to remove scroll bars from code snippets.

juliaelman:
Per feedback from the UX/Visual Design team, there are a few small tweaks to the code snippets. The vertical scroll bars have been removed and the "Copy" button has been re-styled.

I've also added in an absolute reference to the arrow images for now for the accordion, as we still have not come to a final conclusion for using Font Awesome vs... ?? would rather do this for now and fix as we come to a solve.

Screenshots below of the design fixes:

screen shot 2015-07-07 at 3 59 32 pm

PR #183, opened 2015-07-09
Fix scrolling issue for code snippets.

### LINKED ISSUE #181: Scrolling abruptly stops when scrolling over <pre>
@carodew found a bug, as outlined by the gif below. @angel nor myself were able to reproduce the issue, but @maya confirmed this to be an issue. 

w7eqdsdelw

juliaelman:
The issue that causes the scrolling is that there is not a set height on `pre`, as it was removed in #177.

Fixes #181.


maya:
Added in a couple of linebreaks. Otherwise, lgtm :+1:

PR #189, opened 2015-07-11
Add updated zip file for footer

maya:
This adds an updated zip file for download.

This is what the file contains:

screen shot 2015-07-10 at 5 13 51 pm


juliaelman:
Nice work @maya! A few notes and a question in regards to the index.html file viewing:

- It looks like `{{ site.base_url }}` needs to be changed on the image assets
- Do we want to include the `Back to top` functionality in the zip download? //cc @carodew
- `head.html`l file is not being included in the output making the page unstyled, see screenshot below:

screen shot 2015-07-13 at 9 16 44 am

maya:
I think the idea was that they'd be including the component in their own project and need update the image URL to wherever they included it, as well as assets. Open to feedback @carodew @colinpmacarthur

colinpmacarthur:
@maya: It would be great if, when developers download the ZIP and open the .html file, the images render without further modifications to the code. They may need to change the image URL's to reference their own assets later. Does that make sense?

maya:
@juliaelman @colinpmacarthur added head to the footer here: 638a8bd

juliaelman:
+1 we probably don't need the "Download Zip" text and button here. See screenshot below.

screen shot 2015-07-13 at 11 49 01 am

maya:
@juliaelman I was going off the UX wireframe. 

See:
screen shot 2015-07-13 at 9 14 21 am

maya:
Ohh you mean on the footer.html file.

maya:
@juliaelman done in 3224f31.

juliaelman:
LGTM +1

PR #192, opened 2015-07-13
Fix arrow image for download footer zip file

maya:
This fixes the arrows in the zip download file. They weren't appearing before and now are.


juliaelman:
Looks good - thanks for the fix @maya!

PR #194, opened 2015-07-13
Add usa prefix to grid-box

maya:
This adds the `usa-` prefix to `grid-box`, `grid-box-full`, and `width-*` classes.

This also removes grid-item class from annotations as it's unnecessary.

This resolves: https://github.com/18F/govt-wide-patternlibrary/issues/188

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -21,11 +21,11 @@ title: Government Wide Pattern Library
 </div>
 
 <div class="usa-grid-box">
-  <div class="grid-item width-one-half annotation">
+  <div class="width-one-half annotation">
     <h3>Use</h3>
     <p>This is where you'll find basic guidelines about the usage of a component. For example, a minimal palette, clear hierarchy, good information design, and ample whitespace will ensure a voice of authority and expertise in communication.</p>
   </div>
-  <div class="grid-item width-one-half annotation">
+  <div class="width-one-half annotation">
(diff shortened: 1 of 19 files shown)


pages/index.html
```diff
@@ -20,12 +20,12 @@
...
 
-<div class="grid-box">
-  <div class="grid-item width-one-half annotation">
+<div class="usa-grid-box">
+  <div class="width-one-half annotation">
     <h3>Use</h3>
     <p>This is where you'll find basic guidelines about the usage of a component. For example, a minimal palette, clear hierarchy, good information design, and ample whitespace will ensure a voice of authority and expertise in communication.</p>
   </div>
-  <div class="grid-item width-one-half annotation">
+  <div class="width-one-half annotation">
```
msecret: Are we going to prefix the `width-one-half` classes too? They sound like they could potentially be a naming conflict. I'm potentially OK with waiting though.

maya: @msecret Open to adding or waiting too. Maybe it'll be more consistent to just add it tho.

msecret: Yea, and the reason we are doing this now is to have less code to change in the future.

maya: Added in 62d22cb.

juliaelman:
@msecret assigning the final review of this pull request to you - thanks!

Issue #193, opened 2015-07-13
Unintended show/hide toggle for footer topics


colinpmacarthur:
If you (on a desktop)...
1. shrink your window to mobile width, and
2. then re-expand it to a more regular desktop width
...you can click the "Topic" headers to show/hide what's underneath them.

![](http://g.recordit.co/vsWGpH0i1z.gif)

Proposed functionality: when there's no accordion, clicking the topic header's should not show/hide their child links.

juliaelman:
Fixed with #202

PR #202, opened 2015-07-14
Fix for disabling click events on window resize (small to large)

juliaelman:
This pull request fixes the interaction bug found in #193.

### DIFF
--- a/assets/js/components.js
+++ b/assets/js/components.js
@@ -62,17 +62,21 @@ $(function() {
     accordion($(this));
   });
 
-
   var footerAccordion = function() {
     if (window.innerWidth < 600) {
-        $('.usa-footer-big nav ul').addClass('hidden');
 
-        $('.usa-footer-big nav').on('click', 'h3', function() {
-          $(this).parent().removeClass('hidden')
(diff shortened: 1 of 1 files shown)


assets/js/components.js
```diff
@@ -62,17 +62,21 @@ $(function() {
...
     if (window.innerWidth < 600) {
-        $('.usa-footer-big nav ul').addClass('hidden');
 
-        $('.usa-footer-big nav').on('click', 'h3', function() {
-          $(this).parent().removeClass('hidden')
-          .siblings().addClass('hidden');
-        });
+      $('.usa-footer-big nav ul').addClass('hidden');
+
+      $('.usa-footer-big nav h3').bind('click', function() {
```
msecret: I saw this yesterday. Technically, this event should only be bound if it hasn't been bound yet, or you'll bind two events. I guess we could just unbind and rebind:

`$('.usa-footer-big nav h3').unbind('click')`

juliaelman:
@msecret fix made, although really not liking this solve. Really wanting to spend some time to modularize the accordion functionality in general and want to keep this in mind as we build out the JS infrastructure.

PR #203, opened 2015-07-14
Add name block to form components

maya:
This adds the name block form component to the library.

This resolves: https://github.com/18F/govt-wide-patternlibrary/issues/198

This is what it looks like:
screen shot 2015-07-14 at 1 37 04 pm

### DIFF
--- a/_components/form-blocks.md
+++ b/_components/form-blocks.md
@@ -16,19 +16,19 @@ title: Forms Blocks
   <form class="usa-form-one-third">
 
     <label for="title">Title</label>
-    <input class="input-small" type="text" name="title">
+    <input class="usa-input-small" type="text" name="title">
 
-    <label for="first-name">First Name <span class="additional_text">Required</span></label>
+    <label for="first-name">First Name <span class="usa-additional_text">Required</span></label>
     <input class="" type="text" name="first-name">
 
     <label for="middle-name">Middle Name</label>
     <input class="" type="text" name="first-name">
 
-    <label for="last-name">Last Name <span class="additional_text">Required</span></label>
+    <label for="last-name">Last Name <span class="usa-additional_text">Required</span></label>
     <input class="" type="text" name="first-name">
 
     <label for="suffix">Suffix</label>
-    <input class="input-small" type="text" name="suffix">
+    <input class="usa-input-small" type="text" name="suffix">
 
   </form>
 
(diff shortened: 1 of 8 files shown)


_components/form-blocks.md
```diff
@@ -12,7 +12,26 @@ title: Forms Blocks
 <h2>Name</h2>
 
 <div class="preview">
-  <!-- Add HTML markup for example here -->
+
+  <form class="usa-form-one-third">
+
+    <label for="title">Title</label>
+    <input class="input-small" type="text" name="title">
```
msecret: These classes should be usa prefixed, like `input-small` and `additional_text`.

_components/form-blocks.md
```diff
@@ -12,7 +12,26 @@ title: Forms Blocks
...
+
+    <label for="first-name">First Name <span class="additional_text">Required</span></label>
+    <input class="" type="text" name="first-name">
```
msecret: I think you can remove the `class=""`

assets/_scss/components/_forms.scss
```diff
@@ -0,0 +1,17 @@
...
+  }
+}
+
+.additional_text {
+  float: right;
+  font-style: italic;
+  font-weight: normal;
+}
+
+.usa-form-one-third {
```
msecret: Is there a reason this can't be on the form class itself? I feel like this insinuates that you can do form-two-third, or form-one-half, but in reality you can't, which might be confusing to somebody using the library.

assets/_scss/elements/_inputs.scss
```diff
@@ -51,6 +51,7 @@ label {
     top: 3rem;
     bottom: 1.5rem;
   }
+  max-width: 38rem;
```
msecret: Might want to make this a variable if it's being used twice.

assets/_scss/components/_forms.scss
maya: Bc other forms have a different width like one half on the address block. What do you think we should do for this issue?

assets/_scss/elements/_inputs.scss
maya: Done in  b3cf166

assets/_scss/components/_forms.scss
maya: Leaving as is after discussing there are multiple form widths.

_components/form-blocks.md
maya: Done in 01e9f01.

_components/form-blocks.md
maya: Done in c9e9546.

Issue #201, opened 2015-07-14
Create coded version of the "Sign In" block


juliaelman:
**Description**
We currently only have a placeholder image version of the sign in block. This issue to to complete the work to create an actual coded, and responsive, version of the visual design wireframes (see layout below).

**To Do**
- [x] Create html and styles (responsive) based off of current structure
- [x] Run HTML Codesniffer and review for accessibility quality and make any necessary changes
- [x] Submit pull request when code is ready for review
- [x] Once reviewed and merged into staging, run by @angel, @carodew, @colinpmacarthur for final approval
- [x] Iterate/make changes and re-submit PR with any fixes prior to pushing to production 

![form_blocks_login_ui_v1](https://cloud.githubusercontent.com/assets/103496/8680586/55a33aa4-2a30-11e5-914e-09fcfb158977.png)

juliaelman:
Approved and ready! Nice work team :rocket:

PR #205, opened 2015-07-14
Add address form component

maya:
This adds in the address form block. 

This resolves: https://github.com/18F/govt-wide-patternlibrary/issues/197

This is what it looks like:

**Large screen**:
screen shot 2015-07-15 at 3 43 06 pm

**Small screen**:
screen shot 2015-07-15 at 3 43 22 pm

PR #206, opened 2015-07-14
Signup form

msecret:
As described in #201


assets/js/components.js
```diff
@@ -57,6 +57,19 @@ function accordion($el) {
   return new Accordion($el);
 }
 
+function togglePassword($el) {
+  var fieldSelector =  '#' + $el.attr('aria-controls'),
```
juliaelman: Since we're not using `#` as the selector, this does not seem to be functioning correctly. Should we change the actual selector to match up accordingly?

msecret: The selector should end up as `#password` as aria-controls = `password`. What aspect isn't working? I'm expecting it to change the password field to a normal text field and back to show and hide what you typed into it.

juliaelman: +1 gotcha! I missed where the `aria-controls` were located and was not entirely sure how the accessibility portion was integrated in. Thank you for clarifying!

assets/_scss/elements/_typography.scss
```diff
@@ -148,9 +148,13 @@ dfn {
...
 }
 
 .usa-image-text {
   margin-top: 0;
-}
\ No newline at end of file
+}
+
+.usa-drop_text {
+  margin-bottom: 0;
```
juliaelman: @maya should this be `.usa-drop-text` instead?

maya: I don't think we need this class with the legend styling from the other PR in #205

maya:
Added a few style updates. lgtm. :+1:

PR #212, opened 2015-07-16
Ssn input

msecret:


msecret:
@carodew @angel @colinpmacarthur , I was having trouble understanding how to implement certain parts of this, such as the UI when you show the SSN. In that UI, it shows that the UI will fill in placeholder for the rest of the SSN that the user hasn't entered yet: `123-__-____`. This is a little weird in code because when the user wants to go back to edit their ssn, they'll have to clear away the placeholder text.

I'm guessing you wanted to clear the placeholder text when they focus back on the ssn input? I was also wondering if this was meant to be a more fancy interface that fills in the placeholders as they type the ssn. Either way, it might be good to discuss quickly.

maya:
@msecret we aren't implementing placeholder text. I think the `123-__-____` was to show when a user was typing in their SSN.

colinpmacarthur:
Maya's right. I think the idea is that the `123-__-____` is an input mask, not placeholder text. When a user starts typing, their SSN number is automatically into dashed chunks. Before they type, there should be nothing in the box.

@carodew - correct me if I'm wrong?

angel:
I think it's an input mask too.

carodew:
Yep – intended to be an input mask.


There are a lot of notes in the UX wireframes for each component that probably need to be carried over to the GitHub issues that track the components being built, not just the final UI design (which doesn't contain those notes). Something we should probably consider adding to our workflow. @juliaelman is that something Colin and I should track and take care of or something whoever sets up the issue should do?

juliaelman:
@carodew +1 on adding the UX wireframes along with the visual design when opening issues. We can address later if it seems to get cluttered.

maya:
lgtm :+1:

juliaelman:
Looks good!

While we did discuss how to handle external JS libraries and implementing them into the pattern library, this seems like an okay enough short term implementation of politespace. We definitely need to consider moving to something more sustainable and how we can eventually create a separate space between the pattern library website and the pattern library itself.

Issue #124, opened 2015-07-02
Feedback: Include a recommended width of the search box and make it as wide as possible


afeijoo:
The 1/3 and 1/4 grids both appear to be pretty short. 

Make the search box *at least* 27 characters wide, so it can accommodate multiple words without obscuring parts of the user's query. Source: http://www.nngroup.com/articles/top-ten-guidelines-for-homepage-usability/

If only a portion of the text is visible at a time, it makes it difficult to review or edit what you’ve typed. In general, search boxes are better too wide than too short, so that users can quickly review, verify and submit the query.

colinpmacarthur:
Thanks! I’ve added this suggestion to our project workflow.

Issue #123, opened 2015-07-02
Feedback: Use only one indicator to search within the search box


afeijoo:
Displaying both a magnifying glass icon on the left and a button on the right (magnifying glass or 'search') is confusing. Searchers typically expect to see only the right-aligned option. The magnifying glass is most commonly used.

Examples: 
* http://www.bing.com/search?q=test
* https://duckduckgo.com/?q=test&ia=meanings
* https://www.google.com/#q=test

colinpmacarthur:
Thanks! I've added this to our project workflow.

PR #214, opened 2015-07-17
Add search bar component

maya:
This adds the search bar :flashlight: component to the library :books:

This addresses: https://github.com/18F/govt-wide-patternlibrary/issues/196

*Note: The way this is coded, search medium is the default search bar.

This is what it looks like:

**Large screen**:
screen shot 2015-07-21 at 8 23 36 pm

**Medium screen**:

screen shot 2015-07-21 at 8 54 00 pm

*Search big transforms to medium

**Small screen**:

screen shot 2015-07-21 at 8 22 29 pm

*Big and medium (default) transform to small

**In action**: :movie_camera:


maya:
Just got the :+1: on 508 from @nickbristow

assets/_scss/components/_search.scss
```diff
@@ -0,0 +1,102 @@
...
+
+    .usa-search-submit-text {
+      font-size: $h3-font-size;
+    }
+  }
+
+  input[type="search"] {
+    width: calc(100% - 8.5rem);
+  }
+}
```
juliaelman: Add newline.

assets/css/styleguide.scss
```diff
@@ -377,4 +377,12 @@ header[role="banner"] {
...
   padding-top: 3em;
+}
+
+// Search bar grid --------- //
+
+.preview-search-bar {
+  .usa-grid-box {
+    padding: 0;
+  }
 }
```
juliaelman: Add newline.

juliaelman:
Nice work! Made a few comments.

Is there a specific reason why we are using `usa-search` instead of `usa-searchbar` for the classes and mixins? Might be good to be more explicit and stay consistent with the naming conventions we are using.

carodew:
Hmmm, some initial thoughts. 

I think we need to remove the :mag_right: placeholder icon in the search_big search bar so we don't have more than one search indicator (re: #123 ). 

Seems to me that when the three bars are all the same width, they should be the exact same design (e.g. medium screen view – why would two have "Search" and one have :mag_right: ? I think we should pick one and go with it. 

When it gets very narrow before the first breakpoint, the search text input gets too narrow to be usable, as noted in issue #124 : 

screen shot 2015-07-22 at 10 33 46 am

Seems like the breakpoint to full width on all three should be earlier to try to preserve the 27 characters wide target (any thoughts on how we can translate that width into a useful measurement, e.g. ems?). 

I'd like to chat with @colinpmacarthur and @angel to see if we can come up with a better wireframe/mock up to illustrate some of these ideas and the responsive behavior before you guys try to interpret this initial mind dump into FE action.

maya:
@carodew thanks for your feedback! thats related to the grid that we put them in. As you can see as soon as it gets to the next breakpoint it switches to the next view. If you want to change the grid system and breakpoints, that would be best to do when we tackle the grids which is not in this sprint.

maya:
@juliaelman good point! If we want to include the "bar", I feel that @msecret may recommend we use usa-search_bar a la the 18F CSS style guide.

assets/_scss/components/_search.scss
maya: Done in e2ffbdc.

assets/css/styleguide.scss
maya: I don't think we need a new line if the new selector is directly following the parent selector with no properties or directives before it.

maya:
@carodew the reason they are different sizes is be they are in a grid system and we are using 1/2, 1/3, and 1/4 container widths that they are in. A user does not have to use them in the grid system, if they don't then they all be the same width, like they are in the next breakpoints.

carodew:
Just chatted with the design folks. Seems like one thing will be easy to fix – removing the placeholder :mag_right: from the Search Big: 

screen shot 2015-07-22 at 11 09 55 am

All of the responsive / grid questions I think are more complicated and raise bigger questions, so we are still chewing on that. Please stand by.

juliaelman:
@maya +1 on changing to `usa-search_bar`

maya:
@juliaelman done in 12dc35a.

maya:
@carodew done in 8888659.

_components/search-bar.md
```diff
@@ -4,9 +4,59 @@ type: component
...
 ---
 
-<div class="preview">
-  <!-- Add HTML markup for example here -->
-  <img src="{{ site.baseurl }}/assets/img/static/Search_Bar_UI_v1.png">
+<div class="preview preview-search-bar">
+
+  <div class="usa-grid-box">
+    <div class="usa-width-one-half">
+      <form action="#" method="post" class="usa-search usa-search-big">
```
msecret: We didn't put actions and methods for any of our other forms right? I think we should stay consistent and do it for all of them or none of them.

_components/search-bar.md
```diff
@@ -4,9 +4,59 @@ type: component
...
-<div class="preview">
-  <!-- Add HTML markup for example here -->
-  <img src="{{ site.baseurl }}/assets/img/static/Search_Bar_UI_v1.png">
+<div class="preview preview-search-bar">
+
+  <div class="usa-grid-box">
+    <div class="usa-width-one-half">
+      <form action="#" method="post" class="usa-search usa-search-big">           
+        <fieldset>
+          <legend class="usa-sr-only">Search</legend>
```
msecret: I don't know what `sr-only` means. Is it an abbreviation?

assets/_scss/components/_search.scss
```diff
@@ -0,0 +1,98 @@
+@mixin usa-search-submit-small {
+  background-image: url('../img/search.png');  
+  background-image: url('../img/search.svg');  
+  background-position: 50%;
+  background-repeat: no-repeat;
+  width: 4.5rem;
+}
+
+@mixin usa-search-submit-default {
+  @include media($mobile-up) {
```
msecret: Doesn't the mobile-up breakpoint just include everything?

_components/search-bar.md
maya: sorry, that stands for screen reader. So the legend will be viewable to only screen readers and off screen for everyone else.

assets/_scss/components/_search.scss
```diff
@@ -0,0 +1,98 @@
...
+
+.usa-search {
+  max-width: none;
+}
+
+.usa-search_bar {
+  position: relative;
+}
+
+button[type="submit"].usa-search-submit {
```
msecret: This looks like another selector problem but, why not use `.usa-search-submit`?

_components/search-bar.md
maya: Correct. Would you recommend we remove it here?

assets/_scss/components/_search.scss
```diff
@@ -0,0 +1,98 @@
...
+    bottom-right-radius: 0;
+    right: none;    
+    top-right-radius: 0;    
+  }
+  float: left;
+  height: 3.3rem;
+  width: calc(100% - 4.5rem);
+}
+
+.usa-search {
```
msecret: I'm confused why this rule is separate from the rule above, which is just a generic input of type search. why do they have to be separate?

assets/_scss/components/_search.scss
```diff
@@ -0,0 +1,98 @@
...
+  position: absolute;
+
+}
+
+button[type="submit"].usa-search-submit-big {
+  @include media($tablet-up) {
+    @include usa-search-submit-default;
+  }
+
+  @include media($tablet-up) {
```
msecret: Can't you put this in the media query aabove and remove this one?

assets/_scss/components/_search.scss
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+button[type="submit"].usa-search-submit-small {
+  @include usa-search-submit-small;
+}
+
+@include media($mobile-up) {
+  input[type="search"] {
+    width: calc(100% - 8.5rem);
```
msecret: I think the 8.5rem, 4.5rem values should be placed in variables at the top of the file and used throughout. I think it might make it a little more clear, and would make maintenance easier in the future.

assets/_scss/components/_search.scss
maya: Done in acfc87f.

assets/_scss/components/_search.scss
maya: Good point since it's used multiple times in the file.

carodew:
Update for everyone – after chatting with folks we think we should make the change to the big search (removing the placeholder icon) and push to staging. For the responsive behavior, with @juliaelman's blessing I've made cards to account for the work in revising that for both UX and FE teams in a future sprint.

assets/_scss/components/_search.scss
maya: This was switched back in d677129 bc it caused issues.

assets/_scss/components/_search.scss
maya: Added in 82d5e0d, but values aren't working. They aren't being replaced.

assets/_scss/components/_search.scss
maya: Bc we are defining these default values on `button[type="submit"]` in our button.scss file and it is not working without this addition.

assets/_scss/components/_search.scss
maya: Resolved in 1cdfb56.

assets/_scss/components/_search.scss
msecret: Thats what I thought, so just checking.

I'm still thinking through if we should fix that problem, but its fine for now.

_components/search-bar.md
msecret: Yea lets remove

assets/_scss/components/_search.scss
msecret: What issues did it cause?

_components/search-bar.md
maya: Done in 07bd6c6.

assets/_scss/components/_search.scss
maya: The width I defined in the second media query is not appearing when added inside the first. I tried putting it before and after the include and no dice.

Basically the large button is not getting its larger width in large screen sizes.

assets/css/styleguide.scss
maya: As @msecret says... "only when you have "{ }" do you need space after"

assets/_scss/components/_search.scss
maya: Resolved in 546a4fd.

assets/_scss/components/_search.scss
```diff
@@ -0,0 +1,102 @@
...
+  float: left;
+  height: 3.3rem;
+  width: calc(100% - #{$usa-btn-small-width});
+}
+
+.usa-search {
+  max-width: none;
+}
+
+.usa-search_bar {
```
msecret: This should actually be `usa-search-bar` because the bar in this case is a sub element of the search. Is that the correct thinking?

maya: What are we considering "search" is?

assets/css/styleguide.scss
juliaelman: @maya @msecret we've been adding new lines to the end of every file and think it'd be good to stay consistent in this practice.

maya: Sorry didn't realize that's what you meant. Done in a46c293.

assets/_scss/components/_search.scss
juliaelman: @maya `search` could be in relation to many things such as, search results, search autocomplete, etc. etc. In this case, I agree with @msecret and we should change to `usa-search-bar`. Thanks!

maya: Done in 4eb245f.

juliaelman:
+1 nice work team!

PR #215, opened 2015-07-17
Sign in styling fixes

msecret:
Sets the password field to default to password, changes "show password" to sans as asked by design team, increases font size on that sans font a bit because it was looking really small.


maya:
lgtm

PR #217, opened 2015-07-20
Removing unused pages template code and content

blacktm:
Since this is just a simple Jekyll site, and so the “go” script, Gemfiles, and excludes in `_config.yml` aren’t necessary for hosting on 18F Pages, which is almost identical to GitHub Pages (with some additional plugins enabled). Removing unused code/content will simplify development and reduce barriers to contribution. Here's our live [consulting page](https://github.com/18F/consulting) without this stuff for reference.


juliaelman:
Spoke with @blacktm and will be moving over to using `make` for compilation. This work will take some re-organization of our Travis CI instance, as well as some of our contributing docs. 

Moving this into Sprint 5.

blacktm:
Related: ["Why Use Make"](http://bost.ocks.org/mike/make/) from the creator of D3.js

juliaelman:
After discussing with the team, we all agreed that this is not a top priority. We are closing this pull request for now and will re-address as needed in the future. //cc @maya @msecret @blacktm

Issue #195, opened 2015-07-14
Create coded version of "Contact" block


juliaelman:
**Description**
We currently only have a placeholder image version of the contact block. This issue to to complete the work to create an actual coded, and responsive, version of the visual design and ux wireframes (see layout below).

**To Do**
- [x] Create html and styles (responsive) based off of current structure
- [x] Run HTML Codesniffer and review for accessibility quality and make any necessary changes
- [x] Submit pull request when code is ready for review
- [x] Once reviewed and merged into staging, run by @angel, @carodew, @colinpmacarthur for final approval
- [x] Iterate/make changes and re-submit PR with any fixes prior to pushing to production

**UX Wireframe**
![contact us_ux_v1](https://cloud.githubusercontent.com/assets/955558/8777696/98eeaf28-2ec7-11e5-9d41-50328cfad721.png)

**Visual Design Wireframe**
![contact-form_ui_v1](https://cloud.githubusercontent.com/assets/103496/8680411/2a31641e-2a2f-11e5-83cc-d93e2783d792.png)

maya:
Completed #218

PR #218, opened 2015-07-21
Add contact form

blacktm:
This adds the contact form to the library.

This addresses #195 and https://trello.com/c/4FAwqejU

### DIFF
--- a/_components/form-blocks.md
+++ b/_components/form-blocks.md
@@ -185,7 +185,7 @@ title: Forms Blocks
       
       <div class="usa-input-buttons-inline">
         <input type="submit" value="Submit">
-        <a class="usa-button-cancel" href="javascript:void(0)">Cancel</a>
+        <a href="javascript:void(0)">Cancel</a>
       </div>
       
     </fieldset>
--- a/_components/form-blocks.md
+++ b/_components/form-blocks.md
@@ -178,7 +178,7 @@ title: Forms Blocks
         <input id="contact-email" name="contact-email" type="text">
       </div>
       
-      <div class="">
+      <div>
         <label for="contact-comments">Comments</label>
         <textarea id="contact-comments" name="contact-comments" rows="4" cols="50"></textarea>
       </div>
--- a/_components/form-blocks.md
+++ b/_components/form-blocks.md
@@ -162,8 +162,31 @@ title: Forms Blocks
 <h2>Contact Form</h2>
 
 <div class="preview">
-  <!-- Add HTML markup for example here -->
-  <img src="{{ site.baseurl }}/assets/img/static/Contact-Form_UI_v1.png">
+  
+  <!-- <img src="{{ site.baseurl }}/assets/img/static/Contact-Form_UI_v1.png"> -->
+  
+  <form>
+    
+    <fieldset>
+    
+      <legend>Contact us</legend>
+      <p>Email us <a href="javascript:void(0)">info@agency.gov</a>, give us a call at (800)-CALL-GOVT, or send us a message below.</p>
+      
+      <label for="contact-name">Name <span class="usa-additional_text">Optional</span></label>
+      <input id="contact-name" name="contact-name" type="text">
+      
+      <label for="contact-email">Email Address</label>
+      <input id="contact-email" name="contact-email" type="text">
+    
+      <label for="contact-comments">Comments</label>
+      <textarea id="contact-comments" name="contact-comments" rows="4" cols="50"></textarea>
+      
+      <input type="submit" value="Sign in"><a href="javascript:void(0)">Cancel</a>
+      
+    </fieldset>
+    
+  </form>
+  
 </div>
 
 <div class="usa-grid-box">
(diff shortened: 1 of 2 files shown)


_components/form-blocks.md
```diff
@@ -162,8 +162,31 @@ title: Forms Blocks
...
+      <label for="contact-name">Name <span class="usa-additional_text">Optional</span></label>
+      <input id="contact-name" name="contact-name" type="text">
+      
+      <label for="contact-email">Email Address</label>
+      <input id="contact-email" name="contact-email" type="text">
+    
+      <label for="contact-comments">Comments</label>
+      <textarea id="contact-comments" name="contact-comments" rows="4" cols="50"></textarea>
+      
+      <input type="submit" value="Sign in"><a href="javascript:void(0)">Cancel</a>
```
msecret: Is this supposed to be an input or a link?

msecret:
In general, this is looking good

blacktm:
@msecret Sorry, which part?

_components/form-blocks.md
msecret: I think I was confused by line 184, because theres both an input and a link, but now I see that there should be both. I'd recommend moving the link onto the next line.

assets/_scss/components/_forms.scss
```diff
@@ -129,3 +148,14 @@ input.usa-input-medium {
   font-style: italic;
   font-weight: normal;
 }
+
+.usa-input-buttons_inline {
```
maya: @msecret do we want to a dash rather than an underscore here since inline a modifier?

msecret: The `inline` sounds like a modifier so I would say a dash works better.

assets/_scss/components/_forms.scss
```diff
@@ -129,3 +148,14 @@ input.usa-input-medium {
...
+  button[type="submit"],
+  input[type="submit"] {
+    display: inline;
+  }
+  
+  .usa-button-cancel {
```
msecret: I think this would be more robust if it used something like:

```
button:last-of-type,
input:last-of-type {
```

That would mean the user could put the buttons in any order they please and wouldn't be confined to needing a `usa-button-cancel`

_components/form-blocks.md
```diff
@@ -162,8 +162,36 @@ title: Forms Blocks
...
+        <input id="contact-email" name="contact-email" type="text">
+      </div>
+      
+      <div class="usa-input-grid-large">
+        <label for="contact-comments">Comments</label>
+        <textarea id="contact-comments" name="contact-comments" rows="4" cols="50"></textarea>
+      </div>
+      
+      <div class="usa-input-grid-medium usa-input-buttons_inline">
+        <input type="submit" value="Sign in">
```
maya: Can you change the value to "submit"?

blacktm: D'oh! Fixed.

maya: Thanks! Feel free to include the commit in the comment too, just so we can trace it.

For example, fixed in a2ce15e :)

blacktm: Got it, will do.

assets/_scss/components/_forms.scss
maya: Thanks... @blacktm can you update to use a dash here?

blacktm: Yep, fixed in 4c60e94.

maya:
There's 1.5rem too much padding on the top margin of the comments field label. What's the best way to remove this?

assets/_scss/components/_forms.scss
maya: I think that any link style (non-button) element will want this margin, regardless if it's the last or not.

assets/_scss/components/_forms.scss
```diff
@@ -129,3 +152,14 @@ input.usa-input-medium {
   font-style: italic;
   font-weight: normal;
 }
+
+.usa-input-buttons-inline {
```
maya: It seems like this could be potentially abstracted further into .usa-input-inline? What do you think @msecret @blacktm Is there anything unique to it being a button?

assets/_scss/components/_forms.scss
msecret: Thats a good point, although maybe the first one won't.

assets/_scss/components/_forms.scss
msecret: I'm fine with this as is.

maya:
Just adding that I fixed the issue with spacing / grids in 769428e.

assets/_scss/components/_forms.scss
maya: :+1:

assets/_scss/components/_forms.scss
maya: if it follows a button the first one will. On it's own, no.

maya: Adding margin to the right of the button might be a fix that could apply to everything.

maya: Added a more robust solution here: 14273bb

PR #230, opened 2015-07-24
Change checked outline to blue

maya:
This changes the outline to blue when it's checked.

This is what it looks like:

screen shot 2015-07-24 at 1 30 17 pm


angel:
:+1:

PR #247, opened 2015-07-28
Dropdowns

msecret:


maya:
Same comment I left on the closed PR: https://github.com/18F/govt-wide-patternlibrary/pull/246#issuecomment-125777664

msecret:
@maya I thought we said we were doing browser defaults?

Also: I don't think it's a good thing if the `<select>` with no classes added is broken, which is what is currently happening.

maya:
Yes, we are using browser defaults on the dropdown items but the actual select element should match the style of our inputs.

msecret:
Does that mean just to get a `<select>` working a class has to be added to it?

maya:
Or we could apply that styling to the `<select>` element directly?

maya:
Looking good visually. This does not need a fieldset, just a label and id on the select element.

nickbristow:
@maya is correct. Fieldset is only necessary when you're adding extra instructions (legend) && grouping inputs.

maya:
lgtm :+1:

Issue #240, opened 2015-07-28
Create coded version of Alerts


juliaelman:
**Description**
We currently only have a placeholder image version of alerts. This issue is to complete the work to create an actual coded, and responsive, version of the visual design and any UX wireframes (see layout below).

**To Do**
- [x] Create html and styles (responsive) based off of current structure
- [x] Run HTML Codesniffer and review for accessibility quality and make any necessary changes
- [x] Submit pull request when code is ready for review
- [x] Once reviewed and merged into staging, run by @angel / @mollieru and @carodew / @colinpmacarthur for final approval.
- [x] Iterate/make changes and re-submit PR with any fixes prior to pushing to production

**Visual Design**
![alerts_ui_v1](https://cloud.githubusercontent.com/assets/955558/8933240/c21c1b08-350f-11e5-9304-9d127c6e693b.png)

PR #248, opened 2015-07-29
Add alerts

maya:
This adds alerts to the library :ballot_box_with_check: :x: :heavy_exclamation_mark: :information_source:

This addresses #240. 

This is what it looks like:

**Large screens**:
screen shot 2015-07-29 at 12 02 31 pm

**Small screens**:
screen shot 2015-07-29 at 12 15 05 pm


_components/alerts.md
```diff
@@ -5,7 +5,72 @@ title: Alerts
 ---
 
 <div class="preview">
-  <!-- Add HTML markup for example here -->
+
+  <div class="usa-alert-success usa-sans">
+    <div class="usa-alert-icon">
+    <svg width="52" height="52" role="img" aria-label="success">
```
msecret: I think this should be indendted another step

assets/_scss/components/_alerts.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.usa-alert-text {
+  margin: {
+    bottom: 0;
+    top: 0;
+  }
+}
+
+.usa-alert-success {
+  @extend .usa-alert;
```
msecret: I actually think it's better to not do this and just include the `.usa-alert` class on each element. I remember a case where I tried to do exactly this for Peace corps, and I had to switch back for some reason (I can't remember the reason right now).

maya: Ok seems like it could be safer. Updated here: 66b4a4b.

_components/alerts.md
maya: Fixed in 8b6035e.

_components/alerts.md
```diff
@@ -5,8 +5,42 @@ title: Alerts
 ---
 
 <div class="preview">
-  <!-- Add HTML markup for example here -->
-  <img src="{{ site.baseurl }}/assets/img/static/Alerts_UI_v1.png">
+
+  <div class="usa-alert usa-alert-success usa-sans">
```
msecret: Is there a reason why we can't put the usa-sans rule in usa-alert?

I feel like I see that class everywhere.

maya: Sure, done in 6cb74bc.

Issue #237, opened 2015-07-28
Copy updates for various use and accessibility guidelines


juliaelman:
There are a few places on the site where some copy updates need to happen. Big thanks to @maya for pointing those out!

**Address form**
- [x] Add politespace note bc it's being used in zip code (addressed via #242)
- [x] We're saying users can type abbreviation for state but we have a dropdown with full names (value is abbreviation but not visible)

**Search bar**
- [x] "The magnifying glass has been shown to be almost universally recognized by users as an indicator of search functionality, and doesn’t need to be paired with the word "Search.""

— but you need to provide screen reader text

**Accordions**
- [x] "Accordions are a list of headers that can be clicked to hide or reveal additional content."

— Are they tho? ours are not headers, they are list items

- [x] "a minimum of 2 mm"

— this is a website, why are we using mm?

**Footers**
- [x] "Important contact information should be limited to general email or phone numbers, which should be clickable links to dial from a mobile phone."

— our phone numbers are not currently clickable

- [x] "Physical addresses should live on contact pages users can navigate to from the accordion links"

— This is not clear. Are we saying that users must provide a link to the contact page in the footer? Just say "footer links"

- [x] "When collapsed into an accordion"

— explicitly state this is only in small screens

colinpmacarthur:
@maya re: 
> "The magnifying glass has been shown to be almost universally recognized by users as an indicator of search functionality, and doesn’t need to be paired with the word "Search."" — but you need to provide screen reader text

Is there a particular place we need to provide the screen reader text?

maya:
Inside the button itself like we are doing.

![screen shot 2015-07-29 at 1 47 07 pm](https://cloud.githubusercontent.com/assets/5249443/8969412/1685496a-35f8-11e5-9ede-79042d7b9eb8.png)

I don't know if there's a better way to state this cc @nickbristow

juliaelman:
Fixed via #250

PR #250, opened 2015-07-30
Copy updates for various use and accessible guidelines

colinpmacarthur:
To resolve issue #237.

### DIFF
--- a/_components/accordions.md
+++ b/_components/accordions.md
@@ -59,10 +59,9 @@ title: Accordions
 <div class="usa-grid-box">
   <div class="usa-width-one-half">
     <h3>Use</h3>
-    <p>Accordions are a list of headers that can be clicked to hide or reveal additional content.</p>
     <ul>
       <li>
-        They are best used when users only need a few specific pieces of content within a page.
+        Accordions are best used when users only need a few specific pieces of content within a page.
       </li>
       <li>
         Accordion headers can serve as a mini-outline for the content on a page.
--- a/_components/accordions.md
+++ b/_components/accordions.md
@@ -79,7 +79,7 @@ title: Accordions
         Clicking anywhere on the header (not just on the icon) triggers expansion or collapse of that section.
       </li>
       <li>
-        Any interactive element within each section should be placed a minimum of 2 mm from the header to avoid accidentally triggering the wrong interaction.
+        Any interactive element within each section should be placed a minimum of 10 pixels from the header to avoid accidentally triggering the wrong interaction.
       </li>
       </ul>
   </div>
--- a/_components/accordions.md
+++ b/_components/accordions.md
@@ -79,7 +79,7 @@ title: Accordions
         Clicking anywhere on the header (not just on the icon) triggers expansion or collapse of that section.
       </li>
       <li>
-        Any interactive element within each section should be placed a minimum of 10 pixels from the header to avoid accidentally triggering the wrong interaction.
+        Make sure interactive elements within the collapsable region are far enough from the headers that users don't accidentally trigger a collapse. (The exact distance depends on the device.)
       </li>
       </ul>
   </div>
--- a/_components/form-blocks.md
+++ b/_components/form-blocks.md
@@ -107,7 +107,6 @@ title: Forms Blocks
     <p>As you customize this form, ensure it continues to:</p>
     <ul>
       <li>Label the optional ones. Users can infer that all the others are required.</li>
-      <li>Let users type their state's abbreviation when they reach the "state" drop down menu.</li>
       <li>Support both 5 and 9 digit zips. Some addresses require a nine digit ZIP code. The input mask should be "#####-####"" so that the text is properly formatted, regardless of whether a user enters a 5 or 9 digit zip code.</li>
     </ul>
   </div>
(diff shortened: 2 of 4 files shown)


colinpmacarthur:
@carodew: Can you double check these edits? Thanks!

_components/accordions.md
```diff
@@ -80,7 +79,7 @@ title: Accordions
         Clicking anywhere on the header (not just on the icon) triggers expansion or collapse of that section.
       </li>
       <li>
-        Any interactive element within each section should be placed a minimum of 2 mm from the header to avoid accidentally triggering the wrong interaction.
+        Any interactive element within each section should be placed a minimum of 10 pixels from the header to avoid accidentally triggering the wrong interaction.
```
carodew: So the tricky thing here is when screens have differing pixel density. I agree that the 2mm is a confusing measurement and not very useful, but it's based on the size of an average finger/thumb tip (touch size). 10px on a low resolution screen is smaller than 10 px on a high resolution screen. See http://www.lukew.com/ff/entry.asp?1085 

I wish there was an easier way to provide guidance on this. Maybe we just need to be more vague and say "The average adult fingertip is between 8-10 millimeters in diameter; make sure any interactive elements within the collapsable region are far enough from the headers so that users don't accidentally trigger a collapse."

???

_components/form-blocks.md
```diff
@@ -107,7 +107,6 @@ title: Forms Blocks
     <p>As you customize this form, ensure it continues to:</p>
     <ul>
       <li>Label the optional ones. Users can infer that all the others are required.</li>
-      <li>Let users type their state's abbreviation when they reach the "state" drop down menu.</li>
       <li>Support both 5 and 9 digit zips. Some addresses require a nine digit ZIP code. The input mask should be "#####-####"" so that the text is properly formatted, regardless of whether a user enters a 5 or 9 digit zip code.</li>
```
carodew: Should we refer to the note in accessibility about Politespace? Or is that junking up the content too much?

_components/search-bar.md
```diff
@@ -64,7 +64,7 @@ title: Search Bar
     <h3>Use</h3>
     <ul>
       <li>Allow the search bar to be as wide as possible, but a minimum of 27 characters wide. This allows users to enter multiple search terms without their inputs being obscured. The more users can see their search terms, the easier it is to review, verify, and submit their search query.</li>
-      <li>The magnifying glass has been shown to be almost universally recognized by users as an indicator of search functionality, and doesn’t need to be paired with the word "Search."</li>
+      <li>The magnifying glass has been shown to be almost universally recognized by users as an indicator of search functionality and doesn’t need to be paired with the word "Search" (except for screen readers, see the accessibility guidelines).</li>
```
carodew: Suggestion for word-smithing to be shorter: 
"The magnifying glass is almost universally understood and doesn’t need to be visually paired with the word "Search." However, be sure to provide text for screen readers, as noted under Accessibility."

_components/search-bar.md
```diff
@@ -77,6 +77,9 @@ title: Search Bar
       <li><a href="{{ site.baseurl }}/components/#forms-blocks">accessibility guidelines for form templates</a> and</li>
       <li><a href="{{ site.baseurl }}/elements/#inputs">the accessibility guidelines for form controls</a>.</li>
     </ul>
-    <p>There are no other specific accessibility guidelines for this form template.</p>
+    <p>We also recommend:</p>
+    <ul>
+      <li>Even when the search button contains an icon and no text, make sure the word "search" is still inside the <button> element. You can hide this text by absolutely positioning it to -999em (as we have here). This descriptive text tells screen reader users the button's purpose.
```
carodew: Always include the word "search" inside the `<button>` element for screen readers. You can hide this text with `position: absolute; left: -999em;` (as we have here).

carodew:
I added some in-line comments. Other than that looks good, thanks for tackling this!

_components/accordions.md
juliaelman: Rather than associate measurements with thumb prints, it might be best to say 10px for low resolution and give another measurement for higher res?

_components/form-blocks.md
juliaelman: I think it would be good to be consistent with that note, so as not to question why it's in one place and not another.

_components/search-bar.md
juliaelman: +1

_components/search-bar.md
juliaelman: Agreed.

@colinpmacarthur we also want to be sure to add `<code>` around parts of the copy that are related to coding examples.

_components/search-bar.md
colinpmacarthur: Addressed in 13999f2

_components/form-blocks.md
colinpmacarthur: Addressed in 5c5f14c.

_components/accordions.md
colinpmacarthur: This one is tricky. The exact pixel recommendation varies based on device's pixel density and touch precision (which vary substantial device to device). In 5c5f14c, I made this a general principle using @carodew's suggestions. Thoughts?

colinpmacarthur:
Thanks @carodew and @juliaelman! I've made changes in response to your comments.

juliaelman:
LGTM!

colinpmacarthur:
WoW!

PR #257, opened 2015-08-03
Increase tables margins

maya:
This increases the margins in the tables to 1.5rem.

This is what it looks like:
screen shot 2015-08-03 at 11 20 12 am


juliaelman:
+1

Issue #244, opened 2015-07-28
Create coded version of Range Slider


juliaelman:
**Description**
We currently only have a placeholder image version of the Range slider. This issue is to complete the work to create an actual coded, and responsive, version of the visual design and any UX wireframes (see layout below).

**To Do**
- [x] Create html and styles (responsive) based off of current structure
- [x] Run HTML Codesniffer and review for accessibility quality and make any necessary changes
- [x] Submit pull request when code is ready for review
- [x] Once reviewed and merged into staging, run by @angel / @mollieru and @carodew / @colinpmacarthur for final approval.
- [x] Iterate/make changes and re-submit PR with any fixes prior to pushing to production

**Visual Design**
![range_slider_ui_v1](https://cloud.githubusercontent.com/assets/955558/8937960/996209f6-3529-11e5-86f8-080d5ddbb314.png)

juliaelman:
Switching assignee from @blacktm to @maya, who has begun the work here: https://github.com/18F/govt-wide-patternlibrary/pull/259

maya:
this got a +1 from @nickbristow

juliaelman:
Pending final approval from @angel and/or @mollieru.

angel:
+1!

PR #259, opened 2015-08-03
Add range slider

maya:
This adds the range slider to the library using the HTML5 range input.

This addresses #244 

This is what it looks like:
screen shot 2015-08-03 at 4 04 06 pm

Live :movie_camera: 

Note: range inputs will degrade to text inputs in IE9 and below.


assets/_scss/elements/_inputs.scss
```diff
@@ -138,3 +138,98 @@ input[type="radio"]:disabled + label::before {
...
+  border: 1.8px solid $color-grey-600;
+  height: 20px;
+  border-radius: 15px;
+  background: #eeeeee;
+  cursor: pointer;
+  margin-top: -6.5px;  
+  width: 20px;
+}
+
+input[type=range]:focus::-webkit-slider-runnable-track {
```
msecret: Are we going to be able to replace these with generics when we add auto-prefixer? Or are these stuck as browser prefixes?

maya: I'm not positive how autoprefixer works with selectors, I know it'll be able to replace the attributes.

msecret: Do you know what browsers this is going to work on?

maya: http://caniuse.com/#feat=input-range

assets/_scss/elements/_inputs.scss
```diff
@@ -139,3 +139,97 @@ input[type="radio"]:disabled + label::before {
...
+  height: 2rem;
+  border-radius: 1.5rem;
+  background: #eeeeee;
+  cursor: pointer;
+  margin-top: -.65rem;  
+  width: 2rem;
+}
+
+input[type=range]::-moz-range-thumb {
+  background: #eeeeee;
```
msecret: Why aren't these background values using variables?

maya: Fixed in 9a1f8a6.

PR #263, opened 2015-08-04
Add top margin to labels

maya:
This adds more margin on the top of the label so it's the same amount of space above and below. There's more space on the bottom bc of descenders in the font. Since we're using uppercase font, this evens it out.

This also adds a little more margin to the right in case several labels are side by side.

This is what it looks like:
screen shot 2015-08-04 at 4 16 56 pm

### DIFF
--- a/assets/_scss/elements/_labels.scss
+++ b/assets/_scss/elements/_labels.scss
@@ -3,6 +3,7 @@
   border-radius: $border-radius;
   color: $color-white;
   font-size: $small-font-size;
+  margin-right: .1em;
   padding: {
     bottom: 0.1rem;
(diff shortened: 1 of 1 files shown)


assets/_scss/elements/_labels.scss
```diff
@@ -3,11 +3,12 @@
   border-radius: $border-radius;
   color: $color-white;
   font-size: $small-font-size;
+  margin-right: .1em;
```
msecret: Why are the margin values em but the padding values rem?

maya: Removed in margin-right in 4effb07.

Issue #199, opened 2015-07-14
Create coded version of the "Password Reset" block


juliaelman:
**Description**
We currently only have a placeholder image version of the password reset block. This issue to to complete the work to create an actual coded, and responsive, version of the visual design wireframes (see layout below).

**To Do**
- [x] Create html and styles (responsive) based off of current structure
- [x] Run HTML Codesniffer and review for accessibility quality and make any necessary changes
- [x] Submit pull request when code is ready for review
- [x] Once reviewed and merged into staging, run by @angel, @carodew, @colinpmacarthur for final approval
- [x] Iterate/make changes and re-submit PR with any fixes prior to pushing to production

![form_blocks_password_reset_ui_v1](https://cloud.githubusercontent.com/assets/103496/8708995/68fa9560-2b0d-11e5-8f5a-bdf3edef7938.png)

carodew:
I'm adding this wireframe with annotations describing the interactivity of the password reset form. Note – we decided as a team to not worry about validation until the UX and visual design teams are able to incorporate some of the feedback about alerts/inline validation. However it would be good to add the checkmark validation within the "Password must" box. Some notes on that: 

* We aren't recommending any specific requirements for passwords, that's beyond our scope. These are just common examples. 
* The box should be able to serve as a static, informational box if someone wants to implement it without the interactivity. 
* Each criteria listed should be easily modifiable by whoever is implementing it. E.g., if a person's agency says passwords must be at least 12 characters long, they could modify that line. 

![password-reset-form](https://clou.githubusercontent.com/assets/10144074/8725795/4a786198-2ba4-11e5-8342-ff01e1e7d0e3.png)

Ping me if you have questions. Note – the wireframe is showing a 2-column layout while the UI design is showing 1 column. 2-column is for wider screens/layouts and 1-column for narrower. Working with visual design to see if we can get a UI mock up of the 2-column version.

juliaelman:
Waiting on final approval

//cc @angel @carodew @colinpmacarthur @mollieru

carodew:
LGTM!

juliaelman:
Approved from @angel via Slack - thanks team!

PR #264, opened 2015-08-05
Validator for password reset

msecret:
This includes code that can perform validations on regexes defined
through data attributes. It requires the user to write regexes on the
input element that can match up with the line items in a checkbox style
list.

This was done this way to ensure all configuration could be done easily
from within HTML, requiring less to be done in JS.

Refs #199

### DIFF
--- a/_components/form-blocks.md
+++ b/_components/form-blocks.md
@@ -296,8 +296,8 @@ title: Forms Blocks
       <input id="password" name="password" type="password"
         class="js-validate_password"
         data-validate-length=".{8,}"
-        data-validate-uppercase="b"
-        data-validate-numerical="^.*[a-zA-z].*\d.*"
+        data-validate-uppercase="[A-Z]"
+        data-validate-numerical="\d"
         data-validation-element="#validation_list">
       <p class="usa-form-note">
         Password strength
--- a/_components/form-blocks.md
+++ b/_components/form-blocks.md
@@ -284,16 +284,21 @@ title: Forms Blocks
         <div class="usa-alert-body">
           <h3 class="usa-alert-heading">Passwords must:</h3>
         </div>
-        <ul class="usa-check_list">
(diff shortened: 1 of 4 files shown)


_components/form-blocks.md
```diff
@@ -284,16 +284,21 @@ title: Forms Blocks
...
           <li>Another requirement</li>
         </ul>
       </div>
 
       <label for="password">New Password</label>
-      <input id="password" name="password" type="password">
+      <input id="password" name="password" type="password"
+        class="js-validate_password"
+        data-validate-length=".{8,}"
+        data-validate-uppercase="b"
```
juliaelman: I think this regex validation got missed?

juliaelman:
Looks good! There are some issues with the validation (see https://github.com/18F/govt-wide-patternlibrary/pull/264#discussion_r36466377) and movie below:

kiglnqzfw6

msecret:
@juliaelman part of me feels like I want it to work like that, in that we don't want them to use a password that's only numbers. I read it as it should have at least 1 number, but shouldn't be only numbers.

juliaelman:
@carodew @colinpmacarthur thoughts on https://github.com/18F/govt-wide-patternlibrary/pull/264#issuecomment-128516091 ?

maya:
Wouldn't it not validate if the first one (have at least 1 uppercase character) wasn't met though? I think as soon as someone types a number that should check off. Otherwise there's no way to know and it seems confusing.

carodew:
I agree with Maya and Julia. That specific requirement doesn't state that the password can't be all numbers, so I think we should stick with the literal interpretation. If there is a requirement for alpha characters that should be captured in a separate line.

colinpmacarthur:
+1 to Carolyn

juliaelman:
Cool - @msecret does that sound okay to you or do you need further clarification?

juliaelman:
:+1: nice work!

PR #269, opened 2015-08-07
Updating accessibility copy for form blocks, search bar and inputs.

### LINKED ISSUE #243: 508 Copy Feedback
@juliaelman 
## Inputs

Current `All form control tags should have an associated label. The label’s value for the for attribute should exactly match the value for the input’s id attribute. For example, the input with id="favorite-pie" will always have a label with for="favorite-pie". This way screen readers are able to perceive the relevant content.`

The second sentence is a bit odd maybe `The label's for attribute value should match the related input's id attribute.`

I would also work in something to the effect of `This id value must be unique to the entire page.`

I would also add a section on color contrast. `If the color scheme of the buttons are adjusted ensure all states have a minimum contrast ratio of 4.5:1 (for small text, 3:1 for large) for all states of the button. This includes default, hover, selected, disabled.`

## Form blocks 

When explaining fieldsets, I think mentioning that `a single legend is always required for fieldset` and that you can `embed multiple fieldsets and legends` for more complex forms would be helpful. Also maybe provide the example of `a question with radio button options for answer is a good example of fieldset and legend. The whole group is wrapped in a fieldset with the question itself being the legend`

It might also be worth mentioning that keeping your form blocks in a vertical pattern is ideal for accessibility reason. Basically, limited vision can make it hard to scan from right to left. 

### Reset password

Using aria-describedby is a good way to associate the password strength information with the field

## Search bar 

`Even if the search bar isn't using a label, the form field should include a label for screen reader users`

`The search button itself should be a submit button for the form to reduce the number of keystrokes required to use the form.`

juliaelman:
Adding updates per @nickbristow feedback.

Fixes #243

### DIFF
--- a/_elements/inputs.md
+++ b/_elements/inputs.md
@@ -9,9 +9,10 @@ title: Inputs
 <p>As you customize form controls from this library, be sure they continue to meet the following accessibility requirements:</p>
 
 <ul>
-  <li>All  form control tags  should have an associated label. The label’s value for the <code>for</code> attribute should exactly match the value for the input’s <code>id</code> attribute. For example, the input with <code>id="favorite-pie"</code> will always have a label with <code>for="favorite-pie"</code>. This way screen readers are able to perceive the relevant content.</li>
+  <li>All form control tags should have an associated label. The label's for attribute value should match the related input's <code>id</code> attribute and should also be unique to the entire page. For example, the input with <code>id="favorite-pie"</code> will always have a label with <code>for="favorite-pie"</code>. This way screen readers are able to perceive the relevant content.</li>
   <li>Any additional information – such as required, optional, or example text – should be wrapped within the label tags. For example: <code>&lt;label for="name"&gt;Favorite Pie &lt;span&gt;Optional&lt;/span&gt;&lt;/label&gt;</code>. This way screen readers know what additional information is related to each field.</li>
   <li>Do not replace <code>&lt;input&gt;</code> tag-based form controls with styled <code>&lt;div&gt;</code> tags or JavaScript that "fake" form controls. Form controls not written in semantic HTML are very difficult for screen readers to handle.</li>
+  <li>If the color scheme of the buttons are adjusted ensure all states have a minimum contrast ratio of 4.5:1 (for small text, 3:1 for large) for all states of the button. This includes default, hover, selected, disabled.</li>
 </ul>
 
 <p>If you are a building a form with multiple controls, also consider the <a href="{{ site.baseurl }}/components/#forms-blocks">accessibility guidelines in the “Common Form Templates” section</a>.</p>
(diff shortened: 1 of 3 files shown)


_elements/inputs.md
```diff
@@ -9,9 +9,10 @@ title: Inputs
 <p>As you customize form controls from this library, be sure they continue to meet the following accessibility requirements:</p>
 
 <ul>
-  <li>All  form control tags  should have an associated label. The label’s value for the <code>for</code> attribute should exactly match the value for the input’s <code>id</code> attribute. For example, the input with <code>id="favorite-pie"</code> will always have a label with <code>for="favorite-pie"</code>. This way screen readers are able to perceive the relevant content.</li>
+  <li>All form control tags should have an associated label. The label's for attribute value should match the related input's <code>id</code> attribute and should also be unique to the entire page. For example, the input with <code>id="favorite-pie"</code> will always have a label with <code>for="favorite-pie"</code>. This way screen readers are able to perceive the relevant content.</li>
```
colinpmacarthur: Does "label's" need an apostrophe?

_elements/inputs.md
```diff
@@ -9,9 +9,10 @@ title: Inputs
...
   <li>Any additional information – such as required, optional, or example text – should be wrapped within the label tags. For example: <code>&lt;label for="name"&gt;Favorite Pie &lt;span&gt;Optional&lt;/span&gt;&lt;/label&gt;</code>. This way screen readers know what additional information is related to each field.</li>
   <li>Do not replace <code>&lt;input&gt;</code> tag-based form controls with styled <code>&lt;div&gt;</code> tags or JavaScript that "fake" form controls. Form controls not written in semantic HTML are very difficult for screen readers to handle.</li>
+  <li>If the color scheme of the buttons are adjusted ensure all states have a minimum contrast ratio of 4.5:1 (for small text, 3:1 for large) for all states of the button. This includes default, hover, selected, disabled.</li>
```
colinpmacarthur: Comma after ensure?

colinpmacarthur:
Few minor suggested edits, but looks good to me otherwise! Thanks!

_elements/inputs.md
juliaelman: Hrmmm... maybe before ensure so that it reads:

"If the color scheme of the buttons are adjusted, ensure all states..."

_elements/inputs.md
juliaelman: Nope - thanks and it should also be `input` instead. Fixed here: b0f3bb2

_elements/inputs.md
colinpmacarthur: Sounds good to me!

colinpmacarthur:
Changes all look good to me. Thanks @juliaelman!

PR #282, opened 2015-08-11
Small design nudgest to password reset

msecret:
As suggested by design team.

Also includes an aria-labled by for accessibility.


juliaelman:
:+1: from me and @nickbristow

PR #283, opened 2015-08-11
A quick implementation of autoprefixer

jeremiak:
[Autoprefixer](https://github.com/postcss/autoprefixer) is a tool for automatically adding browser prefixes to CSS code. There are a number of ways to implement the tool, but because we've already got Jekyll in place as our build tool the [`octopress-autoprefixer` gem](https://github.com/octopress/autoprefixer) makes this exceedingly easy to get going.

This PR has the following changes:

0. The gem has been added to the `Gemfile` and `Gemfile.lock`
0. `_config.yml` has been configured to use that gem in the build

No subsequent special behavior is required for the tool to start working, `./go serve` will ensure that `autoprefixer` is used.

This is not the only way to add `autoprefixer` to the project, but it seemed like the easiest to try it out.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,3 +4,4 @@ gem 'jekyll'
 
 gem 'rouge'
 gem 'redcarpet'
+gem 'octopress-autoprefixer'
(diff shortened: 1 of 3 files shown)


juliaelman:
@jeremiak looks good! Merging this in and we'll be updating the code base with remove the current prefix values - THANK YOU!

PR #287, opened 2015-08-13
Add new color variables

maya:
This adds in our new color variables to the site :rainbow:

### DIFF
--- a/assets/_scss/core/_variables.scss
+++ b/assets/_scss/core/_variables.scss
@@ -18,51 +18,51 @@ $font-serif:        'Georgia', 'Times New Roman', serif;
 $font-normal:       400;
 $font-bold:         700;
 
+
 // Color
-$color-primary:         #0071bc;
-$color-primary-darker:  #00609f;
-$color-primary-darkest: #000000;
+$color-primary:               #0071bc;
+$color-primary-darker:        #205493;
+$color-primary-darkest:       #112e51;
 
-$color-white:           #ffffff;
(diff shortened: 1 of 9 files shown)


maya:
I want to align the variables better, but some are really short and some are really long so you lose track of them bc they're so far apart.

assets/_scss/core/_variables.scss
```diff
@@ -18,7 +18,52 @@ $font-serif:        'Georgia', 'Times New Roman', serif;
...
+$color-green-light:        lighten($color-green, 20%); // #649967
+$color-green-lighter:      lighten($color-green, 60%); // #94bfa2
+$color-green-lightest:     lighten($color-green, 80%); // #e7f4e4
+
+$color-cool-blue:          #205493;
+$color-cool-blue-light:    lighten($color-cool-blue, 20%); // #4773aa
+$color-cool-blue-lighter:  lighten($color-cool-blue, 60%); // #8ba6ca
+$color-cool-blue-lightest: lighten($color-cool-blue, 90%); // #dce4ef
+
 $color-grey-50:     #fafafa;
```
msecret: Are we going to get rid of the rest of these colors once we remove the references in the code?

If so, could you add a "deprecated" comment?

maya: Do you think there's any value for leaving in the grays?

maya: Removed after discussion

Issue #295, opened 2015-08-13
Update Buttons layout/design to v2 


juliaelman:
**Description**
This card is a part of a series of cards to update the latest style of the buttons to version 2. Below are the updated visual designs for reference:

![buttons_ui_v2](https://cloud.githubusercontent.com/assets/955558/9252386/048fcdf2-41a6-11e5-9ae4-75f5d047005f.png)

maya:
Completed in #310 #301

PR #301, opened 2015-08-14
V2 buttons

### LINKED ISSUE #286: Button classes don't work on <a> elements
There are some really cool button classes in the library (i.e. `usa-button-big`) but the classes don't work properly if applied to an element that isn't a `<button>`. I want to apply these button appearance classes to `<a>` elements so that I can easily use the UI component to go to another page.

Here are some screenshots, where the classes applied to the element are `usa-button-outlined usa-button-big` and only the element type is changed:

0. Using a `<button>` element: http://up.jeremiak.com/cDna
0. Using a `<a>` element: http://up.jeremiak.com/cDtI

Currently, my workaround is to use a JavaScript event handler to change the document location `onClick`

msecret:
v2 button styles …
I removed the button-secondary-{hover,active} styles because I couldn't
figure out why they were ever necessary if they're attached to the base
button class anyway.

Refactored button code …
- All button styles now also apply to `.usa-button` as requested in
- Refactored selectors so there's less duplication.

Fixes #286, #295

screen shot 2015-08-14 at 3 48 41 pm


maya:
We're currently using `usa-button-*` convention. Should we use  `usa-btn-*` instead? It's much terser, saves space, and quicker to type. On the other hand, I don't think we're using abbreviations in other places.

Just checked pivotal labs and salesforce and they use `btn`. Gov.uk uses the spelled out `button`.

maya:
The button-secondary-{hover,active}  were in there in order to remove chaining, like we chatted about earlier. They are intended for use in order to display that state and provide helpers for JS.

maya:
Can you make sure these button styles also work for anchor links?

maya:
Adds feathered box shadow for focus state in 9e065fe & addfe8f.

This is what it looks like:
screen shot 2015-08-14 at 5 08 45 pm

maya:
b3be08b updates the border radius variable to the new design's 3px. 

This is what it looks like:

screen shot 2015-08-17 at 8 23 55 am

_elements/buttons.md
```diff
@@ -8,26 +8,55 @@ title: Buttons
...
+    <button class="usa-button-gray usa-button-active">Sign up</button>
+    <button class="usa-button-gray usa-button-hover">Sign up</button>
+  </div>
 
-  <h3>Outlined Button</h3>
-  <button class="usa-button-outlined" type="button">Button</button>
+  <div class="button_wrapper">
+    <button class="usa-button-outlined" type="button">Sign up</button>
+    <button class="usa-button-outlined usa-button-active">Sign up</button>
+    <button class="usa-button-outlined usa-button-hover">Sign up</button>
```
maya: I thought about this, and think that the present tense `outline` is better and more consistent with how we write up our classes.

_elements/buttons.md
```diff
@@ -8,26 +8,55 @@ title: Buttons
...
+  </div>
+
+  <div class="button_wrapper" style="background: #323a45">
+    <button class="usa-button-negative" type="button">Sign up</button>
+    <button class="usa-button-negative usa-button-active">Sign up</button>
+    <button class="usa-button-negative usa-button-hover">Sign up</button>
```
maya: This is based on the outline style, so where is the `outline` class?

maya: Can you change `negative` to `inverse`?

_elements/buttons.md
```diff
@@ -8,26 +8,55 @@ title: Buttons
...
 
-  <h3>Outlined Button</h3>
-  <button class="usa-button-outlined" type="button">Button</button>
+  <div class="button_wrapper">
+    <button class="usa-button-outlined" type="button">Sign up</button>
+    <button class="usa-button-outlined usa-button-active">Sign up</button>
+    <button class="usa-button-outlined usa-button-hover">Sign up</button>
+  </div>
+
+  <div class="button_wrapper" style="background: #323a45">
```
maya: Can you remove the inline style?

maya: The dark wrapper extends too long. Can you make sure it extends up to the third button, not the full width of the container?

Here's the design:

dark wrapper

assets/_scss/elements/_buttons.scss
```diff
@@ -29,78 +32,123 @@ input[type="image"] {
...
-    background-color: #e0e0e0;
+    &:hover,
+    &.usa-button-hover {
+      background-color: $color-primary-alt-dark;
+    }
+
+    &:active,
+    &.usa-button-active {
+      background-color: $color-primary-alt-darkest;
+      color: $color-white;
```
maya: Can you remove the focus and active state?

screen shot 2015-08-14 at 6 20 55 pm

_elements/buttons.md
```diff
@@ -8,26 +8,55 @@ title: Buttons
   <!-- Add HTML markup for example here -->
 
   <h3>Primary Buttons</h3>
-  <button>Static</button>
-  <button class="usa-button-active">Active</button>
-  <button class="usa-button-hover">Hover</button>
+  <div class="button_wrapper">
+    <button>Sign up</button>
+    <button class="usa-button-active">Sign up</button>
```
maya: Can you add back in the button text which states it’s current state (static, hover, active)? Without this, users will not know what the 3 colors are for.

msecret: In the design, they all say "Sign up", so I purposefully removed them.

_elements/buttons.md
msecret: There's no inverse styles that aren't also outline, so why not have it like this?

maya: Exactly, bc inverse button are reusing styles from outline button, we should have this clearly shown with the class names as well.

_elements/buttons.md
maya: So as a user on the site, how do you know what state you're looking at?

maya: @msecret I just talked to @angel and agreed to add the button states added back in. 

Here is a picture for reference to the previous state:
screen shot 2015-08-15 at 12 21 48 pm

_elements/buttons.md
maya: Also note, this is more consistent with BEM methodology, since inverse is modifier of outline.

_elements/buttons.md
msecret: This is an issue with the design, and I'm waiting to get the new layout design like we had talked about previously.

_elements/buttons.md
msecret: I'm not sure you understand me: theres never a button that's just inverse and not outlined.

_elements/buttons.md
maya: What about the design do you anticipate changing? What we have should be the the final ones to my knowledge.

msecret:
Regarding `usa-btn-*` I have no strong opinion. I do feel we should table this now and focus on getting to MVP.

_elements/buttons.md
maya: I understand you. I don't see why we would not include outline from the class bc it clearly indicates it is a version of outline.

maya:
lgtm :+1:

PR #304, opened 2015-08-17
Add margin to labels

maya:
This adds a right margin to labels so that if there are multiple labels on the same line they have space between them.

This is what it looks like:
screen shot 2015-08-17 at 9 21 04 am

Without it:
screen shot 2015-08-17 at 9 23 16 am


msecret:
I'd prefer a less specific rule such as:

```
.label + .label {
  margin-left: .5rem;
}
```

but this is probably fine.

PR #305, opened 2015-08-17
v2 tables

msecret:
I removed the bordered table class because I really couldn't see a use
for it. This mean's the default `<table>` will be the bordered one which
you can override with unbordered.

Updated color variables.

### DIFF
--- a/assets/_scss/elements/_table.scss
+++ b/assets/_scss/elements/_table.scss
@@ -1,39 +1,36 @@
 table {
+  border-spacing: 0;
   margin: 2em 0;
   min-width: 100%;
-  
+
   thead {
+    tr {
+      background-color: $color-gray-lightest;
+    }
+
     th {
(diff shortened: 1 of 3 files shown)


assets/_scss/elements/_table.scss
```diff
@@ -1,43 +1,40 @@
...
-  border-spacing: 0;
-
-  thead {
-    tr {
-      background-color: #eeeeee;
-    }
-  }
-}
-
 .usa-table-borderless {
```
maya: Do we like the class `-borderless` or prefer something else (for example clean, alt, minimal, subtle)?

juliaelman: +1 on borderless

juliaelman:
This looks good to me!

@maya any other changes?

maya:
lgtm :+1:

PR #306, opened 2015-08-17
Add homepage

maya:
This adds the homepage to the site.

Note: I've moved the current homepage content that has all of our elements listed on a single page to a new `/all/` page.


juliaelman:
Talked with @maya and here is the work that needs to be completed before this PR can be reviewed:
- Homepage artwork from @angel 
- Bug on the buttons #313 
- Icons and arrows need to be added to the buttons

maya:
Talked to @angel and artwork will not be part of this sprint.

_includes/head-home.html
```diff
@@ -0,0 +1,67 @@
...
+
+<!-- CSS
+================================================== -->
+  <link rel="stylesheet" href="{{ "/assets/css/google-fonts.css" | prepend: site.baseurl }}">
+  <link rel="stylesheet" href="{{ "/assets/css/homepage.css" | prepend: site.baseurl }}">
+  <link rel="stylesheet" href="{{ "/assets/css/pygments-theme.css" | prepend: site.baseurl }}">
+
+<!-- Scripts
+================================================== -->
+  <script src="{{ site.baseurl }}/assets/js/vendor/jquery-1.11.3.min.js"></script>
```
msecret: This shouldn't be in the head, it should be before the closing body tag.

_includes/head-home.html
```diff
@@ -0,0 +1,67 @@
...
+  <meta name="description" content="{{ page.description | xml_escape }}">
+  <meta property="og:description" content="{{ page.description | xml_escape }}" />
+{% else %}
+  <meta name="description" content="{{ site.description | xml_escape }}">
+  <meta property="og:description" content="{{ site.description | xml_escape }}" />
+{% endif %}
+
+  <link rel="canonical" href="{{ site.url }}{{ page.url }}" />
+
+<!--[if lt IE 9]>
```
msecret: This shouldn't be in the head, it should be before the closing body tag.

assets/_scss/core/_grid.scss
```diff
@@ -99,9 +99,17 @@
...
 
+  // Specificies end of a row.
+  // Required if grid-box contains multiple rows.
+  // Required if browser does not support :last-child
   .usa-end-row {
     @include omega();
   }
+
+  // Fixes grid alignment when grid-items have varying heights.
+  .usa-end-row + .grid-item {
```
msecret: I thought we weren't really going to use end-row anymore because it doesn't work for mobile?

pages/index.html
```diff
@@ -1,39 +1,167 @@
...
+      <nav>
+        <a class="usa-banner-link-top" href="https://github.com/18F/usfwds">View on GitHub</a>
+      </nav>
+      <div class="usa-banner-content" id="main-content">
+        <h1>U.S. Federal Web Design Standards</h1>
+        <h2 class="usa-lead">Open source UI components, visual style guide and content style guide for US federal government websites.</h2>
+      </div>
+        <a class="usa-button usa-button-big usa-button-secondary" href="{{ site.baseurl }}/getting-started">View the standards</a>
+        <div class="usa-button-block">
+          <a class="usa-button usa-button-big usa-button-outline-inverse">Download the components</a>
```
msecret: What would this be? Like the scss or something?

pages/index.html
```diff
@@ -1,39 +1,167 @@
...
 
-<p>This web design standards is intended to provide reference for designers, developers, as well as anyone else interested in the design and frontend principles that make up the building blocks of federal government websites.</p>
+<section class="usa-section usa-section-alt">
+  <div class="usa-grid">
+    <div class="usa-intro">
+      <h2>Tools for creating beautiful online experiences for the American people.</h2>
+      <p class="intro-text">Built and maintained by a team of US Digital Service + 18F designers and developers, this resource is built on the highest standards of 508 compliance, reuses best practices of existing style libraries and modern web design to guide us in creating beautiful and easy-to-use online experiences for the American people.</p>
+    </div>
+    <div class="usa-width-one-half">
+      <a href="#">
```
msecret: This link will currently take the user to the top of the page. Is that desired?

assets/_scss/core/_grid.scss
maya: End-row has always worked. Last-child is giving us issues in old IE. Unless I'm missing something?

_includes/head-home.html
maya: Agree, fixed in another PR: https://github.com/18F/usfwds/pull/452

_includes/head-home.html
maya: Agree, fixed in another PR: #452

assets/_scss/core/_grid.scss
maya: Actually I think what it is is we're not using the grid-item class. Is that right?

assets/css/homepage.scss
```diff
@@ -0,0 +1,233 @@
...
+    bottom: .5rem;
+    top: .5rem;
+  }
+}
+
+.usa-disclaimer-stage {
+  float: right;
+}
+
+.usa-intro {
```
msecret: Don't we already have classes to do stuff like this? Like the grid system or the usa-content?

pages/index.html
maya: @msecret Yeah the assets when it's all been packaged by us.

assets/css/homepage.scss
```diff
@@ -0,0 +1,233 @@
...
+}
+
+.usa-button-block {
+  display: inline-block;
+  vertical-align: top;
+}
+
+.usa-cta {
+  clear: both;
+  display: block;
```
msecret: Whats "cta"?

This seems like it's applied only to divs. Why does it need a display block?

It seems a little strange to have a class just for clearing things.

pages/index.html
```diff
@@ -1,39 +1,167 @@
...
+        <div class="usa-img-secondary">
+          <img src="{{ site.baseurl }}/assets/img/home/homepage_illustrations_ui_components.png" alt="UI Components">
+        </div>
+        <h3>UI Components</h3>
+      </a>
+      <p>Common web interactions (buttons, forms, navigation, etc.) with reusable and downloadable code</p>
+    </div>
+    <div class="usa-width-one-half usa-end-row">
```
msecret: Should we not use `usa-end-row` anymore because it might break different breakpoints?

Although technically, I think a one-half will be fine, it might not be good practice to have this in our code.

pages/index.html
maya: @msecret Good point. For visual style guide, it should link to that page. But for UI components, which page should it link to? Getting started? cc @carodew @colinpmacarthur @mollieru

assets/_scss/core/_grid.scss
msecret: Yea that's right.

assets/css/homepage.scss
```diff
@@ -0,0 +1,233 @@
...
+    position: absolute;
+    left: 0px;
+    top: 0px;
+    background: white;
+    outline: 0;
+    @include transition(top .1s ease-in, background .5s linear);
+  }
+}
+
+.usa-disclaimer {
```
msecret: I remember seeing this already in styleguide.scss. Is this duplicated?

assets/css/homepage.scss
maya: Yes, this was a remnant form before we had disclaimer in there. Removed in d7c787e.

assets/css/homepage.scss
```diff
@@ -0,0 +1,233 @@
...
+  float: right;
+  font-weight: $font-bold;
+
+  &:visited {
+    color: $color-white;
+  }
+}
+
+.usa-banner-content {
+  max-width: 48rem;
```
msecret: Why do we have a random max-width here? Why aren't we conforming to our grid system or using a variable for max-width?

assets/css/homepage.scss
maya: cta stands for "call to action" - it's the div around the call to action buttons that tells people to view the standards. Is there another approach you'd recommend to make sure it clears?

assets/_scss/core/_grid.scss
maya: Removed in da62ef6.

assets/css/homepage.scss
maya: This is the max-width needed to accomplish the design of the main banner text content. We are not using the grids bc we do not want it to flex and change the text per line as you update the width of the site. It needs to be the same width until it drops to a very small size.

pages/index.html
maya: @msecret We don't have any rules to remove the margin-right on the one-half grid item. If you remove `usa-end-row`, it breaks the grid to another line.

pages/index.html
maya: @msecret fixed in 144e745.

assets/css/homepage.scss
maya: I removed `usa-disclaimer` in d7c787e from your other comment. 

Skipnav is not currently in the library site. It's in styleguide.scss, which we're not using on the homepage. And since we don't offer a header / navigation where we may choose to add it, it needs to be here again. Let me know if there's anything else duplicate that needs to be removed.

assets/css/homepage.scss
maya: @msecret the only other thing that works is.. I can remove the container div and apply this to the button:

```clear: both;
display: table;```

Would you prefer this instead?

File: assets/css/homepage.scss (hunk 10 above)
Comment (msecret) [inline assets/css/homepage.scss]: I still find it very problematic that we can't even build our own site with our own pattern library. I personally think the design should change so this can happen.

File: assets/css/homepage.scss [hunk 11]
```diff
@@ -0,0 +1,220 @@
...
+      margin-top: 0;
+    }
+
+    h3, p {
+      overflow: hidden;
+    }
+  }
+
+  .usa-width-one-half-top {
+    @media (min-width: $medium-screen) {
```
msecret: Should this be using @include media?

assets/css/homepage.scss
maya: There are reasons to use both... Using grids is good if you want all the items to be proportionate with each other. Using max width is good if you want it to have a more consistent experience and reading experience. For the banner and design, max-width feels more appropriate.

If I really want to use grids and have a good reading experience, I would use: http://fittextjs.com/, mostly for the heading, but that seems unnecessary and bloated for this case. And a good experience can be achieved with max-width, so that's why I think we should use it.

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+  .usa-cta {
+    margin-bottom: 1.6rem;
+  }
+
+  .usa-img-example {
+    margin-bottom: 5rem;
+  }
+}
+
+a > h3 {
```
msecret: I don't think you're supposed to wrap block level elements in inline elements. So it should always be `<h3><a>` rather the `<a><h3>`.

assets/css/homepage.scss
maya: @msecret yes, done in fac1b17.

assets/css/homepage.scss
maya: @msecret that is part of a link block on line 47 of index.html

screen shot 2015-09-10 at 10 28 11 am

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+  color: $color-white;
+
+  h2,
+  h3,
+  .usa-lead {
+    color: $color-white;
+  }
+
+  .usa-width-one-half {
+    h3 {
```
msecret: This rule seems kinda strange. Why not just remove margin-top on all h tags? What clear does this seem to be overriding?

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+      clear: none;
+      margin-top: 0;
+    }
+
+    h3, p {
+      overflow: hidden;
```
msecret: What is this going to do?

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+h1, h2, h3, h4, h5, h6 {
+  color: $color-primary-darker;
+  font-family: $font-serif;
+}
+
+h1, h2 {
+  margin-top: 0;
+}
+
+.usa-lead {
```
msecret: We already have a lead text class, why is this different?

assets/css/homepage.scss
maya: @msecret That's just meant to target the h3's on the dark section to achieve this design:

screen shot 2015-09-10 at 10 35 33 am

The other h3's need top margin.

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+a:hover {
+  border-bottom: 0;
+}
+
+.usa-img-circle {
+  float: left;
+  margin-right: 3rem;
+}
+
+.usa-standlast {
```
msecret: What does standlast mean?

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+  p {
+    font-size: $h3-font-size;
+  }
+}
+
+.usa-link-external {
```
msecret: So is this a link that goes off the site? Why would a link like that need a larger bottom margin?

assets/css/homepage.scss
maya: It's how you get the alignment for the images and text.

w/o:
screen shot 2015-09-10 at 10 36 52 am

with:
screen shot 2015-09-10 at 10 37 06 am

assets/css/homepage.scss
msecret: What about using `usa-grid`? Shouldn't that clear content above?

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+
+.usa-section-examples {
+  @include media($medium-screen) {
+    .usa-width-five-sixths {
+      @include shift(1);
+    }
+  }
+}
+
+.usa-section {
```
msecret: Just as a note: this sounds like a thing we'd want in the main library.

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+  padding: {
+    bottom: $section-padding;
+    top: $section-padding;
+  }
+
+  h2 {
+    font-size: 3rem;
```
msecret: Is this a non-standard font size? Like, is there an existing h that matches with this size?

We have very specific typography with various sizes to use, why are we overriding something here to something that isn't standard?

assets/css/homepage.scss
maya: standlast represents any section that you want to call out with larger paragraph text.

I got the idea from standfirst on future learn's pattern library: https://www.futurelearn.com/pattern-library/standfirst

"Standfirst is a short introductory paragraph used to draw attention to specific information related to page content. Standfirst is displayed in larger font size than body text, on Gainsboro grey background. It is typically accompanied by a relevant call to action button. The elements within the standfirst component are center aligned."

But since it wasn't an introductory paragraph, I changed it to standlast. I could have called it "contribute" bc that's the section it's for but I wanted a way to make it more agnostic so it could be used for other things. I'm open to calling it something else though if you have recommendations.

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+  }
+}
+
+.usa-button-block {
```
msecret: I feel like I've seen something like this before. Maybe `.button_wrapper`? Can that be re-used rather the creating another class thats seems similar?

assets/css/homepage.scss
maya: Yes, it's for the links in the example sites that go to external websites. Without margin-bottom it is too tight:
screen shot 2015-09-10 at 10 44 13 am

assets/css/homepage.scss
```diff
@@ -0,0 +1,220 @@
...
+  h1 {
+    color: $color-white;
+
+    @media (min-width: $medium-screen) {
+      font-size: 5.6rem;
+      padding-top: 4.5rem;
+    }
+  }
+
+  .usa-text-small {
```
msecret: Why are we overriding existing `.usa-text-small`? I'd consider this to be something to question with design. We should be using our existing styles.

assets/css/homepage.scss
maya: Yes, we do have lead text, but if you look at what the properties are doing, you'll see I'm applying a different style (margin and color) for lead text inside a dark banner background.

pages/index.html
msecret: I see what's going on here, a problem with our grid or in general, how the neat grid works.

Technically if you have two one halfs next to each other, the second one should not have the margin right, meaning it won't break to the next line. This isn't happening here because you have `<div class="usa-intro">` first under `usa-grid`. This causes the first `one-half` to have the margin-right 0. 

One way to fix would be to move `usa-intro` out of this `usa-grid` div and into it's own above. Although I'm curious to hear your general thoughts on this.

We do have to be careful about `usa-end-row` because if we want to do interesting things with the grid at various breakpoints, it could break that functionality because it's forcing where the grid breaks.

assets/css/homepage.scss
maya: Using `usa-grid` works on the first two sections, but when I tried using it on the section with 4 things, this happened:

screen shot 2015-09-10 at 10 50 35 am

Seems like the table is needed.

assets/css/homepage.scss
maya: Agree, we just haven't built out any templates that would use this.

assets/css/homepage.scss
maya: @msecret yes, removed in 0f86c14. This was made before we did typography, so an artifact from before.

assets/css/homepage.scss
maya: Agree, it was before we had `usa-text-small` in the library. Removed in 6172b86. But keeping in the `text-align: center;` bc I don't think that should be component-wide, but just for the homepage.

maya:
@msecret I moved skipnav into it's own scss file here: a5ee334.

pages/index.html
maya: @msecret I don't feel like we should require people to use a new div wrapper. If either works without doing that, why not just go with it? Seems like it'll be a nuisance for simple layouts like the one I made. 

Feels like this is a situation in which we have to decide if we want to offer flexibility or strict rules. Is there any downside with the way I have it? I tested by adding a few more grids like the way I have it and it still works fine.

assets/css/homepage.scss
maya: @msecret I found your comment, it didn't disappear. Anyhow fixed in a5ee334.

assets/css/homepage.scss
msecret: So to me, this seems more like "link-example" then "external". I think when I saw external, I thought you were styling all external links on the site, which is why I was confused. Does "link-example" make sense to you?

assets/css/homepage.scss
msecret: Considering it's used elsewhere, I think the name is OK. Although my rule for naming: if more then two people are confused by the name, I usually change it.

assets/css/homepage.scss
msecret: Ok I see it's called `usa-font-lead` in the library, is that correct? Should this also be called `usa-font-lead` or is there another `usa-lead` that I'm missing?

I guess what I'm trying to get at is removing duplicated classes that seem like they could be the same.

assets/css/homepage.scss
maya: @msecret Check out the design below... We'll be adding in this external link icon, so that's why I think the name should be external.

screen shot 2015-09-10 at 1 28 21 pm

assets/css/homepage.scss
maya: Haha, point taken! Again completely random coming up with it, just my best guess for the time. If you think of something better tell me and we can change it.

assets/css/homepage.scss
msecret: I would use a different abstraction to get this interface working. Like a class you apply to these h3 and p tags to do this rather then just applying this to the h3 and ps themselves. Then you could use it for different things from h3 and p. Or maybe even a class that gets applied to the same element that usa-width-one-half is on.

assets/css/homepage.scss
maya: You're right, they should be the same. One of those pre-typography artifacts again. The one issue is we define the `usa-font-lead` size as 22px in the library, but the homepage uses 20px.

assets/css/homepage.scss
msecret: So I guess what I'm thinking is that I'd expect a dark background to only have to change the background color and the text to a white (which you do have here). Why would a dark background have to modify positioning? 

Aren't we also using like a "section" class? It would make more sense to me if that changed positioning rather then something thats more cosmetic, like a dark background thing.

maya: Yeah, but this specific section has a unique layout from the rest, it just happens to be dark. I can add another class to the section and target it that way, but what would be a good name to describe it?

assets/css/homepage.scss
maya: @msecret What name would we use to describe aligning the text flush to the image, so it doesn't wrap?

pages/index.html
msecret: So what about this case:

In a year from now everybody gets a super large monitor and we decide to expand the whole pattern library and add another breakpoint for mega size screen. At this mega breakpoint we change 1/2 so it's actually 1/4 so we can get 4 by side. This won't work with a `usa-end-row` because we're forcing it to always break at 2 by 2. Theres also current cases where using an end row with 1/4 widths will break on smaller breakpoints, and that's happening in our current code.

So what I don't want to see: 1. a user looks at our code, see's where using this method for a grid with 1/2 items and then use the same method for a grid with 1/3 items and have it break on them. 2. creating code that can't adapt to changes in the future.

assets/css/homepage.scss
maya: @msecret like we could call it "Principles" but that is so specific and won't be able to use for other things?

assets/css/homepage.scss
msecret: Whats different about this section which would break it?

maya: It's the way in which we're styling the images next to the text.

assets/css/homepage.scss
maya: @msecret fixed in d1384db. I'm going with calling it `usa-graphic-list` which should be modular and easy to use for other things.

assets/css/homepage.scss
maya: @msecret fixed in d1384db. I'm going with calling it usa-graphic-list which should be modular and easy to use for other things.

assets/css/homepage.scss
maya: Fixed in 4c9e498.

pages/index.html
maya: @msecret updated in ca4e324.

assets/css/homepage.scss
maya: @msecret changing the grids fixed this issue, so removed the CSS in: f686b3b.

assets/css/homepage.scss
maya: @msecret `.button_wrapper` and `.usa-button-block` have different styles and uses so I would not reuse.

assets/css/homepage.scss
msecret: Yea I think the icon is perfect, I think the styles for the font size and the margin bottom should be in a separate class like 'link-example'. 

I also just realized something: why don't we just wrap this link in an h tag of appropriate size? I think that would fix what I was talking about.

maya: @msecret ok addressed in af8b02c.

assets/css/homepage.scss
msecret: I would rarely suggest using js for styling like that. I still don't understand why grids wouldn't be used here as I think they're consistent with the framework and don't require extra CSS, with somewhat random values, to be written.

maya: @msecret I think that's a valid point and something to look at down the road. It looks good and works for now, and since we are also running out of time, I think we should move forward with this. Also I just checked both options with @angel and she said this looks fine for now.

PR #317, opened 2015-08-19
Update typography

maya:
This updates the typography on the website to use our new font variables and defaults for sizes, font family, line height, and blue color on the usfwds site headings.

Here is what it looks like:
screen shot 2015-08-19 at 9 12 18 am

screen shot 2015-08-19 at 9 14 12 am

### DIFF
--- a/_components/alerts.md
+++ b/_components/alerts.md
@@ -45,11 +45,11 @@ title: Alerts
 
 <div class="usa-grid">
   <div class="usa-width-one-half">
-    <h3 class="usa-usfwds-heading">Use</h3>
+    <h3 class="usa-heading">Use</h3>
     <p>This is the usage content for the example.</p>
   </div>
   <div class="usa-width-one-half">
-    <h3 class="usa-usfwds-heading">Accessibility</h3>
+    <h3 class="usa-heading">Accessibility</h3>
     <p>This is the accessibility content for the example.</p>
   </div>  
 </div>
--- a/_components/alerts.md
+++ b/_components/alerts.md
@@ -45,11 +45,11 @@ title: Alerts
 
 <div class="usa-grid">
   <div class="usa-width-one-half">
-    <h3>Use</h3>
+    <h3 class="usa-usfwds-heading">Use</h3>
     <p>This is the usage content for the example.</p>
   </div>
   <div class="usa-width-one-half">
-    <h3>Accessibility</h3>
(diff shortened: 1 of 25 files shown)


juliaelman:
@maya looks good! Why are we using both `usa` and `usfwds` in our naming conventions for class names? Seems cleaner to choose just one and would lean towards solely using `usfwds`.

maya:
@juliaelman good question! I thought I should specify that this is a style _only_ for the styleguide and not for components themselves. Do you think we're safe to just use `usa-heading`? I thought users might want that class for something else. I also thought about `usa-library-heading`.

juliaelman:
@maya changing to `usfwds` makes the most branding sense to me and would recommend sticking in that wheelhouse. 

On that note, if we decide to go this route we'd want to change all of the `usa` naming conventions to use `usfwds` instead. It doesn't make a lot of sense to separate out the way we name classes for the design standards website and the design standards themselves. Users will be most interested in the code we provide and not as much on the code that we use for the site itself.

maya:
@juliaelman in terms of `usa` vs `usfwds` I would stick with usa bc it's shorter, easier to read, easier for new people to create new classes (we already asking them to prefix it), and bc we're already using it on everything.

A lot of other pattern library's have styles for the library itself. I was wondering the best way to convey that or if we should just start with using `usa-heading`.

carodew:
I don't normally have an opinion on the code, but I'd vote for sticking with `usa` for the reasons @maya stated, and also because that abbreviation has a meaning to users. `usfwd`s doesn't really, at least not as strongly. And it's hard to type.

carodew:
Also U! S! A! U! S! A! U! S! A!

maya:
+1 @carodew  :us:  :us:  :us:

juliaelman:
@carodew @maya YES! Agreed on all fronts and vote taking out `usfwds` completely. Once that is completed, this PR it good to go.

maya:
@juliaelman fixed in 6f0061a.

juliaelman:
+1

msecret:
So it seems wrong to me that we're applying `usa-heading` to every single h element. Is the a reason we don't just make the h elements headings by default?

maya:
@msecret we're not applying it to every single heading element. We're just applying it to h2 and h3 in the library site. Users of the library will have default base colors for all of their headings.

maya:
More accurately, we're applying it to certain h2 and h3s in the site.

PR #320, opened 2015-08-20
Updating to Jekyll 2.5.3 and removing auto-prefixer gem.

juliaelman:
After some discussion with the rest of the front end team, we decided to remove auto-prefixer to our workflow for MVP. It seemed to cause other issues with our local dev environment, especially after rolling backwards to Jekyll 2.4.0. 

After merging, devs will need to run `bundle update` for this change to go into effect.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,8 @@
 source 'https://rubygems.org'
 
-gem 'jekyll', '2.4.0'
+gem 'jekyll'
 
 gem 'celluloid', '0.16.0'
 
 gem 'rouge'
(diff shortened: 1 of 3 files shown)


maya:
@juliaelman should this also roll back the celluloid fix we did in #318?

juliaelman:
@maya sadly, no. I believe this is b/c of the way the build process is setup and how Travis CI is running it. My recommendation would be to merge this in and then remove celluloid specificity.

maya:
+1

PR #325, opened 2015-08-20
V2 search

msecret:
Some restructuring to make class work without as many classes as we had
before. This is being done because developer feedback says that things
are a little too complicated.

- Removed the wrapper div around the whole input. I found this wasn't
needed and simplifies.
- I removed a bunch of classes from the example. The classes are still
there and work, I'm thinking about adding them back in to ensure we
give devs all teh tools they might need.
- Now the only classes are for input, submit and submit-text and then
the modifiers. Theres no medium modifier as thats the default.
- Jeremia mentioned I could also target aria roles for rules, like
instead of [type="submit]. Don't need to do this yet.


_components/search-bar.md
```diff
@@ -8,55 +8,40 @@ title: Search Bar
 
   <div class="usa-grid">
     <div class="usa-width-one-half">
-      <form class="usa-search usa-search-big">           
-        <fieldset>
```
maya: Why did we remove fieldsets and legends? Did you check with @nickbristow about this?

msecret: I thought we didn't need fieldsets for a single field? I thought that was even what you told me. Anyway, here is what @nickbristow said: 

"its basically when you need to group fields with a title. or like, question answer, so question is a legend for a fieldset of radio buttons"

Here's what Sasha said:
"I am pretty sure as long as the label is associated with the input then no"

maya: @msecret I think it's more than just a single field bc we have a button in the group as well. 

These were previous approved by @nickbristow with fieldsets so if we're removing it, we need to run this by him and make sure it's ok.

maya: Did nick say we don't need it for the search component?

nickbristow:
Its good!

maya:
lgtm :+1:

PR #332, opened 2015-08-21
Add fix for button text color on visited links

### LINKED ISSUE #313: Visited anchor link buttons issue
When the link was visited the text is blue, not white.

screen shot 2015-08-18 at 3 13 41 pm
screen shot 2015-08-18 at 3 13 37 pm

maya:
This adds a fix for visited links with the button class. 

This fixes #313.

This shows that the text is white and how the new class overrides the visited selector.

screen shot 2015-08-21 at 3 38 03 pm
screen shot 2015-08-21 at 3 39 02 pm

### DIFF
--- a/assets/_scss/elements/_buttons.scss
+++ b/assets/_scss/elements/_buttons.scss
@@ -174,3 +174,12 @@ button,
     box-shadow: initial !important;
   }
 }
+
+// Fix for visited links
+
+a {
+  &.usa-button,
+  &.usa-button-primary {
+    color: $color-white;
+  }
+}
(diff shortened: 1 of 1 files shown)


assets/_scss/elements/_buttons.scss
```diff
@@ -174,3 +174,12 @@ button,
     box-shadow: initial !important;
   }
 }
+
+// Fix for visited links
+
+a {
+  &.usa-button,
```
msecret: Why can't we just add a `:visited` rule to the selectors above?

maya: Like this? e93153e

PR #333, opened 2015-08-22
Add new header and footer

maya:
This adds the new header and footer into the site. 

:warning: :construction: This needs #332 to get buttons to work properly before it can get merged. :warning: :construction: 

This is what it looks like:

Header:

screen shot 2015-08-21 at 5 05 57 pm

Footer:

screen shot 2015-08-21 at 5 06 29 pm


_includes/navbar.html
```diff
@@ -1,33 +1,37 @@
...
-        <a href="{{ site.baseurl }}/about/">About</a>
-      </li>
-      <li>
-        <a href="https://github.com/18F/usfwds/" target="_blank">
-          View on Github
+  </div>
+  
+    <nav class="usa-site-navbar">
+      <div class="usa-grid">
+      <a href="#" id="menu-btn">&#9776; MENU</a>
```
msecret: Why does this button have an id?

_includes/navbar.html
```diff
@@ -1,33 +1,37 @@
...
+      <div id="logo">
+        <a href="{{ site.baseurl }}/" accesskey="1" title="Home" aria-label="Home">
+          <h1 class="usa-header-logo-title">U.S. Federal Web Design Standards</h1>
```
msecret: What's `.usa-header-logo-title` doing? I can't find it being used in the current codebase.

_includes/navbar.html
```diff
@@ -1,33 +1,37 @@
...
         </a>
+      </div>
+      <ul class="usa-button-list usa-unstyled-list">
```
msecret: I think `usa-button-list` should use the unstyled-list mixin.

_includes/footer.html
```diff
@@ -0,0 +1,9 @@
+<div class="usa-grid">
+  <div class="usa-site-footer-content usa-width-one-half">
```
msecret: So the `usa-site-footer` is just for the actual styleguide?

My one concern here is if users look at the actual markup on the styleguide to learn how to do things and they come across a class like that.

_includes/navbar.html
maya: For the slideout to function? I didn't change this, just copied this from how it was already.

_includes/navbar.html
maya: I added the mixin in https://github.com/18F/usfwds/pull/331 - so when that gets merged I'll add it here.

_includes/footer.html
maya: Correct, `usa-site-footer` is just for the styleguide, that's why I added "site". 

I think we are going to have special styles just for the styleguide so we should agree on a naming convention to identify that. The class they'll need to look at is when they're looking at the footer component and see `usa-footer`.

Do you have another suggestion you think we should use?

_includes/navbar.html
maya: I think I added it bc I wanted to style it later for mobile, but removing for now: 53a5246

_includes/footer.html
jeremiak: I guess I don't understand the need to have a seperate bit of CSS for the style guide (perhaps beyond doing some margin fixing for presentation inside another grid). Ultimately, I'm going to be looking through the source of the documentation site to determine the class structure for elements and I don't think I'm going to remember that distinction all the time.

When I look at other CSS frameworks they generally use the framework itself for the documentation.

maya: @jeremiak I agree that ideally we wouldn't need this. However, I've seen other pattern libraries that use `*-library` or `*-styleguide` to make this distinction, so I don't think it's true for every case. I would vote move forward bc it's crunch time and we can open an issue to change this when we have more time. 

Also, note: we haven't done v2's of the footer yet so this *could* help.

msecret: Can we make the simple change so this includes `*-styleguide` in the class name? I'll also merge once conflicts are resolved.

maya: Cool, ya I'll change it to `*-styleguide`. Will let you know when it's ready to go.

maya: @msecret styleguide class added in 56f4a2e.

Merge conflict resolved in 43418ae.

PR #342, opened 2015-08-25
V2 accordions change in behavior

msecret:
This makes it so when an already opened item is clicked it will close
that item, not do nothing. The result of the attr if json parsed so
that its becomes an actual boolean.

The width on the ul was changed because before the content div was
setting the width, so when none are displaying it breaks the layout.

### DIFF
--- a/assets/_scss/components/_accordions.scss
+++ b/assets/_scss/components/_accordions.scss
@@ -5,6 +5,7 @@
     margin: 0;
     overflow: hidden;
     padding: 0;
+    width: 100%;
 
     li {
       background-color: $color-gray-lightest;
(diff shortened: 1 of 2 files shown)


maya:
@msecret the behavior is working but are you working off of the latest accordions? There should be 2 versions: borderless and bordered. Another thing that's missing is the new hover styles.

maya:
Also there should be a `<div class="preview">...</div>` wrapped around all examples which appears to be missing here. The reason why is bc I'm going to need to grab this class with JS to output our markup for the code examples.

msecret:
I moved this here as there was a problem with the tracking on this branch: https://github.com/18F/usfwds/pull/351

maya:
Closing in favor of: https://github.com/18F/usfwds/pull/351

PR #351, opened 2015-08-25
V3 accordions

msecret:
This makes it so when an already opened item is clicked it will close
that item, not do nothing. The result of the attr if json parsed so
that its becomes an actual boolean.

The width on the ul was changed because before the content div was
setting the width, so when none are displaying it breaks the layout.


maya:
looks good :+1:

PR #374, opened 2015-08-26
V2 grids

msecret:
Read the commit messages for an explanation of things.

### DIFF
--- a/_components/search-bar.md
+++ b/_components/search-bar.md
@@ -19,7 +19,7 @@ title: Search Bar
   </div>
 
   <div class="usa-grid">
-    <div class="usa-width-one-third">
+    <div class="usa-width-one-fourth">
       <form class="usa-search">
         <label for="search-field">Search Medium</label>
         <input type="search" id="search-field">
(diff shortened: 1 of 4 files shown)


colinpmacarthur:
@msecret: I pulled this down my local machine and need to chat with @carodew about the search bar (know that's a high priority component). Thoughts arriving shortly.

colinpmacarthur:
@msecret: @carodew and I reviewed the new grid and its impact on the high priority components. Everything looks good to us.

The one change we'd like to make is removing the small and medium search bars all together. At the moment, they are too narrow at screen widths larger than 1200 px. I think one search bar will do fine for MVP and figuring out how to make the smaller ones work with the grid is going to be hard. Let me know if I should open a separate issue for this.

cc: @mollieru

maya:
@colinpmacarthur can u open a new issue or trello card about removing the search bars? Also did u check that we are not using any of the other variations in other components? Another alternative is to display them in 1/2 rather the the 1/4 so it doesn't get too small.

cc: @msecret

maya:
nice work @msecret, i'm still wondering how this will work when we have more complex layouts, but we can review when we have more time down the road.

PR #378, opened 2015-08-27
Add code snippets

maya:
This adds code snippets to the library site, displaying HTML markup of the components below each component.


_includes/scripts.html
```diff
@@ -3,3 +3,4 @@
   <script src="{{ site.baseurl }}/assets/js/vendor/respond.js"></script>
   <script src="{{ site.baseurl }}/assets/js/vendor/rem.min.js"></script>
   <script src="{{ site.baseurl }}/assets/js/main.js"></script>
+  <script src="{{ site.baseurl }}/assets/js/vendor/prism.js"></script>
```
msecret: Vendor libs should go before our own, so this should be moved to the line before main.js

assets/js/main.js
```diff
@@ -3,6 +3,36 @@ $(function(){
     $('.sidenav, .overlay').toggleClass('is-visible');
     e.preventDefault();
   });
+
+  var generateCodeSnippets = function(content, previewBox) {
+
+    var self = this;
+
+    this.parseCode = function(previewBox) {
```
msecret: I don't understand the point of having theses functions as instance members on the generateCodeSnippets function if we're not using that as a class. Why wouldn't this just be:

```js
function parseCode(previewBox) {

}
```

Same goes for render.

assets/js/main.js
```diff
@@ -3,6 +3,36 @@ $(function(){
...
+      var sampleCode = $('<div>');
+      $(sampleCode).html($(previewBox).html())
+      $(sampleCode).find('.is-peripheral').remove();
+      return sampleCode;
+    }
+
+    this.render = function(previewBox, sampleCode) {
+
+      var sampleCodeBox = $('<div class="usa-accordion usa-code-sample"><ul class="usa-unstyled-list"><li><button class="usa-button-unstyled" aria-expanded="false" aria-controls="collapsible-0">Toggle code</button><div id="collapsible-0" aria-hidden="true" class="usa-accordion-content"><pre><code class="language-markup"></code></pre></div></li></ul></div>');
```
msecret: This long string is really hard to process. I'd suggest breaking it up onto newlines and using tabs to organize like actual html. Like so:

```js
var sampleCodeBox = $(''+
  '<div class="usa-accordion usa-code-sample">' +
    '<ul class="usa-unstyled-list">' +
      '<li><button ...etc
  
```

It's not perfect, but it'll improve the readability a little.

maya: @msecret definitely, can u commit this to the branch?

_includes/scripts.html
maya: Done in 7e25cbb.

assets/js/main.js
maya: @msecret This was to keep everything contained within the generateCodeSnippets function. Is this a problem?

msecret: Declaring a function within generateCodeSnippets would contain it as well, so I'm still confused it's being declared on this rather the just being declared. Maybe to make unit testing easier? Maybe just add a TODO comment to rethink the structure of it.

assets/js/main.js
```diff
@@ -3,6 +3,36 @@ $(function(){
...
+    this.parseCode = function(previewBox) {
+      var sampleCode = $('<div>');
+      $(sampleCode).html($(previewBox).html())
+      $(sampleCode).find('.is-peripheral').remove();
+      return sampleCode;
+    }
+
+    this.render = function(previewBox, sampleCode) {
+
+      var sampleCodeBox = $('<div class="usa-accordion-bordered usa-code-sample"><ul class="usa-unstyled-list"><li><button class="usa-button-unstyled" aria-expanded="false" aria-controls="collapsible-0">Code</button><div id="collapsible-0" aria-hidden="true" class="usa-accordion-content"><pre><code class="language-markup"></code></pre></div></li></ul></div>');
```
msecret: I think my comment was lost, but could you clean up this huge string by breaking it up into lines and tabs?

maya:
TODO: check why we are having these functions as instance members on the generateCodeSnippets function if we're not using that as a class. 

Possible reasons: 
- keep everything contained within the generateCodeSnippets function
- make it easier for unit testing

assets/js/main.js
maya: @msecret done in 5f55e89.

PR #380, opened 2015-08-27
Remove borders from type, colors

msecret:
As defined in
https://trello.com/c/DDVfZ94q/297-1-component-page-revise-page-layout-to-match-v2-designs

Used a modifier class on preview to ensure js hooks work.

Removed trailing whitespace.

### DIFF
--- a/_layout-system/grids.md
+++ b/_layout-system/grids.md
@@ -3,7 +3,7 @@ layout: default
 title: Grids
 ---
 
-<div class="preview">
+<div class="preview preview-no_border">
 
   <h3 class="usa-heading">Grid</h3>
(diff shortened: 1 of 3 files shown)


maya:
@msecret we need to get rid of the other padding styles as well so it's flush.

screen shot 2015-08-27 at 4 23 46 pm

Also I think @mollieru asked us to remove one of the Grids headings.

maya:
I fixed the above in 213c921, 4d4733a.

maya:
@msecret the changes are not visible for grids bc `preview-no_border` is added to grids.md in collections, however there's now markup added directly to grids.html in pages, and `preview-no_border` is not there.

PR #382, opened 2015-08-27
Add typography

maya:
The adds the typography section of the library, as well as correct typography throughout the site.

Adds the typeface intro, style pairings, typesetting, links, and lists.


assets/_scss/elements/_typography.scss
```diff
@@ -175,3 +177,15 @@ dfn {
 .usa-drop_text {
   margin-bottom: 0;
 }
+
+.usa-background-dark {
+  background-color: #333B46;
+
+  p {
```
msecret: What about span, or form elements?

assets/_scss/elements/_typography.scss
```diff
@@ -175,3 +177,15 @@ dfn {
 .usa-drop_text {
   margin-bottom: 0;
 }
+
+.usa-background-dark {
+  background-color: #333B46;
```
msecret: Why is this not a variable?

maya: The color was not one of our variables, but I'll check with @mollieru

assets/_scss/elements/_typography.scss
maya: I can add span too, which form elements?

assets/_scss/elements/_typography.scss
maya: Confirmed with @angel to use `$color-gray-dark` - done in 717868c.

assets/_scss/elements/_typography.scss
maya: @msecret I added span in aea5292. If you let me know which form elements you think should be added, then I can add it as well.

maya:
Font guidance will need updating, but let's review as is and get the ball rolling so others can use the newest typography. @msecret

assets/css/styleguide.scss
```diff
@@ -858,3 +883,126 @@ code[class*="language-"], pre[class*="language-"] {
...
+  .usa-background-dark {
+    display: inline-block;
+    padding: {
+      left: 1em;
+      right: 1em;
+    }
+  }
+}
+
+.usa-link-visited {
```
msecret: Why would this rule be here? Isn't it in the pattern library?

assets/css/styleguide.scss
```diff
@@ -858,3 +883,126 @@ code[class*="language-"], pre[class*="language-"] {
...
+    font-family: $font-serif;
+    font-size: $h3-font-size;
+  }
+
+  .usa-font-lead-alt {
+    font-size: $h5-font-size;
+  }
+}
+
+.sans-style {
```
msecret: Is this a pattern library style? Why are these sizes being modified? Shouldn't we be showing the users the exact font sizes they should be getting when they use the library?

msecret:
How is the performance.png being used? The image stays at the same level of performance so how is it showing different performance?

maya:
@msecret re: the performance.png - vizdee is working on custom graphics for that and what's there now is placeholder.

assets/css/styleguide.scss
maya: @msecret the reason why font sizes are modified is bc there are different sizes when serif and sans serifs are the base fonts. 

I think the correct way to do this is provide a different typography.scss and variables.scss depending on which family they choose. This was just done for now and to get it up quickly, but let me know if you think there's a better way to do this.

assets/css/styleguide.scss
maya: @msecret are you saying we should add this class along with `a:visited` in typography.scss?

msecret: Yea, if it's being done here right? Is there a reason this should only be in the styleguide?

assets/css/styleguide.scss
maya: Or we can provide the correct values in the documentation for each family?

assets/css/styleguide.scss
maya: It brings into question how many color helpers we want to provide. For example, developers can add this background color in however they want. 

These are all the colored backgrounds we are showing as possible background colors to use:

screen shot 2015-08-31 at 9 53 28 am

The last thing we decided was to _not_ provide the color helpers.

assets/css/styleguide.scss
msecret: Yea that could get complicated for the user to set up but I understand what you're doing now.

assets/css/styleguide.scss
maya: oops, I thought this was a different question

maya: Done in: dff445b.

assets/css/styleguide.scss
```diff
@@ -858,3 +883,122 @@ code[class*="language-"], pre[class*="language-"] {
...
+
+  h5 {
+    font-size: rem(16px);
+  }
+
+  h6 {
+    font-size: rem(13px);
+  }
+
+  .usa-display {
```
msecret: Why wouldn't display and lead be in the pattern library?

maya: @msecret they are, on line 143-158 of typography.scss - these are the unique "other font pairings" as well as "robust" style with the `usa-lead-alt`.

Issue #277, opened 2015-08-10
text cursor position doesn't stay where it should within the input field (iPhone 6 running iOS 8.4)


carodew:
Steps to recreate: 

1. Have an iPhone 6 running iOS 8.4 
2. Insert the cursor into a text field. This puts focus on the field and triggers the mobile keyboard to appear. 
3. Scroll up an down on the page. The cursor will visually stay in the same spot even as the elements on the page move with the scroll. 

![iphone-screenshot1](https://cloud.githubusercontent.com/assets/10144074/9184365/ba18d338-3f82-11e5-82d4-2fcb7e6f5b69.PNG)

maya:
@carodew is this on all input fields or just that component?

carodew:
It's on all text input fields.

carodew:
Here's a video showing the behavior: https://drive.google.com/open?id=0B1Wf-KSFneMUQWk0TTB4eElveE0

vgvg:
This appears to be a bug with Safari on iOS itself. Found some references to similar behavior: https://bugs.webkit.org/show_bug.cgi?id=138201

It seems to be a symptom of having an input box inside of a container with overflow set to scroll. I was able to replicate the bug with just a very basic div with overflow: scroll, and an input in the middle of it and have not been able to find any clean css ways to fix it.

The two ways that would work to remove this bug are to remove the use of overflow scrolling on the main body of the page and either 1) have the header be sticky some other way, 2) let the header scroll along with the page on iOS. Are these options something we would want to do?

maya:
@vgvg awesome, thanks! I agree that changing the header to sticky and changing the way the page is laid out is a good possible solution. I may look into this today.

I think this is also related to #275, so changing this may knock out both bugs.

vgvg:
@maya yeah I just looked at that bug, changing overflow on a container and using the entire page's scrolling would help fix #275

colinpmacarthur:
Replicated this bug on iPad 2 running iOS 8.4

vgvg:
PR #399 should have fixed this one.

carodew:
Received a request to access the video. Since I can't share the original video directly, here's a gif. Note that this issue has been resolved, I'm just providing it for the curious. 

![runaway cursor iphone 1](https://cloud.githubusercontent.com/assets/10144074/10730897/2baaa17a-7bc9-11e5-9f0f-5c2b4e49aeb5.gif)

psr1919plus21:
[...]

Issue #275, opened 2015-08-10
Weird scrolling behavior on iPhone, iPad


maya:
When viewing the site on my iPhone 5, the scroll doesn't work properly. It feels kind of stiff, not sure if there is some scroll hijacking going on due to JS.

@carodew also mentioned experiencing some strange behavior as well.

juliaelman:
@carodew @maya would one of you mind testing out some video recording of this interaction? Using either with an desktop emulator or a video camera to display what you mean by "weirdness" :)

maya:
This is a video to indicate the issue which is also experienced on desktop: 

http://cl.ly/cFNv

maya:
Is this related to momentum based scrolling? Overflow?

carodew:
Adding a preview of a video showing the scrolling behavior. I think it's a momentum-based scrolling issue (not sure). On other websites, a swipe up or down with a finger will result with the scroll 'easing out' for a second or so after the swipe finishes. On our site the scrolling stops immediately when the swipe ends, making it feel throttled. 

iPhone 6 running iOS 8.4, Safari (not sure which version). 

https://drive.google.com/a/gsa.gov/file/d/0B1Wf-KSFneMUSWU0TXhoMGJwMlU/view?usp=sharing

maya:
cc @vgvg

maya:
This may have something to do with we are scrolling within a div as opposed to a page.

.main-content has an overflow: auto

<img width="261" alt="screen_shot_2015-08-25_at_11 35 53_am" src="https://cloud.githubusercontent.com/assets/5249443/9475830/11a844ac-4b1e-11e5-88a6-eecac6bce175.png">

So when you scroll the content, you are scrolling within that div. We'd need to remove all that, make the nav position: fixed, and change how the page is laid out.

colinpmacarthur:
I replicated this behavior on an iPad 2 running iOS 8.4.

vgvg:
PR #399 should have fixed this.

sujithuix:
Add `-webkit-overflow-scrolling: touch;` to the scroll div. The scroll will become smooth then.

PR #383, opened 2015-08-28
[WIP] Using fixed header instead of overflow on main container

vgvg:
@maya this is one minimally invasive approach to trying to fix the iOS scrolling issues and change the page to use a fixed header with scrolling on the full document vs scrolling on the main container. Still just a WIP but thought I'd get some eyes on it for feedback. Please feel free to tag others if you think this is worth pursuing. Seems to fix both #275 and #277.


assets/css/styleguide.scss
```diff
@@ -43,9 +43,31 @@ a#skipnav {
   }
 }
 
+/* Fixes anchor position when header is fixed */
+:target:before {
+  content: "";
+  display: block;
+  height: 75px;
+  margin-top: -75px;
```
vgvg: This weird css is here to make sure anchor tags line up with the bottom of the header, and not with the top of the browser since that means the h2 gets covered when header is fixed. Another approach is to use negative margin and an offsetting positive padding on the element itself.

vgvg:
We could also just choose to not have a sticky header, which would simplify and remove the need for this code and some of the workarounds in here.

assets/css/styleguide.scss
```diff
@@ -43,9 +43,31 @@ a#skipnav {
...
+  @include media($medium-screen + $width-nav-sidebar) {
+    height: $site-top;
+    margin-top: -$site-top;
+  }
+}
+
 .usa-site-header {
   background-color: $color-white;
   border-bottom: 1px solid $color-gray-light;
+  position: fixed;
```
msecret: Mobile sometimes has issues with fixed positioning. Have you tested on mobile?

assets/js/main.js
```diff
@@ -8,8 +8,23 @@ $(function(){
...
+  /* :target already adds spacing to line up the page correctly
+   * prevent double space if current page target already = new target
+   */
+  if (!anchor.is(":target")) {
+    headerOffset = ($('.usa-site-header').first().outerHeight());
+  }
+  
+  /* Firefox needs html, others need body */
+  $('body,html').animate({
+    scrollTop: anchor.offset().top - headerOffset
```
msecret: Are we animating scrolling?

maya: yes, before `scrollTop` did this for us.

assets/css/styleguide.scss
vgvg: I tested on iOS 8.4 (iPad & iPhone) and it seemed to work well on there. Do not have access to an Android device to test on there.

PR #388, opened 2015-08-28
Fix color page margins

maya:
This updates the width classes for color examples. Before, I used some of the grid classes that were messing with this with the new breakpoints, so made new ones for these examples.

This resolves: https://trello.com/c/Epi94GIq/285-2-color-build-color-page-of-visual-style-section

### DIFF
--- a/_visual/colors.md
+++ b/_visual/colors.md
@@ -68,31 +68,31 @@ title: Colors
   <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus consectetur placerat ultrices.</p>
 
   <div class="usa-grid-full usa-color-row">
-    <div class="usa-width-one-fourth">
+    <div class="color-big">
       <div class="usa-color-short usa-color-primary-alt">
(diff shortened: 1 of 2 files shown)


msecret:
It worries be that we wouldn't be using our own grid classes for simple layout like this. The more we add to styleguide.css, the more we're proving the library is hard to use for making websites. What about these rules mean we can't use the grid classes here? Should the rules be added to the pattern lib?

maya:
The reason why we're not using our own grid was bc this page used a custom design and had unique proportions which unfortunately do not correspond to our 12 column grid system. 

If vizdee wants to change the design to work with our grid system, then we can definitely use our own grid.

maya:
I'd also add that this is the colors page, where the primary purpose is for designers to see the colors being used on the site. 

We are not providing any code or saying that this should be the way to structure a page. It's a pretty bespoke / one-of-a-kind layout which I do not foresee ever being used or needed on a govt site.

PR #389, opened 2015-08-28
Update grids file structure

maya:
I was pretty confused why we changed our file structure to add grid content directly in the html page rather than generate it through jekyll collections like we do on everything else. This changes it back to its former structure.


msecret:
We removed Layout from the MVP so grid is not the top level and only layout system.

maya:
So we will have some markup in pages and some in collections? I worry that this makes the organization inconsistent and will be hard for people to find where things are. 

Also the way it is now we have some markup of grids in collections and pages. We should not have duplication and have grids markup in only one place in the repo.

msecret:
So I updated this based on this issue:

https://trello.com/c/dH0qYKGo/281-1-grids

Do your changes here still ensure:
• Title of page should be 'Grids' (no longer 'Layout System')
• Remove all sub-titles ('Grid' and 'Grid Example')
• Remove the border around the examples
• Remove all of the 'Layout' section of the page.

maya:
@msecret yep, that's absolutely a-ok - my question is more about how it's implemented organizationally, and how we're starting to break things up and it's losing its consistency.

msecret:
So I'm still confused, in this current PR, does it ensure all things in https://trello.com/c/dH0qYKGo/281-1-grids are still working? Because if it does, I'll merge it because the structure doesn't matter to me at this point.

If your PR is changing something so https://trello.com/c/dH0qYKGo/281-1-grids is no longer considered fixed, then I think we should table this as it was a very temporary solution until we get Layout back up.

maya:
@msecret yes it ensures that trello card is still complete. It only changed the underlying file structure.

screen shot 2015-08-31 at 12 01 12 pm

PR #398, opened 2015-08-31
Dropdowns v2

maya:
This adds straight corners to dropdowns.

This is what it looks like:

screen shot 2015-08-31 at 12 15 33 pm


msecret:
Did you check if these dropdowns are working correctly in IE browsers and mobile?

maya:
@msecret not yet, this was just to implement the border radius update. But we can leave this open to fully test it, too if you like.

PR #399, opened 2015-08-31
Scroll entire doc, instead of just main content

vgvg:
Tested in Chorme, FF, Safari. iOS 7.1 & 8.4 on both iPad and iPhone. Seems to work pretty well, much smoother scrolling. Need some help testing on an android device.

(this is basically PR #383, again, but I did something weird with that one)


vgvg:
Also seems to work pretty well on IE11

vgvg:
@colinpmacarthur I'm going to go ahead and merge this if you agree and didn't notice any bugs while you were testing it.

PR #402, opened 2015-08-31
New documentation scheme

msecret:
This is to show what the new documentation scheme looks like. I came across various problems in this pull request and want to get feedback before I proceed with the rest of the documentation snippets.

#### Problem 1: when you use content elements, like `<h>` and `<p>` in an accordion, the top and bottom margins are broken
screen shot 2015-08-31 at 4 32 32 pm
If you placed content with `<h>`s and `<p>`s in an accordion, the top and bottom padding was overwhelming, because it was the margins of the `<h>` and `<p>` tags PLUS the padding in the accordion. To fix this, I removed the padding from the accordion and ensured the content we're putting in there is wrapped in a semantic tag, like `<p>`.

#### Problem 2: max-width on `<p>` tags breaks content in the accordion
screen shot 2015-08-31 at 4 25 08 pm

This brought on another problem: `<p>` tags are confined to a max width, so now the accordion content doesn't go to the full width of the accordion. You should be able to use `<p>` tags in an accordion. This is also an issue for the lead-text class which uses a p tag.
screen shot 2015-08-31 at 4 25 15 pm.

I think we either need to remove the max-width rule or not wrap all our content in `usa-content`. Another option is to modify the max-width rule so we pick exactly what elements receive it, in this case, the accordion. I think this is the worst option because it would be a maintenance nightmare.

Other changes here:
- I brought the `button` rules outside of nested `ul` and `li` in accordion so that I could use an accordion that didn't have a list. This eases development and code for dropdowns such as this, that only have one accordion element. I also added some rules to `button` to ensure it always works.
- I added the ul rule because `<ul>`s after `<h>`s and `<p>`s should not have so much top margin.


_components/accordions.md
```diff
@@ -112,8 +132,12 @@ title: Accordions
...
 
-<div class="usa-grid">
-  <div class="usa-width-one-half">
+<div class="usa-accordion-bordered">
+  <button class="usa-button-unstyled usa-accordion-button"
+      aria-expanded="true" aria-controls="collapsible-0">
+    Documentation
+  </button>
+  <div id="collapsible-0" aria-hidden="false" class="usa-accordion-content">
     <h3>Use</h3>
```
maya: @msecret we need to add in `usa-heading` class to all <h3> headings in the documentation in order to get the blue heading.

maya:
@mollieru @angel see @msecret's screenshot of the max width of text inside the accordion. Is this the desired result?

maya:
@msecret regarding the padding issue. can we keep the padding on the accordion as it was and then remove margin-top on the first heading?

msecret:
@maya I fixed problem 1 by removing the padding on the accordion. I think this is a better fix then hacking out the margin-top on the first heading.

maya:
@msecret is this going to be for all accordions or just the documentation?

msecret:
@maya for all accordions. What I'm recommending is that you don't have blank text in accordion content. You wrap it in a semantic tag, which I think is better practice anyway.

maya:
@msecret right, so using a `<p>` tag will make it look decent?

I think @mollieru would like to ensure that the design they outlined is met:

screen shot 2015-08-31 at 5 01 30 pm

maya:
@msecret I talked with @angel and we want to know if we can remove the max-width on `<p>` text for text inside the accordion?

msecret:
So removing the `<p>` text inside the accordion and on the usa-font-lead text and all the other things that come up will be a maintenance disaster.

Also: if we're already needing to override the max-width on our own elements, it seems very likely to me our users will find instances where they will need to remove it too.

maya:
@msecret so it sounds like we need to go through and only use `.usa-content` when needed?

msecret:
@maya yes I think that's the best option.

angel:
@maya @msecret can we up the max-width to be whatever 2/3 of the grid would be, I think that's about 65rem. As for the padding it would be awesome to keep the margin 30px all around to match the design, but if that causes too many problems then we can go with what you think is best. /cc @mollieru

msecret:
I've thought about the padding, and realized that it would have been better if we were more careful about setting margin-tops on our `<h>` and `<p>` elements by using adjacent selectors. That would be pretty hard to do before MVP, so I might want to find a way to make this work.

But I think you're correct, the parent container should define the padding. I just wish we could do that without having to hack at things.

I don't care what we set the max-width to, I just don't think we can set it at such a broad scope as we are doing now. I think we should change that soon.

maya:
@msecret can you explain what you mean "by using adjacent selectors"?

maya:
Is this the correct PR to change max-width in?

angel:
@maya oh sorry, for some reason I thought that was in this thread. Will find a better place to put it.

msecret:
What should be the next step here? @maya do you want to implement `.usa-content` on a more specific scale as you probably know better where it's supposed to be used? In the meantime, I can figure out a solution for the padding and continue implementing these documentation accordions.

maya:
@msecret I think the `.usa-content` issue should be done in a new PR, not this one. So I'd advise we leave it as is, and the other PR should resolve that issue.

maya:
@msecret looking good! it looks like the documentation accordion is just added to components. I'm not seeing it elements, grid, typography, etc...

maya:
lgtm :+1:

PR #410, opened 2015-09-03
Add new sidenav

maya:
This adds the new sidenav to the site. 

The does not show the active item/parent when clicked.

Lead text is in the jekyll front matter for each component. It's in the visual style page for that one, bc it's the only one that shows on a page.

This is what it looks like:

screen shot 2015-09-02 at 7 45 22 pm

Expanded:

screen shot 2015-09-02 at 5 44 08 pm


colinpmacarthur:
Great work @maya!

@carodew pointed out that the 18F content style guide discourages using ampersands. Any bad effects from changing the few & in the navigation to "and"?

maya:
@colinpmacarthur nothing wrong with that! It'll just drop a few items to the second line rather than keep them all tight on a single line.

colinpmacarthur:
Sounds good!

maya:
@colinpmacarthur done in 3fc9b21.

angel:
@maya can we take 2px off the bottom. I know it's suppose to account for the descenders so then that is centered but I think it looks better when the Capital letter is centered within the box.

maya:
@angel I talked to @mollieru and think it's fine to center based on the whole text bc it makes it more robust, and we can revisit this later.

pages/getting-started.html
```diff
@@ -0,0 +1,45 @@
+---
+permalink: /getting-started/
+layout: styleguide
+title: Getting started
+lead: Intro text on what is included in this section and how to use it. No more than one or two sentences.
+---
+
+<h3 class="usa-heading">For developers</h3>
+
+<p><b>HTML</b> is built with progressive enhancement. All users get critical information and experiences. New browsers get the prettiest experiences, older ones get less pretty, but usable ones. If JavaScript fails, users will still get a robust HTML foundation.</p>
```
msecret: Shouldn't this use `<strong>` rather the `<b>`.

"According to the HTML 5 specification, the <b> tag should be used as a LAST resort when no other tag is more appropriate." - w3c

maya: @msecret they do two different things. They both make text bold, strong gives more emphasis. I think it's even ready louder in screen readers.

_components/form-templates.md
```diff
@@ -344,7 +392,52 @@ title: Forms Blocks
...
+            aria-controls="ssn">
+          Show SSN</a>
+      </p>
+
+    </fieldset>
+
+  </form>
+</div>
+
+<div class="usa-grid">
```
msecret: These are using the old documentation scheme, and I had already updated these.

Am I correctly looking at the code where you fixed the merge conflicts?

maya: @msecret re-added in f1489d0. Let me know if that works.

pages/getting-started.html
maya: @msecret done in 2de6b93.

PR #439, opened 2015-09-04
Replaced correct hex code in color label

jenniferthibault:
The hex code in the label before was for `primary-alt-darkest`.

This fix corrects it for the darkest blue color.
Hope I'm doing this right, it seemed to make it easier on you all then filing an issue.


maya:
Wonderful, thank you @jenniferthibault! you did it right :+1:

PR #440, opened 2015-09-04
Updates the titling for the sidenav page

colinpmacarthur:
Spells out the words "side navigation" and updates the titles to be h3s.

Could you double check @carodew?

### DIFF
--- a/_components/sidenav.md
+++ b/_components/sidenav.md
@@ -7,6 +7,8 @@ lead: Intro text on what is included in this section and how to use it. No more 
 
 <div class="preview">
   
+  <h3 class="usa-heading">1. Single level</h3>
+  
   <div class="usa-grid">
     <aside class="usa-width-one-fourth">
       <ul class="usa-sidenav-list">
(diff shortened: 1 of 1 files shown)


_components/sidenav.md
```diff
@@ -7,8 +7,6 @@ lead: Intro text on what is included in this section and how to use it. No more
 
 <div class="preview">
   
-  <h2 class="usa-heading">Sidenav</h2>
-
```
maya: @colinpmacarthur why did we remove this heading?

colinpmacarthur: @maya: That heading just seemed duplicative, because the page’s “Side navigation” title is just above it. But am I missing something?

mollieru: Chatted w @colinpmacarthur and we decided on the following language to make the headings a little tighter:

1. Single level 
2. Two level
3. Three level

maya: @mollieru wouldn't it be plural (i.e. Two levels, Three levels)?

_components/sidenav.md
```diff
@@ -7,8 +7,8 @@ lead: Intro text on what is included in this section and how to use it. No more
...
+  <h3 class="usa-heading">1. Single level</h3>
```
maya: @mollieru we're not numbering anywhere else on the site where we have multiple versions of a component. Why are we starting this convention? Shouldn't it be consistent with the rest?

maya:
lgtm :+1:

PR #441, opened 2015-09-04
Add design tweaks

maya:
@angel and I spent the afternoon refining visual design tweaks on the site. Just a lot of small things.

As well as: https://trello.com/c/DDVfZ94q/297-1-component-page-revise-page-layout-to-match-v2-designs

### DIFF
--- a/_components/search-bar.md
+++ b/_components/search-bar.md
@@ -49,7 +49,7 @@ lead: Intro text on what is included in this section and how to use it. No more 
   </div>
 </div>
 
-<div class="usa-accordion-bordered">
+<div class="usa-accordion-bordered usa-accordion-docs">
   <button class="usa-button-unstyled usa-accordion-button"
       aria-expanded="true" aria-controls="collapsible-0">
     Documentation
--- a/_components/search-bar.md
+++ b/_components/search-bar.md
@@ -9,11 +9,13 @@ lead: Intro text on what is included in this section and how to use it. No more 
   <div class="usa-grid">
     <div class="usa-width-one-half">
       <form class="usa-search usa-search-big">
-        <label for="search-field-big">Search Big</label>
-        <input type="search" id="search-field-big">
-        <button type="submit">
-          <span class="usa-search-submit-text">Search</span>
-        </button>
+        <div role="search">
+          <label for="search-field-big">Search Big</label>
+          <input type="search" id="search-field-big">
+          <button type="submit">
+            <span class="usa-search-submit-text">Search</span>
+          </button>
+        </div>
       </form>
     </div>
   </div>
@@ -21,11 +23,13 @@ lead: Intro text on what is included in this section and how to use it. No more 
   <div class="usa-grid">
     <div class="usa-width-one-half">
       <form class="usa-search">
-        <label for="search-field">Search Medium</label>
-        <input type="search" id="search-field">
-        <button type="submit">
-          <span class="usa-search-submit-text">Search</span>
-        </button>
+        <div role="search">
+          <label for="search-field">Search Medium</label>
+          <input type="search" id="search-field">
+          <button type="submit">
+            <span class="usa-search-submit-text">Search</span>
+          </button>
+        </div>
       </form>
     </div>
   </div>
@@ -33,11 +37,13 @@ lead: Intro text on what is included in this section and how to use it. No more 
   <div class="usa-grid">
     <div class="usa-width-one-half">
       <form class="usa-search usa-search-small">
-        <label for="search-field-small">Search Small</label>
-        <input type="search" id="search-field-small">
-        <button type="submit">
-          <span class="usa-sr-only">Search</span>
-        </button>
+        <div role="search">
+          <label for="search-field-small">Search Small</label>
+          <input type="search" id="search-field-small">
+          <button type="submit">
+            <span class="usa-sr-only">Search</span>
+          </button>
+        </div>
       </form>
     </div>
   </div>
--- a/_elements/form-controls.md
+++ b/_elements/form-controls.md
@@ -269,8 +269,9 @@ lead: Intro text on what is included in this section and how to use it. No more 
 
   <fieldset>
     <legend>Date of birth</legend>
+      <span class="usa-form-hint usa-datefield-hint" id="dobHint">For example: 04 28 1986</span>
+
     <div class="usa-date-of-birth">
-      <p class="usa-form-hint usa-datefield-hint" id="dobHint">For example: 04 28 1986</p>
       <div class="usa-datefield usa-form-group usa-form-group-month">
         <label for="date_of_birth_1">Month</label>
         <input aria-describedby="dobHint" class="usa-form-control" id="date_of_birth_1" name="date_of_birth_1" pattern="0?[1-9]|1[012]" type="text" value="" maxlength="2">
--- a/_elements/form-controls.md
+++ b/_elements/form-controls.md
@@ -269,7 +269,7 @@ lead: Intro text on what is included in this section and how to use it. No more 
 
   <fieldset>
     <legend>Date of birth</legend>
-      <span class="usa-form-hint usa-datefield-hint" id="dobHint">For example: 04 28 1986</span>
+    <span class="usa-form-hint usa-datefield-hint" id="dobHint">For example: 04 28 1986</span>
 
     <div class="usa-date-of-birth">
       <div class="usa-datefield usa-form-group usa-form-group-month">
--- a/_elements/form-controls.md
+++ b/_elements/form-controls.md
@@ -37,8 +37,6 @@ lead: Intro text on what is included in this section and how to use it. No more 
   <label for="input-success">Text Input Success</label>
   <input class="usa-input-success" id="input-success" name="input-success" type="text">
 
-  <h2 class="usa-heading">Text Area</h2>
-
   <label for="input-type-textarea">Text Area Label</label>
   <textarea id="input-type-textarea" name="input-type-textarea"></textarea>
 
--- a/_elements/form-controls.md
+++ b/_elements/form-controls.md
@@ -50,7 +50,7 @@ lead: Intro text on what is included in this section and how to use it. No more 
     Documentation
   </button>
   <div id="collapsible-0" aria-hidden="false" class="usa-accordion-content">
-    <h4 class="usa-heading">Accessibility</h3>
+    <h4 class="usa-heading">Accessibility</h4>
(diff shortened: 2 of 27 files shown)


_components/search-bar.md
```diff
@@ -9,11 +9,13 @@ lead: Intro text on what is included in this section and how to use it. No more
   <div class="usa-grid">
     <div class="usa-width-one-half">
       <form class="usa-search usa-search-big">
-        <label for="search-field-big">Search Big</label>
-        <input type="search" id="search-field-big">
-        <button type="submit">
-          <span class="usa-search-submit-text">Search</span>
-        </button>
+        <div role="search">
```
msecret: Why was adding this div done here? Accessibility? I remember Bristow saying this was fine as is.

_elements/form-controls.md
```diff
@@ -29,28 +29,28 @@ lead: Intro text on what is included in this section and how to use it. No more
   <input class="usa-input-focus" id="input-focus" name="input-focus" type="text">
 
   <div class="usa-input-error">
-    <label for="input-error">Text Input Error</label>
-    <span class="usa-input-error-message" role="alert">Helpful error message</span>
+    <legend>
```
msecret: Why is this legend needed?

_components/search-bar.md
maya: @msecret this was added bc we were missing the search role on the search form. We cannot add the role to the form tag itself bc form has its own native semantics. This means we can either add it to the wrapper around or inside of form. The reason I didn't add it to the wrapper around it is bc we have grid classes and thought it would be hard to distinguish bc they may not want to, and shouldn't have to use the grid. See: http://adrianroselli.com/2015/08/where-to-put-your-search-role.html

_includes/head.html
```diff
@@ -65,4 +65,5 @@
 
 <!-- Scripts
 ================================================== -->
-  <script src="{{ site.baseurl }}/assets/js/vendor/jquery-1.11.3.min.js"></script>
\ No newline at end of file
+  <script src="{{ site.baseurl }}/assets/js/vendor/jquery-1.11.3.min.js"></script>
```
msecret: I just realized, why is this being included here in the head rather then with the rest of the scripts in the scripts file? The scripts should also be included at the bottom, not the head.

_elements/form-controls.md
maya: So that the error message would be part of the legend. I followed gov.uk's example: http://govuk-elements.herokuapp.com/errors

screen shot 2015-09-08 at 3 45 16 pm

_layouts/styleguide.html
```diff
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
-<html lang="en">
+<!--[if lt IE 9]><html class="lt-ie9"><![endif]-->
```
msecret: Why was this added? I can't find the class being used in this PR.

_includes/head.html
maya: Not sure, I didn't do it. Happy to move it though. Also I'm updating a few script things in a different PR so I can include the move there too. Qq: would this be the first in our scripts?

_layouts/styleguide.html
maya: It's not in this PR but it's in our styleguide.css - we have the class used for the sidebar bc we need a different approach for IE.

assets/_scss/components/_alerts.scss
```diff
@@ -34,9 +34,12 @@
 }
 
 .usa-alert-heading {
+  @include media($medium-screen) {
+    margin-top: .3rem;
+  }
   margin: {
     bottom: .15em;
```
msecret: Why is this measurement in em while everything else is in rem? Shouldn't all of them be rem?

assets/_scss/components/_disclaimer.scss
```diff
@@ -1,16 +1,33 @@
...
   padding: {
     bottom: .5rem;
     top: .5rem;
   }
 }
 
+// This hides the stage note of the disclaimer in anything smaller than 690px
 .usa-disclaimer-stage {
-  @include media($medium-screen) {
+  @include media(690px) {
```
msecret: I don't think we should be using random breakpoints. This is hard to maintain, and right now, impossible to change via variables.

Sometimes I've found things that are not "perfect" at a certain screen size. Sometimes this is better then having code that's hard to maintain.

assets/_scss/components/_footer.scss
```diff
@@ -43,6 +40,13 @@
       right: 2.5rem;
     }
 
+    li {
+      @media (min-width: $medium-screen) {
```
msecret: Why aren't this `@include media()` rules?

assets/_scss/elements/_inputs.scss
```diff
@@ -78,6 +81,17 @@ select {
   background-position: right 1.3rem center;
 }
 
+legend {
```
msecret: When is a legend not under a fieldset? Or is it that a legend is always supposed to be before a fieldset? Either way, it seems weird to have different rules for when a legend is in a fieldset vs when it's not.

assets/css/styleguide.scss
```diff
@@ -499,6 +502,10 @@ h3 + .button_wrapper {
   color: $color-primary-darker;
 }
 
+h3.usa-heading {
```
msecret: Is this needed anymore?

assets/_scss/components/_disclaimer.scss
maya: @msecret this is the exact location where it breaks. Before it used the medium breakpoint and it broke to 2 lines. If we use the large it'll disappear too early. What breakpoint do you think we should use?

assets/css/styleguide.scss
```diff
@@ -308,8 +304,15 @@ a#skipnav {
   }
 
   p, a {
```
msecret: I feel like this is a particularly bad rule to put in styleguide.scss. I get how we're out of time so can you put a comment here saying to investigate further? Like this;

```
TODO investigate why this is required in styleguide.scss.
```

maya: @msecret this was one of those "styleguide-only" styles and was done to accomplish this trello card: https://trello.com/c/cYuS0RJm/447-clean-up-typography-on-mobile-footer

It was added to style our super bare footer, which is only text. Are you saying add a note to switch to use classes instead?

assets/_scss/components/_alerts.scss
maya: @msecret fixed in 0b5a1df.

assets/css/styleguide.scss
maya: @msecret removed in  1f018ed.

assets/_scss/elements/_inputs.scss
maya: That was for the error input. Do you have another idea we could use?

Without it, it looks like this:

screen shot 2015-09-08 at 4 06 27 pm

_components/search-bar.md
msecret: I'd prefer adding it to the wrapper div rather then complicate the code with another div which we were successfully able to remove. But I'll defer to your best judgment.

_includes/head.html
msecret: If you're dealing with scripts in a different PR, this can wait.

_layouts/styleguide.html
msecret: Is this just being used for the styleguide? Are we somehow instructing our users that they need this class to get the library working?

assets/_scss/components/_disclaimer.scss
msecret: You should use whatever existing breakpoint that works the best. We might have to concede that it won't be perfect right now.

assets/css/styleguide.scss
msecret: No I'm saying to add a note to see why it wouldn't be in the main pattern library css.

_components/search-bar.md
maya: @msecret I'm ok with using that, but from the view of someone looking to use this component, they'd see this: `<div class="usa-width-one-half" role="search">` and I'm worried they'd think they'd _always_ need to use a wrapper div. We could put this information in the documentation, but for someone that wants to "grab and go" I'm worried this will be lost.

_layouts/styleguide.html
maya: @msecret As of right now, yes, it is just for the styleguide. But when we get into bug testing, we may need this again as well.

Why would we need to tell users they need it to get the library working? I don't think they'd need this information. None of the components require it.

assets/_scss/components/_disclaimer.scss
maya: Ok switched to large breakpoint in: 63c557a. I think it's better if it disappears early than the bad experience of it breaking to two lines. People can always adjust this breakpoint to something that works better for themselves.

assets/css/styleguide.scss
maya: @msecret done in 5d180bd.

_components/search-bar.md
maya: @msecret my view is that putting it inside the form tag is better practice bc it won't get messed up as easily. The wrapper div around it is more for layout even though you could technically add it there. I think the cost of having an extra div on the inside search form tag is less than the cost of someone missing it / messing it up bc they remove the wrapping grid div. I vote to leave it as is.

assets/_scss/components/_footer.scss
maya: @msecret fixed in 5596f6d.

_includes/head.html
maya: @msecret added to this PR: https://github.com/18F/usfwds/commit/849a3dde9cf1e66bb8b34613bfb5bc1b57be087a

assets/_scss/elements/_inputs.scss
maya: @msecret I'm pretty sure legends can be used without a fieldset. It's just that if you use a fieldset, you must use a legend.

maya: @msecret I wrote a different rule, to resolve this issue here: 5d94448. Basically adding this alternate font size directly to the error message class.

_layouts/styleguide.html
msecret: I was saying if the components required it, then we would need to tell them.

maya: @msecret at this point, it does not.

PR #444, opened 2015-09-08
Add current page to side navigation

maya:
This adds the current page to the sidebar navigation so users can know what page they're in.

This is what it looks like:

screen shot 2015-09-08 at 9 58 57 am


colinpmacarthur:
Yay!

PR #468, opened 2015-09-09
Re-attempt deployment

msecret:
Theres probably more code in here then the original deploy, but I'm out of ideas.


maya:
Seems to be working, so closing this.

PR #471, opened 2015-09-09
Content update to sidenav

mollieru:
Updates SideNav component examples to read:
1. Single level 
2. Two level
3. Three level


maya:
closing via 440

PR #480, opened 2015-09-10
Fix error input

maya:
This fixes issues with alignment being not the same size at mobile. It also fixes some text display issues. It also matches the design more closely.

This resolves: https://github.com/18F/usfwds/issues/464

### DIFF
--- a/_elements/form-controls.md
+++ b/_elements/form-controls.md
@@ -30,7 +30,7 @@ lead: Intro text on what is included in this section and how to use it. No more 
 
   <div class="usa-input-error">
     <legend>
-      <span for="input-error">Text Input Error</span>
+      <span class="usa-input-label">Text Input Error</span>
       <span class="usa-input-error-message" role="alert">Helpful error message</span>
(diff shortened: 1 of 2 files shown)


_elements/form-controls.md
```diff
@@ -30,7 +30,7 @@ lead: Intro text on what is included in this section and how to use it. No more
 
   <div class="usa-input-error">
     <legend>
-      <span for="input-error">Text Input Error</span>
+      <span class="usa-input-label">Text Input Error</span>
```
msecret: Is the `for` needed at all for accessibility reasons?

maya: @msecret I thought it was only if it was on a label, like the others. Do you think it is?

maya:
I just reviewed with @nickbristow and got the :+1: 

cc @msecret

_elements/form-controls.md
maya: @msecret fixed in latest commits.

PR #481, opened 2015-09-10
Adds intro text for individual form controls

carodew:
Does not include range slider intro text. 

Current page template only allows for one large intro text for the entire page, but we have intro content for each component. I just added the p tags in with the appropriate class.

### DIFF
--- a/_elements/form-controls.md
+++ b/_elements/form-controls.md
@@ -331,7 +331,6 @@ c
 </div>
 
 <h2 class="usa-heading" id="range-slider">Range slider</h2>
-<!--<p class="usa-font-lead">placeholder for range slider intro text </p>-->
 
 <div class="preview">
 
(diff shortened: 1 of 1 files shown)


_elements/form-controls.md
```diff
@@ -286,7 +290,7 @@ lead: Intro text on what is included in this section and how to use it. No more
       </div>
     </div>
   </fieldset>
-
+c
```
msecret: Is this supposed to be here?

_elements/form-controls.md
```diff
@@ -327,6 +331,7 @@ lead: Intro text on what is included in this section and how to use it. No more
 </div>
 
 <h2 class="usa-heading" id="range-slider">Range slider</h2>
+<!--<p class="usa-font-lead">placeholder for range slider intro text </p>-->
```
msecret: You don't have to make this a comment, you can just remove it.

_elements/form-controls.md
carodew: fixed in e4fb0b8

_elements/form-controls.md
carodew: removed in 48886ee

PR #487, opened 2015-09-10
Adding accordion documentation

colinpmacarthur:


_components/accordions.md
```diff
@@ -139,29 +139,21 @@ lead: Intro text on what is included in this section and how to use it. No more
     Documentation
   </button>
   <div id="collapsible-0" aria-hidden="false" class="usa-accordion-content">
-    <h4 class="usa-heading">Use</h4>
+    <h4 class="usa-heading">Usability</h3>
```
msecret: Should this be an h4 or an h3? Right now you have `<h4>` as the opening tag and `</h3>` as the closing tag.

_components/accordions.md
```diff
@@ -139,29 +139,21 @@ lead: Intro text on what is included in this section and how to use it. No more
...
+    <h5>When to use</h4>
```
msecret: Same as above but this time h5 and h4.

So this should either be `<h5>When to use</h5>` or `<h4>When to use</h4>` depending on what's needed for the design.

_components/accordions.md
colinpmacarthur: Good catch. Fixed in 47985e2.

_components/accordions.md
colinpmacarthur: Good catch. Fixed in 47985e2.

PR #496, opened 2015-09-11
Fix nav scrolling and anchor positioning

vgvg:
Tested in Chrome, FF, Safari, IE8 & iOS.

This resolves: https://github.com/18F/web-design-standards/issues/467

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -80,7 +80,7 @@ $(function () {
   }
 });
 
-$('.sidenav').delegate('a', 'click', function(e) {
+$('.sidenav').on('click', 'a', function(e) {
   var hashLocation  = $(this).attr('href').split('#')[1]; // long url splitting
   var anchor        = $('#' + hashLocation);
   var scrollTopPos  = calculateAnchorPosition(hashLocation);
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -46,26 +46,66 @@ $(function(){
 
 });
 
-$('.secondary-sidenav-link').click(function(e) {
-  e.preventDefault();
-  var hashLocation = $(this).attr('href').split('#')[1]; // long url splitting
-  var anchor       = $('#' + hashLocation);
-  var headerOffset = 0;
+/* Calculates what scrollTop should be in order to
+ * show an anchor properly under the header
+ * and lined up with the nav like the H1
+ */
+var calculateAnchorPosition = function (hash) {
+  var anchor        = $('#' + hash);
+  var topOffset     = 0;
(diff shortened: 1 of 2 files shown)


vgvg:
@maya here's a cross-browser fix to make sure the nav works well

assets/js/main.js
```diff
@@ -46,26 +46,66 @@ $(function(){
...
+  var scrollTopPos  = calculateAnchorPosition(hash);
+
+  if (scrollTopPos > 0) {
+    setTimeout(function () {
+      $(window).scrollTop(scrollTopPos);
+    }, 1);
   }
+});
+
+$('.sidenav').delegate('a', 'click', function(e) {
```
msecret: I think you can use `.on` now rather then `.delegate`. I just read this on jquery's site:

As of jQuery 1.7, .delegate() has been superseded by the .on() method.
http://api.jquery.com/delegate/

assets/js/main.js
```diff
@@ -46,26 +46,66 @@ $(function(){
...
+  //anchor should now align with first item inside nav
+  return anchor.offset().top - topOffset;
+}
+
+$(function () {
+  var hash          = window.location.hash.substr(1);
+  var scrollTopPos  = calculateAnchorPosition(hash);
+
+  if (scrollTopPos > 0) {
+    setTimeout(function () {
```
msecret: I'm not sure I understand why this is needed. Could you explain? Maybe withe a comment as well.

assets/js/main.js
```diff
@@ -46,26 +46,66 @@ $(function(){
...
+      $(window).scrollTop(scrollTopPos);
+    }, 1);
   }
+});
+
+$('.sidenav').delegate('a', 'click', function(e) {
+  var hashLocation  = $(this).attr('href').split('#')[1]; // long url splitting
+  var anchor        = $('#' + hashLocation);
```
msecret: If you're doing this twice, would it make sense to just pass anchor into `calculateAnchorPosition`?

assets/js/main.js
vgvg: The normal behavior of anchors is to position themselves at the top of the browser window. Because the header is position: fixed, this means that the title of the section is by default covered by the header. This fixes that by setting scrollTop on page load when a user lands on a page with a hash on the url.

assets/js/main.js
vgvg: Yeah, I can clean this up and just do it once inside of that function

vgvg:
@msecret addressed comments in latest push

Issue #482, opened 2015-09-10
"Skip to content" link does not bring focus to content


colinpmacarthur:
**Steps to replicate:**
1. Open http://pages-staging.18f.gov/usfwds
2. Hit tab until the "Skip to main content" link appears.
3. Push return or the space bar.
4. Hit tab again. The focus will not travel to content in the main section of the page, but to the download buttons in the header. If you hit tab repeatedly, you'll be hitting the navigation.

**Ideal behavior:** When you hit tab, the focus would be taken to the next focus-able element in the content of the page.

**Video:**
http://cl.ly/image/1K1x321J0E1r

**Found on:**
- IE 10, Windows 7 on Sauce Labs
- Chrome 44, Mac OS X Yosemite

jpyuda:
This seems to work _inconsistently_ for me in Chrome. Sometimes it scrolls, sometimes it does not.

carodew:
@maya, @msecret, @vgvg – if you know what's going on right away feel free to tackle this as you are able. If it's slightly unclear what is going on and what the fix is, let @jpyuda @colinpmacarthur and I investigate further for you.

maya:
@colinpmacarthur I fixed the link in #498. Correction, you wouldn't hit spacebar, you'd hit enter. Part of the content gets hidden but that's bc we have a fixed header that covers it up. 

I don't think skipnav has anything to do with tabbing, it just skips you to the main content area of the page, visually skipping through navigation, but not even sure if it makes sense bc the content starts immediately, and we have a fixed header.

jpyuda:
It needs to move focus to the beginning of `#main`, not just visually move it. The primary use case is screen reader users (so they don't have to tab through a couple dozen links).

In general, we need to make sure we're moving focus to the correct part of the page when doing things in javascript. Not correctly managing focus is a common accessibility problem.

maya:
@jpyuda are you saying we should implement this with JavaScript? Can you send over an example that does it in the way you describe?

jpyuda:
Not at all! If we just implement it with the standard HTML behavior, we should get the focus shift for "free". More making a note in case there are places where we're overriding that behavior with js.

maya:
@jpyuda I thought this was being implemented in the standard HTML way. Is there something that looks off to you with how it's done?

jpyuda:
I was assuming we were doing something with JS based on the fact that the focus is not correctly moving. Are we sure there's an `id="main"` on every page?

[...]

PR #498, opened 2015-09-11
Fix skipnav

maya:
This fixes the skipnav link. The main content area didn't have an id before and this adds one.

Related to: #482.


assets/_scss/components/_skipnav.scss
```diff
@@ -1,18 +1,19 @@
 .skipnav {
+  @include transition(top 1s ease-out, background 1s linear);
```
msecret: Does a skipnav link really need a transition?

maya: @msecret I don't have a preference. It was just there before. Do you think we should remove it?

PR #505, opened 2015-09-11
Fix validation checkmarks not working

msecret:
Was from a regression here: ff5e04f963dba94ae0d59ba2743a5747e30dc1b9
where we changed the class name.

Not a browser specific bug, didn't work on Chrome either.


maya:
Looks good

PR #506, opened 2015-09-11
Vertical line space in accordions

msecret:
The line space in accordions was broken because every nested last-child
was having it's margins removed, which it requires. This ensures only
the first child last child does.

Fixes https://github.com/18F/web-design-standards/issues/445

### DIFF
--- a/assets/_scss/components/_accordions.scss
+++ b/assets/_scss/components/_accordions.scss
@@ -82,7 +82,7 @@
     margin-top: 0;
   }
 
-  *:last-child {
+  > *:last-child {
     margin-bottom: 0;
   }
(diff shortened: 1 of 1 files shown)


maya:
@msecret will we run into any issues with nested first-child elements for this same reason?

msecret:
@maya I didn't do it because I wanted to keep the amount of code changed to as minimal as possible to lessen the chance that I could create more regressions. Although it would make sense to have that rule.

maya:
@msecret can you add it in just to be consistent? I can't think of a regression we could get from doing this.

maya:
lgtm

PR #511, opened 2015-09-11
Update new name

maya:
This updates the new name in the read.me and config file. 

When this merges you'll have to visit: http://127.0.0.1:4000/web-design-standards/ to view the site locally

TODO: check with @mbland re: config + new 18F Pages to see if url is still needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# U.S. Federal Web Design Standards
+# U.S. Web Design Standards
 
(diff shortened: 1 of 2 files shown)


mbland:
FYI, you can get rid of the `baseurl:` from the `_config.yml` and view the site locally at localhost:4000, and it will still show up as https://pages.18f.gov/web-design-standards/. (This capability was added in 18F/pages#24, by way of 18F/pages#22.)

maya:
@mbland done in f3c6bc4.

PR #512, opened 2015-09-11
Selectivizr update and other IE8 fixes

### LINKED ISSUE #462: Figure out why selectivisr.js is not working
This will allow us to use CSS3 properties, like `last-child`.

See: http://selectivizr.com/

This effects the grids issue: https://github.com/18F/usfwds/issues/428 as well as anything using the grid in **IE8**, like the footers #473.

vgvg:
Fixes #462.

Using the latest from master, Selectivizr 1.0.3b fixes some issues that allow it to work with respond.js also on the page, and seems to fix other IE bugs.

Note that html5shiv has to be in the head for it to work. Also removing the doubly-included JS files (which were actually causing a lot of huge problems IE8, freezing the browser and causing weird paint issues).


vgvg:
@msecret @maya This should get IE8 to work ok. There's still a weird FOUC, but at least things look ok after.

maya:
Great stuff @vgvg! 

Here is an image to show grids working in IE8:

screen shot 2015-09-11 at 4 52 31 pm

_includes/head.html
```diff
@@ -31,9 +31,6 @@
 
 <!--[if lt IE 9]>
```
msecret: Is there a reason for this to be in head at all?

maya: Victor mentioned to me that it would only work if this was in the head.

vgvg: Yeah, it doesn't seem to work if it's not in the head. The entire page layout gets screwy.

PR #516, opened 2015-09-14
Adding Google Analytics event listeners

colinpmacarthur:
This PR adds Google Analytics event listeners to the download buttons, "View on Github" buttons or links, and the "View the standards" buttons on the homepage.

Event listeners are our way of reliably telling when the download buttons and the "View on Github" buttons are clicked. GA doesn't  automatically register *which* site people go to when they leave an Analytics tracked page. And it doesn't automatically track downloaded files either.

I'd also like to track the "View the standards" buttons to see whether the buttons farther down the page are ever clicked. Good knowledge for future iterations.

### DIFF
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -19,10 +19,11 @@
       </div>
       <ul class="usa-button-list usa-unstyled-list">
         <li>
-          <a class="usa-button" href="#">Download</a>
+          <a class="usa-button" href="#" onclick="ga('send', 'event', 'Downloaded framework', 'Clicked download button inside site');">Download</a>
         </li>
         <li>
-          <a class="usa-button usa-button-outline" href="https://github.com/18F/usfwds/" target="_blank">
+          <a class="usa-button usa-button-outline" href="https://github.com/18F/usfwds/" target="_blank" onclick="ga('send', 'event', 'Viewed on Github', 'Clicked View on Github from inside site');"
+>
             View on Github
           </a>
           </li>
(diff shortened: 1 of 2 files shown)


msecret:
Are we sure we don't want to send a pageview rather then a custom event for these? In my GA experience, I remember custom events are hard to dig into with different parameters while a pageview acts like any other pageview.

colinpmacarthur:
Good idea, Marco! I will change accordingly.

colinpmacarthur:
@msecret:

Thinking a little more about this, I think we may want to stick with event listeners.

1) For "view on Github" links, it's a bit strange to force a page load event for external links. At best, it looks weird. At worst, the Analytics view we're assigned actually excludes those URLs, so we don't register that data.

2) For the "view the standards" links, we want to differentiate between clicks several links on the sampe page that lead the same place (the inside of the site), but are different places on the homepage. I don't think we could differentiate them with a forced page view.

3) We could definitely use page views for the "download" links, but then we get into defining some goals in terms of page views and some in terms of the pageviews and some in terms of events. That can get messy too.

Thoughts?

msecret:
If you think we can get enough functionality with events then that's OK by me.

_includes/navbar.html
```diff
@@ -19,10 +19,11 @@
       </div>
       <ul class="usa-button-list usa-unstyled-list">
         <li>
-          <a class="usa-button" href="#">Download</a>
+          <a class="usa-button" href="#" onclick="ga('send', 'event', 'Downloaded framework', 'Clicked download button inside site');">Download</a>
```
msecret: The only comment I'll make about this is that these would be better in `data-` attributes which get parsed more globally with jquery. But This is fine if we want to keep things moving.

PR #560, opened 2015-09-15
Move library site assets into separate folder

maya:
This begins work prepping the assets files for download. 

This removes all assets that are not part of the components out into its own folder. Now everything in `/assets` are used in components and `/assets-styleguide` are assets used in the site itself.

`assets/main.scss` will be the stylesheet which imports all components, neat, and bourbon -- and what we'll recommend our users to use.


maya:
@konklone I removed the link to the dap CDN and replaced it with our local file. Can you review and let me know how this looks?

konklone:
@maya DAP's preferred integration is through the DAP CDN, so that security updates and other bugfixes can be deployed instantly. Is that an issue for the project?

maya:
@konklone that's how it was originally, but @msecret said we should switch to local bc CDN was not preferred. Which is why I reverted the deletion of `dap.min.js` and moved it here.

maya:
@konklone updated to use use the dap.digitalgov.gov hosted DAP code fc1dd07.

PR #606, opened 2015-09-17
Extract base.scss from all.scss to fix icon paths

### LINKED ISSUE #600: Icons missing throughout the standards
**Steps to reproduce**
1. Visit the pages for any of the following components:
- alerts
- accordions
- dropdowns
- search icon
2. There are no icons where there were before.

*Alerts:*
screen shot 2015-09-16 at 5 12 54 pm 

*Accordions:*
screen shot 2015-09-16 at 5 13 13 pm

*Dropdown:*
screen shot 2015-09-16 at 5 13 57 pm

*Search bar:*
screen shot 2015-09-16 at 5 14 28 pm

**Found on**
Chrome 44, OS 10.10

mbland:
Closes #600.

The problem was that assets-styleguide/css/homepage.scss and
assets-styleguide/css/styleguide.scss were compiling assets/_scss/all.scss
inline, which meant that the relative paths that existed before #560 would be
relative to assets-styleguide/css, not assets/css. Furthermore, the absolute
paths introduced in #560 worked locally since `site.baseurl` was empty, but
were broken on https://pages-staging.18f.gov/web-design-standards because they
were not prefixed with `site.baseurl`, i.e. `/web-design-standards`.

The solution was to extract a new assets/_scss/base.scss file containing all
of the variables and mixins needed by all of the .scss files, and only include
base.css directly in the assets-styleguide .scss files. Then, the
assets-styleguide .scss files were updated to import the compiled
assets/_scss/all.scss file as a normal CSS import, i.e.
`@import url('{{ site.baseurl }}/assets/css/main.css');`. Since the
assets-styleguide .scss files are not part of the package intended for reuse,
and already had empty front matter that enabled Jekyll to perform Liquid
template rendering, this seemed a reasonable approach.

I also set `baseurl: /web-design-standards` in `_config.yml` to test the
changes locally, and verified that it also worked after removing `baseurl:`.

The tradeoff here is that the assets-styleguide CSS files are no longer
monolithic, and incur an extra request to import
`{{ site.baseurl}}/assets/css/main.css`. If there is a desire to maintain the
monolithic property, it is possible to use #{}-interpolation to inject
`{{ site.baseurl }}` as a prefix for the absolute paths by assigning it to a
SASS variable, per:

http://sass-lang.com/documentation/file.SASS_REFERENCE.html#import

This wouldn't be that difficult or that ugly, but it seemed the current
approach was worth a try first.

cc: @maya @colinpmacarthur @angel

### DIFF
--- a/assets-styleguide/css/homepage.scss
+++ b/assets-styleguide/css/homepage.scss
@@ -3,9 +3,8 @@
 
 // Imports -------------- //
 
-@import 'lib/bourbon/bourbon';
-@import 'lib/neat/neat';
-@import 'all';
+@import 'base';
+@import url('{{ site.baseurl }}/assets/css/main.css');
 
 // Variables ------------ //
 
--- a/assets-styleguide/css/styleguide.scss
+++ b/assets-styleguide/css/styleguide.scss
@@ -3,9 +3,8 @@
 
 // Imports -------------- //
 
-@import 'lib/bourbon/bourbon';
-@import 'lib/neat/neat';
-@import 'all';
+@import 'base';
+@import url('{{ site.baseurl }}/assets/css/main.css');
 
 // Variables -------------- //
 
--- a/assets/_scss/all.scss
+++ b/assets/_scss/all.scss
@@ -1,19 +1,4 @@
-// Vendor -------------- //
-
-@import 'lib/bourbon/bourbon';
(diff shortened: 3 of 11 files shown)


maya:
Closing in favor of #607.

PR #607, opened 2015-09-17
Move styleguide.scss back to assets to fix icon paths

### LINKED ISSUE #600: Icons missing throughout the standards

maya:
An alternate approach to #606. Just moving styleguide.scss back into `assets/css` so we don't have to break up `all.scss`. All other styleguide-only assets remain in `assets-styleguide`.

I imagine we'd want users to use it in this way too. They'd use `main.scss` in `assets/css`. So I can just remember to not include styleguide.scss in the zip.

`homepage.scss` doesn't seem to be effected staying in `assets-styleguide` so leaving it as is.

Resolves #600.


maya:
Another advantage is we could just include styleguide.scss and potentially include styleguide html and then our users would get their very own styleguide for free.

Issue #576, opened 2015-09-16
[Homepage] outlined button looks taller than the view standards button.


angel:
![image](https://cloud.githubusercontent.com/assets/103496/9895994/244153a0-5beb-11e5-8ede-18a265015a2e.png)

mollieru:
@angel is this on multiple browsers or just 1?

angel:
@mollieru I only checked on Chrome, but looks like Safari:
![image](https://cloud.githubusercontent.com/assets/103496/9912618/6b3d0534-5c5d-11e5-94be-7b6e8a58bc2a.png)
and Firefox as well:
![image](https://cloud.githubusercontent.com/assets/103496/9912657/941d3618-5c5d-11e5-8c3a-a7731c1b2577.png)

maya:
Fixed in #608.

PR #608, opened 2015-09-17
Improve outline buttons

maya:
This updates outline buttons to use an inset box shadow for it's outline, rather than using a border and having to reduce the padding. This ensures the buttons remain the same size no matter the size of the outline someone chooses. 

It's less good for old IE but since we are reducing support, I felt this was a good approach bc of it's increased flexibility and ease of use.

Also this PR fixes the homepage mobile button bug in #576 and https://github.com/18F/web-design-standards/issues/463#issuecomment-140472125 (but not that issue, they are separate problems).


assets/_scss/elements/_buttons.scss
```diff
@@ -103,40 +103,36 @@ button,
...
-    border: 2px solid;
-    border-color: $color-primary;
+    box-shadow: inset 0 0 0 2px $color-primary;
     color: $color-primary;
-    padding: .8rem 1.8rem;
 
     &:hover,
     &.usa-button-hover {
-      border-color: $color-primary-darker;
+      box-shadow: inset 0 0 0 2px $color-primary-darker;
```
msecret: Is there a way to store `inset 0 0 0 2px` in a variable to use in all the places rather the repeating it?

maya: @msecret good point, do you think we should store in variables.scss or the top of the buttons file?

msecret: Top of buttons file

maya: @msecret done in c88892c.

PR #612, opened 2015-09-17
update home page content

carodew:


pages/index.html
```diff
@@ -139,7 +139,7 @@ <h3 class="usa-graphic-list-heading">Reuse reuse reuse.</h3>
 <section class="usa-section usa-section-alt usa-section-examples">
   <div class="usa-grid">
     <div class="usa-intro">
-      <h2>For example...</h2>
+      <h2>For example …</h2>
```
angel: @carodew @mollieru could we say "Examples" or something else that avoids the ellipses all together? I think it would look cleaner that way. /cc @maya

pages/index.html
```diff
@@ -74,7 +74,7 @@ <h2 class="usa-font-lead">Open source UI components and a visual style guide for
   <div class="usa-grid">
     <div class="usa-intro">
       <h2>How to use these tools</h2>
-      <h3 class="usa-font-lead">Use a little, use a lot, use it all.<br>Here’s how the web design standards may help you...</h3>
+      <h3 class="usa-font-lead">Use a little, use a lot, use it all.<br> Here’s how the web design standards may help you …</h3>
```
maya: Please remove the space before the ellipses.

carodew: I believe there is supposed to be a space before and after an ellipses. @angel—do you want to weigh in?

pages/index.html
carodew: Ok with me if ok with @mollieru

pages/index.html
mollieru: Trying to think of something a little catchier/more exciting, but falling short...

I'm ok w calling this 'Examples' instead.

angel: @carodew @mollieru @maya According to http://www.thepunctuationguide.com/ellipses.html "Each period should have a single space on either side, except when adjacent to a quotation mark, in which case there should be no space."

pages/index.html
carodew: changed in  920d155

pages/index.html
```diff
@@ -40,7 +40,7 @@ <h2 class="usa-font-lead">Open source UI components and a visual style guide for
 <section class="usa-section usa-section-alt">
   <div class="usa-grid">
     <div class="usa-intro">
-      <h2>Tools for creating beautiful online experiences for the American people.</h2>
```
maya: Why does this have a period? It's not a complete sentence.

pages/index.html
```diff
@@ -40,7 +40,7 @@ <h2 class="usa-font-lead">Open source UI components and a visual style guide for
...
+      <h2>Tools for creating beautiful online experiences for the American people. </h2>
```
maya: Why does this have a period? It's not a complete sentence.

pages/index.html
```diff
@@ -25,12 +25,12 @@
       </nav>
       <div class="usa-banner-content" id="main-content">
         <h1>U.S. Web Design Standards</h1>
-        <h2 class="usa-font-lead">Open source UI components and a visual style guide for U.S. federal government websites.</h2>
+        <h2 class="usa-font-lead">Open source UI components and visual style guide to create consistency and beautiful user experiences across U.S. federal government websites.</h2>
```
maya: Why is there a period if this is not a complete sentence?

pages/index.html
maya: Sounds like we're ok with the space before.

pages/index.html
carodew: Per Writing Lab clarification—phrases that are not complete sentences in this context do not get periods. I will update this PR!

carodew: Fixed in f4394c7

PR #613, opened 2015-09-17
Attempt to fix focus state on accordion hover.

msecret:
An important rule has to be used because theres another one on
"unstyled" button. We should address this at some point, possibly
now.

The other larger problem is the `<ul>` on the accordion has a
overflow: hidden on it. This means the the box shadow won't show
up on accordions using the `<ul>`. Is there a strong reason why we
need overflow:hidden on the accordion `<ul>`?

Fixes part of https://github.com/18F/web-design-standards/issues/534

### DIFF
--- a/assets/_scss/components/_accordions.scss
+++ b/assets/_scss/components/_accordions.scss
@@ -51,7 +51,7 @@
     width: 100%;
 
     &:focus {
-      box-shadow: $focus-shadow !important;
+      box-shadow: $focus-shadow;
     }
 
     &:hover {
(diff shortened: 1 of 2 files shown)


maya:
@msecret this just produces a shadow below the header, not all around:

screen shot 2015-09-17 at 11 43 37 am

msecret:
@maya read the description in the PR.

maya:
@msecret I don't know why overflow: hidden was used.

maya:
Looks good.

PR #614, opened 2015-09-17
Add typography tweaks

### LINKED ISSUE #570: [Password Reset] Text input labels and button should be sentence case

### LINKED ISSUE #572: [Search bar] Change labels to be h6
Right now the labels look like it's part of the component, but our search bars to do not have labels that are part of the component. We should change those labels to h6s to differentiate that.

angel:
Changed the example headings for various components to be h6s instead of h3s, which fixes #572, and updated some form labels to be sentence case, which fixes #570.

### DIFF
--- a/_components/search-bar.md
+++ b/_components/search-bar.md
@@ -13,7 +13,7 @@ lead: Intro text on what is included in this section and how to use it. No more 
     <div class="usa-width-one-half">
       <form class="usa-search usa-search-big">
         <div role="search">
-          <label for="search-field-big">Search big</label>
+          <label class="usa-sr-only" for="search-field-big">Search big</label>
           <input type="search" id="search-field-big">
           <button type="submit">
             <span class="usa-search-submit-text">Search</span>
@@ -29,7 +29,7 @@ lead: Intro text on what is included in this section and how to use it. No more 
     <div class="usa-width-one-half">
       <form class="usa-search">
         <div role="search">
-          <label for="search-field">Search medium</label>
+          <label class="usa-sr-only" for="search-field">Search medium</label>
           <input type="search" id="search-field">
           <button type="submit">
             <span class="usa-search-submit-text">Search</span>
@@ -45,7 +45,7 @@ lead: Intro text on what is included in this section and how to use it. No more 
     <div class="usa-width-one-half">
       <form class="usa-search usa-search-small">
         <div role="search">
-          <label for="search-field-small">Search small</label>
+          <label class="usa-sr-only" for="search-field-small">Search small</label>
           <input type="search" id="search-field-small">
           <button type="submit">
             <span class="usa-sr-only">Search</span>
--- a/_components/search-bar.md
+++ b/_components/search-bar.md
@@ -6,11 +6,13 @@ lead: Intro text on what is included in this section and how to use it. No more 
 ---
 
 <div class="preview preview-search-bar">
+
+  <h6>Search big</h6>
+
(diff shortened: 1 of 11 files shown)


_components/search-bar.md
```diff
@@ -6,11 +6,13 @@ lead: Intro text on what is included in this section and how to use it. No more
...
 
 <div class="preview preview-search-bar">
+
+  <h6>Search big</h6>
+
   <div class="usa-grid">
     <div class="usa-width-one-half">
       <form class="usa-search usa-search-big">
         <div role="search">
-          <label for="search-field-big">Search Big</label>
```
maya: inputs must have labels, please re-add.

_components/search-bar.md
```diff
@@ -20,11 +22,12 @@ lead: Intro text on what is included in this section and how to use it. No more
     </div>
   </div>
 
+  <h6>Search medium</h6>
+
   <div class="usa-grid">
     <div class="usa-width-one-half">
       <form class="usa-search">
         <div role="search">
-          <label for="search-field">Search Medium</label>
```
maya: inputs must have labels, please re-add.

_components/search-bar.md
```diff
@@ -34,11 +37,12 @@ lead: Intro text on what is included in this section and how to use it. No more
     </div>
   </div>
 
+  <h6>Search small</h6>
+
   <div class="usa-grid">
     <div class="usa-width-one-half">
       <form class="usa-search usa-search-small">
         <div role="search">
-          <label for="search-field-small">Search Small</label>
```
maya: inputs must have labels, please re-add.

angel:
@maya added the labels back and class="usa-sr-only" to keep them hidden but accessible!

maya:
@angel sorry just looked at this again. You added a h6 to the search bar and are visually hiding the label. This doesn't make sense. We should just use the label as it's title and remove the extraneous h6.

angel:
@maya I think we always want the label hidden. The h6 is just there as part of the preview title. Not part of the component itself. I would defer to @carodew @colinpmacarthur @mollieru if that doesn't make sense. Thanks!

colinpmacarthur:
@angel: Could you post a screenshot showing the decision we need to make? Thanks!

maya:
@colinpmacarthur they could potentially look identical visibly. This question is related to the markup. In this PR, we are visually hiding the label and adding an extraneous heading. I suggested to remove the heading and leave the label visible.

Unless you think that someone should always have the label hidden, I would not hide it. So that's what we;re asking you.

colinpmacarthur:
Got it. 

I'd argue for keeping the label and not including the h6 here for a couple reasons:
- as it is @angel's latest commit with both the h6 and the label, a screen reader will say the same name twice, right?
- as @maya points out, we have to keep that label for accessibility reasons. So, if we don't want redundancy, we keep the label, right?

But! My opinion is probably not the most important one here. @mollieru and @carodew should probably chime in on this one, too.

carodew:
Personally, I am not going to lose sleep over this, and I think either way is probably fine, and we just need to make a call. 

I would lean towards treating the label in the way we want people to use it in their sites, which sounds like we want it accessibly hidden, right? Otherwise, users will have to either hide it themselves or might remove it entirely if they don't want to show it, which would not be accessible.

maya:
OK will leave it as is with the label visually hidden bc that's how we anticipate people using it on their sites.

Issue #530, opened 2015-09-15
[Sidebar] Focus does not move to link target when using via keyboard secondary link in sidebar


jpyuda:
To reproduce:

1. Open any style guide page that has second-level navigation (i.e. Form Templates)
2. Tab to a link in the sidebar (i.e. Address Form)
3. Hit enter - the page will scroll to that anchor, but focus stays in the sidebar. It should move to the anchor.

jpyuda:
The same thing we did for the skip navigation link should work for this -- just need to generalize it.

maya:
Fixed via #621.

PR #621, opened 2015-09-17
Address keyboard navigation focus for #530

vgvg:
@maya Had to find a way to set focus and tabindex after the animation, re-using the skipnav code didn't work for the left nav items.

### DIFF
--- a/assets-styleguide/js/styleguide.js
+++ b/assets-styleguide/js/styleguide.js
@@ -88,6 +88,13 @@ $(function () {
   }
 });
 
+//capture that the enter key was used to "click"
+$('.sidenav').on('keydown', 'a', function (e) {
+  if (e.which === 13) {
+    $(this).data('keypress', true);
+  }
+});
+
 $('.sidenav').on('click', 'a', function(e) {
   var hashLocation  = $(this).attr('href').split('#')[1]; // long url splitting
   var scrollTopPos  = calculateAnchorPosition(hashLocation);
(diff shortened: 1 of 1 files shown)


maya:
cc @msecret for JS review.

assets-styleguide/js/styleguide.js
```diff
@@ -88,6 +88,13 @@ $(function () {
   }
 });
 
+//capture that the enter key was used to "click"
+$('.sidenav').on('keydown', 'a', function (e) {
+  if (e.which === 13) {
```
msecret: 13 should probably be a constant declared at the top of the function or file so people know what it is:

```
var ENTER = 13;
```

assets-styleguide/js/styleguide.js
```diff
@@ -114,6 +121,17 @@ $('.sidenav').on('click', 'a', function(e) {
       } else if (window.location.hash !== newHash) {
         window.location.hash = newHash;
       }
+    },
+    done: function () {
```
msecret: This seems potentially problematic if the focus happens after an animation is done. If a user is using the keyboard to navigation because they're not browsing the site visually, then this pause before the focus could be confusing. Although technically, I've never heard any official guidance on this. What do you think?

assets-styleguide/js/styleguide.js
vgvg: done

assets-styleguide/js/styleguide.js
vgvg: Yeah, I agree. I didn't notice too much of a delay since the animation is set to 200 ms so done fires super fast, and doing the focusing before breaks the animation. Maybe it's even better to just skip the animation entirely if the keyboard was used?

msecret: I think this would require more research. We have an accessibility expert on staff, I'll try and ask him today

maya:
@vgvg this stopped working for me but I remember it working yesterday.

assets-styleguide/js/styleguide.js
nickbristow: I did a bit of digging about accessible smooth scrolls and tested a couple with the screen reader. At most, focusing after the animation would cause a minor inconvenience, and even that is a bit of a stretch. 

As long as its clear, that its a link targeting another section on the page, and you move focus to the new area after the animation, it should be good. 

I don't think disabling smooth scrolling for keyboard navigation is necessary.

PR #626, opened 2015-09-17
Add link underline lists

### LINKED ISSUE #563: [Components Pages] Links in lists are not underlined
Links in lists should be underlined since there is surrounding text as stated in our list rules on the site. See Getting Started page is an example of where the links should be underlined.

maya:
This gives lists the underline style. It'll only effect links inside `usa-content-list` since non-text related lists should not have this style.

This resolves #563.

### DIFF
--- a/assets/_scss/elements/_typography.scss
+++ b/assets/_scss/elements/_typography.scss
@@ -118,7 +118,8 @@ dfn {
 
 // Custom typography
 
-p, li {
+p, 
+.usa-content li {
   a {
     text-decoration: underline;
   }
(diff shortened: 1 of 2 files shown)


assets/_scss/elements/_typography.scss
```diff
@@ -118,7 +118,8 @@ dfn {
 
 // Custom typography
 
-p {
+p, 
+.usa-content li {
```
msecret: Can we use `usa-content-list` instead? It will be a less global rule and that seems like a good use case.

maya: Done in cf5b185. Good call! Totally forgot about that class.

Issue #591, opened 2015-09-16
[Homepage] Principles section on mobile is not consistently displayed


angel:
On Chrome on my iPhone 5s, when I go to the homepage and then scroll down to the Principles section, I see this:
![image](https://cloud.githubusercontent.com/assets/103496/9913564/591dc186-5c62-11e5-9a81-869200497654.png)
![image](https://cloud.githubusercontent.com/assets/103496/9917498/ca7f81a2-5c76-11e5-8b32-92901eaaaf3a.png)
![image](https://cloud.githubusercontent.com/assets/103496/9917502/ceec9298-5c76-11e5-885d-3bc423965491.png)

It should look like this instead:
![homepage_principles](https://cloud.githubusercontent.com/assets/103496/9913576/64ff75c6-5c62-11e5-8aef-60c156806a46.png)

carodew:
@mollieru reports there is another section that looks more broken, screenshot to be added.

mollieru:
@angel can you add a screenshot of this whole section (not just the first two principles)?

angel:
@mollieru @carodew added two more screenshots above.

PR #627, opened 2015-09-17
Change graphic list to use grid instead

maya:
This updates the graphic list on the homepage to use our grid instead of the technique I was using. 

The problem is that the middle grid setting breaks this from looking usable cc @msecret 

This would resolve: #591

BIG :heavy_check_mark: :ok: 

screen shot 2015-09-17 at 3 36 02 pm

MEDIUM: :x: :fire: 

screen shot 2015-09-17 at 3 31 12 pm

SMALL:  :heavy_check_mark: :ok: 
screen shot 2015-09-17 at 3 36 10 pm

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -107,27 +107,35 @@ title: U.S. Web Design Standards
     </div>
   </div>
   <div class="usa-grid">
-    <div class="usa-width-one-half usa-width-one-half-top">
-      <img class="usa-img-circle"  src="{{ site.baseurl }}/assets-styleguide/img/home/homepage_illustrations_best_easiest.png" alt="Make the best thing the easiest thing">
-      <h3 class="usa-graphic-list-heading">Make the best thing <br>the easiest thing</h3>
-      <p class="usa-graphic-list-text">The web design standards are designed to provide designers and developers easy-to-use tools to most effectively deliver the highest quality government websites to the American people.</p>
+    <div class="usa-width-one-sixth usa-width-one-half-top">
+      <img src="{{ site.baseurl }}/assets-styleguide/img/home/homepage_illustrations_best_easiest.png" alt="Make the best thing the easiest thing">
     </div>
-    <div class="usa-width-one-half usa-width-one-half-top">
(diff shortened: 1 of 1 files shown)


msecret:
@maya you might want to try having the 1/6 widths go to 1 column at medium, but only if that doesn't break other things like the footer. Otherwise, at this point, we might have to look at applying a fix to just this contents medium breakpoint.

maya:
@msecret how about we just have the medium breakpoint for the footer and any other components that need it? That will give us the most flexibility and less having to go and override things.

msecret:
@maya the footer and other components are official examples of using the library. I think users would expect the grid to work and look like those use-cases more then the homepage (which already has a lot of custom CSS anyway). I think we should keep the medium rules on the components to ensure users don't have unexpected experiences with the grid because they're not using a "footer" class or something.

maya:
@msecret I'm not so much concerned that this is the homepage, what I'm concerned about is I see my use of the grid as a very normal thing that I'd expect our users to do as well.

msecret:
But I also see the way the grid is being used for the footer as a likely use case. How do we know which will be used more?

maya:
@msecret I think a good general rule is to offer the less complex thing as default and when you add complexity to do it on a per case basis.

msecret:
I don't think that rule applies in this situation, but before we go really far into this, did you try  having the 1/6 widths go to 1 column at medium?

maya:
@msecret you mean as an override?

maya:
@msecret do you mean it should span 2 columns bc 2/12 = 1/6 ?

maya:
@msecret By going with 2 columns, that fixes the size, but the margins still get killed bc it's the 3rd one:

screen shot 2015-09-17 at 4 37 02 pm

And sure, we could override this too, however we shouldn't expect our users to keep up with all of this.

maya:
@msecret probably gonna close this and do it a different way.

maya:
Closing via #630.

As @norrishung once said, "the grid has its limits, and you have to know when to not use it :smile:"

PR #642, opened 2015-09-18
Changing header button wording

colinpmacarthur:
At @mollieru's request, changing the buttons in the heading to match the original design files:
- "Download" changed to "Download code"
- "View on GitHub" to "View code on GitHub"

Background: There are links to download various types of things throughout the site (palettes, fonts, design files, etc.) and it’s not clear what these “download” button leads to vs. the others. 

The original design files also include icons in those buttons. I'm not sure how we're handling icons within buttons (or icons, in general) and am certainly not qualified to make that decision. If someone has suggestions, I'm happy to revise the PR.

### DIFF
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -20,12 +20,12 @@
       </div>
       <ul class="usa-button-list usa-unstyled-list">
         <li>
-          <a class="usa-button" href="#" onclick="ga('send', 'event', 'Downloaded framework', 'Clicked download button inside site');">Download</a>
+          <a class="usa-button" href="#" onclick="ga('send', 'event', 'Downloaded framework', 'Clicked download button inside site');">Download code</a>
         </li>
         <li>
           <a class="usa-button usa-button-outline" href="https://github.com/18F/usfwds/" onclick="ga('send', 'event', 'Viewed on Github', 'Clicked View on Github from inside site');"
 >
-            View on GitHub
+            View code on GitHub
           </a>
           </li>
       </ul>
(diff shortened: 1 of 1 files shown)


_includes/navbar.html
```diff
@@ -20,12 +20,12 @@
...
       <ul class="usa-button-list usa-unstyled-list">
         <li>
-          <a class="usa-button" href="#" onclick="ga('send', 'event', 'Downloaded framework', 'Clicked download button inside site');">Download</a>
+          <a class="usa-button" href="#" onclick="ga('send', 'event', 'Downloaded framework', 'Clicked download button inside site');">Download code</a>
         </li>
         <li>
           <a class="usa-button usa-button-outline" href="https://github.com/18F/usfwds/" onclick="ga('send', 'event', 'Viewed on Github', 'Clicked View on Github from inside site');"
 >
-            View on GitHub
+            View code on GitHub
```
maya: Can we keep this to "View on Github"? and then just put the assets in a link on the readme? The reason is this should be our main link to the repo so it's not needed.

colinpmacarthur: Fine with me. (I can see how "View code on GitHub" seems very redundant.) Thoughts @mollieru? Or @carodew?

carodew: I'd prefer "View on GitHub." Code seems implied. If it's not adding something meaningful I'll always opt for the pithier version.

maya: Agreed. Done in 0f55741.

Issue #564, opened 2015-09-15
[Components Page] Padding should be 30px around between the element and example box.


angel:
Right now the top and bottom padding does not match the sides.

Below are some examples:
#### Buttons:
![image](https://cloud.githubusercontent.com/assets/103496/9892983/f2f259e2-5bc8-11e5-9e0e-5b69f3bc7b92.png)

#### Labels: 
![image](https://cloud.githubusercontent.com/assets/103496/9892995/01d03466-5bc9-11e5-850d-f12d92965558.png)

#### Alerts: 
![image](https://cloud.githubusercontent.com/assets/103496/9892316/71a26eb4-5bc2-11e5-9db1-bc9baae25e9c.png)
![image](https://cloud.githubusercontent.com/assets/103496/9892320/7ae9415a-5bc2-11e5-941a-6b339b9abd41.png)

#### Tables:
![image](https://cloud.githubusercontent.com/assets/103496/9893019/325308b6-5bc9-11e5-9ca4-998fd8f089b6.png)

#### Accordions:
![image](https://cloud.githubusercontent.com/assets/103496/9892974/d58ae45a-5bc8-11e5-887c-2ba7caf6de45.png)
![image](https://cloud.githubusercontent.com/assets/103496/9892977/df4bb0d2-5bc8-11e5-85c8-e836abc85a7a.png)

elarenee:
@jovchinnikova and @elarenee are working on this!

bradnunnally:
@elarenee Awesome! Let me know how it goes and if you have any questions.

maya:
Closed via #1274.

PR #651, opened 2015-09-18
Add space to website preview boxes

angel:
Updated spacing to be a little more consistent for the preview boxes to address this #564.

An example:
image

Resolves: #564.

### DIFF
--- a/docs/_components/accordions.md
+++ b/docs/_components/accordions.md
@@ -7,7 +7,7 @@ lead: Accordions are a list of headers that can be clicked to hide or reveal add
 
 <div class="preview">
 
-  <h6 class="styleguide">Borderless</h6>
+  <h6 class="preview-first">Borderless</h6>
 
   <div class="usa-accordion">
     <ul class="usa-unstyled-list">
--- a/docs/_components/accordions.md
+++ b/docs/_components/accordions.md
@@ -7,7 +7,7 @@ lead: Accordions are a list of headers that can be clicked to hide or reveal add
 
 <div class="preview">
 
-  <h6>Borderless</h6>
+  <h6 class="styleguide">Borderless</h6>
 
   <div class="usa-accordion">
     <ul class="usa-unstyled-list">
--- a/docs/_components/alerts.md
+++ b/docs/_components/alerts.md
@@ -7,7 +7,7 @@ lead: Alerts keep users informed of important and sometimes time-sensitive chang
 
 <div class="preview">
 
-  <div class="usa-alert usa-alert-success styleguide">
+  <div class="usa-alert usa-alert-success preview-first">
     <div class="usa-alert-body">
       <h3 class="usa-alert-heading">Success Status</h3>
       <p class="usa-alert-text">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod.</p>
--- a/docs/_components/alerts.md
+++ b/docs/_components/alerts.md
@@ -7,7 +7,7 @@ lead: Alerts keep users informed of important and sometimes time-sensitive chang
 
 <div class="preview">
 
-  <div class="usa-alert usa-alert-success">
+  <div class="usa-alert usa-alert-success styleguide">
     <div class="usa-alert-body">
       <h3 class="usa-alert-heading">Success Status</h3>
       <p class="usa-alert-text">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod.</p>
--- a/docs/_components/footers.md
+++ b/docs/_components/footers.md
@@ -7,7 +7,7 @@ lead: Footers serve site visitors who arrive at the bottom of a page without fin
 
 <div class="preview">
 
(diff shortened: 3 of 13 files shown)


assets/css/styleguide.scss
```diff
@@ -321,11 +325,27 @@ $site-top:           124px;
...
+  fieldset,
+  h6:first-child,
+  h6.usa-heading-alt:first-child {
+    margin-top: 3rem;
+  }
+
+  ul,
+  ol {
+    margin-bottom: 0;
+  }
```
maya: This is effecting the actual lists. If you need to change component-wide spacing, it should be done in list.scss. In general we should not be messing with components in styleguide.scss.

assets/css/styleguide.scss
```diff
@@ -321,11 +325,27 @@ $site-top:           124px;
...
     bottom: 2em;
   }
-  padding: 1em $site-margins;
+  padding: 0 $site-margins $site-margins;
   border: 1px solid #eeeeee;
   .is-peripheral {
     opacity: .2;
   }
+
+  fieldset,
```
maya: Why are we changing fieldset margins in styleguide.scss?

assets/css/styleguide.scss
```diff
@@ -321,11 +325,27 @@ $site-top:           124px;
...
+  h6:first-child,
```
maya: Can you add a class instead? That would be a less global rule.

assets/css/styleguide.scss
```diff
@@ -832,7 +852,7 @@ h3 + .button_wrapper {
   padding-bottom: .8rem;
 }
 
-h6.usa-heading-alt {
+h6, h6.usa-heading-alt {
```
maya: Prefer to use a class if we need to change layout bc of the styleguide. Unless it's a global rule. Then it should be done in typography.scss.

assets/css/styleguide.scss
```diff
@@ -321,11 +325,27 @@ $site-top:           124px;
     top: 2em;
     bottom: 2em;
   }
-  padding: 1em $site-margins;
+  padding: 0 $site-margins $site-margins;
```
maya: Can you break this out to: 

```padding: {
  bottom: something;
  left: something;
  right: something;
  top: something;
}```

File: assets/_scss/elements/_typography.scss [hunk 6]
```diff
@@ -39,7 +39,7 @@ h1, h2, h3, h4, h5, h6 {
   font-family: $font-serif;
   line-height: $heading-line-height;
   margin: {
-    top: 1.5em;
+    top: 2.3em;
```
maya: This should be in alphabetical order.

assets/css/styleguide.scss
```diff
@@ -321,11 +325,27 @@ $site-top:           124px;
     top: 2em;
     bottom: 2em;
```
maya: this should be in alphabetical order.

assets/css/styleguide.scss
```diff
@@ -321,11 +325,27 @@ $site-top:           124px;
...
+
+  ul,
+  ol {
+    margin-bottom: 0;
+  }
+
+  > a {
+    margin-top: 3rem;
+    display: inline-block;
+  }
```
maya: What does this effect and why do we need this?

angel: @maya I had taken out the top margin on the preview box trying not to change the margin of the elements themselves, since some elements had margin tops to begin with. For the h6s they have a 4rem margin normally but the first one should be only 3rem. The > a was just for the link within that preview so that would only take that margin top when it was in a preview and not effect the a itself. ul and ol I wanted to take the margins out just for that box, but agreed that it should not effext the component itself. Will look into using a class this weekend.

assets/css/styleguide.scss
angel: Added a styleguide class

angel:
@maya made some updates based on your feedback, let me know if I missed anything. Thanks!

maya:
@juliaelman there's still a lot of `styleguide` classes in this. Were you going to swap it for `preview-first`?

_elements/form-controls.md
```diff
@@ -198,7 +198,7 @@ lead: Form controls allow users to enter information into a page.
 
 <div class="preview">
 
-  <fieldset class="usa-fieldset-inputs usa-sans">
+  <fieldset class="usa-fieldset-inputs usa-sans styleguide">
```
maya: I don't know if we should add a presentation class to our code samples.

maya:
@juliaelman I don't mind adding `preview-first` to a heading before the actual code, but this also adds `preview-first` to the actual code and don't want people to copy that over.

maya:
How about adding a class to the `preview` div and trying to target the the selector through that? That will keep our code samples much cleaner.

maya:
Closing for now as this is not being actively developed and will have some new design work coming in to the website. #564 will remain open to address at a later point.

PR #652, opened 2015-09-20
Skip calculateAnchorPosition if hash is empty string

### LINKED ISSUE #529: Javascript error on page load in Firefox
The error is:
`Empty string passed to getElementById().`

vgvg:
Fixes #529


maya:
I can confirm that the JS error is gone in FF.

PR #660, opened 2015-09-21
update search bar accessibility guidelines

carodew:


_components/search-bar.md
```diff
@@ -63,14 +63,11 @@ lead: A block that allows users to search for specific content if they know what
...
-    <p>As you customize this form template, ensure it continues to follow the:</p>
     <ul class="usa-content-list">
-      <li><a href="{{ site.baseurl }}/components/#forms-blocks">accessibility guidelines for form templates</a> and</li>
-      <li><a href="{{ site.baseurl }}/elements/#inputs">the accessibility guidelines for form controls</a>.</li>
+      <li>As you customize this form template, ensure it continues to follow the <a href="{{ site.baseurl }}/components/#forms-blocks">accessibility guidelines for form templates</a> and the <a href="{{ site.baseurl }}/elements/#inputs">accessibility guidelines for form controls</a>.</li>
     </ul>
-    <p>We also recommend:</p>
     <ul class="usa-content-list">
-      <li>Always include the word "search" inside the <code>&lt;button&gt;</code> element for screen readers. You can hide this text with <code>position: absolute; left: -999em;</code> (as we have here).
+      <li>Always include the word "search" inside the <code>&lt;button&gt;</code> element for screen readers. You can visually hide this text with <code>position: absolute; left: -999em;</code>.
```
maya: I would say they should use our screen reader only class or mixin and move this into implementation.

carodew: I don't mind being specific on which class or mixin to use, but a) I'll need you to tell me specifically which class or mixin to use, as I don't know what that is and b) I think it's good to mention in accessibility as well, as it's an important concern in that area.

maya: The class is `usa-sr-only`. The mixin is `@include sr-only;` both found in assets/_scss/core/uttilities.scss.

carodew: added in 35923ab

PR #662, opened 2015-09-21
Adding browser support language

mollieru:
Additional short paragraph about which browsers we support + why


maya:
@mollieru I believe @jpyuda is working on this section.

Issue #579, opened 2015-09-16
[Accordians] iPhone 5/Safari: Text in containers lacks right inner margin in some cases


jjoteal:
*Steps to reproduce*
Open a 'Documentation' accordion on iPhone5/Safari

*Expected behavior*
Right margin between text and accordion border

Screenshot:
![image](https://cloud.githubusercontent.com/assets/12192706/9908925/e25dfcda-5c63-11e5-800e-8f2ac6023fbd.png)

mollieru:
@jjoteal what version of iOS are you running?

jjoteal:
8.4.1

mollieru:
i have 8.3

carodew:
I'm able to replicate this using an iPhone 5 simulator, but not on my iPhone 6 (which is also running 8.4.1).

Very strange. 

To simulate: 

1. Open Chrome developer tools:
2. Toggle "Device Mode" to on and select "Apple iPhone 5" from device list. 
3. Navigate to the accordion page and scroll down to the accessibility guidelines. 

You'll see this, which matches the issue above: 

<img width="366" alt="screen shot 2015-09-17 at 4 30 06 pm" src="https://cloud.githubusercontent.com/assets/10144074/9945109/649dbab4-5d59-11e5-8bdb-21c57d67e442.png">

carodew:
I think the issue is in the phrase `aria-controls=‘collapsible-#` because the part that says `controls=‘collapsible` does not wrap, which is forcing the entire text area to be wider than it should be.

maya:
@carodew are you hitting refresh in the emulator? I did not experience this in emulator or phone for most. I checked every one and only saw this effecting dropdowns.

Also is @jjoteal looking at our most up to date code on staging?

carodew:
It's only on iPhone 5 and only on sections where there is a code snippet within the documentation that is not wrapping, and forcing the text block to be wider than the accordion. E.g. on the Accordion Accessibility documentation. I noticed it in a couple other places, which I can list if it's helpful. 

I can verify I did refresh and I am on staging.

carodew:
Arguably this could be a medium, but it looks bad so we put it as a high. Happy to revisit during triage.

maya:
@carodew I was testing on an iPhone 5

jpyuda:
I can't replicate this on an iPhone 5S with 8.4.1

carodew:
wbr – word break opportunity might be a fix.

[...]

PR #663, opened 2015-09-21
Add word-break opportunities

jpyuda:
Adding optional word-break spots using `<wbr>` inside of `<code>` tags
to address issue #579.


maya:
@jpyuda I'm still seeing this issue on dropdowns:

img_0152

jpyuda:
Added another `<wbr>` to dropdowns.

(Note: I don't think we'll likely keep up with adding these tags to new content, and ultimately I'm okay with that. That content will still scroll horizontally if somebody is stuck.)

maya:
Looks good. I agree, it will be hard to remember to keep adding these in.

PR #670, opened 2015-09-21
Add zip file of assets (code)

maya:
This adds the zip file of the assets to the site in the `assets/releases` folder.

This is waiting on button text guidance from @carodew and @colinpmacarthur around including the version number on the:

1. homepage button and/or small text below 
2. download button in the navbar


maya:
@alexose good call. I have hidden files hidden by default so didn't see these. I only see .DS_STORE, not  __MACOSX. When I delete it, it keeps coming back.

alexose:
We'll make sure that whatever build process we use in the future strips them out automagically :)

maya:
We will not be adding in version number to the site.

PR #643, opened 2015-09-18
Override default `baseurl:` with /designstandards

mbland:
cc: @mollieru @maya

PR #671, opened 2015-09-22
Override default `baseurl:` with /designstandards

mbland:
Supersedes #643 by merging into `18f-pages-staging` rather than `18f-pages`. That means that the new staging URL will be https://pages-staging.18f.gov/designstandards/, and once `18f-pages-staging` is merged into `18f-pages`, the 18F Pages URL will be https://pages.18f.gov/designstandards/.

cc: @maya @mollieru

### DIFF
--- a/_config_18f_pages.yml
+++ b/_config_18f_pages.yml
@@ -1,0 +1,1 @@
+baseurl: /designstandards


maya:
@mbland When I run the server now, it runs from `http://127.0.0.1:4000/`. Is this the expected behavior? Or is it also supposed to run from `http://127.0.0.1:4000/designstandards/`?

mbland:
Yes, that's expected. If you want to run it at http://localhost:4000/designstandards/, run it with `bundle exec jekyll serve --config _config.yml,_config_18f_pages.yml`.

maya:
Great, thanks for the clarification! lgtm :+1:

PR #677, opened 2015-09-22
Adds examples to font pairings

colinpmacarthur:
Per @mollieru's request. Includes links to screenshots for sites that are not public.


maya:
:+1:

PR #680, opened 2015-09-22
Updated the font size and line height of the intro text on the homepage

angel:


assets-styleguide/css/homepage.scss
```diff
@@ -154,6 +154,8 @@ h1, h2 {
 
 .intro-text {
   font-family: $font-serif;
+  font-size: $base-font-size;
```
maya: @angel this is already set to 1.7rem I believe

angel: @maya good catch, removed!

maya:
lgtm, @mollieru says that this follows our own guidance.

PR #684, opened 2015-09-22
Add example imgs pairings

angel:
Added example images of font pairings like this:
image

The images were a little small, so they are blurry. @mollieru if you want to re-save the images. I can update them.


maya:
Images are blurry. We should resolve.

Is this the focus we want? the image is a link but doesn't have a focus state:

screen shot 2015-09-22 at 3 10 09 pm

angel:
@maya unlinked the images. Waiting on higher res images.

angel:
@maya yup, images have been updated and the class has been removed. Let me know if I missed anything else. Thanks!

mollieru:
@maya is this good to be merged or do you need another dev reviewer?

_visual/typography.md
```diff
@@ -75,7 +75,10 @@ order: 01
             <p><span class="usa-label-big">Fast</span></p>
             <p>Ideal number of fonts. Will allow for optimal page load performance.</p>
             <h6 class="usa-heading-alt">Example</h6>
-            <p><a href="{{ site.baseurl }}/assets-styleguide/img/epa-emanifest-screenshot.png">EPA eManifest  (screenshot of non-public site)</a></p>
+            <p>
```
maya: Do we need to wrap these in `<p>` tags?

angel: @maya they were in p tags for the links that were merged before so I just added the images. It gives consistent spacing so I left it in. Let me know if you think we should take it out.

maya: Ok no it's ok, thanks @angel!

Issue #644, opened 2015-09-18
Browser support language for Getting Started


konklone:
@maya asked me to suggest some language for the Getting Started page re: IE8 and browser support. Here's an initial suggestion:

> These standards are designed to support both older and newer browsers through progressive enhancement, and officially support Internet Explorer 9 and up. Internet Explorer 8 and below generally see quite low usage today, and most agency websites should be able to safely begin support at Internet Explorer 9.

maya:
cc @jpyuda

konklone:
Here's a slightly longer version that acknowledges the security aspect:

> These standards are designed to support both older and newer browsers through progressive enhancement, and officially support **Internet Explorer 9 and up**. Internet Explorer 8 and below generally see very low usage today, and most agency websites should be able to safely begin support at Internet Explorer 9. Because IE8 is the last version of IE available on Windows XP, beginning support at IE9 also allows an agency to more easily adopt [modern security standards](https://https.cio.gov/sni/).

mollieru:
Thank you @konklone !!

konklone:
Thank **you**!

maya:
Added in https://github.com/18F/web-design-standards/pull/662. Thanks @konklone!

PR #692, opened 2015-09-23
Fix radio buttons and checkboxes in IE8

robinwhittleton:
In IE8 no checkboxes or radio buttons are visible. This patch overrides the styles and shows the native ones again. They don’t look quite so snazzy, but at least IE8 users can access services that use them.

screen shot 2015-09-23 at 16 30 18
screen shot 2015-09-23 at 16 30 47

### DIFF
--- a/assets/_scss/elements/_inputs.scss
+++ b/assets/_scss/elements/_inputs.scss
@@ -126,6 +126,14 @@ legend {
 input[type="checkbox"],
 input[type="radio"] {
   @include sr-only();
+
+  .lt-ie9 & {
+    border: 0;
+    float: left;
+    margin: .4em .4em 0 0;
+    position: static;
+    width: auto;
+  }
 }
 
 input[type="checkbox"] + label,


maya:
@robinwhittleton this is really great, thanks! 

I was actually meaning to add that in myself, but toward the end of the project we decided to officially support IE9 and up (See #644). However, this definitely does not hurt, so will share with the team. 

Many thanks! We :heart: GDS.

robinwhittleton:
Yep, I saw your official support schedule. A small fix never hurts though :)

(in other news, your radios / checkboxes also don’t work in iOS5, i.e. iPad 1s. We fixed that with a horribly hacky `onclick=""` on each `label`, but I haven’t done a PR for that as it’s pretty ugly, but let me know if you want to know more)

maya:
@robinwhittleton could you alphabetize the attributes so it follows our CSS coding styleguide: https://pages.18f.gov/frontend/css-coding-styleguide/format/

robinwhittleton:
No problems, I’ve amended my original commit to alphabetise the properties and added an extra line above to make it consistent with your styleguide.

PR #713, opened 2015-09-24
Update footer colors to use our variables

maya:
This makes a few updates to the footer:

- This updates the footer colors to use our color variables rather than hex codes. 
- This resolves color contrast issue with links on the footer and its background color.
- This uses a base/black link against the darker gray background.
- This lightens our `$color-gray-lightest` variable by 1% so that default blue links will pass color contrast requirements on this background.


alexose:
:+1:

PR #719, opened 2015-09-24
Add svg and higher res png arrow icons

maya:
This adds SVG with PNG fallback for arrows on mobile footers. This is what we do on all other icons and previously the icons were very pixelated.


alexose:
:+1:

PR #723, opened 2015-09-24
Optimize SVGs

shawnbot:
This PR optimizes all of the non-font SVGs using [svgo](https://github.com/svg/svgo).


carodew:
Hey @shawnbot ! When this is no longer WIP can you re-add the `to-be-triaged` tag?

msecret:
This looks good. Do we need a process documented for when we add new SVGs?

Actually now that I think about it, I think we'd eventually move SVG optimization into the build process. It's something I've done with Peace corps before

PR #728, opened 2015-09-25
Fix dropdown bug

maya:
This fixes the dropdown arrow bug which was huge bc we switched out the icons to SVG and need to declare the size here. Didn't realize that icon was used in another place besides the footer.


alexose:
:+1: 

Looks good-- Sorry I missed it in the first place!  This will teach me to grep more carefully..

PR #738, opened 2015-09-28
Ensure :focus shadow on sidenav is above :hover fill

andrewliebchen:
Its a really minor thing, but it bothers me that the hover `background-color` in the sidebar overlaps the `box-shadow` of the previous `:focus`-ed element:

screenshot 2015-09-28 16 13 20

This little change ensures that the `box-shadow` is always above the hovered elements:

screenshot 2015-09-28 16 15 01

Great work! So excited to see this. Looking forward to dive in when I have more time.

### DIFF
--- a/assets/_scss/components/_sidenav.scss
+++ b/assets/_scss/components/_sidenav.scss
@@ -29,6 +29,11 @@
       text-decoration: none;
     }
 
+    &:focus {
+      position: relative;
+      z-index: 1;
+    }
+
     &.usa-current {
       color: $color-primary;
       font-weight: $font-bold;


cademarkegard:
:+1:

hanibash:
:+1:

maya:
@andrewhughey thanks for this! I just want to look into if there's a better way to manage z-index with Sass rather than just use `z-index: 1` 

See: https://css-tricks.com/handling-z-index/

andrewliebchen:
Thanks @maya. At [OpenGov](http://opengov.com/), we use a combination method of a map variable to store the `z-index` key value pairs for each type of component, and a mixin to call that value in your components. 

I reserve `z-index` >= 10 for local components...since `z-index` is relative to it's nearest relative parent, I don't stress about it too much _within_ components. I'm plenty fussy about the components themselves.

maya:
@andrewliebchen good point! I think ultimately our active state will end up as a bold and blue link style so not sure if that focus state will stay. 

See:

screen shot 2015-09-30 at 2 08 38 pm

@vgvg implemented this originally.

maya:
Realizing we still need this for tabbing, even with the new future style.

maya:
Thanks @andrewliebchen!

andrewliebchen:
Yay! Glad to help...looking forward to seeing what else I can do.

PR #744, opened 2015-09-29
Remove link to golang site from README

liquiddandruff:
The `./go` file is a ruby script, not a golang script, so I have removed references to golang for clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ $ brew install ruby
 ### Installation
 
 Now that you have verified that you have Ruby installed, clone and run the
-following go commands to initialize and serve the library locally.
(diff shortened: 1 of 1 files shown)


Sequoia:
:+1: this was very confusing. :smile_cat:

mbland:
Thanks for the catch! But can you actually push a commit to replace `go script` with `[go script](https://github.com/18F/go_script)`?

mbland:
(This also reminds me I should update the `./go` script to use the new `go_script` gem.)

liquiddandruff:
I've updated it with the correct link, now it makes sense. Thanks!

PR #746, opened 2015-09-29
Fix error with regards to aria-hidden in accordion

SahAssar:


maya:
I'm pretty sure this is correct, can you also review this @nickbristow?

nickbristow:
@maya yes, `(aria-hidden = true) ~= (display: hidden, invisible, offscreen)`

So this pr addresses the issue.

maya:
Thanks @SahAssar! :+1:

PR #748, opened 2015-09-29
changing link for ./go scripts

dyanarose:
pardon me if this is incorrect, but I don't believe the ./go scripts are golang?


mbland:
Thanks for the catch, but this was also reported in #744.

PR #749, opened 2015-09-29
Change alertdialogue to alertdialog

dsturley:
Per [ARIA Spec](http://www.w3.org/TR/wai-aria/roles#alertdialog) the role is `alertdialog` and not `alertdialogue`.


maya:
Thanks for this! Can you review @nickbristow @colinpmacarthur @carodew?

carodew:
Lgtm.

maya:
@dsturley thanks for catching this! :+1:

Issue #745, opened 2015-09-29
Bower


robertmain:
It would be really nice to have a `bower.json` file in this repo. That way, people could just add it to their bower file removing the need to clone the entire repo.

matyasfodor:
:+1:

mollieru:
Thanks for the input! We've been discussing adding this as well. We'll keep you posted!

Scotchester:
:+1:

kiblerdude:
:+1:

andrewliebchen:
NPM too.

robertmain:
npm isn't really for client side dependencies though

andrewliebchen:
It is though. React and React components are distributed on npm. So is Bourbon, which is a dependency if this library. It's not tough to publish to both bower and npm.

matthewjewell:
:+1:

jeremiak:
Hi all, love the feedback! I like where this is going and completely agree that we should make the design standards a bit easier to leverage in your projects through dependency management tools (i.e. `bower` or `npm`). However, 18F projects don't use bower as a matter of policy (18F's front end policies and guidance can be found in our [Front end guide](https://pages.18f.gov/frontend/recommendations/dependency-management/)).

I'm going to close this issue, as I did wth #760, in favor of #767 (making the standards available through `npm`) because it strives for the same goal of making it easier to include the design standards in your project and get started using them.

PR #760, opened 2015-09-30
Create bower.json

robertmain:
Created bower.json so that people using gulp can easily have gulp add it to the sass path and use the mixins as normal i.e 
```scss
@import  'elements/buttons'
```

ref #745

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,16 @@
+{
+  "name": "web-design-standards",
+  "description": "The U.S. Web Design Standards is a library of open source UI components and a visual style guide for U.S. federal government websites.",
+  "version": "4.2.5",
+  "main": [
+    "./assets/_scss/**/*.scss",
+  ],
+  "authors": [
+    "18F (https://playbook.cio.gov/designstandards/)"
+  ],
+  "homepage": "https://playbook.cio.gov/designstandards/",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:18F/web-design-standards.git"
+  }
+}


robertmain:
ref #745

jeremiak:
Hi @robertmain - thanks so much for opening a PR and getting involved. I like where this is going and completely agree that we should make the design standards a bit easier to leverage in your projects through dependency management tools. However, 18F projects don't use `bower` as a matter of policy (18F's front end policies and guidance can be found in our [Front end guide](https://pages.18f.gov/frontend/recommendations/dependency-management/)).

I'm going to close this issue in favor of #767 (making the standards available through `npm`) because it strives for the same goal of making it easier to include the design standards in your project and get started using them.

robertmain:
Hi @jeramiak

Thanks for that, I do have a couple of points though..

1. In issue #745 @mollieru mentioned that bower was something being considered. I think a little communication between authors of this project wouldn't go amis so that contributors (such as myself) know what's going on

1. Not everybody uses bower. I realise that you don't like it, but that doesn't mean that everybody feels the same way. In fact I would go as far as to say that bower (in my opinion) is more suited to front end dependency management than npm due to the lack of nested dependencies (although I acknowledge this has been fixed somewhat in later versions of npm). Seriously, it's one file that a few kb in size - why not include it in your repo for folks that want to use it?

1. Tooling. Currently I use gulp to compile all my dependencies into one file (as many people do). That relies heavily on [main-bower-files](https://github.com/mauricedb/gulp-main-bower-files). Whilst such a solution may exist for npm, I'm not currently aware of such a thing.

PR #774, opened 2015-10-02
Closes #763.

### LINKED ISSUE #763: [Sidebar nav] Mobile experience
On iOS 8 with Chrome the mobile sidebar menu has varying success. When the nav is expanded sometimes swiping up and down scrolls the nav items and other times the page content scrolls and nav incorrectly stays fixed.

Most people are right handed. Putting the menu button in the top left corner of the screen is furthest from the dominate thumb's natural position to expand the nav.  It may even require two hands to use that UI control position depending on the scale of device to the user's hand.

Taitrnator:
if you guys do decide on changing putting the nav on the right side, I went ahead and swapped the relevant styles here.

### DIFF
--- a/assets/css/styleguide.scss
+++ b/assets/css/styleguide.scss
@@ -123,7 +123,7 @@ $site-top:           124px;
     display: none;
   }
   display: inline;
-  float: left;
+  float: right;
   margin-top: -4px;
   padding-top: 1.5rem;
   padding-bottom: 1.5rem;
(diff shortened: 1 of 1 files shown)


mollieru:
Thanks so much! We're considering the options here -- stay tuned!

maya:
Refers to #783.

maya:
@Taitrnator this is going to go into our V2 site design, and we'll add it on to another PR. Thanks again!

PR #776, opened 2015-10-02
Add alert role to error alert

### LINKED ISSUE #727: ARIA attributes in code samples
Not sure whether you guys talking about this or not but there's a discrepancy between the accessibility documentation and the code samples in some elements. (https://playbook.cio.gov/designstandards/alerts/)

Should attributes like aria-roles be placed in code samples by default so it encourages people to use them? (especially if they are copying and pasting)

maya:
This adds `role="alert"` to the error alert.

This resolves #727.

Do we need to add anything in the documentation stating we only added `role="alert"` to error alerts and not the others (i.e. success, warning, info)?

### DIFF
--- a/_components/alerts.md
+++ b/_components/alerts.md
@@ -21,7 +21,7 @@ lead: Alerts keep users informed of important and sometimes time-sensitive chang
     </div>
   </div>
 
-  <div class="usa-alert usa-alert-error">
+  <div class="usa-alert usa-alert-error" role="alert">
     <div class="usa-alert-body">
       <h3 class="usa-alert-heading">Error Status</h3>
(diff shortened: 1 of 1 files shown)


carodew:
I thought we needed more research to make a decision on how to move forward with this?

maya:
@carodew I talked to @nickbristow and @andrewmaier and they both agreed with this route.

maya:
@carodew how do you plan to user test this? I can change this to a [WIP] PR if you like.

carodew:
I didn't realize @nickbristow and @andrewmaier had reviewed this. If that's the case, then I'm ok with merging this. @colinpmacarthur and I are starting to discuss the next steps for user research now, so there isn't a formal plan yet.

alexose:
Looks good.  I agree that only "error" alerts should get the alert role based on what MDN says.

Issue #764, opened 2015-10-01
Months and Days should limit input to two characters.


imthatoneguy:
You should reject dates like the month '8318801' inside of say a birthday field.

Kireyin:
That and validate only if numbers are inputed.

maya:
@imthatoneguy @Kireyin The `maxlength` attribute is not supported for number inputs. Do you have any other suggestions? I've included a min and max attribute in #777.

maya:
Resolved in #777.

PR #777, opened 2015-10-02
Update date of birth inputs

maya:
This updates the date of birth inputs so that we include min and max attributes and values for each input. It removes the `maxlength` attribute as it's not supported for number inputs, just text inputs.

This is related to #764.


nickbristow:
🙌

alexose:
LGTM.  Although Chrome doesn't highlight invalid numbers the way Firefox does.  Not a show-stopper-- just something I learned :)

Issue #557, opened 2015-09-15
[Footer] Medium Footer 'primary link' sending users to top of page in Safari


mollieru:
Click 'Primary link' -- you'll be taken to the top of the page. 

http://screencast-o-matic.com/watch/coQo3ifdzV

*Expected behavior:*
Blank link shouldn't take user anywhere

*Tested in*
OX 10.8
Safari 6.2

maya:
Closed via #778.

PR #778, opened 2015-10-03
Update link values to use hashes #

### LINKED ISSUE #743: Remove "javascript:void(0)" href values
These should be made into real links or at least "#" (as the universal and arguably better alternative)

maya:
This changes our links to use `#`'s instead of `javascript:void(0)`. This adds an event handler to prevent the default behavior of links jumping to the top of the page.

This resolves #743 and #557.

Thanks @dylanvalade!


dylanvalade:
Glad it worked out.  Thanks for the quick resolution!

alexose:
Looks good!

PR #780, opened 2015-10-03
[WIP][IDEA] Using Jekyll Entries to Compile valid Data.json v1.1 

JJediny:
## PR ONLY for comments and feedback 

**Looking for issues/conflicts/ideas on the approach to create a lightweight distributed means for any agency/program/office/effort/event to generate a data.json file/URL that Data.gov can point to and register**

Code itself needs some work to address filtering for null/blank infrequently-used metadata fields that would otherwise drag performance... 

Inspired by @state-hiu-dev  https://github.com/state-hiu/hiu.state.gov... Which used YAML Frontmatter to generate valid GeoJSON :+1: 

* https://github.com/state-hiu/hiu.state.gov/tree/master/products/_posts
* https://raw.githubusercontent.com/state-hiu/hiu.state.gov/master/products/products.json

...builds...

* http://state-hiu.github.io/hiu.state.gov/products/products.json

Part of a broader effort to provide vetted/production ready shared services that automate registration of new data to Data.gov...
https://github.com/DistributedOpenUnifiedGovernmentNetwork


gbinal:
Hey @JJediny.  

So - I totally agree with the usefulness of this.  Taking the same model (laid out [here](https://github.com/GSA/Very-Simple-API)), I got started building this [here](https://github.com/project-open-data/simple-catalog-generator).  

I definitely think a number of agencies would use this if we iron it out, but regardless, I recommend we take this over to [Project Open Data](https://github.com/project-open-data/simple-catalog-generator), since I think the goal is pretty distinct from the web design standards.

pjdufour:
Put this together:  https://github.com/pjdufour/datajson-builder.  Check out the live site at http://pjdufour.github.io/datajson-builder and http://pjdufour.github.io/datajson-builder/data.json.

I put in some example data from State's data.json at http://state.gov/data.json.

I think using Jekyll to build YAML posts into a data.json could be immensely useful.  This lightweight approach would enable agencies to leverage the great social aspects (issues, PRs, etc.) of GitHub to manage publishing of open data.  CKAN could harvest directly from this site.

gbinal:
Made a note about these ideas over at https://github.com/project-open-data/simple-catalog-generator/issues/3.

JJediny:
Thanks @gbinal @pjdufour ! Completely agree that the move be made over to [Project Open Data](https://github.com/project-open-data/simple-catalog-generator/issues/3) Always intended that to be the case - just wanted to proof the concept with the web-design-standards as the "reuse it yourself" approach to the frontend/design/layout/components of the web-design-standards combined together with a data.json entry as a Jekyll `page` as a way to provide both aspects. 

Will move the conversation over to project-open-data/simple-catalog-generator#3.

Issue #793, opened 2015-10-06
tabindex="0" on input not needed


goetsu:
the code example on https://playbook.cio.gov/designstandards/form-controls/#checkboxes and https://playbook.cio.gov/designstandards/form-controls/#radiobuttons doesn't need tabindex="0" as input are already a focusable element

maya:
Thanks for the catch. We'll work on getting this removed.

PR #798, opened 2015-10-06
Remove tabindex="0" on inputs

maya:
This removes `tabindex="0"` on radio buttons and checkboxes bc it's not needed - they are already focusable. 

Resolves: #793.


vgvg:
:+1:

PR #801, opened 2015-10-06
force the ruby version when using RVM/rbenv

afeld:
Just makes switching into the projects slightly more convenient.

https://rvm.io/workflow/projects


mbland:
Just bumped the version in the README, too. @maya, ready to do the honors?

PR #804, opened 2015-10-07
doc: typo encompases to encompasses

reggi:


maya:
Thanks for the catch :+1:

Issue #803, opened 2015-10-06
[Documentation] Color codes for secondary-darkest and secondary-dark are swapped


timbonicus:
Looking at the [palette documentation](https://playbook.cio.gov/designstandards/visual-style/#palette), it shows secondary-darkest as hex #cd2026, and secondary-dark as #981b1e. These are backwards, as evidenced by the swatches above using the opposite hex code.

maya:
Thanks for catching this!

carodew:
This will be fixed in our next push to production, thanks!

maya:
Fixed in #808.

PR #808, opened 2015-10-07
Fix incorrect hex codes

maya:
This swaps the hex codes of secondary-darkest and secondary-dark so that we display the correct hex codes.

Resovles #803.


vgvg:
:+1:

PR #819, opened 2015-10-11
Fixed #818 - Added options for military and territory mail codes

### LINKED ISSUE #818: Standard template for entering a U.S. mailing address should include Overseas Military Mail codes and U.S. territories
```
<option value="AL">AL - Alabama</option>
<option value="AK">AK - Alaska</option>
<option value="AS">AS - American Samoa</option>
<option value="AZ">AZ - Arizona</option>
<option value="AR">AR - Arkansas</option>
<option value="CA">CA - California</option>
<option value="CO">CO - Colorado</option>
<option value="CT">CT - Connecticut</option>
<option value="DE">DE - Delaware</option>
<option value="DC">DC - District of Columbia</option>
<option value="FM">FM - Federated States of Micronesia</option>
<option value="FL">FL - Florida</option>
<option value="GA">GA - Georgia</option>
<option value="GU">GU - Guam</option>
<option value="HI">HI - Hawaii</option>
<option value="ID">ID - Idaho</option>
<option value="IL">IL - Illinois</option>
<option value="IN">IN - Indiana</option>
<option value="IA">IA - Iowa</option>
<option value="KS">KS - Kansas</option>
<option value="KY">KY - Kentucky</option>
<option value="LA">LA - Louisiana</option>
<option value="ME">ME - Maine</option>
<option value="MH">MH - Marshall Islands</option>
<option value="MD">MD - Maryland</option>
<option value="MA">MA - Massachusetts</option>
<option value="MI">MI - Michigan</option>
<option value="MN">MN - Minnesota</option>
<option value="MS">MS - Mississippi</option>
<option value="MO">MO - Missouri</option>
<option value="MT">MT - Montana</option>
<option value="NE">NE - Nebraska</option>
<option value="NV">NV - Nevada</option>
<option value="NH">NH - New Hampshire</option>
<option value="NJ">NJ - New Jersey</option>
<option value="NM">NM - New Mexico</option>
<option value="NY">NY - New York</option>
<option value="NC">NC - North Carolina</option>
<option value="ND">ND - North Dakota</option>
<option value="MP">MP - Northern Mariana Islands</option>
<option value="OH">OH - Ohio</option>
<option value="OK">OK - Oklahoma</option>
<option value="OR">OR - Oregon</option>
<option value="PW">PW - Palau</option>
<option value="PA">PA - Pennsylvania</option>
<option value="PR">PR - Puerto Rico</option>
<option value="RI">RI - Rhode Island</option>
<option value="SC">SC - South Carolina</option>
<option value="SD">SD - South Dakota</option>
<option value="TN">TN - Tennessee</option>
<option value="TX">TX - Texas</option>
<option value="UT">UT - Utah</option>
<option value="VT">VT - Vermont</option>
<option value="VI">VI - Virgin Islands</option>
<option value="VA">VA - Virginia</option>
<option value="WA">WA - Washington</option>
<option value="WV">WV - West Virginia</option>
<option value="WI">WI - Wisconsin</option>
<option value="WY">WY - Wyoming</option>
<option value="AA">AA - Armed Forces Americas</option>
<option value="AE">AE - Armed Forces Africa</option>
<option value="AE">AE - Armed Forces Canada</option>
<option value="AE">AE - Armed Forces Europe</option>
<option value="AE">AE - Armed Forces Middle East</option>
<option value="AP">AP - Armed Forces Pacific</option>
```

IanLee1521:


msecret:
This looks fine from a code perspective. @carodew @colinpmacarthur, feel free to close if you're good to go on this.

carodew:
@colinpmacarthur do you think that adding the state abbreviations to each option will cause issues for screen readers?

colinpmacarthur:
Hmm. I'm not completely sure, but that's definitely a concern. A question for our accessibility folks?

IanLee1521:
Would it be better if I removed the state abbreviations? I was just following along the proposed change from #818. Presumeably this must be a solved problem on some form somewhere? :)

maya:
@IanLee1521 I think we were debating that very thing! Not sure if we know which is best practice, w/ or w/o abbreviations - since the value is the abbreviation too.

carodew:
@IanLee1521 I checked with some of our accessibility experts about the screen reader experience and they all thought the abbreviations were fine, and probably even a benefit (since a user could skip the option more quickly). So I think we're ok to merge!

maya:
@carodew I was talking with my critique group and noticed we've never seen abbreviation before. Also, how is adding an abbreviation helpful for usability. Isn't hearing "Ohio" easier to understand than "O-H"? I think it would be good to do some usability testing before merging. Or sticking with w/o abbreviations until we've done testing.

Also, our label is "state" but we're also including military territories. So do we need to change the label?

carodew:
We had initial user testing indicate that people had trouble typing in the state when the state shared first letters with other states. For example, typing in "New" brought up New Jersey but users had to revert to mouse clicks or other navigation types to get to New Mexico or New York. Hence why I thought abbreviations might solve that problem. But as it is non-standard I'm happy to add that to our testing backlog to further validate that assumption. And in fact it might be easier to have a different solution altogether and not use a dropdown. A text box with type-ahead might solve all of these issues and _might_ be easier to use.

maya:
@carodew you mention users had to type a state. How would they type on a select input? It's just a dropdown.

carodew:
They didn't have to type. But for most it was their first choice of interaction rather than opening the dropdown and scrolling.

maya:
So you saw them trying to type in the select input?

carodew:
Yes, as a way to jump to the correct option.

maya:
Interesting, I didn't know that people did that. It would be interesting to see if they thought they were typing in it like a text field or knew they could jump to different options.

maya:
@carodew do you have any thoughts around:

>Also, our label is "state" but we're also including military and territory codes. So do we need to change the label?

IanLee1521:
@maya I've used select boxes this same way. For example for me, I'll often start typing 'C' as a way to jump in a select box to 'California'. Or when I lived in Connecticut, hitting 'C' three times would often accomplish the right thing ('C' => 'California', 'CC' => 'Colorado', 'CCC' => 'Connecticut'.

Similarly for month selectors, though there it is tougher because sometimes the selector is by spelling, e.g. 'August' and sometimes it's '08' (and then sometimes just '8'...)

IanLee1521:
I should add that I fully understand that I'm typing in a text field. As a quick example, if you use the [w3schools TryIt for the `<select>` tag](http://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select), if you give focus to the select field (I clicked in the result area and hit `TAB` until it was in focus). You can get to the 'Saab' option just by typing 'S'.

In case it matters, this is in Chrome on OSX, not sure it it's an OS / Browser dependent feature. :)

acki:
Definitive using here that keyboard option too - on Firefox / OS X, but works also on other OS and Browser.
@maya Maybe you name it "State / Military / Territory"? With state alone the military user maybe didn't know they could select their codes.

carodew:
FWIW, I am not entirely convinced that a drop down is best solution for a state selector, although it is one that people are very familiar with. It seemed to be more challenging on desktop/laptops than text boxes, but on mobile, text boxes seemed more difficult.

It might be better as some type of text box with type-ahead suggestions (as a progressive enhancement), but I'm not sure. It's an area I'm hoping we can do more research into in the future, including how best to accommodate military addresses.

bradnunnally:
After reviewing everything and doing some research on accessible form design and such I'd like to explore the following options. None of these are meant to say "this is the direction" rather some prototypes we could build out to test with. 

1. Use a combo-box design pattern that allows for both selection and type ahead, where the format of the data set is STATE (XX). This would allow for both full state input as well as the abbreviation. 
2. Use a combo-box where the data list is just the name of the state/territory/military location, and there is hidden meta data which would allow someone to type in the shorthand of a state. E.g. CA for California. This metadata would be screen reader available though to ensure accessibility.
3. Look into using APIs like that [USPS Verify An Address](https://www.usps.com/business/web-tools-apis/welcome.htm) tool to assist users in correcting any accidental errors. (Can an API connection like this be included in the component design or is that best left up to the implementation team using the standards?)
4. Do some outreach with people that frequently send mail to either a territory or military address to validate if the label State makes sense? Or if another label or additional microcopy is needed. 
5. Can we move the zip code entry up in the order of fields to pre-fill city and state, while providing the option to correct any errors if the wrong city is pre-filled? (The flow here would be Address Line 1 & 2 -> Zip Code -> Verify/Correct City & State)

I'll put some thought to this in the coming days and post a few prototypes that illustrates these different scenarios.

jpyuda:
I'm a big fan of doing ZIP code first. There's some secondary research that supports it as a pattern, as well.

API verification of address is probably a worthwhile recommendation, and we could even provide code snippets, but is something that individual implementors would have to do for themselves because of how USPS runs the API.

As far as the original question: I don't really know what the frequency and pattern for military mailing addresses is. I suspect the need varies quite a bit by agency.

juliaelman:
@IanLee1521 thank you for your contribution and helping us move forward for the work on #818! We appreciate it.

Closing this pull request for now and will take a look again once we surface some more findings from the research outlined in the above comments.

PR #827, opened 2015-10-14
Updating title for Forgot username

burgwyn:
Correcting text


maya:
Still showed a merge conflict so superceded in #829. Thanks for catching this!

PR #841, opened 2015-10-19
Update zip

alexose:
I incremented wds-design-v0.8.1.zip even though nothing has changed.  Hopefully that makes sense.


maya:
lgtm :+1:

PR #842, opened 2015-10-19
Change overflow to auto on styleguide sidenav

maya:
This makes it so the scrollbar only appears when the height of the sidebar navigation is not tall enough for the content. Otherwise, we just see a nice, solid line.

When it's tall enough:

screen shot 2015-10-19 at 2 01 23 pm

When it's short:

screen shot 2015-10-19 at 2 01 34 pm

h/t to @shawnbot for figuring this out!


alexose:
:+1:

Issue #552, opened 2015-09-15
[Homepage] Fix text link under Contribute on Github button


KHarshawat:
The spacing of the link under the `Contribute on Github' button looks a little weird. Consider the following options: 

- add more padding between it and the button above
- put the link inline with the button so it reads more like a sentence:

![image](https://cloud.githubusercontent.com/assets/6335546/9888733/a253ae86-5bc4-11e5-8537-71e97199ae5f.png)

maya:
Resolved via #845.

Issue #844, opened 2015-10-20
expand feedback options 


gbinal:
This issue may be a duplicate but if not it'd be great to add an option for those who want to get in touch but aren't comfortable using github.  Email seems the easiest solution to add, but there are other good options, I'm sure.

brittag:
This email address would be nice to mention <a href="https://github.com/18F/web-design-standards#got-feedback">in the README</a> as well.

gbinal:
+1 - expanding the pull request to do that

PR #845, opened 2015-10-20
Adding email links

gbinal:
In support of #844.

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -187,6 +187,6 @@ title: U.S. Web Design Standards
     <div class="usa-cta">
       <a class="usa-button usa-button-secondary" href="{{ site.repos[0].url }}">Contribute on GitHub</a>
     </div>
-    <a href="{{ site.baseurl }}/getting-started" onclick="ga('send', 'event', 'Viewed the standards', 'Clicked View the Standards link at the bottom of the homepage');">Or view the standards</a>
+    <a href="mailto:uswebdesignstandards@gsa.gov">Or drop us an email</a>
   </div>
 </section>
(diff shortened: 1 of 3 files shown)


pages/index.html
```diff
@@ -187,6 +187,6 @@ <h3 class="usa-example-heading usa-link-external">
     <div class="usa-cta">
       <a class="usa-button usa-button-secondary" href="{{ site.repos[0].url }}">Contribute on GitHub</a>
     </div>
-    <a href="{{ site.baseurl }}/getting-started" onclick="ga('send', 'event', 'Viewed the standards', 'Clicked View the Standards link at the bottom of the homepage');">Or view the standards</a>
+    <a href="mailto:uswebdesignstandards@gsa.gov">Or drop us an email</a>
```
maya: Do we want to include any GA handler events in this? cc @colinpmacarthur

carodew:
Can we write out the full email address? Many people don't have their primary email client set up on the computer and clicking on the email link opens an email client they don't use and haven't set up an account with. Writing out the email address means they can also copy/paste into their preferred email  client.

gbinal:
We definitely could.  It's a bit lengthy though and may not look good.  I don't see it as a big problem but am fine either way.

maya:
I've updated the footer to state the full email address.

What do you guys prefer for the homepage on the bottom?

screen shot 2015-10-21 at 11 54 51 am
screen shot 2015-10-21 at 11 54 34 am

mollieru:
@maya the second option is compliant w our link styles (https://playbook.cio.gov/designstandards/visual-style/#links) so lets go w that one!

maya:
@mollieru done in b078614. Also, adding the extra space resolves #552.

Issue #737, opened 2015-09-28
Font rendering on Contribute section of main page


kendaleiv:
Font rendering seems a bit off on the **Contribute** section of the main page with Windows 10 in Chrome and Firefox (Chrome on left, Firefox on right).

Perhaps there's a way to improve it?

![image](https://cloud.githubusercontent.com/assets/1012917/10150633/7f14fe44-660e-11e5-80b2-a68b23a29566.png)

mollieru:
Thanks for the feedback! We'll check it out. Stay tuned.

maya:
Resolved via #848.

Issue #802, opened 2015-10-06
Update Source Sans Pro fonts to the latest version


miguelsousa:
The Source Sans Pro font files in `assets/fonts` are [version 1.050, which was released in April 2013](https://github.com/adobe-fonts/source-sans-pro/releases/tag/1.050R). The [current versions](https://github.com/adobe-fonts/source-sans-pro/releases) are 2.020 (uprights) and 1.075 (italics), and were released last month.

Updating to the latest version of the fonts will likely fix issue #737, reduce the fonts' footprint, and get the assets in sync with the latest and greatest.

I'm one of the maintainers of the [Source Sans Pro repository](https://github.com/adobe-fonts/source-sans-pro) and I'm willing to submit a PR to update the fonts. **But before I do any work, I'd like to know if there's any interest in getting the fonts updated?** I ask this because the original font files will require some amount of post-processing to get them down-sampled to match the same set of characters and features that is supported by the current files in `assets/fonts`. In other words, the PR is not as simple as submitting the original font files (although I can certainly do that, but it will come with a performance hit).

maya:
@miguelsousa thanks so much for reaching out! 

We're very interested in using the latest and greatest fonts (the current ones are just what was offered via Google Fonts). We'd want to consider performance impact as well, so optimizing it like you mentioned sounds like a good idea.

A few questions:
- Can you let us know what the file size difference would be?
- What's new in those files?

Thanks again!

miguelsousa:
@maya happy to learn that you welcome this improvement.

The file size will be comparable to the current one. I won't know the exact number until I work on it, but the new font files may actually come in slightly smaller because we took out the license's full text from the fonts.

In terms of differences, the design got a bit more polish (intermediate weights are better interpolated; the diacritics were resized), and the TrueType hints were redone from scratch (a [comment from a TrueType expert](https://twitter.com/mickduggan/status/649151849732419584) was actually what prompted us to reach out). The font family received other improvements, such as the addition of Greek and Cyrillic, but these are not relevant for your project.

Feel free to assign me to this issue. I'll report back once I have more news.

maya:
@miguelsousa I think we can only assign issues to collaborators/team members on the project. However, feel free to submit a PR and we'll be happy to review it.

mollieru:
@miguelsousa awesome, let us know when you've got something for us to look and we will merge it in! Thanks so much for keeping us up to speed :ok_hand:

miguelsousa:
[...]

Issue #756, opened 2015-09-30
Font should support international characters


maciejsaw:
Currently used webfonts doesn't support international characters.
As a user I may have international characters in my name or address so I want to type it correctly.

Suggested solution:
To use the full character set, update your API link to include the 'latin-ext' subset.
https://developers.google.com/fonts/docs/getting_started#Subsets

maya:
Thanks we'll look into this.

msecret:
We should pay attention to performance when we make this change. Essentially do a before and after to determine the impact.

maya:
The standards now has support for international characters with Source Sans Pro: 

<img width="590" alt="screen shot 2015-10-21 at 12 16 24 pm" src="https://cloud.githubusercontent.com/assets/5249443/10647521/a9b0e47e-77ed-11e5-9267-f7d7278e9a1e.png">

miguelsousa:
@maya that's not quite right. The original Source Sans Pro fonts support Greek and Cyrillic (along with extended Latin), but the subset font versions I submitted as part of issue #848 do not support Greek nor Cyrillic. [The glyph/character set supported by the subset fonts is shown in this PDF](https://github.com/miguelsousa/source-sans-pro-subset/blob/master/sourcesanspro-regular-webfont.pdf).

maya:
Ah ok, thanks for the clarification @miguelsousa

miguelsousa:
Here's the complete list of (Unicode) characters supported by the subset fonts:
```
0020 SPACE
0021 EXCLAMATION MARK
0022 QUOTATION MARK
0023 NUMBER SIGN
0024 DOLLAR SIGN
0025 PERCENT SIGN
0026 AMPERSAND
0027 APOSTROPHE
0028 LEFT PARENTHESIS
0029 RIGHT PARENTHESIS
002A ASTERISK
002B PLUS SIGN
002C COMMA
002D HYPHEN-MINUS
002E FULL STOP
002F SOLIDUS
0030 DIGIT ZERO
0031 DIGIT ONE
0032 DIGIT TWO
0033 DIGIT THREE
0034 DIGIT FOUR
0035 DIGIT FIVE
0036 DIGIT SIX
0037 DIGIT SEVEN
0038 DIGIT EIGHT
0039 DIGIT NINE
003A COLON
003B SEMICOLON
003C LESS-THAN SIGN
003D EQUALS SIGN
003E GREATER-THAN SIGN
003F QUESTION MARK
[...]

PR #848, opened 2015-10-21
Update Source Sans Pro fonts

miguelsousa:
This is to address issue #802.

The collective size of the new files is 546KB, while the previous version was 595KB. On the surface, this doesn't seem that big of a size reduction, but the new fonts actually support for more characters (which addresses issue #756, I believe), and they also support true small caps and old style numbers.

I did not generate new SVG files because those "fonts" are only needed for [**very old** browsers](http://caniuse.com/#feat=svg-fonts). I think they should be removed from this project. [WOFF](http://caniuse.com/#feat=woff), and soon [WOFF2](http://caniuse.com/#feat=woff2), are more broadly supported and have a smaller footprint.


maya:
This is awesome - thanks so much @miguelsousa!

I think we can probably create a new issue about removing SVG fonts.

This will also resolve #737.

maya:
@miguelsousa Quick question: What are old style numbers?

What do you recomend as the best place to view all available characters, glyphs, etc for Source Sans Pro?

jpyuda:
Old-style numbers have ascenders and descenders much like lowercase letters. They were traditionally used in running copy, while using lining numerals for tables. (when you want all the numbers to line up)

http://theworldsgreatestbook.com/wp-content/uploads/2011/11/figureStyles1.gif illustrates it well.

miguelsousa:
@maya 
> I think we can probably create a new issue about removing SVG fonts.

I can take care of that, if you want.

> What do you recomend as the best place to view all available characters, glyphs, etc

[This PDF](https://github.com/miguelsousa/source-sans-pro-subset/blob/master/sourcesanspro-regular-webfont.pdf) is probably the best place. Keep in mind that this is showing the subsetted fonts that were submitted in the PR and not the original fonts. The original fonts have a lot more glyphs and support a lot more languages. If you come across characters that are not supported by the current subsetted fonts, let me know and I'll generate new subset fonts with broader support.

I will also try to remember to update the fonts here by submitting a PR whenever a new version of [Source Sans Pro](https://github.com/adobe-fonts/source-sans-pro/releases) comes out. I've created a [dedicated repository](https://github.com/miguelsousa/source-sans-pro-subset) to make that easy and reproducible.

PR #851, opened 2015-10-21
Update color contrast guidance - issue #757

### LINKED ISSUE #757: contrast of disabled form input
on https://playbook.cio.gov/designstandards/form-controls/ the accessibility guidelines state : 
"If you adjust the color scheme of the buttons, ensure a minimum contrast ratio of 4.5:1 (for small text, 3:1 for large) for all states of the button. This includes default, hover, selected, and disabled."

WCAG 2.0 doesn't require minimum contrast ratio for disabled interface component because otherwise user can't know it's disabled cf : WCAG 2.0 1.4.3 criteria "Incidental: Text or images of text that are part of an inactive user interface component, that are pure decoration, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement."

maya:
This updates color contrast guidance to remove disabled from our list in our accessibility guidelines as items that must meet AA color contrast. 

Resolves #757.

cc @nickbristow


nickbristow:
Good to go

PR #858, opened 2015-10-22
Link to Sass

fureigh:
Link the word "Sass" to its project page for consistency with Bourbon and the rest.


fureigh:
Same as #856 but with regard to the correct branch. :)

PR #861, opened 2015-10-23
Remove SVG fonts

miguelsousa:
SVG fonts are only needed for [**very old** browsers](http://caniuse.com/#feat=svg-fonts). [WOFF](http://caniuse.com/#feat=woff), and soon [WOFF2](http://caniuse.com/#feat=woff2), are more broadly supported and have a smaller footprint.


maya:
lgtm :+1:

maya:
@miguelsousa I was just chatting with another team member - did you consider the support for SVG fonts on mobile browsers, i.e. non-chrome android browsers?

miguelsousa:
@maya Mobile browsers support support at least on of these formats: [TTF](http://caniuse.com/#feat=ttf), WOFF, WOFF2.
iOS 4.2 added support for TTFs; until then SVGs were required. TTFs have been supported by Android since version 2.2.
I'm quite confident that removing the SVGs is safe.

PR #862, opened 2015-10-23
Update text about Source Sans Pro's origin and license

miguelsousa:


brittag:
Tiny copy suggestions: Should "The Source Sans Pro font files" and "The Merriweather fonts" have matching phrasing (either "font files" or "fonts")? Does it make sense to simplify "a custom subset" to just "a subset"?

miguelsousa:
Good suggestions. I'll make those changes.

maya:
@brittag in all of our other text we use the format: "The files in `/path/to/file` ..."

Is it ok to deviate from that in this instance?

brittag:
@maya I'd say it's fine! The goal of that phrasing is just to help readers identify which files have which licenses. Maybe "The Source Sans Pro font files" instead of "The Source Sans Pro fonts" would be more consistent with the rest of the list.

maya:
Yes, I like that @brittag.

maya:
@miguelsousa could you update the text from "fonts" to "font files" so it's consistent with the rest of the list?

So like this:

"The Source Sans Pro font files..."

"The Merriweather font files..."

miguelsousa:
@maya sure. I also gave some more thought to the removal of the word "custom" and I think it should be in there, because the font files were actually modified from the original versions. The original fonts have more characters than the version that is now part of this project, and this project does not include all the styles that are part of the original Source Sans Pro.
@brittag and @maya what do you think about using, "The Source Sans Pro font files in `assets/fonts` are a customized subset of..."

brittag:
Sure, that makes sense to me!

miguelsousa:
OK, all the changes we've discussed have been made.

maya:
:+1:

Issue #866, opened 2015-10-28
Alerts should use color variables


maya:
The alerts are not using our color variables and they should be.

h/t @msecret

colinpmacarthur:
:+1:

PR #867, opened 2015-10-28
Change alert colors to use variables - issue #866

maya:
This changes the alert background colors to use our variables. It uses the lightest version of each color. You can see below that the colors are slightly different.

This is what it looks like now (using variables):

screen shot 2015-10-27 at 8 44 04 pm

This is what it looked like before (not using variables):

screen shot 2015-10-27 at 8 44 12 pm

Resolves: #866


mollieru:
@maya lgtm!

PR #871, opened 2015-10-29
[ website ]: Remove older mobile meta tags

andrewhughey:
I removed the older mobile meta tags form head.html and head-home.html. They specifically target Palm, Blackberry, and PocketPC devices. Seems pretty unnecessary these days. The viewport tag is all we need.

### DIFF
--- a/_includes/head-home.html
+++ b/_includes/head-home.html
@@ -2,12 +2,7 @@
 ================================================== -->
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
-
-<!-- Mobile Specific Metas
-================================================== -->
-  <meta name="HandheldFriendly" content="True">
-  <meta name="MobileOptimized" content="320">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
 
 <!-- Title and meta description
 ================================================== -->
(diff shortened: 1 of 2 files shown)


jpyuda:
Not sure we want to remove these. Blackberries are still heavily used among Congress and gov't employees, so seems a worthwhile thing to retain. Palm is not used much by anybody but as far as I know, that `meta` tag is not harmful to any other devices so I don't mind the slightly-higher download size to better support the odd Palm user.

andrewhughey:
I dunno. Including these tags just feels like front-end voodoo. 

These tags pre-date responsive design and are **vendor specific** (Avantgo/Palm browsers and Microsoft) that saw some limited and **unknown adoption** in other mobile browsers. I hate to be including them just for completeness sake, especially if we aren't targeting or testing on those devices. Firing up a couple BB emulators just gets me a blank screen for the site; what's it look like on real devices?

A few references:

* [The viewport metatag](http://davidbcalhoun.com/2010/viewport-metatag/)
* [Layout Meta Tag](https://msdn.microsoft.com/en-us/library/bb431690.aspx): MSDN article from 4/2010 written for Windows mobile 6.5

colinpmacarthur:
We need to a bit more research into what we need to do to support government Blackberries, but will come back to this when we figure that out. Thanks!

rogeruiz:
Closed in light of #1175

PR #875, opened 2015-11-01
Removed references to nonexistent _site directory.

alex:


maya:
Thanks @alex!
 
The `_site` directory is jekyll related and appears when you bring the site down locally and run it. The reason we had this in our documentation like this was bc we compile our Sass to CSS using jekyll, but since we are manually taking a copy of `main.css` from `_site/assets/css` and putting it in `assets/css` it seems like we should change this so it's clearer and more consistent.

alex:
@maya interesting, perhaps that section should indicate you need to run jekyl first?

maya:
@alex I'm going to close this out for now in favor of opening issue #885.

alex:
:+1:

PR #881, opened 2015-11-11
grammatical fix for an vs a on word sound

garrypolley:
I believe should use "a" instead of "an" because the pronunciation of the word usable.


maya:
Correct, thanks @garrypolley!

PR #890, opened 2015-11-18
adding a section to list examples of re-use

gbinal:
It's very helpful to be able to point to examples of these being implemented.  I'd suggest keeping that list here.


jenniferthibault:
As someone who wants to use the standards on a few projects, I have also looked for (and asked someone about the existence of!) this type of list so that I can see what other people are doing with the various elements.

maya:
Resolved via https://github.com/18F/web-design-standards/issues/895.

PR #894, opened 2015-11-20
Update "Getting Started" to refer to zip file

### LINKED ISSUE #885: Rethink developer documentation reference to _site/assets
We are guiding people to download assets from the `_site` directory although people need to run the `./go serve` command (which runs jekyll) to get this locally. The reason we're doing this is bc you have to run jekyll in order to get compiled CSS. The assets folder includes everything except compiled CSS. 

Also we currently have all assets as a zip file, so should we update the instructions to have them get it from there?

References #875.

maya:
This changes the "Getting Started" section to instruct people to download the assets ZIP file.

This resolves #885.

Note: we're linking to the specific version, so we'll need to remember to update this on each release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,9 @@ Include the Web Design Standards CSS and JavaScript files in each HTML page in y
 
 Refer to these files by adding a `<link>` and a `<script>` element into your HTML pages and also include the font font:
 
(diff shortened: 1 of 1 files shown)


vgvg:
Much more clear. :+1:

PR #900, opened 2015-11-30
Change name to draft standards

maya:
This changes the name from U.S. Web Design Standards to Draft U.S. Web Design Standards.

This is what the homepage will look like:

screen shot 2015-12-02 at 2 40 53 pm

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -15,14 +15,6 @@ exclude:
 
 permalink: pretty
 
-repos:
-- name: U.S. Web Design Standards
-  description: Main repository for the U.S. Web Design Standards.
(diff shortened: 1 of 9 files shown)


_config.yml
```diff
@@ -1,7 +1,7 @@
 # Markdown config
 markdown: redcarpet
 highlighter: rouge
-name: U.S. Web Design Standards
+name: U.S. Web Design Draft Standards
```
maya: Do we need to change this below in "repos" as well on line 18?

maya:
Update: change to Draft U.S. Web Design Standards

PR #902, opened 2015-12-01
Add package.json for npm support

### LINKED ISSUE #767: Make NPM installable.

jeremiak:
Added the `package.json` so that the library can be installed through `npm`, which should close #767. I included `node-sass` as a dependency, which is run after the library is installed as a dependency for another project. This enables the CSS to be compiled using a Javascript dependency instead of requiring Ruby/Jekyll.

The difference number in the `package.json` is bumped to `0.8.3` because using release tags to install from Github is really easy in `npm`. Since there is already a packaged up `0.8.2` release, the `0.8.3` (or higher) release will enable users to be able to install by adding something like the following to their `dependencies`:
`"uswds": "git@github.com:18F/web-design-standards.git#v0.8.3"`

It can also be published to the npm registry at some point to enable something like `$ npm install uswds`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 _site/
 .sass-cache/
+node_modules/
 .DS_store
 .*.swp
--- a/README.md
+++ b/README.md
@@ -32,6 +32,18 @@ Refer to these files by adding a `<link>` and a `<script>` element into your HTM
 <script src="/path/to/your/assets/js/components.js"></script>
 ```
 
+### Install with NPM
+
+If you have `node` installed on your machine, you can use `npm` to install the Web Design Standards. In your `package.json` list the following under `dependencies`:
+
+```json
+  "uswds": "git@github.com:18F/web-design-standards.git#v0.9.0"
+```
+
+On subsequent runs of `npm install`, the package will be installed in `node_modules` under `uswds`. The CSS will be compiled after it is installed, and the final CSS file will be named `uswds.css`. You should be able to find it in `node_modules/uswds/assets/css/uswds.css`.
+
+*Note:* You might get an [`npm` warning related to `lodash`](https://github.com/18F/web-design-standards/pull/902#issuecomment-161076213), but you can generally ignore it.
+
 ## Setup for your local environment
 
 ### Requirements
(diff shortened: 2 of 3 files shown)


juliaelman:
Looks good and thanks for the contribution @jeremiak! A few things I am noting:

- Please add yourself to the contributors list in the `package.json` file! :smile: 
- Please add documentation for users to follow on how this new method of compiling works. I think the following file would be a great place to add it: https://github.com/18F/web-design-standards/#getting-started
- I received the following warning after running `npm install`:

```
npm WARN deprecated lodash@1.0.2: lodash@<3.0.0 is no longer maintained. Upgrade to lodash@^3.0.0.
```

Noting here for reference more than anything, but wanted to just note it for now. I don't see this as a blocker for merging in. Maybe add `lodash 3.10.1` to the list to avoid any confusion to users and/or this error form appearing. 
- This pull request bumps the release to 0.8.3. Not entirely sure what categorizes a release bump and look to @maya for her thoughts on this.

maya:
Thanks for reviewing @juliaelman! Following semantic versioning, the last number ("3" in 0.8.3) is for patches and middle number ("8" in 0.8.3) is for features. 

I think this would be considered a new feature, thus bumping it to 0.9.0. What do you think?

juliaelman:
@maya agreed on being a `v0.9.0` release, but not sure we'd want to jump to that without hitting some of other bugs first. Might be good to create a release branch that this merges into instead of merging into `18f-page-staging` //cc @jeremiak 

@jpyuda what are some other priority bug fixes that you think need to be completed before the `v0.9.0` feature release?

jeremiak:
Thanks for the review/feedback @juliaelman. I've gone ahead and added myself as a contributor and updated the getting started section of the README. I put in the `v0.9.0` tag so I think this PR should not be merged until that release is pretty much imminent.

maya:
@juliaelman Minor versions may also include patch level changes, so this does not need to wait on anything else. 

@jeremiak since we may have other changes coming in, maybe we can hold off on adding the tag so it includes the most recent changes. Typically I'll add release tags right before pushing to production. Do we also need to update the package.json to `0.9.0`?

juliaelman:
@maya @jeremiak I went ahead and made that update to `package.json` :+1: 

I also created a v0.9.x branch and merged these changes locally into that branch, rather than have Jeremia submit a new pull request. I'll be adding a tag for the `v0.9.x` release and will look to the core team to help build out a road map to this release. 

I've also created the following tag to point any updates to that are targeted for this release: https://github.com/18F/web-design-standards/releases/tag/v0.9

In regards to closing #767, I think this is part of the solution and that we'll want to add this as an installable package on npm before officially closing.

Closing this pull request and thanks again @jeremiak for your contribution! :tada:

jeremiak:
Wonderful, thanks @juliaelman!!

Issue #571, opened 2015-09-16
[Password Reset] Space between confirm password input box and button should be the same as the login component.


angel:
![image](https://cloud.githubusercontent.com/assets/103496/9894743/2cd9e626-5bdc-11e5-8a65-dffd71796316.png)
vs. 
![image](https://cloud.githubusercontent.com/assets/103496/9894748/375f6cce-5bdc-11e5-97a2-b87b07cf5825.png)

juliaelman:
Issue resolved via #903, see screenshots below for verification.

![screen shot 2016-05-20 at 12 07 20 pm](https://cloud.githubusercontent.com/assets/955558/15434140/7f193440-1e83-11e6-873d-91af04b7f0ab.png)

![screen shot 2016-05-20 at 12 07 48 pm](https://cloud.githubusercontent.com/assets/955558/15434139/7f1575b2-1e83-11e6-99be-0f956b2624bc.png)

PR #903, opened 2015-12-02
Added the same margin-bottom on password inputs

helenvholmes:
Went with personal preference of adding a button margin to all inputs since it was a decision, but removing the ```margin-bottom``` on all inputs would be as easy as deleting line 30 in ```assets/_scss/elements/_inputs.scss```.

### DIFF
--- a/assets/_scss/components/_forms.scss
+++ b/assets/_scss/components/_forms.scss
@@ -23,7 +23,7 @@ form {
     margin-bottom: 1.5em;
   }
 
-  input[name="password"] {
+  input[name="password"], input[name="confirmPassword"] {
     margin-bottom: 1.1rem;
   }
 }


maya:
Thanks for bringing this up @helenvholmes!

The issue #571 you referenced is about the space between the input field and button, but it looks like this PR is about the space between the input and the "Show password" link.

I noticed on forms which wrap inputs and labels in a `div` we get a margin that is too large, so I think we should hold off on applying that style. This way the input margin will be consistent in different use cases.
screen shot 2015-12-01 at 9 26 53 pm

Changing `input[name="password"]` to `input[type="password"]` will get the margins consistent on the Reset Password and Sign In form. However, it seems it would be safer to apply a top margin to the "Show password" link instead.

helenvholmes:
Noted! Is it still worth making the change to the top margin on the "Show password" link to make those consistent? I can go ahead and do that.

maya:
Sure, it doesn't hurt! :smile:

helenvholmes:
You'll see this in the commit, but I ended up adding the margin-bottom to ```name[confirmPassword]``` instead of a top-margin to the "show password" link to stay consistent what's already been done :blush:

assets/_scss/components/_forms.scss
```diff
@@ -23,7 +23,7 @@ form {
     margin-bottom: 1.5em;
   }
 
-  input[name="password"] {
+  input[name="password"], input[name="confirmPassword"] {
```
msecret: First off, does this work with just the attribute selector?

```
[name="password", [name="confirmPassword"]
```

If it doesn't work, then having the tag in there as well is fine, but was hoping we could move to having just attribute selectors across the site to fix specificity problems.

Also, does adding bottom margin to these two password inputs work in different configurations of form elements? Such as when you could have password inputs before another form element?

Generally, I think it would be more successful if all these inputs used top margins and adjacent sibling selector, so we could do something more like this:

```
form > * + *
```

but we're a while away from that so if the two things I bring up are looked at, I'm OK with this change in the mean time.

maya: @msecret would you want `[name="password", [name="confirmPassword"]` nested inside `form`?

maya: > Also, does adding bottom margin to these two password inputs work in different configurations of form elements? Such as when you could have password inputs before another form element?
Yes, I just tried and since we have a default `margin-top: 3rem` applied to labels which go before inputs and since margins collapse it works fine. @msecret 

I think that a better method in the long run would be adding `margin-top` to the show password link. But we can just add a note about this to change later.

msecret: @maya it probably doesn't need to be nested.

maya: @msecret it may be good to keep it in a form to ensure people markup forms correctly. We can always update the specificity issue later down the line, so going to go ahead with this for now.

maya:
@helenvholmes thanks for the help! :tada: :dolphin:

PR #907, opened 2015-12-03
Update tables to use a table header in each row

### LINKED ISSUE #905: accessible html tables
I work for FCIC/OCSIT as an Accessible developer/designer an noticed a few issues in your responsive tables. 

The “scope” attribute on the “td” element is obsolete. Use the “scope” attribute on a “th” element instead. All <th> elements should generally always have a scope attribute. While screen readers may correctly guess whether a header is a column header or a row header based on the table layout, assigning a scope makes this unambiguous.
```<tbody>
<tr>
<td scope="row"> Declaration of Independence</td> [this TD attribute should be a TH scope="row"]
<td>Statement adopted by the Continental Congress declaring independence from the British Empire.</td>```


Below is an example that we us for tables.
example: http://promotions.usa.gov/platform/register-vote.html

maya:
This changes the first row item in `tbody` to `th` since scope on a `td` is being deprecated in HTML5.

Resolves #905.

### DIFF
--- a/_elements/tables.md
+++ b/_elements/tables.md
@@ -19,22 +19,22 @@ lead: Tables show tabular data in columns and rows.
     </thead>
     <tbody>
       <tr>
-        <td scope='row'>Declaration of Independence</td>
+        <th scope='row'>Declaration of Independence</th>
         <td>Statement adopted by the Continental Congress declaring independence from the British Empire.</td>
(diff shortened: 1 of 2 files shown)


maya:
This got the :+1: from @nickbristow.

nickbristow:
🚢

Issue #840, opened 2015-10-19
Make clearer about "Accessibility out of the box"


nickbristow:
I'm a bit concerned that the "Accessibility out of the box" could be a bit confusing. 

Specifically that there is work to be done to make it accessible, they have to follow the guidance. 

Should we update the paragraph below to reference this?

Just a thought.

msecret:
Yea I've thought about this. What's making the framework accessible is mostly the HTML code snippets which include correct HTML attributes. There's also some custom JS code for tabbing and ensuring the HTML is set up so tabbing works down the page. If a developer changes the HTML, which they potentially could, then accessibility might not work.

maya:
Yeah I think that sounds like a good idea. Feel free to draft something up and we'll take a look @nickbristow.

@msecret I think that point is something we can add into either the Getting Started or individually per component in the docs.

coreycaitlin:
Okay, it sounds like the key things to address here are:

- "Everything you need" implies that they won't have to worry about it at all, which isn't true.
- The great thing about the standards is that at least people won't be swimming upstream to make sites accessible — it's about starting with an accessible framework, rather than putting it off until the very last step.

And it would be nice to make the heading a gentle imperative phrase, so it's more parallel with the other three call-outs.

Because this is such a small slice of text, I'm putting it here for discussion, and will make the change and issue a PR if it looks good. Here's what I've got so far:

> Start with accessibility

> These guidelines focus on accessibility at every step of the design process, setting you on the path to building a usable, accessible site. From colors to code, these tools help you meet high standards of accessibility and conform to Section 508 Standards.

Let me know if this solves things, or if you have any questions or notes.

maya:
@coreycaitlin this is really great, thanks! 

What do you think @nickbristow @msecret @mollieru?

nickbristow:
I like it @maya

mollieru:
Cool! Good ideas all! 

One thought: can we come up w another way of saying 'Start with accessibility'?  The goals of this are two fold: that we put a premium on accessibility from the get go, and that we wanted to it easier for teams to build accessible sites.  'Start with...' doesn't quite capture this for me. Gonna noodle on some idea...

Also...devil's advocate: have we heard that this is in fact confusing to people? or are we surmising?  I'm down to change it for the sake of clarity but just wondering what the feedback has been...

coreycaitlin:
[...]

PR #909, opened 2015-12-04
Updated explanation in accessibility callout

coreycaitlin:
Clarified the language under "Accessibility out of the box" to clarify, per #840: the standards help set folks on the right path to accessibility, but don't necessarily guarantee accessibility — people will need to continue being aware of accessibility as they change the code or create content.

I left the header as-is, because it's pretty clear. We can always revisit to make it a verb phrase.


mollieru:
looks good!

PR #913, opened 2015-12-07
Remove presentation classes from grid code samples

### LINKED ISSUE #906: Styleguide presentation classes in code samples
In some cases, classes only intended for presentation of the styleguide are being shown in the code sample markup. This is bc our code samples are live and automatically generated by the actual markup. This should be removed so people can directly copy and paste the code samples without getting these extraneous classes.

One place where this occurs is in the grids. You'll see classes `usa-grid-example` and `usa-grid-example`.

screen shot 2015-12-03 at 10 55 46 am

One reason why this happens is bc we have two differently styled examples in a single preview box. So we can't target these wthout adding a class inside our around the markup.

Ideally, this is what users would see in the code sample:

```html
  <div class="usa-grid">
    <div class="usa-width-one-whole">1/1</div>
  </div>
  <div class="usa-grid">
    <div class="usa-width-one-half">1/2</div>
    <div class="usa-width-one-half">1/2</div>
  </div>
  <div class="usa-grid">
    <div class="usa-width-one-third">1/3</div>
    <div class="usa-width-one-third">1/3</div>
    <div class="usa-width-one-third">1/3</div>
  </div>
```

maya:
Opening this PR to discuss a possible solution for #906. This adds a new code sample box below the first grid example in order to remove extraneous classes. I needed to add a class surrounding the grid and the only way to do this was to add it on a new preview box.

This is what it will look like...

Closed:

download

Opened: 

download-1

Resolves #906.


colinpmacarthur:
This makes sense to me and should probably be our approach moving forward, but @carodew should also chime in.

carodew:
I agree with removing the presentation classes. I'm not entirely sure why we need two code boxes on the page, or why we would need to present both grids – one with the fraction widths as text and the other with the example text. Wouldn't one code snippet be enough? And probably the one with fraction widths be easier to manipulate?

colinpmacarthur:
I'm not totally sure, but I think that the two code snippets are so that we can display the code for each example without displaying the surrounding "presentation" markup. If they're the same example, whatever headers or other elements separate them are also included in the markup. Thoughts @maya?

@mollieru might remember more about the intention of the two examples. Maybe so one can see how the grid handles div with different amounts of text?

maya:
>  I'm not entirely sure why we need two code boxes on the page -@carodew 

This was a solution for the issue we had. The question is, why is there presentation classes in the first place? The answer is bc the two examples are styled differently. So by breaking it out into two boxes, I can add the class on the code sample box which will be hidden from the user.

An alternative to keep them in the same box would be to write the code manually, i.e. not auto-generating the snippet like in all the other examples.

maya:
Why there's two... I think the first example is more symbolic and the second is meant to show how it would look in the real world, with text. I'm sure @mollieru can say more.

carodew:
I understand why there are two examples, I'm not sure why each one needs a code snippet.

maya:
@carodew I think it's just consistent to how we've been doing it, that each live example has a corresponding code snippet to go along with it.

maya:
This got the :+1: from team. Will open a new PR against `v0.9.x` branch.

Issue #916, opened 2015-12-08
Gray hex value used instead of color variables


maya:
We use a color `#757575` in a few instances for hint text and outlines of inputs (radio buttons and checkboxes). `#757575` is not a color variable we offer, but is the lightest gray that can be used which passes color contrast requirements. The closest variable would be `$color-gray` which is slightly darker.

Should we use one of our color variables or add a new variable for this gray?

Here is what they look like side-by-side:

<img width="340" alt="screen shot 2015-12-07 at 4 27 17 pm" src="https://cloud.githubusercontent.com/assets/5249443/11644333/e4037b96-9cff-11e5-9ba5-23e68ec59801.png">

mollieru:
These shades are different enough that I'm wondering if we should add the #757575 to the color palette.  Rationale: looking at the two grays in context, #757575 is light enough that when next to our black, it recedes to support a bit more distinguishable visual hierarchy.  See for example the difference below:

![image](https://cloud.githubusercontent.com/assets/3680494/11787889/1fa2a7b0-a25c-11e5-96bb-bf324ce8431a.png) 
![image](https://cloud.githubusercontent.com/assets/3680494/11787896/25fae014-a25c-11e5-97be-0650edf6f952.png)

But I'm curious what other folks think about this -- @jenniferthibault @jehlers @rtwell @jjoteal ?

mollieru:
If so, thinking we might name is `gray-medium` -- thoughts/ideas on this welcome, too.

jenniferthibault:
I'd be curious to know why you then need the darker grey. Could #757575 replace #5b616b ? I don't know your use cases for #5b616b , but I see the benefit of #757575 with black in the screenshots above.

Drakenhart:
Would it be for designer flexibility@mollieru?
This project has a nice, wide range of color choices. I could see it being made and added color just for designer choice.

jenniferthibault:
An alternative! If you do need multiple greys....

Your existing grey #5b616b is a cool blue grey, while #757575 is a grey with much more red in it, which is why they feel like they are mismatched when put next to each other. To keep it more in-palette, try a lighter blue-based grey to match your ##5b616b.

Here's what I mean.
Current:
![screen shot 2015-12-14 at 12 40 20 pm](https://cloud.githubusercontent.com/assets/11636908/11788597/e376b688-a25f-11e5-86e2-a3b41ae30df4.png)

[...]

PR #917, opened 2015-12-08
Refactor Sass in elements

maya:
This refactors Sass for files in `assets/_scss/elements` and follows 18F Front-end coding styleguide: https://pages.18f.gov/frontend/.

This includes removing nested properties (in favor for writing them out), alphabetizing, correct nesting order, etc...

**TODO**
- [x] change color hexcode to color variable - #916


assets/_scss/core/_utilities.scss
```diff
@@ -88,7 +88,7 @@
     @include media($medium-screen) {
       float: left;
       margin-right: 2em;
-      margin-top: .5em;
+      margin-top: 0.5em;
```
msecret: Why are these em rather then `rem`?

assets/_scss/elements/_inputs.scss
```diff
@@ -147,19 +141,19 @@ input[type="checkbox"] + label::before,
...
+  box-shadow: 0 0 0 1px $color-gray-medium;
   content: '\a0';
   display: inline-block;
   height: 1.8rem;
-  line-height: .8;
-  margin-right: .6em;
-  text-indent: .15em;
-  vertical-align: .2em;
+  line-height: 0.8;
+  margin-right: 0.6em;
```
msecret: There's a mix of `em` vs `rem` units in this file and I'm not sure if its' intentional.

assets/_scss/core/_utilities.scss
maya: This PR was mostly about consistent formatting and didn't look at `em` vs `rem`.

assets/_scss/elements/_inputs.scss
maya: The main goal of this PR was consistent formatting. I didn't look into deeply into rem vs em. In this instance, it could be that the actual checkboxes are rem bc it's more defined, but the margins are ems as it's more related to spacing in proportion to text.

msecret:
I think an issue should be opened to check the codebase for consistency with em/rem units.

PR #921, opened 2015-12-10
Refactor Sass in components

maya:
This refactors Sass for files in `assets/_scss/components` and follows 18F Front-end coding styleguide: https://pages.18f.gov/frontend/.

This includes removing nested properties, alphabetizing, correct nesting order, etc...

### DIFF
--- a/assets/_scss/components/_alerts.scss
+++ b/assets/_scss/components/_alerts.scss
@@ -1,7 +1,4 @@
 .usa-alert {
-  @include media($medium-screen) {
-    background-size: 5.2rem;
-  }
   background-color: $color-gray-lightest;  
   background-position: 1rem 2rem;
   background-repeat: no-repeat;
@@ -9,11 +6,13 @@
   margin-top: 1.5em;
   padding: 1em;
 
+  @include media($medium-screen) {
+    background-size: 5.2rem;
+  }
+
   ul {
-    margin: {
-      bottom: 0;
-      top: 1em;
-    }
+    margin-bottom: 0;
+    margin-top: 1em;
   }
 }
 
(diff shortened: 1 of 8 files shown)


assets/_scss/components/_alerts.scss
```diff
@@ -9,11 +6,13 @@
   margin-top: 1.5em;
   padding: 1em;
 
+  @include media($medium-screen) {
+    background-size: 5.2rem;
```
msecret: Why is an `rem` unit being used here and an `em` down there? Should the `ul` margin top be changed based on a parent font size?

assets/_scss/components/_disclaimer.scss
```diff
@@ -1,23 +1,23 @@
...
+
+  @include media($small-screen) {
+    font-size: $h5-font-size;
   }
 }
 
 // This hides the stage note (i.e. alpha, beta, etc) of the disclaimer for 
 // anything smaller than large screen sizes
 .usa-disclaimer-stage {
+  display: none;
```
msecret: Does it need to appear to screen readers regardless of size? I'm actually not sure what a mobile screen reader would publish its size as.

maya: A screen reader on a desktop would read it and on a mobile device would not. I don't know what the priority of the stage text is for screen reader users.

assets/_scss/components/_alerts.scss
maya: This `rem` unit is for the icon image, where as the rest applies for text. I don't think there's a need to have a huge icon image if there's larger text. Correct, in this case the margin top would change depending on the parent text size.

PR #922, opened 2015-12-11
Remove copyright year and symbol from licenses and attribution

### LINKED ISSUE #920: Update licenses and attributions in 2016
Just a reminder to update any licenses/attributions to include year 2016 in January [here](https://github.com/18F/web-design-standards#a-few-parts-of-this-project-are-not-in-the-public-domain)

maya:
This removes the copyright year and © symbol as it is not legally required. See http://www.copyright.gov/circs/circ03.pdf for more information.

Resolves #920.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,31 +48,31 @@ Questions or need help with setup? Feel free to open an issue here [https://gith
 
 ### A few parts of this project are not in the public domain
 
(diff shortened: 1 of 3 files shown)


vzvenyach:
Just curious, is there a design benefit to removing it?

maya:
@vzvenyach yes, it prevents us from having to update it each year. Since it appears in multiple places, this will save a lot of work. Also, it will take time for products to update it, so it would require us to check to see if they updated it. It may be strange to have one thing with 2015 and another with 2016 bc it wasn't updated by them.

vzvenyach:
:+1:

PR #924, opened 2015-12-11
Alphabetize Sass includes

maya:
This moves Sass `@include`s to `assets/css/main.scss` and removes `assets/_scss/all.scss`. It also alphabetizes `@include`s in each import block.

TODO
- [x] update npm so it gets Sass includes from the right location


msecret:
So was the purpose of the all.scss file to make it so you don't have to repeat that large list of imports in 3 places?

maya:
I think so, bc now I see it and it's a pretty long list. Do you think it's valuable to keep for that reason? I just don't like the idea of an all import for `main.scss` bc I want to see everything that's being imported and want to get rid of an extra step, but maybe for the other places it makes sense?

msecret:
I think it does make sense to only have one list. I'm not sure it should be called "all" though, it should be called something like "web_design_standards".

maya:
@msecret I think `web_design_standards ` is too verbose and we often change our name. Ideally I think we'd use `wds` or `uswds`. On the other, you can still use these includes and it's still using the web design standards even if you don't use all of them. So in a way, `all` is semantically accurate. 

If we bring back the `all` file, how about we just add the `all` imports to styleguide and and home. I like listing them all in `main.scss` for the reasons in my previous comment.

msecret:
We have to consider how an outside user would be using this. If they wanted to import the sass for the library, wouldn't they be importing from this file? If that's the case, we should name it something related to the standards, like `wds` or `uswds` so it's easy to understand where the import is coming from.

For instance, if I download bootstrap, then I get a bootstrap.less file with I require as `"bootstrap"`.

It sounds like we don't need the main file. Whats the purpose of keeping that around?

maya:
I agree that a shorter name like `wds` or `uswds` is the way to go, but I'd hold on this until our name is more defined or we have a chance to discuss it with the team. Currently, the project is called the Draft U.S. Web Design Standards, so technically it would be `duswds`. Keeping what we have, at least for now, will make it more resilient to another name change.

We have `main.scss` inside `assets/css` because it compiles out to `main.css` in that location.

maya:
@msecret fixed merge conflict above. FYI we must keep variables first in the import order of Core as some files use variables and need them defined.

maya:
@msecret I changed the imports again to how it was before, until we have time to assess this further. So this PR only includes alphabetizing import sections (excluding core bc it requires a specific order).

PR #925, opened 2015-12-11
Update change log

maya:
This updates the change log to include changes for the next release `v0.9.0`.

TODO:
- [x] add all changes made before release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## [0.9.0] - 2015-12-??
+## [0.9.0] - 2015-12-21
 ### Added
 - Add package.json for npm support
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Remove copyright year and sign from license and attributions
 
 ### Fixed
+- Add the same margin bottom on all password inputs
 - Refactor Sass in Core, Elements, and Components
 - Fix design assets zip file name
 - Fix tables by using table headers for accessibility
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Change explanation in accessibility callout
 - Change name of project to draft standards
 
+### Removed
+- Remove copyright year and sign from license and attributions
+
 ### Fixed
-- Refactor Sass in Core
+- Refactor Sass in Core, Elements, and Components
 - Fix design assets zip file name
 - Fix tables by using table headers for accessibility
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add package.json for npm support
 
 ### Changed
+- Alphabetize Sass `@include`s
 - Change explanation in accessibility callout
 - Change name of project to draft standards
 
 ### Removed
+- Remove postinstall script, users will have to compile Sass on their own
 - Remove copyright year and sign from license and attributions
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.9.0] - 2015-12-??
+### Added
+- Add package.json for npm support
+
+### Changed
+- Change explanation in accessibility callout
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.9.0] - 2015-12-21
+### Added
+- Add package.json for npm support
```
msecret: @jeremiak did we completely figure out npm support? Like could I currently do npm install and `@import 'wds'` and it'll work?

jeremiak: Well, we're using it on Federalist but we run the build task on our side and then we move that CSS around. See [scripts here](https://github.com/18F/federalist/blob/master/package.json#L86)

I'm curious, what would you expect if you were to `import 'wds'`?

msecret: yea I guess its two different use cases.

`import "wds";` would import the scss file which you then compile. There's also the use case of like `require('wds')` using webpack, which could be the CSS. We'd potentially have a different npm module for the Sass files, I've seen this done before. 

For when we're publishing this to npm, I don't think you should have to build it to get the CSS, which is what federalist is doing now. You should npm install it, and the CSS should be compiled somehwere, and then I think there's even something like a "main-css" directive you can put in package.json to point to it.

Also, what is the plan to publish this on npm?

jeremiak: I like the approach that Bootstrap uses:

* There will be a `css` and a `scss` key in the `package.json` that each point to a file. The `main` key in the `package.json` will point to the main javascript for the library. This means that if you `require('uswds')` you will get the main JavaScript file, not the CSS.
* To get to the `css` and `scss` keys programatically one could `require('uswds/package.json').css` or `require('uswds/package.json').scss`

I don't think we should commit compiled code to Github, instead using `npm prepublish` scripts and maybe even having Travis build and publish when appropriate.

maya: @jeremiak: @alexose mentioned liking having pre-built files in the repo so people can clone and get started right away. He also mentioned D3 as a good example that does this. cc @msecret

mbland:
Since I've scrubbed in to try to help @maya with this, I've run into some questions for @jeremiak and @msecret.

So when we talk about "npm support", we're talking about being able to write browser-only JavaScript that can `require('uswds')`, ostensibly after browserifying it? We're _not_ talking about being able to `require('uswds')` from an arbitrary Node program or the Node REPL and having it work without errors? For example:

```js
// npm.js
// Compiled using: browserify npm.js -o index.js
'use strict';
require('uswds');
```

And then in the HTML file:
```html
<script src="./assets/js/vendor/jquery-1.11.3.min.js"></script>
<script src="./index.js"></script>
```

This seems to behave in my browser; whereas `require('uswds')` in the Node REPL dies horribly.

And these programmatic `require('uswds/package.json').css` or `require('uswds/package.json').scss` directives aren't actually standardized yet, but proposed standards, correct? I saw that [bootstrap uses the `less` and `style` tags](http://getbootstrap.com/getting-started/#download-npm), but gives no guidance on how they're used. Then I found that the [`parcelify` npm](https://www.npmjs.com/package/parcelify) supposedly uses `style`, but again, that requires the user knowing to use it. Also:

- http://stackoverflow.com/questions/32037150/style-field-in-package-json
- This hints that some npms exist to overload the `@import` directive: http://techwraith.com/your-css-needs-a-dependency-graph-too/, namely:
  - [resin](https://www.npmjs.com/package/resin)
  - [rework-npm](https://www.npmjs.com/package/rework-npm)

Point being, if we add these keys, we'll have to document how to use them. So basically, the `package.json` needs updated to something like this, plus some documentation, and we're good to go:

```json
  "main": "assets/js/components.js",
  "scripts": {
    "build-sass": "node-sass assets/_scss/all.scss > assets/css/uswds.css",
    "prepublish": "npm run build-sass build"
  },
  "assets": "assets/",
  "css": "assets/css/uswds.css",
  "scss": "assets/_scss/all.scss",
```

jeremiak:
Cool, I'll take a look at the PR @mbland but also wanted to answer your question. In my opinion the benefit of having the library on NPM is that it can be a declared dependency to ease installation and deployment. If you are `var wds = require('uswds')`-ing then I would assume that `wds` is an object from the `index.js` file in that repo and be the JavaScript components of the library.

As far as those other keys, they're definitely not yet standards with regards to their semantic meaning or place in `package.json`. But :+1: for shipping it with documentation, as it enables flexibility for any arbitrary build tool chain.

As an example, [on Federalist](https://github.com/18F/federalist/blob/master/package.json#L86-L92) we use the `build-sass` script to compile the CSS.

mbland:
No, I get the benefit of npm packaging. I was checking whether my understanding of how it would be used is (now) correct, that it would be `require`d by browser scripts only. This after pulling a fair amount of hair out trying to get it to load without throwing a stack trace in the REPL.

PR #931, opened 2015-12-16
Remove presentation classes from grid code samples

### LINKED ISSUE #906: Styleguide presentation classes in code samples

maya:
This is the same PR as #913. This adds a new code sample box below the first grid example in order to remove extraneous classes.

Resolves #906.


assets/css/styleguide.scss
```diff
@@ -442,24 +444,28 @@ $site-top:           124px;
...
-  background: $color-grid-dark;
+.grid-text {
+  .usa-grid {
+    background: $color-grid-dark;
 
-  > * {
-  background: $color-grid-light;
-  color: #000;
-  }
+    > * {
```
msecret: This is hard to see in gh, but shouldn't the lines below be indented in one more?

maya: @msecret I just checked.. it's indented with the background property which is one level in from `.usa-grid` so I think that's correct:

screen shot 2015-12-18 at 10 55 48 am

maya: @msecret Ooops, looked at the wrong line. Fixing....

maya: Fixed in 5ccaffe.

PR #940, opened 2015-12-23
Fix homepage image from being too large in IE

### LINKED ISSUE #863: example images on landing page are too large, messes w/ page width
can I just start by saying that I *love* these standards? such rad work all around.

anyways, to the issue - there's a lotta whitespace to the right of the content, due to the size of the images in the examples down the page

screen shot 2015-10-23 at 2 11 34 pm

maya:
Found the issue of images inside of links being too large on the homepage, which was found and fixed for Firefox in #863, also exists for IE.

Even though the container the image is in has a defined width, it does not obey its width to fill the container. Instead, it exceeds that width and appears at it's full size.

This fix adds a max-width on the image link. Safari and Chrome will fill the width of the container it's in. Firefox and IE will only do this for direct parent containers. Since this is a grandparent container, it does not obey it.

This screenshot shows this fixes the issue in IE (tested in IE9 and IE10):

screen shot 2015-12-22 at 5 01 18 pm

Fixes #863.


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Remove copyright year and sign from license and attributions
 
 ### Fixed
+- Fix homepage image links that were too large in Internet Explorer [styleguide-only]
```
juliaelman: We should be specific on the version of Internet Explorer that was having an issue. Please add which versions of IE this is effected by and solves for.

juliaelman:
Looks good! Added note above about IE versioning, but otherwise looks good to me.

CHANGELOG.md
maya: @juliaelman I didn't put the IE version bc it effects all versions.

juliaelman: @maya hrmmm...  all versions meaning IE[X] - ? only the one's we support? Looks like you mentioned testing only on IE9 - 10.

maya: @juliaelman Yes all versions of IE, I explicitly tested IE9-11 and assuming it effects below IE9 as well.

maya: @juliaelman actually, just checked and it doesn't effect IE Edge. I can explicitly state IE11 and below.

maya: Updated in 0ed253d.

juliaelman: @maya I still think we need to be explicit and state IE9 - 11, as we have not tested other versions of IE outside of those and don't want to make any assumptions.

maya: @juliaelman fixed in c9cdb5f.

juliaelman:
@maya :+1: merge away!

PR #942, opened 2015-12-27
Fix list bullets on older systems

skonzem:
The list bullet character ("•") in this file is non-ASCII. On older systems that are not set to a UTF-8 locale, this character will be rendered strangely.

This appears to be the only .scss file with non-ASCII characters that does not already have a charset declaration. The others are all in Bourbon or Neat. (grep -rlP '[^\x00-\x7f]' . | grep scss | xargs head -n 1)

### DIFF
--- a/src/stylesheets/elements/_list.scss
+++ b/src/stylesheets/elements/_list.scss
@@ -31,7 +31,7 @@ ul li {
   display: table-row;
 
   &:before {
-    content: "•";
+    content: "\2022";
     display: table-cell;
     padding-right: 0.4em;
   }


maya:
@skonzem thanks so much for opening this PR! We'd definitely love bullets to appear correctly as widely as possible.

Quick question: what was the older system you found that this issue was happening in?

skonzem:
Just an older Linux box set to ISO-8859-1.

skonzem:
To be clear, this is a problem that happens during the "jekyll build" phase. According to the SASS Reference (http://sass-lang.com/documentation/file.SASS_REFERENCE.html#encodings):

"Sass follows the CSS spec to determine the encoding of a stylesheet, and falls back to the Ruby string encoding."

My guess is that I'm falling back to Ruby's string encoding, which defaults to my system locale.

maya:
@skonzem sorry for the delay! Can you describe how you were ingesting the files into your project?

Thanks!

skonzem:
Nothing too fancy. I copied the assets directory into a new Jekyll site without modification.

maya:
@skonzem thanks for the background!

I chatted with @rogeruiz and we think it may be good to declare this globally. We're debating whether to add it to `_base.scss` or `_all.scss`. Adding to `_all.scss` is good if someone intends to use it as is, and unchanged, and it adds it to the top of the file as the first thing. Adding it to `_base.scss` is good, if someone wants to use something other than all the imports. For example, if someone wants to use just the accordion, we'd advise them to also include everything in Core which includes `_base.scss` plus the accordion SCSS.

Thoughts @msecret?

skonzem:
@maya I think this setting may be per-file. It seems to affect how the file gets read into Jekyll. I just tried reverting this change and adding it to `assets/_scss/all.scss` and `assets/_scss/components/_base.scss`, but I got the same problem as before.

maya:
@skonzem but the ultimate file you're using in your project is a single CSS file (compiled from `assets/_scss/all.scss`), correct?

maya:
Actually, it's `assets/css/main.scss` that imports all the files from `assets/_scss/all.scss` into `_site/assets/css/main.css`.

rogeruiz:
@skonzem what version of Ruby are you using?

skonzem:
@maya Yes, I'm actually importing `main.css`. I just tried adding the line to `assets/css/main.scss`, and that also did not work.

skonzem:
At least in the Bourbon and Neat code included here, the `charset` declaration is everywhere, as shown by [this search](https://github.com/18F/web-design-standards/search?p=1&q=charset&utf8=%E2%9C%93). Perhaps that is the way to declare it universally?

@rogeruiz Ruby 2 something (meaning after a lot of the unicode changes). Forgive me for not being more specific.

msecret:
I don't necessarily see anything wrong with adding it to every component and element file, besides the labor in doing so, as @skonzem mentioned.

maya:
@msecret should we accept this PR to resolve the only known issue, or wait until we add it to all?

skonzem:
I'd be happy to amend this PR or submit a follow-up.

skonzem:
@maya @msecret I notice there's now a merge conflict due to the new directory structure. Since this PR needs revision anyway, I'm seeking guidance on whether to apply the change to the rest of the files while I'm at it.

I'm also not 100% clear how to run the bleeding edge version in another site. Would I just copy over the src directory, as I used to copy over the assets directory? In the readme, I only see instructions for using the release copies of the assets, and for running the Design Standards site.

rogeruiz:
Hey @skonzem, Since #978 and #980 have landed on `18f-pages-staging`, we will need to rebase this branch onto `18f-pages-staging`. Once you've done that, the `src` directory contains the files that were once in `assets`. You can add the `charset` line to all of the source files. Thank you!

The project now uses `gulp` to automate various tasks around bundling and compiling if you want the compiled distribution files you can follow these steps.

```shell
cd path/to/repo
npm install
gem install bundle scss_lint
gulp # which will give you a list of available commands
gulp build # which will build the `dist` and `dist-gem` directories.
```

There will be more documentation coming soon as it's on our roadmap for the next release.

tysongach:
No harm in declaring the `@charset` (keep in mind it has to be _first_ element in the stylesheet, per: https://developer.mozilla.org/en-US/docs/Web/CSS/@charset)…but instead we could (should?) just escape that special character:

```diff
- content: "•";
+ content: "\00b7";
```

I usually use this handy table to find the entity I’m looking for: https://brajeshwar.github.io/entities/

maya:
Thanks @tysongach! I like that idea too, seems like the simplest fix. That way we can add `@charset` later on when we decide to. @skonzem can you check if escaping the special character works for you?

skonzem:
@tysongach @maya It's actually the bullet character, \2022, but yes that approach works. Adding the charset still seems like a good idea given that the developers are using modern computers. The advantage of leaving the character as-is is that it's more readable.

Gulp may be a challenge for my antiquated machine. I could submit an untested PR, or we could file this as an issue and let someone else take over.

rogeruiz:
@skonzem If you rebase this branch, I can test and merge this PR for you so you don't have to pull down and run gulp. Let's use the \2022 entity instead of the character for now. Since it's only one character. We can revisit this if and when we introduce more characters.

skonzem:
@rogeruiz Thanks for volunteering to test! I've updated the PR to the \2022 entity.

rogeruiz:
Took a look at this on my local machine, and the escaped entity gets converted to a bullet character because of my `utf-8` system. So having the original entity there works as expected. Thanks for your contribution @skonzem!

PR #948, opened 2016-01-11
Enable npm packaging, including .css and .scss

mbland:
Per the discussion on #925. Still needs documentation regarding how to use the `css` and `scss` keys. For reference, here's some source material for the forthcoming documentation:

- http://stackoverflow.com/questions/32037150/style-field-in-package-json
- http://techwraith.com/your-css-needs-a-dependency-graph-too/
- https://www.npmjs.com/package/rework-npm
- https://www.npmjs.com/package/resin
- https://www.npmjs.com/package/parcelify

cc: @maya @msecret @jeremiak

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,10 @@
   "main": "assets/js/components.js",
   "scripts": {
     "build-sass": "node-sass assets/_scss/all.scss > assets/css/uswds.css",
-    "prepublish": "npm run build-sass build"
+    "prepublish": "npm run build-sass"
   },
   "assets": "assets/",
+  "style": "assets/css/uswds.css",
   "css": "assets/css/uswds.css",
   "scss": "assets/_scss/all.scss",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -2,13 +2,14 @@
   "name": "uswds",
   "version": "0.9.0",
   "description": "Open source UI components and visual style guide for U.S. government websites",
-  "main": "package.json",
+  "main": "assets/js/components.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


jeremiak:
:+1: 

I didn't know about `parcelify` before, nice find @mbland!

package.json
```diff
@@ -2,13 +2,14 @@
   "name": "uswds",
   "version": "0.9.0",
   "description": "Open source UI components and visual style guide for U.S. government websites",
-  "main": "package.json",
+  "main": "assets/js/components.js",
   "scripts": {
     "build-sass": "node-sass assets/_scss/all.scss > assets/css/uswds.css",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "prepublish": "npm run build-sass build"
```
msecret: Isn't it just supposed to be `npm run build-sass`?

msecret:
Why wouldn't we use a `style` property as well as `css` and `scss`?

package.json
mbland: Good catch. I'd had a separate "build" target previously that I'd since taken out. Done.

mbland:
@msecret Added a `style` key. PTAL.

maya:
Does this PR need to be opened against `v0.9.x` rather than `18f-pages-staging`?

msecret:
LGTM

maya:
Chatted with @msecret, we can cherry-pick the commits ourselves into `v0.9.x`. Thanks @mbland!

maya:
Added into our `v0.9.x` release branch: https://github.com/18F/web-design-standards/tree/v0.9.x - thanks @mbland!

PR #956, opened 2016-01-13
Remove frontmatter from assets zip file

### LINKED ISSUE #955: Downloadable assets main.scss file contains frontmatter
In the downloadable assets in `assets/css/main.scss`, file contains frontmatter from jekyll. This should be removed so users do not get this.

This is what's currently in the file:

```
---
# Front matter comment to ensure Jekyll properly reads the file.
---

// Place list of your partial imports below, located in assets/_scss.

@import 'all';
```

maya:
This removes Jekyll frontmatter from `assets/css/main.scss` in the downloadable assets zip file. 

Resolves #955.


msecret:
How do we ensure this doesn't happen again? Do we have instructions on how to generate the zip file that we could add this to?

maya:
@msecret good question. Since the `main.scss` file doesn't change, it should not happen again. When we get to automating the creation of the zip file, we'll need to instruct it to use just this.

PR #958, opened 2016-01-14
_footer.scss uses $medium instead of $medium-screen

BrianDiggs:
The current version of the file generates a final CSS snippet of 

```css
@media screen and (min-width: min-width 600px 6) {
  .width-one-sixth {
    float: left;
    display: block;
    margin-right: 2.35765%;
    width: 14.70196%; }
    .width-one-sixth:last-child {
      margin-right: 0; } }
```

The problematic part is the `min-width: min-width 600px 6`. Changing `$medium` to `$medium-screen` makes it `min-width: 600px`. I admit that I am new to SCSS programming, but that seems like the right thing by analogy to the other calls.

### DIFF
--- a/assets/_scss/components/_footer.scss
+++ b/assets/_scss/components/_footer.scss
@@ -56,7 +56,7 @@
 }
 
 .width-one-sixth {
-  @include media($medium) {
+  @include media($medium-screen) {
     @include span-columns(2);
   }  
 }


maya:
@BrianDiggs thanks so much for catching this! I've reviewed the section of `assets/_scss/components/_footer.scss` and can't figure out why that code block is there in the first place:

```sass
.width-one-sixth {
   @include media($medium) {
      @include span-columns(2);
    }  
  }
```

It appears to repeat what is currently in `assets/_scss/core/_grid.scss` (line 29) already and is not nested in any footer selector. For those reasons, I'd vote to remove that code, cc: @juliaelman @rogeruiz @msecret

juliaelman:
@maya you are correct! These lines can be removed and are already part of the grid.

@BrianDiggs thank you for your contribution and thoughts around making the Draft Web Design Standards even better! :rocket:

msecret:
I agree with removing teh code

maya:
@BrianDiggs closing this PR in place of #981. 

Thanks again for bringing this to our attention :tada:

PR #962, opened 2016-01-22
Update `npm` workflow instructions

rogeruiz:
These changes are related to the documentation surrounding the `npm` instructions. I've included instructions on how to deal with the use-case of wanting to work locally with the web-design-standards as a dependency in another project. I'm currently using this technique to ensure that the package is installable via npm and that the instructions for using it can be updated. The structure of the sample project is:

```
index.html, package.json, assets/start.js, assets/style.scss
```


rogeruiz:
Seeing some errors around dependencies. I'm going to fix these and that means adding [Browserify]() to the project in order to include other `npm_modules` along with updating the prepublish step to build the `components.js` file. Thoughts on how to best do this would be appreciated. Here's a glimpse of issues happening.

screen shot 2016-01-22 at 6 50 09 pm

rogeruiz:
I've since built out a little `npm`-friendly front-end starter kit that I think would be a good example of how this project would be ingested via `npm`. [You can see it here](https://github.com/rogeruiz/npm-front-end-starter). Those dependencies can be fixed on a few different ways, but I'll be opening a separate issue for that one.

This updated documentation is good to go, I think. If you check out that project I linked to above, you can follow the _Install with `npm-link`_ instructions in this PR and test what it would be like to require the `uswds` package into a project.

rogeruiz:
I've added some more documentation around using this package with `npm`. Because jQuery is a dependency of the code in `components.js`, I've included instructions that point users to use the included jquery file within the `vendors` directory. I've also added the `.npmignore` file that is needed since this project also contains the website and we don't necessarily want to publish that to `npm`.

README.md
```diff
@@ -34,21 +34,71 @@ Refer to these files by adding a `<link>` and a `<script>` element into your HTM
...
-  "uswds": "git@github.com:18F/web-design-standards.git#v0.9.0"
+```
+cd path/to/project-using-npm
+npm install --save uswds
 ```
 
-On subsequent runs of `npm install`, the package will be installed in `node_modules` under `uswds`. The CSS will be *not* compiled after it is installed. Instead, you are able to compile it however it fits into your project.
+This will add the Web Design Standards as a dependency for your project and save
+it to your `package.json` file. The package will be installed in `node_modules`
+under `uswds`. The CSS file(s) are generated on the `npm-prepublish` hook and can
```
msecret: Doesn't pre-publish mean that the task will be run before they even install it. Why would the user need to know this information?

README.md
```diff
@@ -34,21 +34,71 @@ Refer to these files by adding a `<link>` and a `<script>` element into your HTM
...
+found in the following directory: `./node_modules/uswds/assets/css`.
 
-An `npm` task called `build-sass` is supplied, though you are not required to use it, that will compile the Sass to a CSS file named `uswds.css` (in `node_modules/uswds/assets/css/uswds.css`).
+An `npm` task called `build-sass` is supplied, though you are not required to
```
msecret: Why would the user ever need to know this?

README.md
```diff
@@ -34,21 +34,71 @@ Refer to these files by adding a `<link>` and a `<script>` element into your HTM
...
+use it, that will compile the Sass to a CSS file named `uswds.css` (in
+`node_modules/uswds/assets/css/uswds.css`).
 
-If you simply want to compile the Sass you could do something like:
+Note: You might get an [`npm` warning related to `lodash`](https://github.com/18F/web-design-standards/pull/902#issuecomment-161076213), but you can generally ignore it.
```
msecret: I like this idea of saying this. It can be scary to get warnings, although they are common in node installs.

README.md
```diff
@@ -34,21 +34,71 @@ Refer to these files by adding a `<link>` and a `<script>` element into your HTM
...
+This error is related to a dependency found in `node-sass`.
 
-`cd node_modules/uswds && npm run build-sass`
+#### Install with `npm-link`
```
msecret: I think these instructions should be moved to a larger different section in the docs thats dedicated to developing the standards. This ensures theres no confusion about which install method to use when. If it's too much to restructure right now I'm fine with this for now.

README.md
rogeruiz: If the user of the standards would like to integrate the standards into their own asset-pipeline, they may want to know how it is generated. But I do agree that it might be too much for the readme and the project in it's current state. I'll be updated this section to fix this along with the `prepublish` comment above.

maya:
@rogeruiz can you add "Draft" in front of "Web Design Standards" in all the documentation you added?

README.md
rogeruiz: I've updated this to be in the `CONTRIBUTING` doc for now.

rogeruiz:
Thanks @maya. I've added _Draft_. I'll make sure to be aware of it going forward.

README.md
rogeruiz: I think this line was added to explain why the CSS file isn't in the repo. A user that clones the project won't know where the CSS file is since it's only generated when it's published to `npm`.

rogeruiz:
The packaged up assets directory is now better aligned with the downloads ZIP file that the website contains. I've updated npm to ignore the `styleguide.css` file along with the `releases/` directory. @msecret Could you take a look at this when you get a chance?

maya:
@rogeruiz did you also make sure to modify the `assets/main.scss` file so it removes frontmatter?

maya:
Update to above comment: @rogeruiz will remove that file (`assets/_scss/main.scss`) as `node-sass` compiles the Sass into `uswds.css` so it's not necessary.

rogeruiz:
@msecret @maya I've updated the documentation and the `npmignore` around the topic of what files should be included and how user's can use the library once it's installed via `npm install`. I think this should be good to publish to `npm` and closes #767 along with #948.

README.md
```diff
@@ -34,21 +34,56 @@ Refer to these files by adding a `<link>` and a `<script>` element into your HTM
...
+under `uswds`.
+
+Refer to these files by adding a `<link>` and a `<script>` element into your HTML pages:
 
-On subsequent runs of `npm install`, the package will be installed in `node_modules` under `uswds`. The CSS will be *not* compiled after it is installed. Instead, you are able to compile it however it fits into your project.
+```
+<link rel="stylesheet" href="./node_modules/uswds/assets/css/uswds.css">
+<link rel="stylesheet" href="./node_modules/uswds/assets/css/google-fonts.css">
+<!-- Include jQuery from the `uswds` project before `components.js` -->
+<script src="./node_modules/uswds/assets/js/vendor/jquery-1.11.3.min.js"></script>
```
msecret: Just thinking into the future:

eventually we'll want to move all JS in USWDS into one output file. If we suggest this now, how do we ensure people that followed these instructions don't include jquery twice (once with this script, and another time within the uswds.js file)?

README.md
```diff
@@ -34,21 +34,56 @@ Refer to these files by adding a `<link>` and a `<script>` element into your HTM
...
+This will add the Draft Web Design Standards as a dependency for your project and save
+it to your `package.json` file. The package will be installed in `node_modules`
+under `uswds`.
+
+Refer to these files by adding a `<link>` and a `<script>` element into your HTML pages:
 
-On subsequent runs of `npm install`, the package will be installed in `node_modules` under `uswds`. The CSS will be *not* compiled after it is installed. Instead, you are able to compile it however it fits into your project.
+```
+<link rel="stylesheet" href="./node_modules/uswds/assets/css/uswds.css">
+<link rel="stylesheet" href="./node_modules/uswds/assets/css/google-fonts.css">
```
msecret: Doesn't this file include 8 fonts? When we discussed fonts originally we decided to limit the default to 3 or 4 or 5 (can't actually remember what we agreed on) to limit performance problems. Should we not be recommending people to include all 8 fonts if they don't need them.

On my cloud.gov project where I imported USWDS, I ended up creating my own fonts file with just 3 `@font-face` links included.

README.md
rogeruiz: Yeah, I agree on the future thinking. I'm not quite sure the best way to recommend this. Above this line, I add the requirements about `jQuery` and `$` needing to be on the window object.

README.md
rogeruiz: Taking a look at the output of `uswds.css` there are no references to fonts.

maya: @msecret Doesn't the `@font-face` declaration just specify what fonts can be used and how to use them and doesn't add page weight until they are declared in CSS?

rogeruiz:
I've updated this PR to be rebased off of `18f-pages-staging` now with the latest production code at it's base. I've run through it as well, and aside from some updates in the README around the version number I think we are good on documentation at this step.

@msecret @juliaelman @maya I don't think that this is ready to review because of how closely it is tied to the discover / _design_ phase of [the other related card](https://trello.com/c/3J13uLIH/34-as-a-master-builder-i-want-to-be-able-to-know-what-dependencies-are-required-to-successfully-work-with-the-wds-and-i-want-the-op).

I'm going to close this PR for now and reopen it pointing to the appropriate branch, but the contents of the `README` here contains the direction for the npm installation instructions I'm going for. I need to split this apart a little more right now in different issues and update those two cards.

PR #969, opened 2016-01-28
Publishing `uswds` package to NPM

rogeruiz:
This series of patches updates the documentation around installing the `uswds` package via NPM. It include instructions about how to use the package _in it's current state_ with an `npm`-based asset pipeline and how to also use the package with static HTML files. There are instructions and caveats pointed out around dependencies.

This change does not need to be merged / released on to `18f-pages` / production because it doesn't affect the website. It must be merged onto `18f-pages-staging` as that is the default branch for the project. The documentation that user's will see when they navigate to our repository via `npmjs.org` will need to match what gets published to the `npm` registry.

The version number for the package will also need to be updated. I would consider this a patch version update because the are no changes to the actual code but rather the documentation. The minor version update should happen when the updates concerning splitting the repository and dependency management / documentation is merged in the project.

> https://trello.com/c/4MWOWON6/30-as-a-master-builder-i-want-to-be-able-to-install-the-web-design-standards-easily-into-my-project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@ node_modules/
 .DS_store
 .*.swp
 npm-debug.log
+
+# Ignore generated files
+#
+assets/css/uswds.css
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ _site/
 node_modules/
 .DS_store
 .*.swp
+npm-debug.log
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,16 @@
+_components/
+_elements/
+_includes/
+_layout-system/
+_layouts/
+_visual/
+assets-styleguide/
+pages/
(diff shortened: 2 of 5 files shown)


rogeruiz:
@msecret and I are in agreement that we will hold off on actually publishing the package to `npm` until #978 and #980 land on `18f-pages-staging`. We reviewed the code and docs yesterday and all should be good to go. This PR does need to be merged into `18f-pages-staging` before #980 can be merged in.

Could someone merge this in please? CC: @maya @juliaelman

juliaelman:
@rogeruiz :hand: happy to help!

rogeruiz:
@msecret This PR has been rebased onto `18f-pages-staging` to remove any of the conflicts post the refactor of #974.

PR #978, opened 2016-02-02
Reorganizing repository file/folder structure

juliaelman:
The work in this pull request reorganizes the files and folders to help assist with automation of the project. A few things to note:

- The Draft WDS website pages now live in the `/docs` folder.
- `_config.yml` has a new variable called `version` for us to be able to update the release version around the site more efficiently.
- The `Gemfile` now includes the the `web_design_standards_gem` instead of relating the the actual files in the repo: 

Fixes #974

PR #981, opened 2016-02-02
Remove unneeded code from footer

maya:
This removes unneeded code from the footer in `assets/_scss/components/_footer.scss`. 

This is instead of: https://github.com/18F/web-design-standards/pull/958

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -55,12 +55,6 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


rogeruiz:
Looks like this was fixed before #978 landed. We'll need to rebase the branch onto `18f-pages-staging` and resolve any code conflicts, @maya.

maya:
@rogeruiz rebased!

PR #984, opened 2016-02-04
Fix asset compilation for /docs + remove style folders

juliaelman:
This pull request fixes the issue of asset compilation for the new docs structure, as well as removes and moves files/folders into a cleaner and more accurate structure.


rogeruiz:
Looks good :+1:

PR #989, opened 2016-02-06
Reduce repetition by using Bourbon mixins

tysongach:
This PR introduces a few refinements to the stylesheets to reduce repetition by using some of Bourbon’s mixins. Some of these were already being used, but this expands the use of them.

1. The `margin()` and `padding()` mixins allow you to wrap up nested `padding` blocks. This now let’s you define repeated values only once, e.g. when parallel sides need the same value, like in this example:

    ```diff
    - padding: {
    -   bottom: $section-padding;
    -   top: $section-padding;
    - }
    + @include padding($section-padding null);
    ```

2. Similarly the `position()` mixin can reduce repetition and take multiple lines of CSS and wrap them into a one line, connected syntax:

    ```diff
    - position: absolute;
    - right: 0;
    - top: 0;
    + @include position(absolute, 0 0 null null);
    ```

    One nice feature of the `position()` (as well as `margin()` and `padding()`) is that it “unravels” shorthand syntax:

    ```diff
    - @include position(fixed, 0 0 0 0);
    + @include position(fixed, 0);
    ```

    …which saves you from having to write this in CSS:

    ```css
    position: fixed;
    top: 0;
    right: 0;
    bottom: 0;
    left: 0;
    ```

### DIFF
--- a/docs/assets/css/homepage.scss
+++ b/docs/assets/css/homepage.scss
@@ -30,10 +30,7 @@ p {
 }
 
 .usa-section {
-  padding: {
-    bottom: $section-padding;
-    top: $section-padding;
-  }
+  @include padding($section-padding null);
 }
 
 .usa-button-block {
@@ -55,10 +52,7 @@ p {
   background-color: $color-primary-darkest;
   background-image: none;
   color: $color-white;
-  padding: {
-    bottom: 3.7rem;
-    top: 2rem;
-  }
+  @include padding(2rem null 3.7rem);
 
   .usa-label {
     display: table;
@@ -198,10 +192,7 @@ a {
 
 .usa-example-heading {
   font-weight: $font-normal;
-  margin: {
-    bottom: 1.6rem;
-    top: 0;
-  }
+  @include margin(0 null 1.6rem);
 }
 
 .usa-section-examples {
@@ -212,7 +203,7 @@ a {
   .usa-example-link {
     margin-top: 5rem;
     max-width: 980px;
-    
+
     &:first-of-type {
       margin-top: 0;
     }
--- a/docs/assets/css/styleguide.scss
+++ b/docs/assets/css/styleguide.scss
@@ -15,33 +15,27 @@ $site-top:           124px;
 // Navigation ------------- //
 
 .usa-site-header {
+  @include position(fixed, 0 null null 0);
   background-color: $color-white;
   border-bottom: 1px solid $color-gray-light;
-  position: fixed;
-  left: 0;
-  top: 0;
   width: 100%;
   z-index: 1;
-  
+
   a {
     border-bottom: none;
   }
 
   .usa-site-navbar {
     @include media($medium-screen + $width-nav-sidebar) {
+      @include margin(2.7rem 0 1rem);
       height: 6rem;
-      margin: {
-        bottom: 1rem;
-        top: 2.7rem;
-      }
     }
     margin-top: 4px;
 
     .logo {
       @include media($medium-screen + $width-nav-sidebar) {
-        padding-left: $site-margins;
-        text-align: left; 
-        padding-top: 0;
+        @include padding(0 null null $site-margins);
+        text-align: left;
         width: auto;
       }
       float: left;
@@ -77,9 +71,9 @@ $site-top:           124px;
       float: right;
       display: none;
       margin-top: -5px;
-      
-      li { 
-        display: inline; 
+
+      li {
+        display: inline;
         font-family: $font-sans;
 
         &:last-child .usa-button {
@@ -96,7 +90,7 @@ $site-top:           124px;
     font-size: $base-font-size / 1.25;
     border-bottom: 1px solid #616161;
     font-weight: 100;
-    .us-official { 
+    .us-official {
       margin: 0;
     }
     .stage {
@@ -106,11 +100,11 @@ $site-top:           124px;
       padding-top: 5px;
     }
     @include media($medium-screen + $width-nav-sidebar) {
-      .us-official { 
-        margin-left: 177px; 
+      .us-official {
+        margin-left: 177px;
       }
-      .stage { 
-        float: right; 
+      .stage {
+        float: right;
         padding: 0;
       }
     }
@@ -122,11 +116,10 @@ $site-top:           124px;
   @include media($medium-screen + $width-nav-sidebar) {
     display: none;
   }
+  @include padding(1.5rem null);
   display: inline;
   float: left;
   margin-top: -4px;
-  padding-top: 1.5rem;
-  padding-bottom: 1.5rem;
   color: $color-white;
   background-color: $color-primary;
   font-size: 1.5rem;
@@ -138,7 +131,7 @@ $site-top:           124px;
     color: $color-white;
     background-color: $color-primary-darker;
   }
-  
+
   &:visited {
     color: $color-white;
   }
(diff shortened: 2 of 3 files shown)


rogeruiz:
Hey @tysongach Thanks for this. I wonder if there are any other areas we can leverage Bourbon more.

tysongach:
@rogeruiz Of course! I’m the maintainer of Bourbon and I’m more than happy to help with that. I figured I’d start small and work iteratively. We’ve also been working on the next major release of Bourbon which will have some deprecations I can help with, and also a few neat new features.

maya:
I know we sometimes favor readability over terseness here at 18F. I wonder what would be the advantage of using the mixin over shorthand?

`@include padding(20vh null 10px 3em);`
vs
`padding: 20vh 0 10px 3em;`

cc @msecret

tysongach:
@maya Great question! And one we get often. The key difference is the ability to pass `null` in the Bourbon mixin.

Taking the CSS you posted above…

```css
padding: 20vh 0 10px 3em;
```

…`padding-right` is being set to `0` because using CSS shorthand sets _all_ values. This can cause “[Accidental CSS Resets](https://css-tricks.com/accidental-css-resets/)”.

When utilizing Bourbon’s mixin, though…

```scss
@include padding(20vh null 10px 3em);
```

…we’re able to use `null`, which compiles to longhand CSS and “skips” that side:

```css

rogeruiz:
That's a nice feature of Bourbon's mixins to include passing `null` params to prevent resetting margin / padding. 

@tysongach Are the new features anywhere we can review them?

tysongach:
@rogeruiz 5.0 is on [`master`](https://github.com/thoughtbot/bourbon) as of now (here’s a [comparison of 4.2.6 to 5.0.0](https://github.com/thoughtbot/bourbon/compare/v4.2.6...master)).

I’m actually hoping to get the changelog written up this week and release the beta.

The overall theme for 5.0 is clean up and to have a more focused library. We are deprecating all of our prefixing features (we’re encouraging use of Autoprefixer), which I think will affect this project. New features will be light, but include a [`contrast-switch`](https://github.com/thoughtbot/bourbon/commit/379e8c068faa9911fb4016570ff9286a1484ef66) function and a [`hide-visually`](https://github.com/thoughtbot/bourbon/commit/98f9473f444d4148113ff9cb5e7a531a8c8f250d) mixin.

maya:
@tysongach good thing we never used Bourbon's prefixing! :) We kept our eye on "things that will be deprecated" and didn't use them.

>a hide-visually mixin

Interesting, we have an [`sr-only`](https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/core/_utilities.scss#L25) "screen reader-only" mixin as well.

tysongach:
@maya Watch out for rendering implications with your `sr-only()` mixin. The off-screen `position` approach causes the box to be drawn to the size specified, in this `999em`, which can decrease performance. We chose to utilize the `clip` method, for this reason.

Regarding the prefixing mixins: I think a few of them may have made their way into the code :grin::

- [`_skipnav.scss`](https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/components/_skipnav.scss#L2)
- [`_skipnav.scss`](https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/components/_skipnav.scss#L12)
- [`styleguide.scss`](https://github.com/18F/web-design-standards/blob/18f-pages-staging/docs/assets/css/styleguide.scss#L153)
- Maybe more…but a couple quick searches returned these.

rogeruiz:
This is great! I think this is good to merge in. Thanks @tysongach. Let's
great an issue around updating Bourbon and revisit its use throughout the
codebase. We can keep track of the conversation there.

On Tue, Feb 9, 2016 at 11:46 AM, Tyson Gach <notifications@github.com>
wrote:

> @maya <https://github.com/maya> Watch out for rendering implications with
> your sr-only() mixin. The off-screen position approach causes the box to
> be drawn to the size specified, in this 999em, which can decrease
> performance. We chose to utilize the clip method, for this reason.
>
> Regarding the prefixing mixins: I think a few of them may have made their
> way into the code [image: :grin:]:
>
>    - _skipnav.scss
>    <https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/components/_skipnav.scss#L2>
>    - _skipnav.scss
>    <https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/components/_skipnav.scss#L12>
>    - styleguide.scss
>    <https://github.com/18F/web-design-standards/blob/18f-pages-staging/docs/assets/css/styleguide.scss#L153>
>    - Maybe more…but a couple quick searches returned these.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/web-design-standards/pull/989#issuecomment-181951168>
> .
>



-- 
Roger Steve Ruiz
804-539-2255
Innovation Specialist @ 18F
General Services Administration
1800 F St. NW, Washington, DC 20405

tysongach:
@rogeruiz That sounds like a plan! I’m here to help!

:tada:

msecret:
@tysongach just to be clear, the updates in this PR don't require Bourbon be updated, that's for other features?

tysongach:
@msecret Correct.

tysongach:
Bourbon `5.0.0-beta.1` has been released. You can find a changelog here: https://github.com/thoughtbot/bourbon/blob/master/NEWS.md

maya:
:clap: on the release log!

Oh ohs, I was thinking of the [Prefixer mixin](http://bourbon.io/docs/#prefixer) when you mentioned them being deprecated. I wasn't thinking of the transform and transition mixins bc it didn't mention them being deprecated on the site. That's good to know!

Reading the release log, I noticed, "The `em` and `rem` mixins have been deprecated." Those are _definitely_ [in the standards](https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/core/_defaults.scss#L7): So we'll have to make a note to change this.

If I remember correctly, we tried the clip method first but ran into some issues that I can't remember right now. Good ole positioning offscreen was more stable. I'd be curious to know the performance hit it takes.

tysongach:
@maya Thanks!

We are leaving the actual [`prefixer` mixin](https://github.com/thoughtbot/bourbon/blob/master/core/bourbon/addons/_prefixer.scss) in as part of `5.0`, but all of the individual mixins that were wrapped with the `prefixer` (`transition`, `appearance`, etc.) are no more. So, in case anyone wants to make their own, they still can. Might be useful for a particular bleeding-edge prefix before Autoprefixer supports it.

`em` and `rem` have indeed been deprecated, and I imagine we’ll get some heat around that. But simply put: they lied a lot when used throughout stylesheets. There is no way for those mixins to know the true font size (or the true font size of `html` for `rem`) of the parent where it is being used. Therefore—as an example—`em(20)` only outputs a value that is equal to 20px _if_ its parent’s font size is 16px (because we have to hard code 16 as the base in the mixin). And the risk of its parent having a different font size is too high, which leads to `em(20)` not actually necessarily being equal to 20px. I also believe you _should_  have the freedom to change font sizes as you need, and not accidentally create unintended affects to the codebase.

I noticed something similar is already being discussed in #992. For now, it’d be easy enough to copy these two small mixins over as custom mixins in this repo, and just stop relying on Bourbon for them if you do update to 5.0.

PR #993, opened 2016-02-10
Fix a typo

pra85:
Remove extra `and`


rogeruiz:
Good catch @pra85! Thank you for this PR. It looks good to merge in.

PR #997, opened 2016-02-16
Update main property to point to src/start;

rogeruiz:
This patch modifies the `main` property in order to fix issues with
installing this package via `npm`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "uswds",
   "version": "0.8.3",
   "description": "Open source UI components and visual style guide for U.S. government websites",
-  "main": "dist/js/uswds.js",
+  "main": "src/js/start.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


rogeruiz:
I was able to test this in a new project and had everything run correctly via just using `require('uswds')`. Looks like `browserify` was choking on the relative paths in the require when the `dist/js/uswds.js` file was being consumed.

msecret:
Could this also mean that if a project that imported the standards did a `require('jquery')`, browserify would be smart enough to only include it once (if the versions matched up)? Either way this is fine.

rogeruiz:
Yeah, I believe so @msecret. I'm going to do more testing on this and post here `||` submit a patch. I'm not sure which version would take precedence. I'll report back here about whether we'll need documentation to clear this up.

PR #999, opened 2016-02-16
Move vendor libs into npm dependency management system

rogeruiz:
- [x] Force version `1.11.3` of jQuery
- [x] Add @rogeruiz to contributors

Trello card associated with this PR => https://trello.com/c/3J13uLIH/34-as-a-master-builder-i-want-to-be-able-to-know-what-dependencies-are-required-to-successfully-work-with-the-wds-and-i-want-the-op


msecret:
We're not support IE8 anymore, is there a good reason to stay on jQuery 1.x?

*jQuery 2.x has the same API as jQuery 1.x, but does not support Internet Explorer 6, 7, or 8*

msecret:
Also, how are we supposed to maintain that contributor list? What about the public contributors?

rogeruiz:
@msecret Yeah, I'm not sure about the contributor's list. We could add a file. This list is specifically for the `npmjs.org` registry.

As far as jQuery version number, I'm willing to update it to be the latest or something else. I made it `1.11.3` because of what is included in the `docs/assets/js/vendor/` directory.

maya:
Perhaps we chose jQuery 1.x before we settled on IE9 and up for our browser support, which came a bit later.

msecret:
@rogeruiz I'm going to work off this branch to do https://trello.com/c/3J13uLIH/34-as-a-master-builder-i-want-to-be-able-to-know-what-dependencies-are-required-to-successfully-work-with-the-wds-and-i-want-the-op.

I also realized something: we changed our `main` in package.json to point to the src JS, which may means those JS files are now dependencies, not dev dependencies. Or is that wrong?

msecret:
So if the standards was on npm, you'd npm install, and the `require('uswds')`. That require points to `src/js/start.js`, which is unbuilt code that does `require('jquery')`. So in that case, isn't the user of the library using code that wasn't built yet?

rogeruiz:
Hmm. That is true. Since they would need to require it themselves and it would lead to errors. So anything that is JS related will need to be a `dependencies` item.

rogeruiz:
@msecret Running into the following error when I try using `node-sass` on the command line.

```sh
> npm-front-end-starter@1.0.0 build-css /Users/rogerruiz/Developer/npm-front-end-starter
> cat assets/style.scss | node-sass --output-style compressed > dist.css

{
  "formatted": "Error: File to import not found or unreadable: ../../node_modules/bourbon/app/assets/stylesheets/bourbon\n       Parent style sheet: /Users/rogerruiz/Developer/npm-front-end-starter/node_modules/uswds/src/stylesheets/all.scss\n        on line 3 of node_modules/uswds/src/stylesheets/all.scss\n>> @import '../../node_modules/bourbon/app/assets/stylesheets/bourbon';\n   ^\n",
  "message": "File to import not found or unreadable: ../../node_modules/bourbon/app/assets/stylesheets/bourbon\nParent style sheet: /Users/rogerruiz/Developer/npm-front-end-starter/node_modules/uswds/src/stylesheets/all.scss",
  "column": 1,
  "line": 3,
  "file": "/Users/rogerruiz/Developer/npm-front-end-starter/node_modules/uswds/src/stylesheets/all.scss",
  "status": 1
}
```

rogeruiz:
That error above looks like it's related to the file paths being relative to where I run the `node-sass` command.

msecret:
@rogeruiz does the gulp command work?

rogeruiz:
@msecret Well this is within another project that doesn't have `gulp` support. Maybe I'm not testing this properly? I was expecting to be able to include the `uswds` package after installing it using the following code in my `scss` file.

```scss
@import 'node_modules/uswds/src/stylesheets/all';
```

msecret:
@rogeruiz I see what you mean now, you're in a project that's consuming the standards. I have a local project too that doesn't have gulp support, maybe I'll take a look.

rogeruiz:
@msecret Cool. The project I'm using is this one https://github.com/rogeruiz/npm-front-end-starter

msecret:
@rogeruiz I did get it to work by changing the node-sass command to this:

```
node-sass --output-style compressed assets/style.scss dist.css
```

msecret:
I think it is a path problem otherwise. If you use `cat assets/style.scss | node-sass --output-style compressed > dist.css"` then node-sass won't know that you're coming from the `assets` directory. I also did it the cat way and changed the @import to `./node_modules/uswds...` and it worked.

rogeruiz:
I'm still seeing an error here

```sh
> npm-front-end-starter@1.0.1 build /Users/rogerruiz/Developer/npm-front-end-starter
> npm run build-css && npm run build-js


> npm-front-end-starter@1.0.1 build-css /Users/rogerruiz/Developer/npm-front-end-starter
> node-sass --output-style compressed assets/style.scss dist.css

{
  "formatted": "Error: File to import not found or unreadable: ../../node_modules/bourbon/app/assets/stylesheets/bourbon\n       Parent style sheet: /Users/rogerruiz/Developer/npm-front-end-starter/node_modules/uswds/src/stylesheets/all.scss\n        on line 3 of node_modules/uswds/src/stylesheets/all.scss\n>> @import '../../node_modules/bourbon/app/assets/stylesheets/bourbon';\n   ^\n",
  "message": "File to import not found or unreadable: ../../node_modules/bourbon/app/assets/stylesheets/bourbon\nParent style sheet: /Users/rogerruiz/Developer/npm-front-end-starter/node_modules/uswds/src/stylesheets/all.scss",
  "column": 1,
  "line": 3,
  "file": "/Users/rogerruiz/Developer/npm-front-end-starter/node_modules/uswds/src/stylesheets/all.scss",
  "status": 1
}
```

msecret:
@rogeruiz how are you including the wds repo? For instance, I'm using `npm link`ing and ensuring I did `npm install` and `gulp build` in the wds repo.

rogeruiz:
@msecret All I ran in the `npm-front-end-starter` repo was this command. I was downloading it from Github.

```sh
npm install --save git+https://github.com/18F/web-design-standards#update-package-json
```

I checked in the `node_modules/uswds/` directory to make sure it was installing the correct branch. I'm going to investigate this morning and I'll post anything I find here.

rogeruiz:
Seems like doing the above install, didn't run any of the hooks for the `uswds` package to install it's own dependencies. I'm going to look into the `npm` documentation to see why that would be. After going into the `node_modules/uswds/` directory and running an `npm install` I was able to get it to work properly.

```sh
> npm-front-end-starter@1.0.1 build /Users/rogerruiz/Developer/npm-front-end-starter
> npm run build-css && npm run build-js


> npm-front-end-starter@1.0.1 build-css /Users/rogerruiz/Developer/npm-front-end-starter
> node-sass --output-style compressed assets/style.scss dist.css

Rendering Complete, saving .css file...
Wrote CSS to /Users/rogerruiz/Developer/npm-front-end-starter/dist.css

> npm-front-end-starter@1.0.1 build-js /Users/rogerruiz/Developer/npm-front-end-starter
> browserify assets/start.js > dist.js
```

rogeruiz:
I did run into this issue as well, when I ran an `npm install` within the `node_modules/uswds/` directory.

```sh
~/Developer/npm-front-end-starter/node_modules/uswds master*
      ❯ gulp scss-lint                                                                                                                                                                                  ruby-2.2.3
[11:05:48] Using gulpfile ~/Developer/npm-front-end-starter/node_modules/uswds/gulpfile.js
[11:05:48] Starting 'scss-lint'...
[11:05:48] 'scss-lint' errored after 663 ms
[11:05:48] Error in plugin 'gulp-scss-lint'
Message:
    Input file did not exist or was not readable
```

:mag: Investigating this error too before we can :shipit:  :mag_right:

rogeruiz:
@msecret I figured it out. The problem is with the way `npm v2` and `npm v3` handle dependency nesting.

Read more here => https://docs.npmjs.com/how-npm-works/npm3

> While npm2 installs all dependencies in a nested way, npm3 tries to mitigate the deep trees and redundancy that such nesting causes. npm3 attempts this by installing some secondary dependencies (dependencies of dependencies) in a flat way, in the same directory as the primary dependency that requires it.

Here's a screenshot after an `npm install` and the resulting `node_modules/` directory.

screen shot 2016-03-03 at 11 30 49 am


:thought_balloon: My suggestion would be to force `npm v2` since we're looking to have _nested dependencies_ or maybe we actually need to use something like `eyeglass` in order to properly resolve these paths? :thought_balloon:

aredridel:
Use the `resolve` module instead of relying on the relative paths between modules. The assumptions of hard-coding node_modules paths are already broken in npm v2 in the face of `npm dedup`.

rogeruiz:
Thanks for the suggestion @aredridel. The relative path that we're trying to match is within a `scss` file though. Would `resolve` still work in this case?

aredridel:
Perhaps. It may take some wiring to connect up -- you have to tweak how your CSS module looks up files; some preprocessors have had this done. atomify-css was one approach to this, and others have wired it up with sass before I know.

rogeruiz:
Cool. Sounds _possible_ @aredridel at least.

@msecret looks like we'll have to do something about managing those sass dependencies in the `prepublish` hook.

rogeruiz:
Looks good. There is a small change that needs to happen for `npm install [...]` to work. Because of how the package gets built we need to remove `.scss-lint.yml` from the `.npmignore` file.

```diff
diff --git a/.npmignore b/.npmignore
index 2bc134f..e1008f5 100644
--- a/.npmignore
+++ b/.npmignore
@@ -13,7 +13,6 @@ dist-gem/
 .travis.yml
 _config.yml
 _config_18f_pages.yml
-.scss-lint.yml
 Gemfile
 Gemfile.lock
 go
```

rogeruiz:
Looks good to merge in based on the initial scope of just managing dependencies for `Sass`/`CSS`/`JavaScript`.

screen shot 2016-03-04 at 11 44 08 am

Though, we need to update the documentation around `fonts` and `images` in the Readme. There isn't a good way to reference them or copy them over using `npm`. We found a solution in the `ffd-microsite` around this but the specifics would be project based.

Examples using `gulp` for images and fonts:
- https://github.com/18F/ffd-microsite/commit/25a79d31f83f58d2c6037e149f8753beff3bd162#diff-b9e12334e9eafd8341a6107dd98510c9R122
- https://github.com/18F/ffd-microsite/commit/11796376721d92227e6bc97d1b3d9a24b40c5a05#diff-b9e12334e9eafd8341a6107dd98510c9R128

We could have some language in the `README` like:

> The compiled and bundled CSS expect to find `/fonts` and `/img` directories at the root level of the web server. For example: `http://localhost:8000/fonts/merriweather-bold-webfont.woff2` and `http://localhost:8000/img/minus.svg`

:point_up: This would be a great time to standardize on whether want the expectation of having `/fonts` and `/img` being at the root of the web server or having them prefixed with an `/assets` directory. User's bringing the `uswds` package won't have the ability to update these paths without editing the source. This issue was brought up here I believe. https://github.com/18F/web-design-standards/issues/1007

I don't think this can make it into this particular release for `v0.9.0` and can be something that we should iterate on in further releases.

:thought_balloon: @maya @juliaelman @msecret

rogeruiz:
Looks good to be merged in. Because a few PRs ( i.e. #1032 ) related to the release depend on these changes, I'm landing this in `18f-pages-staging`.

PR #1005, opened 2016-02-19
usa-width-five-sixths @ medium breakpoint typo

lboonebah:
@ medium breakboint the usa-width-five-sixths is set to the wrong number of columns. Documentation indicates it should be 4/6 not 5/6.


maya:
LGTM :+1: thanks @lboonebah!

PR #1012, opened 2016-02-26
Remove duplicate clearfix mixin

tysongach:
Bourbon already includes [a `clearfix` mixin](http://bourbon.io/docs/#clearfix) and this is a duplicate. Since these two mixins share the same name, it was confusing which one is actually being used (the latest imported one).

I recommend we remove this duplicate `clearfix` mixin and Bourbon’s will fill its place.

If this `clearfix` mixin being declared is solving a specific problem that Bourbon’s `clearfix` does not (browser support, maybe?), we should namespace this project one so to not having a name conflict.

### DIFF
--- a/src/stylesheets/core/_utilities.scss
+++ b/src/stylesheets/core/_utilities.scss
@@ -5,17 +5,6 @@
   }
 }
 
-// Clearfix helper
-@mixin clearfix() {
-  &:after {
-    content: ".";
-    display: block;
-    height: 0;
-    clear: both;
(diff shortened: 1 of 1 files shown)


msecret:
I don't know of this clearfix doing anything that the bourbon one doesn't do. @maya can you recall anything?

maya:
Not that I'm aware of @msecret. Just one clearfix mixin should be good.

msecret:
I did a quick manual visual test of the code and it looks godo

PR #1015, opened 2016-03-01
Nest list in README

maya:
This corrects a formatting issue on the nested list in the README cc @kategarklavs


kategarklavs:
Many thanks, @maya!

PR #1009, opened 2016-02-22
Update docs to save `uswds` from Github;

rogeruiz:
Fix #1003

Issue #1003, opened 2016-02-18
NPM uploads seem not to be working


edwelker:
Hi USWDS,

I tried to follow the instructions on `master` to install via NPM, but it would seem that the NPM registry is returns the following: 
`npm install --save uswds --registry https://registry.npmjs.org/`

```
npm ERR! node v0.10.24
npm ERR! npm  v3.7.2
npm ERR! code E404

npm ERR! 404 Registry returned 404 for GET on https://registry.npmjs.org/uswds
npm ERR! 404
npm ERR! 404  'uswds' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

and additionally:

https://registry.npmjs.org/uswds returns simply `{}`

maya:
Hi @edwelker, we haven't published it to npmjs.org yet. It will be published in our next `v0.10.0` release. cc @rogeruiz

msecret:
@edwelker, in the mean time you can install the standards through git with the following:

```
npm install --save git+https://git@github.com/18F/web-design-standards.git#18f-pages --registry https://registry.npmjs.org/
```

It should be noted that when we do publish to npm, the library will have a different structure, so some file paths may be different.

edwelker:
Thanks!!   Just wanted to point out that your documentation on master currently says it's available.  :)

PR #1022, opened 2016-03-02
Re-add update docs to save `uswds` from Github; #1009

maya:
Re-adds additions saved over from #1009

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,16 @@ And that’s it — you should be set to use the Standards.
 
 If you have `node` installed on your machine, you can use npm to install the Standards. Add `uswds` to your `package.json` as a dependency or a development dependency:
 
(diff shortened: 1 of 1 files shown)


rogeruiz:
Looks good to me. This is the current state of the `uswds` package since it isn't on the `npmjs.org` registry yet. Since people will be coming to this repo expecting to be able to `npm install ...` this repository, they'll need to know how to do it using the `github...` URL instead of the command that is currently in the repository.

:thought_balloon: @msecret, @juliaelman ? #1003 brings up the issue that `npm install --save[-dev] uswds` doesn't work. I don't believe that that issue would have been opened if the original `README` had had these instructions. These instructions were in `18f-pages-staging` when #1009 landed.

msecret:
LGTM

PR #1026, opened 2016-03-02
Update npm instructions in README

maya:
This updates the npm instructions so the command can run.


rogeruiz:
Looks good to me. Include the `'single-quotes'`will prevent globbing in the terminal :+1:

PR #1030, opened 2016-03-03
Small language update on README

maya:
This updates the language about using the Ruby gem cc @kategarklavs


kategarklavs:
Merging now — thanks @maya!

PR #1036, opened 2016-03-09
Update npm workflow

maya:
This updates the "Using npm" docs. This also removes the version number from the CSS and JS files, while keeping it in the parent directory and zip file. We recompiled the `dist/` directory.


README.md
```diff
@@ -78,6 +82,20 @@ found in the `src/` directory.
     - stylesheets/
 ```
 
+`require('uswds')` will load all of Draft U.S. Web Design Standard's JavaScript onto the page. The uswds module itself does not export anything.
```
kategarklavs: Hey! A few comments:

I'd add "the" before "Draft U.S. Web Design Standard's JavaScript." Also, I'd recommend capitalizing "uswds" in the second sentence (because USWDS is an abbreviation of a proper noun).

package.json
```diff
@@ -6,8 +6,6 @@
   "scripts": {
     "prepublish": "gulp no-gem build"
```
juliaelman: Should we remove the `no-gem` flag here?

README.md
```diff
@@ -78,6 +82,20 @@ found in the `src/` directory.
...
+
+The main Sass (SCSS) source file is here: 
+
+```
+node_modules/uswds/src/stylesheets/all.scss
+```
+
+The non-minified CSS that’s been precompiled is here:
```
juliaelman: Why not point them to the minified versions, as well as the minified/non-minified JS too?

README.md
```diff
@@ -42,13 +41,13 @@ into your HTML pages:
...
 ```html
-<link rel="stylesheet" href="/path/to/your/assets/css/uswds-0.9.0.css">
+<link rel="stylesheet" href="/path/to/your/assets/css/uswds.css">
 ```
 
 Add this before the closing `</body>` tag:
 
 ```html
-<script src="/path/to/your/assets/js/uswds-0.9.0.js"></script>
+<script src="/path/to/your/assets/js/uswds.js"></script>
```
juliaelman: Should we recommend using the minified version?

README.md
```diff
@@ -42,13 +41,13 @@ into your HTML pages:
 Add this to your `<head>` element:
 
 ```html
-<link rel="stylesheet" href="/path/to/your/assets/css/uswds-0.9.0.css">
+<link rel="stylesheet" href="/path/to/your/assets/css/uswds.css">
```
juliaelman: Should we recommend using the minified version?

msecret: So @rogeruiz yesterday explicitly wanted to have the version in the name. Is there a reason why this changed?

README.md
maya: @kategarklavs thanks! done in 8640af8.

README.md
msecret: Another comment here and for :50.

Above we're telling them their folder structure would be like `uswds-0.9.0/css/etc`. Should we include that folder structure here too?

README.md
maya: Added backticks to `uswds` in 6436fe4 rather than capitalizing to match how we talk about it in the code.

README.md
rogeruiz: Well for `JS` a `require('uswds');` is all they would need to use for `npm`. There is no ~~good or~~ standard way of doing this with `CSS` so we are leaving it up for the developer's interpretation.

README.md
rogeruiz: @msecret it changed due to maintaining the documentation using the version number. It's not a good idea to have to update version numbers everywhere whenever in the documentation when we update the package's version. Also, it could be a user pain to have our users update and then break their asset-paths.

rogeruiz: @msecret __second note__

The folder structure is included a little further above those lines :point_up:

README.md
rogeruiz: That sounds like a great idea! @juliaelman

README.md
juliaelman: I agree with Roger on adding the version number to the filenames.

README.md
juliaelman: Sounds good!

README.md
msecret: :+1:

README.md
msecret: So should this line be more like:

`/path/to/your/assets/lib/uswds-0.9.0/cssuswds.css`

rogeruiz: I think `path/to/your/assets/` lets them decide enough about where it should go. but :thought_balloon: welcome. Let's settle on something, `lib` or `vendor`. 

It might make sense for it to look a little more like this.

```
<link rel="stylesheet" href="/path/to/your/assets/css/lib/uswds.css">
<script href="/path/to/your/assets/js/lib/uswds.js"></script
```

rogeruiz:
Giving some more context around the discussion in all the `line-notes`.

- [ ] Update the `Downloads` section of the `README.md` with the minified
  versions.
- [ ] _Possibly_ update the examples in the `Downloads` section with a path
  that includes `lib`.

juliaelman:
@rogeruiz maybe one more thing to add to the checklist:

- [ ] Update the docs to use the minified versions styleguide CSS and JS files.

rogeruiz:
@juliaelman @maya @msecret This is good to be reviewed now with the updated suggestions in the checklist! :tada:

README.md
maya: I think @juliaelman meant to add the minified version in addition to the non-minified version here. Is that correct? I don't think she mentioned updating the example HTML to minified but just for this section.

rogeruiz: @maya @juliaelman 
I see, well for the `npm` instructions there is no way for them to used the minified code using `require('uswds');` They need to use the source files and the entry point is referenced in the `package.json` [here](https://github.com/18F/web-design-standards/blob/18f-pages-staging/package.json#L5). Minifiying the source when using `npm` is something that those specific developers would need to do on their own. _We could definitely add a sentence about that though._ Thoughts?

juliaelman: Sorry for the confusion, but yes. My note was for us to explicitly recommend the minified version when implementing via the downloaded version of the

@rogeruiz that is :100: okay in terms of using `require('uswds');` and having the non-minified version. Maybe something to add to our research to see what our users are wanting/needing for this method of implementation. //cc @carodew

README.md
```diff
@@ -29,26 +29,43 @@ Then, add the `dist` directory files into a relevant place in your code base —
...
+the browser. The examples below recommend using the minified versions.
+
 Refer to these files by adding the following `<link>` and `<script>` elements
 into your HTML pages:
 
 Add this to your `<head>` element:
 
 ```html
-<link rel="stylesheet" href="/path/to/your/assets/css/uswds-0.9.0.css">
+<link rel="stylesheet" href="/path/to/your/assets/css/lib/uswds.min.css">
```
msecret: I think it should be `/path/to/your/assets/css/lib/uswds-0.9.0/uswds.min.css`.

README.md
```diff
@@ -29,26 +29,43 @@ Then, add the `dist` directory files into a relevant place in your code base —
...
 ```
 
 Add this before the closing `</body>` tag:
 
 ```html
-<script src="/path/to/your/assets/js/uswds-0.9.0.js"></script>
+<script src="/path/to/your/assets/js/lib/uswds.min.js"></script>
```
msecret: `/path/to/your/assets/js/lib/uswds-0.9.0/uswds.min.js`

msecret:
LGTM, just had one comment.

README.md
rogeruiz: I don't know if we want to include the folder. We name the folder `uswds-VERSION_NUMBER` so users can keep track of it when it gets unzipped, but we're assuming that they are copying the files out of that folder and putting it somewhere in their project right? not just sticking the whole folder into `/path/to/your/assets/**/*`

README.md
rogeruiz: @msecret https://github.com/18F/web-design-standards/pull/1036#discussion_r55734103

README.md
msecret: If that's what we're expecting, then I agree with you, we should keep the path as this.

README.md
```diff
@@ -29,26 +29,43 @@ Then, add the `dist` directory files into a relevant place in your code base —
...
 
+We offer two versions — a minified version, and an un-minified one. The minified
+versions contain a `.min` suffix in the filename before the file's extension.
+The un-minified versions do not contain the `.min` suffix.
+
+Use the minified version in a production environment or to reduce the file size
+of your downloaded assets. And the un-minified version is better if you are in a
+development environment or would like to debug the CSS or JavaScript assets in
+the browser. The examples below recommend using the minified versions.
+
```
maya: I sort imagined this paragraph would go below the link examples. Thoughts @kategarklavs? I think it's also obvious that minified is `min` so not sure if we need it.

kategarklavs: Hey @maya — I'm totally OK with this comment appearing below the examples (we'll just need to update the last sentence to say, "The examples *above* recommend using the minified version."). As far as the necessity of the explanation goes: My sense is that this explanation is to help out newer users, but if it's overkill, I'm also OK with removing it.

PR #1043, opened 2016-03-11
Fix website task to use package name

rogeruiz:
`dutil.fileName` is a legacy property and was breaking the `website` tasks. Thanks @juliaelman


juliaelman:
WFM - thanks!

PR #1045, opened 2016-03-11
Update npm docs

maya:
This updates the "Using npm" part of the README to remove some unneeded text and update the order of the paragraph.


rogeruiz:
Looks good!

PR #1047, opened 2016-03-13
Remove image type from buttons

tysongach:
- This is not a valid type attribute for button
- Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

### DIFF
--- a/src/stylesheets/elements/_buttons.scss
+++ b/src/stylesheets/elements/_buttons.scss
@@ -11,8 +11,7 @@ $button-stroke: inset 0 0 0 2px;
 button,
 [type="button"],
 [type="submit"],
-[type="reset"],
-[type="image"] {
+[type="reset"] {
   appearance: none;
   background-color: $color-primary;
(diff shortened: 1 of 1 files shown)


rogeruiz:
Hey @tysongach! Thanks for this PR. According to the W3 HTML spec, the `[type="image"]` attribute can be used to select an `x` and `y` position from an image.

> A coordinate, relative to a particular image's size, with the extra semantic that it must be the last value selected and initiates form submission

[Attributes: Input types table](https://html.spec.whatwg.org/multipage/forms.html#attr-input-type)
[4.10.5.1.19 Image Button state (type=image)](https://html.spec.whatwg.org/multipage/forms.html#image-button-state-(type=image))

I think we should keep this in the Standards for now, although we may refactor the CSS if we create a UI component for `input[type="image"]`.

tysongach:
@rogeruiz Whoa, crazy! Never knew…TIL! Thanks for that.

PR #1052, opened 2016-03-14
Update Travis to use old `./go` command

rogeruiz:
Update Travis to use old `./go build` deployment strategy.


rogeruiz:
Seems like it's not necessary to add this and have the site build twice.

PR #1064, opened 2016-03-17
Fix sass-watch bug

rogeruiz:
The original watch tasks around `sass` were too greedy when watching for file changes and would recursively watch themselves for ever. This PR fixes this problem so that `npm start` can successfully watch for subsequent changes to the source directories.

### DIFF
--- a/config/gulp/sass.js
+++ b/config/gulp/sass.js
@@ -1,5 +1,4 @@
 var gulp = require('gulp');
-var gutil = require('gulp-util');
 var dutil = require('./doc-util');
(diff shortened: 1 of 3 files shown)


config/gulp/sass.js
```diff
@@ -1,4 +1,5 @@
 var gulp = require('gulp');
+var gutil = require('gulp-util');
```
msecret: Where does this end up getting used?

rogeruiz: ah ha not any more. I'll remove it.

PR #1074, opened 2016-03-21
Remove google-fonts.css link

ismamz:
The file doesn't exists.


maya:
Thanks so much @ismamz! Looks like @rogeruiz is already on it via #1069.

PR #1078, opened 2016-03-24
Fixed paths in link / script tags in README

IanLee1521:


msecret:
@IanLee1521 we actually purposefully wrote the paths as such, to advocate a convention that files in the `lib` directory should not be modified.

IanLee1521:
hmm, so where does that path show up? What I get from the download tarball doesn't actually have those paths, which is why I thought it was an error.

msecret:
Hey @IanLee1521. I see your point. I think our assumption was that the user would already have or create a new `lib` folder.

In making this change to have the path like this, we had a hypothesis that we'd limit people from the bad practice of editing the src files of the standards by logically separating it into a `libs` folder. I think your confusion shows that it might not be the right solution. Regardless, I want to keep this open as an issue until we can do more research or collect more feedback.

Thanks for writing in! It's very helpful to see what confuses our users.

maya:
Closing this as the path in the example was the intention and will need further research to see if this is confusing for others. Feel free to open an issue if you find this confusing or would like additional documentation.

PR #1085, opened 2016-03-28
make accordion class selector more inclusive

brendansudol:
This PR fixes a bug (https://github.com/18F/web-design-standards/issues/1083) where the accordion was not being initialized in the JS when there was a css class preceding the `usa-accordion` class (i.e., `div class="foo usa-accordion"`)


src/js/start.js
```diff
@@ -52,7 +52,7 @@ $(function() {
     $('[data-' + componentName + ']').politespace();
   });
 
-  $('[class^=usa-accordion]').each(function() {
+  $('*[class*=usa-accordion]').each(function() {
```
msecret: What's the reason for not doing `$('.usa-accordion')`?

brendansudol: the reason i didn't do that is because there is another class to check for as well, namely `usa-accordion-bordered`

msecret: Ah I remember now. 

I'm of the opinion that `$('.usa-accordion,.usa-accordion-bordered')` is a better solution. This is because with our CSS methodology, we could have classes like `.usa-accordion-child` that wouldn't represent the whole accordion, but an *element* of the accordion.

I don't see us creating that many different accordion types, and I think I'd be expected that if you did, you'd expect to have to add a selector for a new accordion type.

brendansudol: agreed

PR #1096, opened 2016-04-05
Input CSS fixes

brendansudol:
This PR addresses https://github.com/18F/web-design-standards/issues/957 and https://github.com/18F/web-design-standards/issues/1039.

1. https://github.com/18F/web-design-standards/issues/957 calls to attention that select inputs are a larger height than other input boxes in Firefox; this is the result of line-height / height quirks (see [here](http://joshnh.com/weblog/line-height-doesnt-work-as-expected-on-inputs/) in a few browsers). By explicitly setting the height property, this issue is resolved. This approach is in line with frameworks like Bootstrap.

2. https://github.com/18F/web-design-standards/issues/1039 outlines an edge case where long text overlaps with the dropdown chevron. To fix this, a right padding is added to select inputs.


src/stylesheets/elements/_inputs.scss
```diff
@@ -1,3 +1,21 @@
+$input-line-height: 1.3;
+$input-border-width: 1px;
+$input-padding-v: 1rem;
+$input-padding-h: 0.7em;
+
```
maya: I tend to follow this guideline for adding in new variables (http://sass-guidelin.es/#variables) in which both criteria should be met:

- the value is repeated at least twice;
- the value is likely to be updated at least once;

Are both of these being met?

dist/css/uswds.css
```diff
@@ -1381,6 +1383,8 @@ textarea {
 }
 
 select {
+  padding-right: 32px;
```
maya: Can you change this value to `3.2rem`? We only use px for detail work when it's 5px or less.

maya:
Thanks @brendansudol! Even though the issues addressed here are both about inputs, we like to stick to 1 PR to 1 issue for ease of reviewing.

src/stylesheets/elements/_inputs.scss
brendansudol: yes, they are used twice -- once for the property declarations and once for the height property computation. i'm not sure i fully understand the second criteria (for example, aren't there certain variables that won't be updated yet are still valuable to have abstracted because they're used throughout?)

brendansudol:
thanks for comments -- breaking these out into two separate PRs :)

Issue #1039, opened 2016-03-10
Dropdown: Select menu text can run into chevron


humancompanion:
In many places the text in our dropdown menus is long enough that it runs into the chevron handle that indicates the field is a dropdown. This cause some usability issues in that the affordance for the dropdown is lost when the downward facing chevron is obscured.

For what it is worth, I fixed the problem in the USAJOBS design system by applying some padding-right to the select element:
https://github.com/USAJOBS/design-system/blob/gh-pages/_scss/element/_select.scss

maya:
That is a good find @humancompanion and we ought to do the same. Thank you!

elarenee:
picked up this issue!

maya:
@elarenee feel free to familiarize yourself with: https://github.com/18F/web-design-standards/pull/1096 - there was an IE regression introduced on that solve that we couldn't figure out.

arielkennan:
Hello! Wanted to share that NYC is running to this too on one of our projects and we'd love a fix. We also ran into the IE11 issue with the artifact. Thanks!

juliaelman:
@arielkennan thank you for your comment and making us aware of this. I am going to add this to our upcoming sprint planning. We appreciate your patience and support!

arielkennan:
Thanks for the quick reply! I'll also share our site once it launches you can see your beautiful work implemented on another .gov.

arielkennan:
A breakthrough - this just in from one of our developers on how we were able to fix it: 

So in my original build I was using additional padding on the left (right for RTL) to make room for the decorative custom arrow so it wouldn't overlap the text. As you saw in IE11 this caused two sets of arrows. IE11's native arrow control was being pushed inside by the padding rather than sitting on top of it. 

Using

`select::-ms-expand {
  display: none;
}
`
I could selectively target IE11 and hide that native control. Then it was just a matter of fixing the background-size definition of the decorative custom arrow so it would work in IE11 and not appear like a tiny dot.

maya:
That's great to hear you were able to fix this, @arielkennan! Linking @brendansudol's fix here which found the "tiny dot" issue so the team can implement the fix: https://github.com/18F/web-design-standards/pull/1097

PR #1097, opened 2016-04-05
Fix select text overlap

brendansudol:
This PR addresses https://github.com/18F/web-design-standards/issues/1039 (an edge case where long text overlaps with the dropdown chevron image). To fix this, a right padding is added to select inputs.


src/stylesheets/elements/_inputs.scss
```diff
@@ -82,7 +82,9 @@ textarea {
   height: 16rem;
 }
 
+// padding right added to avoid text overlapping with bg image
 select {
+  padding-right: 3.2rem;
```
msecret: Although it's already incorrect, could you change this so that every property is in alphabetical order? https://pages.18f.gov/frontend/css-coding-styleguide/format/

maya: What's property order is incorrect?

brendansudol: sure thing @msecret -- thanks for the link; i wasn't sure what convention we followed

dist/css/uswds.css
```diff
@@ -1381,15 +1381,16 @@ textarea {
 }
 
 select {
+  appearance: none;
```
maya: I think it makes more sense to keep this as it was before, since the vendor-prefixed properties begin with a dash and it keeps the ones that begin with letters together. This is what thoughtbot Bourbon does: https://github.com/thoughtbot/bourbon/blob/master/core/bourbon/library/_prefixer.scss#L19

brendansudol: done!

maya:
LGTM 🎉

maya:
@brendansudol one thing I noticed when testing this in IE is it also pushes the dropdown arrow to the left and I see a small artifact too.

Before:
screen shot 2016-04-11 at 2 15 30 pm

After:
screen shot 2016-04-11 at 2 14 37 pm

brendansudol:
Hmm, that's odd -- what IE version is that, @maya?

maya:
@brendansudol I tested in IE 9 and 11 with the same result. Did you find something different?

brendansudol:
You're right -- I see it now 😢 -- I don't think this solution will work in light of this... I think maybe the way to go is that add a white background to the chevron image (instead of the current transparent one); this should have the desired effect of not having the option value text overlapping with the chevron dropdown area... thoughts?

rogeruiz:
Closing this issue for now as it doesn't completely solve the issue in #1039

Issue #1007, opened 2016-02-22
Request: Make CSS background image path links a SASS variable


bbgvisualjournalist:
We're working on a Wordpress theme based on the web standards. 

The relative paths for background images are currently written like this: url('../img/arrow-down.png') 

This assumes that the CSS file is inside of a "/css" folder.

But Wordpress expects to see the CSS in the root folder of the theme (and uses the CSS to help define the theme).

Would it be possible to set the path for images as a SASS variable so that we can reference the images? (So that it's compatible with Wordpress or other systems in the future.)

Example: We need the above path to be:  url('img/arrow-down.png')

Many thanks.

maya:
Hello @bbgvisualjournalist! That's a great question. Dealing with image paths for different asset pipelines is something that's definitely come up before and using a Sass variable to handle this is an interesting idea.

I could see us doing this with adding a Sass variable:

 ```
// Variable

$usa-image-path: ../img/;
```

To use this, we'd use [interpolation syntax](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#interpolation_)

`background-image: url('#{$image-path}arrow-down.png')`

And then you'd only need to worry about setting the `$usa-image-path` variable for your use case.

Thoughts @msecret @rogeruiz @juliaelman?

**Comment by rogeruiz** [comment]
I think this is a good approach @maya. I like the idea of having variables for 
the `value` section. What about having `@mixins` for the `property` part as 
well? Something along the lines of:

```scss
@mixin usa-property-builder ( $prop, $value ) {
    $prop: $value;
}

body.ruby-on-rails {
    @include usa-property-builder( 
        background-color, 
        asset-path('#{usa-image-path}')
    );
}
```

This could go a long way to address some of the issues in 18f/us_web_design_standards_gem#15

msecret:
I ended up using a postcss plugin to do this on cg-style specifically because my code was consuming the css rather then the sass. Might want to consider that as an option on top of the variable if that's going to be a common use case (using the css rather then sass).

msecret:
This was completed here: https://github.com/18F/web-design-standards/pull/1415

ui-coder:
How about this way if using css variables, `--img: url(../img/filename.jpg)` and then `.selector { background: var(--img); }`.

Issue #952, opened 2016-01-13
Unable to override default colors from extended Sass


msecret:
In https://github.com/18F/cg-style, I have a scss file that has a file to override certain variables from the styleguide like color:

https://github.com/18F/cg-style/blob/master/src/css/override_vars.scss#L3
```
$color-primary:              #1F2B38;
$color-primary-darker:       #2C3E50;
$color-primary-darkest:      #1F2C38;
```

I then import the standards and include this file after:

https://github.com/18F/cg-style/blob/master/css/cg_style.scss#L4
```
@import "../node_modules/uswds/assets/_scss/all.scss";
@import "../src/css/override_vars.scss";
```

But when I use the CSS from *cg-style*, the colors of the USWDS components are not modified. For instance, the button color is still `#0071bc`. This includes if I import the overrides before or after the standards.

This is happening because the standards are setting all variables without `!default` in the `variables.scss` file. If this wasn't the case, and only the `defaults.scss` file was used, I would be able to import my override_vars.scss file and it would correctly be overriden.

rossdakin:
@msecret this is great — I have been hoping for a way to override values without modifying the distributed flies (so I can just nuke and replace `/designstandards/` in my project but preserve my customized values).

Two potential issues with this though:
 1. I currently don't `@import "../node_modules/uswds/assets/_scss/all.scss";` (I just link to the precompiled CSS that you distribute) because the import breaks CSS paths to images. This is its own issue, not exacerbated by your change here but one that prevents me from benefitting from your change.
 2. Because I'm not importing `all`, I am importing `variables` so that I can at least reference them in my custom SCSS. This change will kill that file and thus break my code. I wonder if you could leave the file in place but just import `default` into it? Or just tell me "this is in beta -- get off my lawn." Both acceptable :-)

maya:
+1 It would also be nice if we could provide `variables.scss` for people to download on the styleguide site to customize their variables and add the file themselves.

rossdakin:
Also, just to make sure I understand this correctly, it seems like your example:

```scss
@import "../node_modules/uswds/assets/_scss/all.scss";
@import "../src/css/override_vars.scss";
```

…should actually be:

```scss
@import "../src/css/override_vars.scss";
@import "../node_modules/uswds/assets/_scss/all.scss";
```

Looking at the `!default` docs as referenced [here](https://robots.thoughtbot.com/sass-default) (but no longer live?):

[...]

PR #953, opened 2016-01-13
Allow variable overriding by remove variables.scss

msecret:
This allows users of the library to override global variables like color
by importing their variables before the USWDS library. This ensures only
defaulted vars are imported, meaning if the value is already set, such
as by the user, then it won't be overriden. This also ensures the USWDS
components will receive the new colors.

More information can be found here:
https://github.com/18F/web-design-standards/issues/952

Closes 952.

PR #1098, opened 2016-04-06
Implement ability to override variable defaults

msecret:
This allows users of the library to override global variables like color
by importing their variables before the USWDS library. This ensures only
defaulted vars are imported, meaning if the value is already set, such
as by the user, then it won't be overriden. This also ensures the USWDS
components will receive the new colors.

More information can be found here:

Refs #952


maya:
How does this PR relate to #953? 

Also, the title says it resolves #1007 but the description says #952.

rogeruiz:
Closing this ticket for now so we can switch over branches easily. This PR should be reopened and rebased after the `dist/` directory is removed.

PR #1099, opened 2016-04-06
Fix firefox input height

brendansudol:
This PR addresses https://github.com/18F/web-design-standards/issues/957 (a bug report that select inputs have a larger height than other input boxes in Firefox); this is the result of line-height / height quirks (see [here](http://joshnh.com/weblog/line-height-doesnt-work-as-expected-on-inputs/) in a few browsers). By explicitly setting the height property, this issue is resolved. This solution aligns with how other frameworks such as Bootstrap approach this issue.


docs/_scss/elements/_inputs.scss
```diff
@@ -1,3 +1,20 @@
...
+  ($base-font-size * $input-line-height) +
+  ($input-padding-vertical * 2) +
+  (rem($input-border-width * 2))
+);
+
+// truncated to 1 decimal place
+// (i.e., 1.21rem -> 1.2rem)
+$input-height: floor($input-height-exact * 10) / 10;
+
+
```
maya: is this an extra line? @msecret

PR #1100, opened 2016-04-06
Fixes #1093

### LINKED ISSUE #1093: CDN
Please host these on CDN for simple reuse

fulldecent:


jpyuda:
Thank you so much for setting up the CDNJS instance!

Unfortunately, our security practices prevent us from putting this forward as an officially-supported CDN. If we're to offer a hosted version of the standards, we need to be able to vouch for the integrity of all the servers in the chain of delivering that hosted code -- otherwise many government agencies would not be able to rely on it due to various laws.

That said, I think the CDNJS instance is very useful for non-government entities who are using the standards, as well as anybody prototyping/experimenting. We'll continue to think about whether (and how) we can inform the community about non-official tools like this.

alex:
The appropriate way to address this is through Subresource Integrity, a sort-of-new HTML standard that allows you to include the hash of the downloaded javascript, which means you dont' have to trust the CDN for integrity.

If you look at the bootstrap CDN instructions, they include the SRI attributes right in their example, meaning users who take the path of least resistance get all the integrity protection of hosting locally, with the benefits of a CDN: https://getbootstrap.com/getting-started/#download-cdn

alex:
(Current deployment status: http://caniuse.com/#feat=subresource-integrity)

konklone:
Even with SRI, I would still recommend that the USWD not promote the use of a third party CDN for distribution of the USWD code, especially for its intended audience (US government websites). SRI removes much of the security risk, but there is still a privacy impact of sharing online citizen/user activity with a third party.

The primary reason to use a third-party CDNs is performance, but HTTP/2 means that the performance hit for downloading self-hosted USWD assets is minimal, and much faster than the CDN-hosted version when the user doesn't have it cached already. It would be great if USWD actively recommended that government agencies using their assets self-hosted them, with an additional suggestion to enable HTTP/2.

alex:
My personal view is that insisting everyone self-host imposes a non-trivial
barrier to usage (frontend package management is still not outstanding, and
something like bootstrap can be used without needing to care about that
whole thing), and I think lowering the barrier to adoption is particularly
important given the goal of providing commonality across a large and
disparate set of websites.

On Fri, May 20, 2016 at 7:26 PM, Eric Mill <notifications@github.com> wrote:

> Even with SRI, I would still recommend that the USWD not promote the use
> of a third party CDN for distribution of the USWD code, especially for its
> intended audience (US government websites). SRI removes much of the
> security risk, but there is still a privacy impact of sharing online
> citizen/user activity with a third party.
>
> The primary reason to use a third-party CDNs is performance, but HTTP/2
> means that the performance hit for downloading self-hosted USWD assets is
> minimal, and much faster than the CDN-hosted version when the user doesn't
> have it cached already. It would be great if USWD actively recommended that
> government agencies using their assets self-hosted them, with an additional
> suggestion to enable HTTP/2.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/web-design-standards/pull/1100#issuecomment-220741247>
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: D1B3 ADC0 E023 8CA6

joshbruce:
Even though this is closed - it's an interesting conversation.

Correct me if I'm wrong, CDN would offer the ability to pull the compiled CSS and JS, which is fine if no changes to the defaults are made.

Otherwise, developers would have to add more CSS to compensate or override (same with Bootstrap or other CDNs). Whereas, having it as a modular package allows developers to pick and choose more easily. The modularity would not necessarily be lost as long as the full package is still available through `npm`. 

An interesting poll might be to ask those using the Standards how much customization has been performed - if any.

Some questions this thread brought to mind:

What primary purpose of the Standards is being taken? 
- A toolset to be used? 
- An edict to be followed?
- Walking the fine line between the two?
- Or something else?

I believe the writings about the Standards points a certain way - and am more interested in hearing the views of others as this project proceeds. It may also not be the right time to CDN the Standards (still a beta pre-one-oh release). @jpyuda did allude to a future possibility of a hosted version without an external provider (Federally hosted and vouched for - and, honestly, outside my wheelhouse).

konklone:
> My personal view is that insisting everyone self-host imposes a non-trivial
> barrier to usage (frontend package management is still not outstanding, and
> something like bootstrap can be used without needing to care about that
> whole thing), and I think lowering the barrier to adoption is particularly
> important given the goal of providing commonality across a large and
> disparate set of websites.

But SRI means that code can't be updated in place without making client-side updates to signatures, so the CDN can't update the version on behalf of their customers. So, controlling which version of the code you're running is always up to the site owner, and upgrading always takes manual work on the part of the site owner -- I don't see how this would be any more a barrier than self-hosting.

jpyuda:
We're interested in (but not actively working on) something of a middle ground: operating our own CDN, which allows hotlinking from .gov sites only. I'm sure there are quite a lot of intricacies of this (like I said, we're not actively working on or researching this setup) but it feels like a solution that helps to meet all of the needs discussed above.

konklone:
@jpyuda That's a reasonable thing to do, though if you do begin actively working on that, I recommend talking with me and with the Digital Analytics Program. They've moved to a CDN for their JS, at dap.digitalgov.gov, and it is a serious responsibility. If their JS were compromised, it would be very bad, and DAP instituted new processes to handle the increased importance of content integrity. Also, right now there's still a [privacy leak to Symantec](https://github.com/digital-analytics-program/gov-wide-code/issues/38) (that they are working on resolving) that makes me deeply uncomfortable. My point is just that if you centralize things, you have to think harder about things that might be less important when when it's distributed.

And the DAP can't use SRI, because the whole point they are promoting the use of a CDN rather than self-hosting is so they can deploy updates (including security updates) reliably across participating agencies without requiring manual intervention from each agency, and SRI would make that impossible. The USWD might not have that as a goal, but if not, then my comment above applies about the CDN not really lowering the barrier to deployment.

In general, I think HTTP/2 undermines the value of SRI for most websites, and third party CDNs for all websites. First or second party CDNs, like something shared within the government, continue to make sense when they're stood up for the right reasons and managed with appropriate care.

jpyuda:
Oh, we'll certainly reach out and plan sufficiently to ensure we have our ducks in a row on this one.

PR #1115, opened 2016-04-11
Remove type from state dropdown

### LINKED ISSUE #1109: Form Templates - Address Form - State Dropdown
&lt;select id="state" name="state" type="text"&gt;

I might be able to do a PR on this later - but I don't believe 'type="text"' should be there - because it's a dropdown - not an input with a type.

<select id="state" name="state">

jbrucegsa:
Fixes #1109


maya:
LGTM 🎉 

Thanks @jbrucegsa!

PR #1116, opened 2016-04-11
Fixed checkbox name attribute

### LINKED ISSUE #1107: Radio and Checkbox patterns
Curious about the difference in pattern between radio and checkboxes.

Radio

```
 <fieldset class="usa-fieldset-inputs usa-sans">
    <legend class="usa-sr-only">Historical figures 2</legend>
    <ul class="usa-unstyled-list">
      <li>
        <input id="pea-soup" type="radio" checked="" name="soup" value="pea">
        <label for="pea-soup">Elizabeth Cady Stanton</label>
      </li>
      <li>
        <input id="chicken-noodle" type="radio" name="soup" value="chicken-noodle">
        <label for="chicken-noodle">Susan B. Anthony</label>
      </li>
      <li>
        <input id="tomato" type="radio" name="soup" value="tomato">
        <label for="tomato">Harriet Tubman</label>
      </li>
    </ul>
  </fieldset>
```

Checkbox 

```
  <fieldset class="usa-fieldset-inputs usa-sans">
    <legend class="usa-sr-only">Historical figures 1</legend>
    <ul class="usa-unstyled-list">
      <li>
        <input id="apple-pie" type="checkbox" name="apple-pie" value="apple-pie" checked="">
        <label for="apple-pie">Sojourner Truth</label>
      </li>
      <li>
        <input id="key-lime-pie" type="checkbox" name="key-lime-pie" value="key-lime-pie">
        <label for="key-lime-pie">Frederick Douglass</label>
      </li>
      <li>
        <input id="peach-pie" type="checkbox" name="peach-pie" value="peach-pie">
        <label for="peach-pie">Booker T. Washington</label>
      </li>
      <li>
        <input id="disabled" type="checkbox" disabled="">
        <label for="disabled">George Washington Carver</label>
      </li>
    </ul>
  </fieldset>
```

From the checkbox code above, it would be necessary to check for an "is set" equivalent on each checkbox possibility. Alternatively, we could follow the radio pattern - using name - and have it POST as an array. (This is usually the pattern seen in PHP or Ruby.)

Checkbox alt

```
  <fieldset class="usa-fieldset-inputs usa-sans">
    <legend class="usa-sr-only">Historical figures 1</legend>
    <ul class="usa-unstyled-list">
      <li>
        <input id="apple-pie" type="checkbox" name="pie[]" value="apple-pie" checked="">
        <label for="apple-pie">Sojourner Truth</label>
      </li>
      <li>
        <input id="key-lime-pie" type="checkbox" name="pie[]" value="key-lime-pie">
        <label for="key-lime-pie">Frederick Douglass</label>
      </li>
      <li>
        <input id="peach-pie" type="checkbox" name="pie[]" value="peach-pie">
        <label for="peach-pie">Booker T. Washington</label>
      </li>
      <li>
        <input id="disabled" type="checkbox" disabled="">
        <label for="disabled">George Washington Carver</label>
      </li>
    </ul>
  </fieldset>
```

This allows the posting of an array of the check values - then we only have to check for the "pie" variable in the POST data. Again, curious if this was a conscious decision to move a different way. Otherwise, could be an issue.

jbrucegsa:
Fixes #1107

### DIFF
--- a/docs/_elements/form-controls.md
+++ b/docs/_elements/form-controls.md
@@ -135,15 +135,15 @@ lead: Form controls allow users to enter information into a page.
 
     <ul class="usa-unstyled-list">
       <li>
-        <input id="apple-pie" type="checkbox" name="name[]" value="apple-pie" checked />
+        <input id="apple-pie" type="checkbox" name="pies[]" value="apple-pie" checked />
         <label for="apple-pie">Sojourner Truth</label>
       </li>
       <li>
-        <input id="key-lime-pie" type="checkbox" name="name[]" value="key-lime-pie">
+        <input id="key-lime-pie" type="checkbox" name="pies[]" value="key-lime-pie">
         <label for="key-lime-pie">Frederick Douglass</label>
       </li>
       <li>
-        <input id="peach-pie" type="checkbox" name="name[]" value="peach-pie">
+        <input id="peach-pie" type="checkbox" name="pies[]" value="peach-pie">
         <label for="peach-pie">Booker T. Washington</label>
       </li>
       <li>
--- a/docs/_elements/form-controls.md
+++ b/docs/_elements/form-controls.md
@@ -135,15 +135,15 @@ lead: Form controls allow users to enter information into a page.
 
     <ul class="usa-unstyled-list">
       <li>
-        <input id="apple-pie" type="checkbox" name="apple-pie" value="apple-pie" checked />
+        <input id="apple-pie" type="checkbox" name="name[]" value="apple-pie" checked />
         <label for="apple-pie">Sojourner Truth</label>
       </li>
       <li>
-        <input id="key-lime-pie" type="checkbox" name="key-lime-pie" value="key-lime-pie">
+        <input id="key-lime-pie" type="checkbox" name="name[]" value="key-lime-pie">
         <label for="key-lime-pie">Frederick Douglass</label>
       </li>
       <li>
-        <input id="peach-pie" type="checkbox" name="peach-pie" value="peach-pie">
+        <input id="peach-pie" type="checkbox" name="name[]" value="peach-pie">
         <label for="peach-pie">Booker T. Washington</label>
       </li>
       <li>


maya:
Thanks @jbrucegsa! Quick question: is there a reason you're using `name[]` instead of `name`?

Looping in @nickbristow for more 👀

jbrucegsa:
@maya - Always asking good questions that make me learn!

I do it so the checked values get POSTed as an array of the same <code>name</code>. Without it, I usually only get the last element to be checked of that name.

To POST as an array of the same variable: <code>name[]</code>

To POST as a single yes/no toggle-style question, the way we have it now (can be difficult to process server-side; having to check the existence of each potential, but GET could be easier): <code>name</code>

To POST as a keyed array (GitHub profile settings): <code>name[key]</code>

screen shot 2016-04-11 at 8 03 18 pm

Some examples I'm seeing from the W3C go the unique name-value pair route; however, the spec does say checkboxes can share the same name: http://www.w3.org/TR/html401/interact/forms.html#h-17.2.1.

Maybe something in the description notes?

I've been doing this a while and things might have changed. So, if I'm being too old-school, let me know.

**Other resources**

Ruby on Rails
http://stackoverflow.com/questions/4037585/http-post-with-multiple-checkboxes-ror
http://stackoverflow.com/questions/4425176/in-rails-how-to-handle-multiple-checked-checkboxes-just-split-on-the-or

PHP
http://stackoverflow.com/questions/4997252/get-post-from-multiple-checkboxes

docs/_elements/form-controls.md
```diff
@@ -135,15 +135,15 @@ lead: Form controls allow users to enter information into a page.
 
     <ul class="usa-unstyled-list">
       <li>
-        <input id="apple-pie" type="checkbox" name="apple-pie" value="apple-pie" checked />
+        <input id="apple-pie" type="checkbox" name="name[]" value="apple-pie" checked />
```
msecret: Would name in these three places actually be called something like `"pies[]"` as thats technically what this collection is?

msecret:
@jbrucegsa I just had one comment around how to name the `name` variable.

jbrucegsa:
@msecret - Agreed! Sorry about that. :)

docs/_elements/form-controls.md
jbrucegsa: done

maya: No, actually it's a name "Sojourner Truth." We are not displaying pies, it changed but the internal values didn't change along with it.

maya:
I thought we preferred to go bracketless based off of conversations on slack @msecret bc we can't assume users are going to use rails or php and our code shouldn't make any assumptions but provide an abstracted pattern.

jbrucegsa:
@maya & @msecret - I should have numbered those. If we decide to lose the brackets - the second POST concept is correct and we can just revert this merge.

The brackets are appearing to be a common pattern:
http://stackoverflow.com/questions/10560345/can-i-pass-multiple-checkbox-values-with-one-name
http://www.dyn-web.com/tutorials/forms/checkbox/same-name-group.php

I did, however, find this example demonstrating only the same name:
http://www.javascriptkit.com/javatutors/radiocheck2.shtml

However, it does not appear to return an array of *only* the checked checkboxes - still have to loop through.

Again, maybe the description should or could be updated to account for the three variations?

msecret:
@maya the arguments in that conversation were based more off preference. An it's not necessarily only a rails and PHP thing, other frameworks in other languages also potentially have them.

The web design standards is more of a front end framework, so isn't really mean to take a strong decision on backend. I think showing examples like this that are commonly accepted is perfectly fine.

msecret:
That being said, if users find it confusing in the future, we'll address it then.

maya:
I agree that we shouldn't make assumptions, but I think adding the brackets is an assumption. Without them is more of an abstracted view which I think may be a cleaner approach.

I'd like to hear thoughts from @juliaelman @yozlet @jpyuda

jpyuda:
The name attribute has all kinds of implications for accessibility, and I would be very, very hesitant to treat it like an array without knowing what effects that would have on a broad range of assistive technology.

nickbristow:
I did some research, as far as I can tell, AT doesn't seem to be looking at the `name` attribute for anything. We have the fieldset for grouping and proper labels. 

Using the `[ ]`'s should be fine as its just a string value and shouldn't affect accessibility. 

I think this is just a best practice/how do we want to show this example in the clearest way possible. Will people be confused by the brackets? Do we need to explain what the attributes are doing and how they are used?

jbrucegsa:
@jpyuda - That's fair - and interesting. If I get the opportunity, I'll see what I can do on that score.

msecret:
I still agree with @jbrucegsa here. I think this is somewhat of a standard in the dev community and is fine to use in our examples.

juliaelman:
Thanks to everyone for weighing in on this and especially to @jbrucegsa for your contribution! :us: 

Overall, I don't think we can make assumptions on what our dev community wants or needs in regards to how we recommend implementations of a dynamic codebase without the user research backing. I agree with @maya and am going to merge her pull request that removes the `[ ]` from the code examples. If we find our community explicitly asking for this request, and in need of an example, we will review again at that time.

PR #1119, opened 2016-04-13
Remove brackets from name attribute in radio fields

maya:
This removes the brackets added in the name attribute in radio fields in #1116.

This also adds a missing name attribute and value in the disabled radio field.


juliaelman:
LGTM! 

Refs https://github.com/18F/web-design-standards/pull/1116#issuecomment-210045905

PR #1123, opened 2016-04-15
Cloud.gov config

yozlet:
This PR includes the necessary configuration for CircleCI to auto-deploy certain branches to Cloud.gov:

 * `master` and `18f-pages` branches deploy to https://standards.usa.gov/
 * `staging` and `18f-pages-staging` branches deploy to https://wds-microsite-staging.apps.cloud.gov/ (soon to also be available at https://standards-staging.usa.gov/ )

After merging, I'll set up CircleCI to watch this repo. Later, we'll want to remove the `18f-pages(-staging)` branches from this configuration.


msecret:
So this solution doesn't rely on Jekyll and allows us to run build commands on a CI server before deploy? Because that sounds awesome!

Is there a reason for using Circle CI rather then travis? I know we already use Travis, but there's so little being done I'd be happy to move everything over to Circle.

msecret:
Just thought of more things; is this PR going to be where we actually switch over? We'll need to update some documentation probably (like the main repo page) to point to the new URL.

How could we redirect from the current playbook page to the new URL? Should we deal with that in another PR?

maya:
@msecret I don't think this PR is meant to be the final switch over, but just add in the branches we need for cloud.gov. @yozlet says "Later, we'll want to remove the `18f-pages(-staging)` branches from this configuration." so I'm assuming that step will happen after the redirects are set up.

yozlet:
@msecret @maya We don't have to switch over exactly with this PR; that's why it specifies two of each branches, so it starts working before the switch and continues after. I also don't know the best way to do the redirect, but my initial guess is that we
 1. merge this, and start auto-deploying
 2. branch `master` & `staging` off the existing two branches
 3. remove the `18f-pages` branches from the CircleCI config, so it won't push those branches to the new site
 4. commit the redirect pages to the `18f-pages` branches, so that they get pushed to the old site

Does that make sense?

yozlet:
Also, I switched to Circle because my experiences with it have always been better than with Travis, especially when dealing with caching and debugging the build process. (Circle lets you SSH into the build as it's happening to debug it, if needed)

Issue #988, opened 2016-02-05
Collapsible and Dismissible Alert Examples


harrisj:
It would be nice if alerts also included options the following actions:
- Dismissing the alert entirely (with an appropriate dismissal icon on the right)
- Collapsing/expanding the body of the alert (so the title would still be there)

I am not sure if this would mean doing JS for these actions or just providing some guidance on appropriate styling, but it would be useful for me on a current and some future projects.

msecret:
I agree that a component like this would be good from a dev standpoint(interested in hearing from design). The code could almost be re-used for the accordion. In general, it'd be pretty easy to implement for a dev perspective.

mollieru:
This isn't exactly an ideal implementation, but here's an example of one
existing adaptation of collapsable alert:
https://certify.sba.gov/am-i-eligible/app?

On Fri, Feb 5, 2016 at 5:01 PM, Marco Segreto <notifications@github.com>
wrote:

> I agree that a component like this would be good from a dev
> standpoint(interested in hearing from design). The code could almost be
> re-used for the accordion. In general, it'd be pretty easy to implement for
> a dev perspective.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/18F/web-design-standards/issues/988#issuecomment-180584854>
> .
>



-- 

*Mollie Ruskin*
US Digital Service
Office of Management and Budget | Executive Office of the President

joshbruce:
See #1124

bradnunnally:
Closing the related PR and will use the information contained in #1124 when creating the new component.

maya:
This is not on the current product roadmap, closing for now.

PR #1124, opened 2016-04-15
Dismiss button

joshbruce:
For my current project we wanted a way to dismiss alerts. The solution
we found was to include a form in the alert with a button. So, we
automatically got a dismiss button. However, we may want to include an
inverse as well for accessibility and different background color
applications.


jbrucegsa:
ps. I am not setup with AT tools, and not sure who to loop in that part of the conversation.

maya:
@jbrucegsa thanks for this PR! We're still working on coming up with our contribution guidelines for new components which will help us evaluate new features from public contributors. I'll also add a note that we'll need to evaluate variations to existing components.

jhfrench:
@jbrucegsa may I ask why you favor a `button` (and, therefor, `form`) over an `a` tag that has an internal `href` attribute? You probably have a good reason in mind.

My concern is a form-based approach might get tricky when the alert is already a child of a form.

jbrucegsa:
@maya - No worries! So far this seemed to be the only new-new item that we thought could be applied across the USWDS - not just application of design patterns to existing components (like the error handling on form controls).

@jhfrench - Great question! 

tl;dr

You're correct, the W3C does not allow for sub-forms; therefore, putting a dismiss *form* inside another form would not work; however, a button, by itself, could by giving the button a name and value attribute.

As to other deciding factors (some of the thoughts prompted by the question):

Part of choosing a button comes from the USWDS itself: "Use buttons to signal actions." Given dismiss is an action, it seemed appropriate. Placing it in a form is mainly to allow a complete (standalone) non-JS solution - otherwise, we could apply an event override via JS or the aforementioned name and value to the button - something we would probably need to do if the dismiss button was used in conjunction with a modal presentation (see the "download and share" functionality here: http://www.census.gov/popclock/).

Re alert dismissal specifically: The USWDS application of alerts *within* a form seems mainly for extended instructional purposes not served by the inline hints (http://127.0.0.1:4000/form-templates/#password-reset-form); therefore, probably not a good place to use dismiss functionality. Further, dismissing alerts, at least for the program I'm on, is specific to dismissing system alerts; therefore, they do not reside within a form and would be in addition to any forms already displayed.

I added display: inline-block to the class definition, which allows the visual and functionality to be applied to an anchor tag should other developer solutions require it to be placed within a form - though, as I'm thinking about it - if the page refreshes from the link (non-JS), that might disrupt the user's ability toe complete the form. So, the dismiss button is starting to feel more and more like a rarity - like some of the alternative buttons available.

Hope that helps. Thanks again for the question! Let me know if you have more.

p.s. I'm not sure why this change causes CircleCI to not build.

bradnunnally:
Closing PR. Will revisit when #988 is picked up.

PR #1127, opened 2016-04-18
Fix radio and checkbox vertical alignment

maya:
This fixes the vertical alignment of checkboxes ☑️ and radio buttons 🔘 (without resorting to use magic numbers). The issue was brought up in #1105, however this supersedes #1105. 

_**This is what it looks like NOW...**_

Checkboxes:

screen shot 2016-04-18 at 3 59 32 pm

Radio buttons:

screen shot 2016-04-18 at 3 59 39 pm

_**This is what it looks like BEFORE...**_

Checkboxes:

screen shot 2016-04-18 at 4 17 45 pm

Radio buttons:

screen shot 2016-04-18 at 4 17 51 pm

*You can see that all of the radio buttons are to low and the first and third checkboxes (Sojourner and Booker) are too low.


maya:
This PR has design changes @ericadeahl and @carodew. Let me know if it's OK!

carodew:
LGTM!

src/stylesheets/elements/_inputs.scss
```diff
@@ -145,16 +145,16 @@ input[type="radio"] + label::before {
   content: '\a0';
   display: inline-block;
   height: 1.8rem;
-  line-height: 0.8;
+  line-height: 1.8rem;
```
msecret: What if somebody changes their base font sizes, then wouldn't this no longer be aligned? Could this be written so it accounts for that?

maya: Yes, it accounts for that. This only looks at the size of the button not text, it will always be centered.

Here I made the text way bigger and it's still centered:
screen shot 2016-04-19 at 11 25 23 am

msecret:
I checked out the code and rebased this.

Issue #933, opened 2015-12-18
Illustrator color swatch file suggestion


rtwell:
I noticed that the file **web-design-standards-colors.ai** is essentially an empty Illustrator file with the USWDS swatches preloaded. I have 3 suggestions for this file:

1) Swatches in Illustrator should be Global Colors. This radically enhances how designers control color in Illustrator. Adobe has a [7 minute tutorial video here](https://helpx.adobe.com/illustrator/how-to/process-and-global-colors.html).

2) The AI color file could have objects pre-colored for designers to effortlessly load a few colors at a time into their documents instead of importing all of the USWDS colors. I would suggest something like this:
![possibleai](https://cloud.githubusercontent.com/assets/12564977/11902362/c7a96ec2-a576-11e5-839e-30b3fbc733be.png)
Then designers have a choice on how they want to add colors to their AI docs. I have created an AI file with all of these changes—I'm happy to provide if you're interested.

3) Some of the naming conventions used for the colors are counter-intuitive for designers. "Primary-alt" getting grouped with the Secondary colors doesn't make sense to me.

Drakenhart:
Every designer has different patterns of behavior when it comes to their work flow. Not everyone is going to realize that the color swatches are pre-loaded.

I realized this myself a bit late and opted to create my own color boxes.

It makes it easier to have colored objects (circles or rectangles for example) on a Layer inside and Ai document.

So I agree and support @rtwell 's suggestion.

carodew:
I'm all for this. :+1:

carodew:
I think it might also be valuable to include the color contrast pairings for typography in the colors file (or the typography file). Thoughts?

rtwell:
I think that is a great idea!

Drakenhart:
Agreed @carodew that would also be helpful.

carodew:
The original suggestion has been implemented, the current illustrator file has the color palette drawn out on the artboard and global colors. I'll create a separate issue for adding the color contrast pairings.

PR #1129, opened 2016-04-20
Background colors

joshbruce:
By using mixins we decouple dependence on styleguide.css helper
classes. Further, by making classes, designers and developers can
quickly generate accessible background and text color combinations.

See #933

Include @carodew


maya:
Hi @joshbruce, that's a neat idea to tie the background and text color into a mixin. I think that the mixins help to enforce these standards, but reduce flexibility in trying different combinations out. To be clear, this PR does not include the "Colors on a white background" in the example - is that right?

joshbruce:
Hey @maya - Thanks. You are correct - the main set for this PR used the non-white backgrounds found in: https://standards.usa.gov/visual-style/#text-accessibility

However, if we want, more can always be added and, I believe, the mixins would only increase the compiled size if used. If it would be beneficial I can add the white-backgrounds in another PR.

Regarding flexibility - there's a root mixin all the pre-defined mixins use:

@mixin bgcolor($background, $color)

This allows any background and text color combination and would be available for anyone including the utilities.scss to their sass builds. The root mixin also only sets background-color; so, background-image et al can still be applied. Only setting background, I believe, would cause issues.

joshbruce:
ps. @maya - On further reflection - we *could* eliminate the other mixins in favor of *just* the bgcolor() one to reduce confusion and the number of mixins in _utilities.scss. Would need to modify the change done to the _alerts.scss for this PR.

carodew:
As long as this doesn't change the actual color combinations that we recommend for text accessibility, I don't think this needs a separate visual design review. Personally I like the idea of using mixins to make the process easier but still allowing flexibility, although ultimately I defer to the dev team on that one.

joshbruce:
@maya - I'm considering closing this and resubmitting - as I don't think I can do what I want to without doing so.

To increase flexibility even further we could define the root mixin as:

@mixin colors($text: $color-black, $background: transparent)

This work though not tested. This is a better function name semantically for what it does. This also allows us to create a default of transparent with black text. Further, we can set the font color without having a background color. Finally, the mixin maintains the core functionality and flexibility.

Thoughts?

ps. If you know a way for me to update this PR to make that change without having to close and submit anew - please let me know.

msecret:
@joshbruce I think that slightly improves although I think what's here is also fine.

Could you restructure this code based on what we talked about in #1137 , moving the changes from `docs` to `src` and running the build process? Again, sorry for having you do that. We could potentially help out by forking off your fork if you're interested.

maya:
@msecret `usa-bgcolor-*` doesn't comply with our coding styleguide. What do you think would work better?

joshbruce:
@msecret - I think I will want to do a different PR for this one anyway. So, yes, can definitely start switching over to modifying the `/src`.

This PR, as it stands, does not consider the `a` tag, which I believe we would want to. Might be worth reaching out to @carodew again to see if this affects design considerations. Especially since this could apply to links within alerts and the label component.

@maya - That's fair. I will wait to do anything further on this score until hearing back regarding what to name the mixins and if @carodew has anything to add re links.

Sorry I might have jumped the gun on the PR part of this - appreciate the methodical and reflective approach in its consideration.

Would like to leave it open, for now, if possible. Or, convert to an issue. Whichever helps y'all the most with your workflow.

msecret:
@joshbruce it would actually help us a lot to close PRs that aren't being actively worked on because we're working though the large amount of PRs currently submitted.

joshbruce:
@msecret - Know how that goes. Not a problem.

See #1158 

Closing.

PR #1133, opened 2016-04-25
Optimize images

madmanlear:
PNGs are not optimized. This may be a result of the license applied to those images, and if so this PR can be ignored.

- Converted 24-bit PNG files to 8bit with alpha transparency
- Ran through lossless optimization as well

Results in a 25% reduction of total image size with zero noticeable loss in quality.


maya:
Thanks @madmanlear! We have a PR related to this already here: https://github.com/18F/web-design-standards/pull/1106 Have you taken a look at that?

madmanlear:
Gah! I did not, my apologies. I'll look more closely through PRs before submitting. Thanks @maya!

Issue #1075, opened 2016-03-21
Footer window.resize


jflowers45:
On https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/js/start.js
The following code fires an awful lot if you're resizing the window, when in reality it probably only needs to fire once the resize completes. 

$(window).resize(footerAccordion);

Would you be interested in a PR that implemented something like this?
https://css-tricks.com/snippets/jquery/done-resizing-event/

maya:
Hi @jflowers45, thanks for opening this!

Looping in @rogeruiz and @msecret, do you think @jflowers45's approach would serve us well?

msecret:
Hi @jflowers45. I agree that the event is firing too much, but I think the better approach is to find a way to eliminate the JS events altogether. We have a goal in our roadmap of "UI components are comprised of clean, semantic code" which this issue fits in well with.

jflowers45:
@msecret sounds good.  Should I close this issue or leave it open for reference?

msecret:
@jflowers45 let's leave it open for now, I've added it to the correct milestone.

madmanlear:
I added #1134, which addresses this issue, but I agree that the js can be removed altogether. I'm happy to take a pass at a solution.

maya:
Wanted to check in if this was still an issue?

maya:
@el-mapache do you have any idea if this is still an issue?

el-mapache:
@maya I'm not exactly sure what marco's comment above means, but the resize event is being debounced and fires what I think is a totally appropriate amount; I think we can close this.

PR #1134, opened 2016-04-25
Throttle resizer

madmanlear:
The footerAccordion function runs constantly as the window is resized. While this is not a performance issue on such a small scale, I would recommend throttling the function so that it only runs when the resize is complete. This sets a precedent should more complex resize functions be necessary in the future.

- Updated the window resize event callback to clear and set a timeout for the footerAccordion function


maya:
Related to #1075.

msecret:
@madmanlear I saw you were working separately on getting the footer working without JS, can we close this PR in favor of that solution?

madmanlear:
You bet!

Issue #1122, opened 2016-04-14
Accordion content not displayed without JavaScript


msecret:
If Javascript is not available, the accordion content will not be displayed.

Mentioned in #1121 .

jbrucegsa:
See also PR #1066 - comments re same concern.

maya:
Closed via #1224.

Issue #1121, opened 2016-04-14
Accordions Accessibility


scottaohara:
Hi there,

I just wanted to point out a couple of issues I've found with the [accordions example](https://playbook.cio.gov/designstandards/accordions/)

1a. If JavaScript is blocked or unavailable on a device, any accordion content not set to be displayed by default is completely inaccessible. A fix for this would be to dynamically add the aria-hidden* attributes when JavaScript is enabled, but to not include them when JavaScript is not enabled.

1b. Additionally, when JavaScript is disabled, the buttons obviously don't work, and in a no-js solution, these buttons should either not exist, or should actually be rendered as Headings to the accordion panels they are lead-ins to.

2. The accordion panels (.usa-accordion-content) would benefit from having an aria-labelledby attribute. This would require the buttons to have their own unique IDs as well.

*As an aside, resolving the (aria-hidden issue)[https://github.com/18F/web-design-standards/issues/1120] would have implications on this component, and any other component that requires aria-hidden to be set to display none.  For these sorts of examples, aria-hidden would need to be defined as display: none on a per-component basis.  Or, as I mentioned in the aria-hidden issue, these components could instead be updated to use [hidden] set to display none, as that would be the HTML native equivalent of aria-hidden="true" anyway.

Please let me know if there is any further information I can provide to help with this issue.

Thank you

jhfrench:
To take this a little further--consider removing the button element entirely, and instead progressively-enhance `.usa-accordion [aria-controls]` ... something like (quick and dirty in jQuery)

```
$('.usa-accordion')
    //hide all but the first panel
    .find('.usa-accordion-content').not(':first').attr('aria-hidden, true').end().end()
    .find('[aria-controls]').attr('aria-role', '').on('click', function(){
        //show or hide the child panel associated with the clicked panel header
        $('#'+$(this).attr('aria-controls')).toggle(); 
    });
```

This approach provides a usable experience even without JavaScript (but a better experience with it).

jhfrench:
You might also consider incorporating `role="tablist"`, `role="tab"`, and `role="tabpanel"` into the example HTML.

scottaohara:
Agreed with all the above. 

there are also additional keyboard controls that should be implemented here, but I didn't this single issue to balloon out of control.

msecret:
Thanks for all your comments! We've known about issue 1a, I think we have it tracked somewhere (although I can't find it right now so might need to make a new issue).

[...]

Issue #1120, opened 2016-04-13
aria-hidden="true" should not have display: none !important


ducklet:
Hi,

I'm trying to use Font Awesome icons that are purely decorative and sit next to the text a screen reader needs.  Without `aria-hidden="true"`, a screen reader user will hear the same information twice, but `aria-hidden="true"` has been given the css rule `display: none !important`, which totally defeats its purpose.  Can you please fix this?  (Sorry it's a .pdf, but WebAIM recommends `aria-hidden="true"` for these icons: http://webaim.org/presentations/2014/myeducator/POUR.pdf)

Thanks!

ducklet:
cc @edwelker

maya:
Hi @ducklet, thanks for opening this issue! Can you point us to where you're seeing this rule in the code base?

jbrucegsa:
Hi @ducklet, I'm curious about this as I really try to make thing accessible as much as possible. 

Can you show the markup that is causing it to be read by the screen reader?

For example:

<pre><code>
<i class="fa [icon]"></i>
</code></pre> 

or something else.

edwelker:
https://github.com/18F/web-design-standards/blob/18f-pages-staging/src/stylesheets/core/_base.scss#L23

scottaohara:
+1 for this issue.

aria-hidden is quite useful to hide content from assistive technology, but not visually (as ducket points out). 

maybe this should be replaced with the native [hidden] attribute, as that has some older cross browser issues where a display: none!important would be useful.

ducklet:
Hi @jbrucegsa,

The issue is actually not that it will be read by a screen reader, but that I can't use it as intended because the `display: none !important` hides any element it's applied to from sighted users as well.  

Thanks for looking into it.

nickbristow:
This is a good point @ducklet 

We should change how aria-hidden="true" is handled with CSS @maya 

We may need to add to update some things that were being hidden that way.

maya:
Thanks for all the feedback! 

I agree with removing that from the CSS, and updating the instances that were relying on that to hide things in favor of something else (i.e. `hidden` attribute or `display: none!` on a per component basis like @scottaohara mentions in #1121).

maya:
Good news, I searched the Draft Standards and the only place using `aria-hidden="true"` in this way is the accordion component. So if we can solve it there, we can remove this CSS.

jhfrench:
[...]

PR #1137, opened 2016-04-26
no list accordions

### LINKED ISSUE #1136: Accordion semantics
#1120, #1121, #1122 

Not to throw another log on this fire, but:

I was looking at the markup for the accordions and I'm curious about the decision to place them inside a list.

From a dev-perspective it seems the only reason to do it is to be able to add multiple accordions within a single usa-accordion-bordered or usa-accordion div. The reason I say this is that if we place a single accordion inside that those class, **without the li**, the accordion works fine. Almost like the li is used to add a top margin when multiple accordions are at play.

I will see if I can put together a PR for this particular part of the overarching conversation related to accordions. Otherwise, if someone else has the bandwidth before I get to it, by all means.

joshbruce:
Fixes #1136

I am not sure what the conversation around placing each accordion
within a list item was; however, it feels less than semantic to have a
list item containing a button and a div…imagine a text document with a
list item containing an entire chapter of a book, for example.

I’m also curious if a more suitable tag (say, section) would be more
proper, given the specification of a div and section:

http://www.w3.org/TR/html5/grouping-content.html#the-div-element - The
div element has no special meaning at all. … Authors are strongly
encouraged to view the div element as an element of last resort, for
when no other element is suitable.

http://www.w3.org/TR/html5/sections.html#the-section-element - The
section element represents a generic section of a document or
application.

Changing to a section tag could get us toward a no JavaScript solution
to the accordions (#1122 and #1121) - basically, add the buttons and
collapse the content if JS is present.

Could result in “flashing” content though prior to JS kicking in.


maya:
Hi @joshbruce, we'll have to take a deeper look into the semantics of the accordion. Thanks for bringing it up. The team is currently at an offsite, so apologies for the delay.

cc @nickbristow

msecret:
@joshbruce, I commented on #1136, I'm not sure I agree with removing the list aspect from the accordions.

docs/_scss/components/_accordions.scss
```diff
@@ -11,21 +11,6 @@ $accordion-border: 3px solid $color-gray-lightest;
     margin-top: 1rem;
```
msecret: So unfortunately the files in `docs/` are not actually the source files and shouldn't be hand edited. The files in `src` are where the `scss` files should be edited which will then be copied into `docs`.

I'm sorry for this confusion, it was done to limit complications with the Jekyll docs site and we're actually working on fixing it right now. In the meantime, it would help us if you could move your code changes in this file to `src/stylesheets/components/_accordions.scss` and then run the build process with `npm run build` or the more specific gulp tasks.

Let me know if you'd like some help with this. You've been a great contributor so I'd be happy to get these files into the correct form.

joshbruce: Thanks @msecret! 

I thought that's where we were going with gulp.

Would the `gulpfile.js` found here help inform the fix you are looking to make with the gulp build? (https://github.com/BrianGilmanGSA/sam-web-design-standards)

Right now I *cannot* run a gulp build (see #1148) - I can only do `$ ./go serve` to confirm the changes, which doesn't update the `/docs`. If you could help me out on that score, I would be more than happy to make changes and resubmit along with using the `/src` directory for future PRs.

Thanks again.

joshbruce: Sorry, meant the website.js file: https://github.com/BrianGilmanGSA/sam-web-design-standards/blob/gh-pages/config/gulp/website.js

msecret: Sure let's work through it.

What happens when you run `gulp website:serve`? Is there an error you could show me?

joshbruce: I'm going to clone directly from the 18f repo instead of my fork - just to make sure.

1. Cloned 18f/web-desing-standards
1. `$ cd ~/path/to/repo`
1. Completed
1. Ran `bundle update`
1. Completed
1. Ran `npm update`
1. Completed
1. Ran `$ gulp website:serve`
1. Halted with: `jekyll 3.1.3 | Error: File to import not found or unreadable: lib/bourbon. Load path: /path/to/repo/docs/_scss on line 3`

It appears `/src/stylesheets/lib` no longer exists: https://github.com/18F/web-design-standards/tree/18f-pages-staging/src/stylesheets

On the initial clone, `/docs/_scss/lib` does exist: https://github.com/18F/web-design-standards/tree/18f-pages-staging/docs/_scss

When `$ gulp website:serve` removes `/docs/_scss` the `/docs/_scss/lib` directory is removed and not replaced.

Let me know if you experience something different - or, of course, if I'm doing this incorrectly.

msecret: Cool, so instead of `npm update` you should run `npm install`. See what happens after that.

joshbruce: Fascinating.

Still needed the `$ bundle update` - just in case that shouldn't be necessary.

Can you point me to *why* install versus update makes the difference?

Reading the Terminal didn't seem much different.

msecret: @joshbruce first off, did that solution work for you at all?

Based on the documentation on npm, and what I've read, there doesn't seem to be a large difference(http://stackoverflow.com/questions/12478679/npm-install-vs-update-whats-the-difference), I gave you that command just to ensure we're both doing the same thing to make debugging a little easier.

joshbruce: @msecret - Sorry, should have been more explicit.

Yes. The build completed.

I'm not sure what difference there was, but it was enough for `/src/stylesheets/lib` To appear. Not sure why though, as I thought install and update only effected the `/node_modules` directory.

joshbruce: @msecret - Just a quick follow up - I'm a curious soul - sorry for blowing up your inbox. Wanted to show some screenshots to isolate *when* it happens.

Before `$ npm install`

screen shot 2016-05-09 at 5 03 29 pm

After `$ npm install`

screen shot 2016-05-09 at 5 05 04 pm

The point at which the `/src/stylesheets/lib` directory appeared.

screen shot 2016-05-09 at 5 05 18 pm

Looks like `$ npm update` does not run the `$ gulp copy-vendor-sass` - from the npm `prestart` (??), which makes sense. To test that hypothesis - I did the following.

1. `$ bundle update` - to avoid the kramdown error (must be done prior to `$ npm start`.
1. `$ npm update`
1. `$ npm start`

It worked. Which makes sense as calling `$ gulp website:serve` does not appear to call `$ gulp copy-vendor-sass`. If the goal is to *not* call `gulp` directly - then it works as expected.

msecret: Ah @joshbruce I figured it out:

In `package.json` we have a `prepublish` script that executes certain commands after `npm install` (which doesn't make much sense). I guess those have to be run before `website:serve`.

I'll keep this in mind when working on improving the dev workflow this week.

ducklet:
@joshbruce I created https://github.com/18F/web-design-standards/pull/1153 to restore the functionality of `aria-hidden`, which should remain in place.  I believe it makes sense to add the `display: none !important` rule to the `<div>` when the `<button>`'s `aria-expanded` state is `"false"`, but I don't know javascript.  When you re-open/restore this PR would you please be so kind as to add the appropriate code?  I believe this would be the last step in resolving #1120.   

Thanks, and of course kit with any comments/questions, I hope I've thought of everything here :)

Issue #794, opened 2015-10-06
using h4 in footer


goetsu:
https://playbook.cio.gov/designstandards/footers/#big-footer

Currently on the code example you use an h4 element inside an ul element and it's not conforming to html who only allow li element inside ul element. You need to move the h4 before the ul element or use two levels lists.
Furthermore, using h4 may break the document outline if you don't use h1, h2 or h3 before it

maya:
Thanks again.

TODO: add this to the documentation:

`Ensure the heading level starts from highest to lowest so that you do not break the document outline.`

KimberlyMunoz:
I found that http://khan.github.io/tota11y/ was a great tool to test the heading levels when I was doing an accessibility check across our project.

mollieru:
Thanks @goetsu and @KimberlyMunoz -- looking into this now, we'll keep you posted. 

// cc @maya

maya:
Fixed in #1183.

PR #1140, opened 2016-04-29
Remove footer js and update footer list html

madmanlear:
The big footer (with collapsing navigation on small screens) had an unnecessary reliance on javascript. Further, link lists in this footer included an h4 as a direct child of a ul, and this is not allowed.

- Remove javascript requirement for big footer
- Relying instead on css, show all list items with the parent is hovered. This automatically defaults to being triggered on click on touch-only devices
- Wrapped footer h4 in li to comply with semantics

### DIFF
--- a/src/js/start.js
+++ b/src/js/start.js
@@ -56,29 +56,6 @@ $(function() {
     new Accordion($(this));
   });
 
-  var footerAccordion = function() {
-    if (window.innerWidth < 600) {
-
-      $('.usa-footer-big nav ul').addClass('hidden');
-
-      $('.usa-footer-big nav .usa-footer-primary-link').unbind('click');
-
-      $('.usa-footer-big nav .usa-footer-primary-link').bind('click', function() {
-        $(this).parent().removeClass('hidden')
-        .siblings().addClass('hidden');
-      });
-    } else {
-
-      $('.usa-footer-big nav ul').removeClass('hidden');
-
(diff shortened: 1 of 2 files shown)


maya:
Hi @madmanlear, sorry for the delay - the team is currently at an offsite. Thanks so much for submitting this PR!

One issue is that hovering is not accessible by screen readers, so unfortunately won’t be a sufficient solution for us. I think there are usability issues with hovering as well, although @carodew would know more.

We'd love to fix the document outline issue you mentioned about the h4 incorrectly located within a list. This would resolve #794 if it's possible to bring it into another PR, we'd appreciate it.

madmanlear:
Hey @maya, no worries!

Good point about screen readers. Could we target the "speech" media type to ensure that the lists are not hidden to screen readers? Or perhaps even better would be to use the "screen" media type and only hide lists when it's present.

If neither of those are acceptable then we're definitely back to requiring js, as far as I can tell.

Let me know your thoughts and I'll prep a PR for the h4 fix. Thanks!

carodew:
Hovering is not accessible to screen readers or to keyboard users, but it's also more difficult for mouse users who have motor difficulties (think hand tremors). Overall I'd avoid using hover to expose essential interactions on any app, but especially navigation. 

I'm also tagging @bradnunnally since he is taking over the UX role on this project.

madmanlear:
Thank you, @carodew, that makes perfect sense. I'll close this and move the h4 edit to a new PR. Thank you both for your guidance!

PR #1143, opened 2016-04-29
Fix broken JS references

### LINKED ISSUE #1142: jQuery scoping problems
When I add the Draft Standards to a new project, the accordion component doesn't work. 

Here's the error:

```js
uswds.min.js:3 Uncaught ReferenceError: $ is not defined
```

I'm using the JS provided in the `dist` folder. I talked with @msecret and problem is around how jQuery is scoped. 

Even though jQuery is included in the JS file, to get it to work, you have to add jQuery script again before the uswds JS in your document. One fix would be to require jQuery for every component that uses it. 

@msecret said @rogeruiz may have some ideas around this. One issue with that method is if someone is using jQuery on an existing project already.

msecret:
We have to require both jQuery and `toggleFieldMask` in the files that
need them.

Fixes #1142.


juliaelman:
LGTM 👍 thanks for hoping in and making this fix @msecret!

PR #1153, opened 2016-05-09
Fixes issue #1120 by removing display: none from aria-hidden, doesn't…

ducklet:
… impact anything but accordion component which should be addressed separately (see https://github.com/18F/web-design-standards/issues/1121, https://github.com/18F/web-design-standards/issues/1122, et al).

Regardless of how the accordion issue is solved, it is not appropriate for aria-hidden=true to have additional rules; instead, a class with display: none should be used where needed.

Thanks!

### DIFF
--- a/src/stylesheets/core/_base.scss
+++ b/src/stylesheets/core/_base.scss
@@ -19,7 +19,3 @@ body {
     filter: none !important;
   }
 }
-
-[aria-hidden=true] {
-  display: none !important;
-}


maya:
Hey @ducklet! Thanks so much for submitting this. In order for us to not break the accordion component, we'll need to fix that before adding this in.

joshbruce:
@ducklet - Per the request on #1137 - I'm curious, as I don't know if it's possible.

- Can I check out this branch and update this pull request directly? 
- Or, do I need to create another PR entirely?

I would prefer the former as it would be easiest. I've done some very brief Google-fu and saw that it was a feature request, but not finding instructions on doing it.

Maybe @maya knows?

maya:
@joshbruce opening a new PR would be best! I'm going to close this out since it doesn't have what we need in order to accept it, i.e. we can't fix one thing and break another.

**But we agree with the intent of this 💯 .**

PR #1159, opened 2016-05-10
[WDS-1156]: Adds mocha/jsdom specs

el-mapache:
* Adds mocha and JSDOM for unit testing components
* Adds tests for JS components
* Refactors togglesXField components into a single component

There are a million changed filed here from the build process, here are the relevant ones:

* config/gulp/test.js
* gulpfile.js
* spec/
* src/js/start.js
* src/js/components/validator.js
* src/js/components/toggle-input-field.js
* docs/_components/form-templates.md


spec/unit/accordion/template.js
```diff
@@ -0,0 +1,9 @@
+var template = [
```
msecret: An interesting post-MVP idea to explore here: have this file be a standard HTML file or handlebars file. This would bring us closer to having a components HTML separated with the component, in the same way [fractal](https://github.com/frctl/fractal) does.

el-mapache: I would love for this to be an actual HTML or compiled-to-HTML file, although I'm not clear exactly how the consuming client would be able to drop that in to their application without have a build process themselves. I guess that's the questions everyone is asking though :D

PR #1164, opened 2016-05-11
Ms remove dist dir

msecret:
First draft of attempt to remove dist directory

- New folder `docs/doc_assets` holds all js, css, img and font assets that are specific to the docs site and not WDS.
- Attempts to ensure the `dist` directory is build before doing anything with docs/website tasks.
- Copies the js, fonts, and imgs from `dist`to `docs/assets/`
- Copies the js, css, fonts, and imgs from `docs/doc_assets/` to `docs/assets`
- References `src/stylesheets` for `scss` paths, and `homepage.scss` and `styleguide.scss` `@import` `all.scss` from this dir rather then a copied over file.
- Runs the jekyll build, by now everything should be stored in `docs/assets`.
- Ignores `docs/assets` and `dist`.


.gitignore
```diff
@@ -19,9 +19,8 @@ npm-debug.log
 # like to include it in the repository for the time being. The `dist-gem/`
 # directory though will keep being ignored for the time being.
 #
-!dist/
+dist/
 dist-gem/
```
juliaelman: Remove `dist-gem/`, as it has been removed and is no longer maintained. See https://github.com/18F/us_web_design_standards_gem/issues/15#issuecomment-203564667

.gitignore
```diff
@@ -19,9 +19,8 @@ npm-debug.log
 # like to include it in the repository for the time being. The `dist-gem/`
```
juliaelman: Remove this comment about the `dist/` and `dist-gem/` directories.

config/gulp/website.js
```diff
@@ -87,28 +81,54 @@ gulp.task('copy-bundled-javascript', function (done) {
...
+
   });
 
-  copyFonts.on('error', function (error) { done(error); });
+});
+
+gulp.task('make-images-dirs', function(done) {
+  var makeImagesDirs = spawn('mkdir', [
+    '-p',
+    'docs/assets/fonts'
```
juliaelman: Should be `docs/assets/img`

juliaelman:
Nice work and good start! A few notes from my initial pass:

- The following files need updates on where users can now download the compiled version of the standards:
  - `.github/CONTRIBUTING.md`
  - `/docs/pages/getting-started.html`
  - `README.md`
- If I am understanding this correctly, we are using the `dist/` directory as a `.tmp` dir. If so, let's maybe move towards that pathway of the `tmp` holding tank to help build out what we need. It looks like we have a task started for that in `config/gulp/release.js`.
- There was an issue with image paths when building the site locally. I tested out updating the `make-images-dir` locally and I *think* this may resolve the issue.

//cc @rogeruiz would be great to get your 👀  on this too!

msecret:
@juliaelman I'm not actually sure what would need to happen in your second point, `dist` as a `tmp` dir. Also: it's a temp dir in that it doesn't get checked in, but if you're developing locally or download the zip or from npm, it should exist.

rogeruiz:
@juliaelman @msecret I can see the confusion around the `dist/` directory name for the directory now that we're not checking it in. _We could rename it as a `tmp/` directory_. We would also need to update a few things about the build.

- Update file paths for the various copy / build / release commands
- Since the directory is ephemeral, it should be deleted when the build finishes. 🐣 
- Update various documentation to point to the new `tmp/` directory

🐣 : This would keep it inline with being a temporary directory. This would also affect where we store the `uswds-*.zip` file after we build it. _This may not be needed if we decide it's easier to keep the directory around_.

I'm not sure if going the route of updating the name from `dist/` to `tmp/` has any major implications and it mostly comes down to semantics, but it does require us to change a few things about how the package is compiled / bundled, the website is built, and the package is released. The `dist/` / `tmp/` directory is a copy of the distribution files for the package __and__ is used as a temporary directory when building some of the files for the documentation website. Some files are placed in `dist/` compiled and ready to be archived into a zip directory ( I think it's specifically only the CSS styles that are not copied out of the `dist/` directory since the documentation site works with the Sass style files ). [Other files](https://github.com/18F/web-design-standards/blob/18f-pages-staging/config/gulp/website.js#L66-L139) in the same directory are later copied using the `gulp website:build` from the `dist/` directory into the `doc/` directory.

All of the above taken into account, [the `gulp release` process already creates a temporary directory named after the package and the current version number](https://github.com/18F/web-design-standards/blob/18f-pages-staging/config/gulp/release.js#L9-L30). This folder is created, archived, and then deleted. I propose something similar to that for a temporary directory since it communicates that these files shouldn't be relied on.

The original intent around the `dist/` directory was to make sure that if someone was building the project from source, they would have the files compiled / bundled on their local machine in a place where they can verify it's contents.

Because so much of the pipeline is dependent on the `dist/` directory being named so and having `dist/` exist locally for users would want to build the project from source, I think it makes more sense to keep the names of the `dist/` directory as is and follow up on updating the documentation related to where users can now download the compiled version.

juliaelman:
@rogeruiz thanks for your thoughts on the directory structure! I spoke with @msecret about this to get a better understanding of the intended goals are for this pull request. Here are some of the answers:

1) The zip file creation is still going to be a manual process.
2) The `/dist` folder will likely have it's assets copied around.
3) The files included in the `npm` package would stay the same.

Also, thanks all for touching base this morning too! Sounds like we are moving forward, adding some documentation via another PR and working on some fixes.

rogeruiz:
💯

PR #1167, opened 2016-05-13
Fix broken links & typos in README & CONTRIBUTING

yozlet:
Fixes a couple of doc links between the top-level dir and the `.github` dir.


juliaelman:
🎉 nice catch @yozlet!

PR #1179, opened 2016-05-20
[Website] Stop copying of vendor libs around

msecret:
## Description
Removes the build step of copying the CSS vendor libs to the src/stylesheets folder by using a full path to them to node_modules.

## Additional information
So node-sass can import .css files but ruby sass cannot, so I want
to ditch the annoying normalize lib for the scss version.

I'm using the straight node_module paths, which isn't the perfect
solution. Something better but would be much more complicated to
get working on sites that use the standard is to set the includePath
in the sass build path. I'm hesitant to do this now because node-sass
and ruby-sass use these differently so I'm not sure how we'll use them
both.

### Benefits
- A user can install with npm directly from the github repo, `@import` the src files into their project and everything will work
- There's no non-source code in `src`.

### Negatives
- If bourbon, neat or normalize changes where the main scss files are, then the imports will break

TODO
- Remove unnecessary build steps.

### DIFF
--- a/config/gulp/sass.js
+++ b/config/gulp/sass.js
@@ -33,24 +33,6 @@ gulp.task('scss-lint', function (done) {
 
 });
 
-gulp.task('copy-vendor-sass', function (done) {
-
-  dutil.logMessage('copy-vendor-sass', 'Compiling vendor CSS');
-
-  var stream = gulp.src([
-    './node_modules/normalize.css/normalize.css',
-    './node_modules/bourbon/app/assets/stylesheets/**/*.scss',
-    './node_modules/bourbon-neat/app/assets/stylesheets/**/*.scss',
-  ])
-    .pipe(normalizeCssFilter)
-      .pipe(rename('_normalize.scss'))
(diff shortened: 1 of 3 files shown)


juliaelman:
Thanks @msecret! Ran the site locally and it looks like the homepage might not be pulling in `normalize.scss`, as there is a white border showing up around the edge. This does not occur on the interior pages of the Standards.

Screenshot below from Chrome. Also seeing in the latest versions of FF/Safari:

screen shot 2016-05-20 at 4 17 51 pm

rogeruiz:
Are there any implications to building the project locally within another project that installs the `uswds` package? I remember we ran into an issue with that because of `npm@v3`'s flat dependency structure, @msecret.

juliaelman:
After discussing with the WDS core dev team, we're deciding to hold off from moving forward with this for the time being. Due to the way `Bourbon` and `Bourbon Neat` are consumed by the WDS, we're needing to re-think how we would approach this solve.

PR #1160, opened 2016-05-10
Fix aria hidden

joshbruce:
Remove from _base.scss: aria-hidden, display:none
Add class to _accordion.scss: usa-accordion-content-collapsed, display:none
Update accordion preview classes
Update accordion.js: Add remove class to expand and collapse.
Update styleguide.js to display code accordion collapsed on load.

Issue #1128, opened 2016-04-19
Gulp setup fails...


jhfrench:
...on my system with with error "bundle && gulp copy-vendor-sass && gulp build" while attempting to follow the ["Building the project locally with Gulp" instructions](https://github.com/18F/web-design-standards/blob/18f-pages-staging/CONTRIBUTING.md).

Before filing this issue I have:
- [updated my node install](http://stackoverflow.com/a/31520672/1430996) 
- [searched the project's issues for "copy-vendor-sass"](https://github.com/18F/web-design-standards/issues?utf8=%E2%9C%93&q=%22copy-vendor-sass%22) to make sure there isn't already guidance

Selected content from my npm-debug.log:

```
98 error node v5.6.0
99 error npm  v3.8.7
100 error code ELIFECYCLE
101 error uswds@0.9.1 prepublish: `bundle && gulp copy-vendor-sass && gulp build`
101 error Exit status 1
102 error Failed at the uswds@0.9.1 prepublish script 'bundle && gulp copy-vendor-sass && gulp build'.
102 error Make sure you have the latest version of node.js and npm installed.
102 error If you do, this is most likely a problem with the uswds package,
102 error not with npm itself.
102 error Tell the author that this fails on your system:
102 error     bundle && gulp copy-vendor-sass && gulp build
102 error You can get information on how to open an issue for this project with:
102 error     npm bugs uswds
102 error Or if that isn't available, you can get their info via:
102 error     npm owner ls uswds
102 error There is likely additional logging output above.
103 verbose exit [ 1, true ]
```

If you want the whole log file, please let me know how to get it to you (I prefer to not share it here).

msecret:
@jhfrench I think it might help to see more output, as I can't find the reason the command failed here.

Another issue currently with our prepublish hook: it requires ruby and bundler to be installed. Could that be the issue?

jhfrench:
@msecret I don't already have ruby and bundler installed on my system (if needed, can instructions be added to CONTRIBUTING.md?)...but this particular issue sure looks like it's isolated to the Gulp install/instantiation/[pedantically-correct verb].

I *suspect* that some of the dependencies reference deprecated Gulp package versions, and the above is the result of that.

If you [try a fresh install](https://github.com/18F/web-design-standards/blob/18f-pages-staging/CONTRIBUTING.md) within the latest NPM, does it work for you?

madmanlear:
[...]

PR #1192, opened 2016-05-25
[UI Component]: Fix aria hidden

### LINKED ISSUE #1120: aria-hidden="true" should not have display: none !important

### LINKED ISSUE #1181: [UI Component]: use toggleClass for accordions
## Issue type

* UI component

## Description

#1160 - not merged as of this writing.

Causes a change in the JavaScript for toggling the accordions using class names instead of the value of aria-hidden. Might be possible to use jQuery .toggleClass() instead of two function calls (one to remove and one to add). Believe, under the hood, jQuery does the same thing; so, this might fall under refactoring.

Would like to see how high a priority this change would be. And, if there is a performance hit over the current PR.

n/a

joshbruce:
See PR #1160

Fixes #1181 - toggleClass(), within the hide or show functions, causes
double-click requirement - if show-hide functionality is encapsulated,
toggleClass(). Encapsulation would probably reduce code readability.

Fixes #1120


joshbruce:
Adding from #1160:

@msecret, @ducklet, @rogeruiz, @maya

yozlet:
Hey, thanks for this!

I'll leave it to my colleagues to review the rest of the commit, but could you revert the inclusion of `Gemfile.lock`? Those gems are due for an update but we should do that in a separate commit in case it breaks something.

msecret:
Hi @joshbruce, thanks for the PR. What do you mean by "causes double-click requirement"?

src/stylesheets/components/_accordions.scss
```diff
@@ -90,3 +90,7 @@ $accordion-border: 3px solid $color-gray-lightest;
     margin-bottom: 0;
   }
 }
+
+.usa-accordion-content-collapsed {
```
msecret: Is there a reason we need the accordion to have it's own class for display none, or should this be put in a general utility?

maya: What about what's done here: https://github.com/frend/frend.co/blob/gh-pages/_components/accordion/accordion.css

```
.fr-accordion__panel[aria-hidden="true"] {
  visibility: hidden;
  height: 0;
}
```

joshbruce: @msecret - Interesting suggestion. I'm not sure I'm seeing the utility solution that you might be. If we make a `mixin` for `display: none`, wouldn't we still need to `@include` it somewhere that distinguishes the two states? 

@maya - Seems an interesting alternative. After looking at the JavaScript for the Frend accordion, it appears, to overcome setting `height: 0` bye calculating heights for the expanded accordions - might need to test and verify. If so, it would be more JavaScript dependence to make the accordions expand (#1121 & #1122) - unlike the `display: block` route. See [lines 90 through 122](https://github.com/frend/frend.co/blob/gh-pages/_components/accordion/accordion.js).

maya: @joshbruce you're right, there may not be any benefit to the `visibility: hidden` method, that I could think of, but it might require more research. Anyhow, I tried a simple, classless method here: #1196. I'd like to hear what others think.

joshbruce:
@msecret - Good question.

When I used the following with `hide`:

`$content.toggleClass('accordion-content accordion-content-collapsed');`

And the following to `show`:

`$content.toggleClass('accordion-content-collapsed accordion-content');`

It would require hitting the accordion button twice to actually get it to collapse or expand, respectively. Therefore, I may have misinterpreted what `toggleClass()` really did under the hood. It appears it will either add, or remove, a class if the class is set or not; however, it will not change one class name *into* another class name.

joshbruce:
@maya and @msecret - Updated PR, with both the simple (#1196) and utility concepts. Can always rollback and update PR, if desired. However, I do appreciate Maya's simple solution.

joshbruce:
@yozlet - Hey. Sorry, I missed your request for the revert on `Gemfile.lock`. Because I had to start fresh from a new fork - long story - I had to run `$ bundle update` (tried running just `$ npm install` - did not seem to work). See my comment on #1128 

> Ran into this error today when doing
> $ gulp website:serve
> website:serve jekyll 3.1.3 | Error: kramdown
> Not sure why, given that kramdown is not used - unless it is a dependency for red carpet.

And, of course, `$ bundle update` rewrites the `Gemfile.lock`.

In PHP (using Composer) we sometimes ignore `composer.lock` file in `.gitignore` to try and overcome this. 

If there's something we can do to make a fresh clone (or fork) work without needing `$ bundle update` - let me know; otherwise, I'm not sure I can revert the file by itself.

src/stylesheets/components/_accordions.scss
joshbruce: @maya - Saw that. It's  good solution. Incorporated it here. Also tested just doing `height: auto`, which did work - no JS changes needed. I do like the height solution from the screen-reader perspective - though I'm not sure if `visibility: hidden` would cause it to be skipped. Which brings up the question: Do we want accordion content to be hidden from screen-readers?

src/stylesheets/core/_utilities.scss
```diff
@@ -16,6 +16,13 @@
   left: -999em;
 }
 
+// Aria hidden helper
+@mixin aria-hidden() {
```
msecret: I think when I said a global utility, I mean a class, such as `.hidden`. Sorry for the confusion. Although, what might be the benefits of using a mixin like this? It might be a better solution.

src/stylesheets/core/_utilities.scss
```diff
@@ -16,6 +16,13 @@
...
+  &[aria-hiddent=true] {
```
maya: This says "aria-hidden**t**=true" - there is an extra t.

src/stylesheets/core/_utilities.scss
maya: @msecret I'd be careful about calling it hidden, so it doesn't get confused with the `hidden` attribute. Although that might have been just an example or it might be ok?

Also not sure I'd call the mixin `aria-hidden` either bc its not the true function of aria-hidden, just a special use case for this situation. Curious to here what are the benefits of using the mixin is too. Are we reusing this somewhere else?

maya: @msecret: @joshbruce explained the benefits of the mixin in #1196:

>To allow developers to `@include` within any class or element; thereby, allowing any element to opt-in to being toggled based on `aria-hidden` state based on CSS

joshbruce: @msecret & @maya - I think having it as a mixin also establishes a pattern of how to handle `display: none` related specifically to accessibility considerations - similar to the `unstyled-list` mixin.

I don't think this particular pattern is used with any other UI component. I think the problem we were trying to solve initially was that `aria-hidden` shouldn't `display: none` on elements that use that attribute setting. @maya's simple solution solves this problem and limits its scope to the `accordion-content` class. This PR extends the flexibility of that implementation, which may be a pre-optimization (capability with no demand from developers, internal nor external)...??

yozlet:
@joshbruce I believe `git checkout 35bc35b Gemfile.lock` will revert the file to the previous version, but it'll need another commit. It's weird that `bundle install` didn't work, though! Can you try it again after doing the revert? (And whether it works or not, I'd much prefer we leave `Gemfile.lock` as the old version for this PR, but we can update it in another one.)

src/stylesheets/core/_utilities.scss
msecret: Thanks for the explanation @joshbruce , I think you've convinced me that the mixin is a good way to go. I'd agree with @maya that we should come up with a better name for it though.

maya: @joshbruce can you change the mixin name to `accessibly-hidden`?

joshbruce:
@yozlet - Done. Thanks for the instructions there. It appears there is a deletion from what in the current project: "bundled with 1.12.3"

@maya - `@mixin` name changed.

maya:
@joshbruce `accessibly-hidden` please, that says `accessibility-`.

joshbruce:
@maya - Sorry. Thought that was a typo. :)

yozlet:
@joshbruce: Thank you! Nearly done - `go_script` needs to come out too (it was pulled out in the main repo 5 days ago). Sorry, it's a moving target!

ducklet:
@maya @joshbruce sorry I've been mia; I'm starting to feel like the wheels came off the wagon a bit with the aria-hidden/display none rules...  why are we tying `display: none` into `aria-hidden` at all?  Having both is redundant.  Imho it would make more sense to simply create a class with rule `display: none`.  

I'm also a bit confused by the request to use class name `accessibly-hidden`.  Maybe there's another logic that's escaping me, but content with `display: none` is inherently not accessible, to anyone.

I understand the impetus behind not simply removing the `display: none` rule from `aria-hidden`, but I fear we may be dragging this out and adding unnecessary complexity to the accordion fix.  Could we consider just a "use this to hide something until it's exposed" class, be it specific to the accordion or more versatile for other potential future usages?  

Thanks for your consideration and esp @joshbruce your great work and persistence in making this fix happen!  I am able to override the rule for the time being but putting `!important` in our code to override `!important` feels icky :)

joshbruce:
@yozlet - Did a copy-paste from `staging` as of today.

@ducklet - I asked a similar question in the comments and have decided to leave it to the judgement of the Standards team. I believe this particular concept is the notion of an assistive technology consideration driving sighted user's experience. So, instead of `display: none` (sighted experience) being the driving force for assistive technologies, we allow assistive technology state `aria-hidden=true` to drive the visual experience. This would help adopters use this accessibility-first pattern more easily.

I believe the choice for naming is based on the fact that it is directly tied to the `aria-hidden=true` state; therefore, directly related to an accessibility consideration - not just `display: none`.

From looking at the line notes - the additional class was the original solution provided - the concern there was creating "yet another class" for adopters to use and remember. Though this fix does not necessarily remove the learning it - it does keep the class list one smaller than the other solution - while allowing adopters to add the @mixin to an element they so choose; therefore, adding `display: none` to *any* element with `aria-hidden=true`.

I agree that `!important` is icky - and I believe the Standards has attempted to remove its use altogether.

Cheers! Looking forward to seeing how this goes. :)

maya:
@ducklet What @joshbruce stated is correct. The mixin allows us to hide things dynamically based on the `aria-hidden` state -- it's only to be used for specific components/use cases, not for everything.

maya:
These changes look good to me. Thanks again for your help, @joshbruce! 🎉

ducklet:
@maya this is merged but just something to think about moving forward, the problem with the abstract concepts we're now discussing is that bottom line, aria-hidden would never be used with display: none.  WebAim cautions that mis-using ARIA roles can be as bad or worse than not using them, and ignoring the fundamentals breeds bad practices.  I don't mean to harp but I do feel strongly that attaching our own interests to something like this is a slippery slope.

PR #1196, opened 2016-05-26
[UI Component]: Fix aria hidden (simple)

maya:
## Description

Removes `display: none` from `aria-hidden="true"` and moves that to `usa-accordion`.

Fixes: #1120.

### DIFF
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -89,4 +89,9 @@ $accordion-border: 3px solid $color-gray-lightest;
   > *:last-child {
     margin-bottom: 0;
   }
+
+  &[aria-hidden=true] {
+    visibility: hidden;
+    height: 0;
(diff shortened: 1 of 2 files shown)


joshbruce:
@maya - Very nice alternative - didn't think of it myself until your comment. Well done!

@msecret - If I'm thinking about this solution, combined with the utility suggestion from #1192 we would get something like:

```
@mixin aria-hidden() {
  &[aria-hidden=true] {
    display: none;
  }
}
```

To allow developers to `@include` within any class or element; thereby, allowing any element to opt-in to being toggled based on `aria-hidden` state based on CSS, yeah?

maya:
@joshbruce @msecret where else are we using "allowing any element to opt-in to being toggled based on `aria-hidden` state" on the site?

maya:
Will use #1192 instead of this.

PR #1198, opened 2016-05-26
[WDS-1193]: Fixes javascript issues in styleguide

el-mapache:
Fixes javascript errors in styleguide as described in issue 1193: https://github.com/18F/web-design-standards/issues/1193
Use the title line as the title of your pull request, then delete these lines.

## Description

* alias jquery to window.$
* uswds.min.js file above scripts.html to ensure jQuery is available first
* conditionally instantiates plugins we only use on the /form-templates/ page.


maya:
@el-mapache thanks! I just checked and the code sample boxes are there, but not opening on click.
screen shot 2016-05-25 at 9 09 21 am

maya:
The homepage uses some polyfill scripts in `docs/_includes/scripts.html` but likely does not need the rest (styleguide.js, prism.js, or ethnio).

We can either create a scripts-homepage.js with just those, or add them manually to `layouts/homepage.html` cc: @rogeruiz

src/js/start.js
```diff
@@ -85,7 +85,9 @@ $(function () {
     $(this).attr('tabindex', '-1');
   });
 
-  toggleFormInput($('.usa-show_password'), 'Show Password', 'Hide Password');
-  toggleFormInput($('.usa-show_multipassword'), 'Show my typing', 'Hide my typing');
-  validator($('.js-validate_password'));
+  if (/form-controls/.test(window.location.pathname)) {
```
msecret: Just realized something about this:

The code compiled from `start.js` is distributed as part of the WDS separate from the docs site, but this code is assuming you're on the docs site because it's testing for URLs of the docs site. So if a user had the standards on their own site, this if statement wouldn't be called correctly because they won't have a `form-controls` page (it's really confusing knowing how the WDS is separate from the docs site)

What is this check attempting to protect from? Would it make sense to uses classes rather then URL's to test for existence?

rogeruiz: I agree @msecret good catch. This should be targeting `class`es or `id`s related to `form-controls`.

rogeruiz:
I added a small commit to move Ethnio into `analytics.html` which is outlined [in the commit](https://github.com/18F/web-design-standards/pull/1198/commits/aadf2dae204d2a95c46a0f8a8781ca5cf5afa083).

src/js/start.js
el-mapache: The check is to prevent the code from instantiating those plugins if the DOM elements aren't on the page. We could definitely just check individually for the existence of those elements.

I used URLs because it was faster than checking for each element individually, and because I didnt consider that other users might be installing the docs :(

rogeruiz: Well the `start.js` isn't a part of the docs but rather a part of the package. It can be very confusing. Sorry @el-mapache.

el-mapache:
@maya Is it crazy to unify the start.js and styleguide.js files? The code samples accordion isn't opening because that content is dynamically added in the styleguide.js file. But, the accordion plugins are initialized in start.js. 

If unifying these files isn't possible we could put our accordion plugin under an 18F namespace on the window object.

rogeruiz:
@el-mapache It wouldn't be a good idea as the `styleguide.js` code isn't actually a part of the `uswds` package. Currently, the code in the package doesn't instantiate the components it ships. Also the docs don't have any documentation alluding to that fact either, _I don't think_.

juliaelman:
@rogeruiz just checked aadf2dae204d2a95c46a0f8a8781ca5cf5afa083 and looks 👍 ! Makes sense to move the Ethnio script into the analytics include.

src/js/start.js
el-mapache: Ah, so, in that case, we should probably move docs specific code to the styleguide.js file, correct?  Initializing the all the plugins with their respective DOM elements wouldn't relevant to the overall package, just to the docs pages.

el-mapache:
@rogeruiz  That makes sense. So, we should still be able to move the docs page specific plugin initialization from start.js and put it into styleguide.js?

rogeruiz:
@el-mapache absolutely. that would be good to have in this PR to help avoid this in the future. The `docs/**/styleguide.js` and the `src/**/start.js` have very different responsibilities.

rogeruiz:
@el-mapache [This commit](https://github.com/18F/web-design-standards/pull/1198/commits/51b87366846ec7d42623629de36370be8fdebe4a) seems to have reverted all the changes rather than just some of them. Only the jquery needed to be brought back, but I do like the restructuring of `_includes/scripts` to `_includes/doc-scripts`. I have time to pair on this later in the day if you'd like. We should not include the `docs/_scss` directory in this PR as we ~~fought~~ coordinated to remove it in the PR from @msecret #1164.

maya:
Also, the styleguide site and homepage have different JS requirements @rogeruiz @el-mapache - i.e. the homepage doesn't use JS, while the website does.

Issue #1182, opened 2016-05-21
Custom Checkboxes and Radio elements don't appear to be selectable on iOS VoiceOver


slowbot:
## Custom Checkboxes and Radio elements don't appear to be selectable on iOS VoiceOver
While working to implement checkboxes and radio elements in we recently discovered that checkboxes and radios were not selectable on iOS with Voiceover. 

The issue seemed to stem from how the sr-only mixin was using text indent. We opted for setting the opacity of the checkbox/radio to 0 and repositioning it with a negative margin. Let me know your thoughts if anyone has had the same problem and if this solution works for you as well. 

input[type="checkbox"],
input[type="radio"] {
  // @include sr-only();
  opacity: 0;
  position: absolute;
  margin-left: -20px;
}

maya:
@slowbot thanks so much for opening this! Looks like we may be implementing your fix for checkboxes and radio buttons, unless we find something else. 

cc: @nickbristow

PR #1203, opened 2016-05-26
[Checkbox and radio inputs] Fix checkbox and radio inputs from not appearing selectable on iOS VoiceOver

maya:
## Description

Fixes custom checkboxes and radio element issue where they didn't appear on iOS VoiceOver.

Thanks @slowbot for the fix!

Fixes: #1182

cc: @nickbristow


nickbristow:
Reviewed and works now with iOS voiceover. 👍

Issue #353, opened 2015-08-25
[Tables] Example runs off the edge of the screen (multiple devices)


colinpmacarthur:
The example table runs off the edge of the example box:

Android:
![tables run off edge of page](https://cloud.githubusercontent.com/assets/2092076/9480984/869dc9c2-4b54-11e5-9a7c-d01854081603.png)

iPhone:
![table iphone display](https://cloud.githubusercontent.com/assets/2092076/9481039/0a667394-4b55-11e5-979c-f4bfeb5020c5.PNG)

## Devices
- Moto E, Android 4.4, Chrome 44
- iPhone 6, iOS 6
- Samsung S5, Android 5 Lollipop, Chrome

carodew:
We need to decide what the fix is for this.

jpyuda:
Let's make the `<td>` cells white instead of transparent to at least get rid of the border bleeding through.

There may later be additional design changes aiming to fix the margins.

joshbruce:
One option would be to wrap tables in a `div` (or, preferably, different element). See UIKit's "Responsive Table" at the bottom of the page:

http://getuikit.com/docs/table.html

maya:
@jpyuda are you thinking we should make `<td>` cells white for the table component or custom styleguide CSS?

msecret:
@maya the table component. This is something that will come up outside of our docs.

PR #1209, opened 2016-05-31
[Tables] Fix table background overlap

maya:
## Description

Prevents page background from interfering with table background. 

## Additional information

This is what it looks like:

screen shot 2016-05-31 at 12 25 17 pm

Fixes: #353.

### DIFF
--- a/src/stylesheets/elements/_table.scss
+++ b/src/stylesheets/elements/_table.scss
@@ -4,7 +4,7 @@ table {
   min-width: 100%;
 
   thead {
-    tr {
+    th {
       background-color: $color-gray-lightest;
     }
   }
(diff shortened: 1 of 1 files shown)


src/stylesheets/elements/_table.scss
```diff
@@ -4,7 +4,7 @@ table {
   min-width: 100%;
 
   thead {
-    tr {
+    th {
```
msecret: Why does this fix the issue (the `th` should already be nested in a a `tr`) and why was the rule `tr` before?

maya: Because the white background will override the gray table header unless this is changed to `th`.

Issue #345, opened 2015-08-25
Small screen + landscape mode = impossible to scroll because of header height (Moto E, Android 4.4, Chrome 44)


colinpmacarthur:
On a small screen, in portrait mode, with the keyboard open, it's impossible to scroll around the site because the header is too tall:

![impossible to scroll portrait mobile keyboard up](https://cloud.githubusercontent.com/assets/2092076/9476440/c960c224-4b3a-11e5-82b2-693fe249b977.png)

colinpmacarthur:
@carodew and @mollieru:

After discussion with @angel while she was redesigning the header, I propose we close this issue because:
- It only applies on relatively small mobile screen sizes in portrait mode
- The Moto E has the same problem with all mobile websites with sticky headers
- Making the header any slimmer would make the menu button very hard to tap

Thoughts?

carodew:
@colinpmacarthur – do you mean landscape mode, not portrait mode? 

other qqs: does this happen on other devices? 
Were we going to try to find a way to reduce the size of the "This is an official site … currently in alpha" header? Or is that accounted for somewhere else? Or is that a post-MVP kind of thing?

colinpmacarthur:
@carodew: All good questions!

1. Yes, landscape not portrait. So amended above. :)
2. The header in staging has reduced the size of both the header and the official site banner,
 but the problem persists: 
![unnamed 1](https://cloud.githubusercontent.com/assets/2092076/9548644/ae128e1c-4d6f-11e5-9f82-c5bc69122bf0.png)
3. I cannot replicate this problem on my iPhone 6 or iPhone 4 because:
  - The soft keyboard is relatively smaller on iOS
  - iPhone screen sizes are slightly larger
  - iPhone pixel densities are higher

Maybe a good next step would be trying to replicate on your Android phone?

colinpmacarthur:
@mollieru also asked a good related question in a [related Trello card](https://trello.com/c/9nafSh2p/436-1-site-design-design-mobile-header): "does this problem happen on ssa.gov, my.uscis.gov or doi.gov?" It doesn't; those sites don't have sticky headers, so the header doesn't persistently cover a portion of the screen like ours.

carodew:
Android/Safari – similar issue. 

Video: https://drive.google.com/open?id=0B1Wf-KSFneMUY08xWFIzSDhpRXc

How much real estate do we get back by reducing/hiding the header? 

[...]

PR #1212, opened 2016-05-31
[Website]: Fix open keyboard header bug in landscape mode on Android

maya:
## Description

Fixes a bug that made it unable to scroll in landscape mode with an open keyboard on Android.

Note: this will need to be tested, I was unable to test it on a device. 

Aspect ratio from here: http://stackoverflow.com/questions/8883163/css-media-query-soft-keyboard-breaks-css-orientation-rules-alternative-solut

Fixes: #345


donjo:
Tested on a Nexus 5 in Chrome and it worked for me 👍

maya:
@donjo thanks for testing! let's see if I can fix this merge conflict and get this in.

Issue #1178, opened 2016-05-18
update GA tag for standards.usa.gov


gbinal:
As discussed in the #g-analytics channel on slack, it probably makes sense to split out the google analytics for standards.usa.gov from playbook.cio.gov.  

To that end, can you replace `UA-48605964-33` with `UA-48605964-43` throughout the site?  Holler if you have any questions.  Thanks!

maya:
@gbinal will we lose all of our data by switching this?

gbinal:
It won't be integrated into the new `standards.usa.gov` profile, but you won't be losing any data.   

You'll always have access to it via the `playbook` profile.

PR #1217, opened 2016-06-01
Update Google Analytics code

maya:
## Description

Replaces the Google Analytics code for standards.usa.gov so it's separate from playbook.cio.gov. 

Resolves: #1178.

cc: @gbinal


gbinal:
👍

Issue #792, opened 2015-10-06
fieldset legend support on voiceover


goetsu:
fieldset / legend are not supported by VoiceOver on osX or iOS so I think at least you need to have a warning in https://playbook.cio.gov/designstandards/form-controls/ especially for checkbox, radio or date input.

A possible solution is to use aria-describedby on the input for example : 
```html
  <fieldset class="usa-fieldset-inputs usa-sans">
    <legend class="usa-sr-only" id="legend-id-1">Historical figures 1</legend>

    <ul class="usa-unstyled-list">
      <li>
        <input id="apple-pie" name="apple-pie" value="apple-pie" tabindex="0" checked="" type="checkbox" aria-describedby="legend-id-1">
        <label for="apple-pie">Sojourner Truth</label>
      </li>
      <li>
        <input id="key-lime-pie" name="key-lime-pie" value="key-lime-pie" tabindex="0" type="checkbox"  aria-describedby="legend-id-1">
        <label for="key-lime-pie">Frederick Douglass</label>
      </li>
      <li>
        <input id="peach-pie" name="peach-pie" value="peach-pie" tabindex="0" type="checkbox"  aria-describedby="legend-id-1">
        <label for="peach-pie">Booker T. Washington</label>
      </li>
      <li>
        <input id="disabled" disabled="" type="checkbox"  aria-describedby="legend-id-1">
        <label for="disabled">George Washington Carver</label>
      </li>
    </ul>

  </fieldset>


```

nickbristow:
I can't find anything that says fieldset and legend aren't supported, and when I tested it, the legend was read on mine. Using (10.9.5)

Is there a specific version that isn't supported?

KimberlyMunoz:
I've seen a few problems when I've tried testing tables with voiceover and found some articles about iOS's bugs with reading tables, legends and fieldsets in certain versions. 

Here's a video of the legend/fieldset bug:
http://www.tyack.net/2015/01/27/accessibility-frustrations-with-apple-voiceover-on-ios-8-1/

goetsu:
Legend was announced and read when you reach it but the main Accessibility feature of fieldset legend was not supported by VO ios and OS X < 10.11. Normaly when you reach an input inside a fieldset with a legend the content of the legend must be added to the label value. In case of VO it's not
 
http://pauljadam.com/demos/voiceover-ios-html-aria-support.html

nickbristow:
Thanks @goetsu 
[...]

PR #1220, opened 2016-06-02
[Forms] Fix label and fieldset for iOS/OS X Voiceover

maya:
## Description

Fieldset and legend are not supported by VoiceOver on iOS and OS X, so this changes our code so it will work. This is not the ideal way but necessary to make Apple work, bugs have been reported about this issue Apple already.

This also fixes some of our labels that were still reading different pies instead of historical figures.

cc: @nickbristow 

Fixes: #792.

### DIFF
--- a/docs/_elements/form-controls.md
+++ b/docs/_elements/form-controls.md
@@ -131,23 +131,23 @@ lead: Form controls allow users to enter information into a page.
 
   <fieldset class="usa-fieldset-inputs usa-sans">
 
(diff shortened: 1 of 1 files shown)


maya:
@nickbristow can we add some background info about this to our accessibility guidance content?

nickbristow:
@maya 👍

maya:
@nickbristow will be drafting up some documentation.

maya:
Closing in favor of #1221.

PR #1224, opened 2016-06-03
[Accordion]: Make content accessible without JavaScript

shawnbot:
## Description

This PR updates the accordion component to update its buttons' and corresponding content elements' `aria-expanded` and `aria-hidden` attributes (respectively) when initialized. This means that we can strip all of the `aria-expanded="false"` and `aria-hidden="true"` attributes from the HTML, which makes the content accessible in the event that the JS isn't loaded (or until it is). Hopefully, this addresses #1122.

⚠️ Note that the biggest change here is not in the JS, but in the HTML: Users will be expected to update their HTML to remove `aria-expanded="false"` and `aria-hidden="true"` attributes from accordion panels that aren't expanded by default so that their content is accessible without JS.

## Additional information

* The accordion documentation (and all of the accordions in each component's documentation) markup now reflects the best practice of leaving `aria-expanded="false"` and `aria-hidden="true"` attributes out, so that the content is accessible without JavaScript.
* I've updated the accordion documentation to state that the `aria-expanded` and `aria-hidden` attributes are unnecessary now (except in the case of the content area that you wish to be expanded by default), but this could probably use some clarification and/or word-smithing from @kategarklavs.
* The demo HTML now uses the `amendment-n+1` naming convention instead of `collapsible-n`, so that it plays nicely with the the Code and Documentation accordion below. See: #1225
* I've added jsdoc comments to all of the Accordion methods, and updated some of the existing ones.
* The Accordion JS component also now returns `this` from its `show()`, `hide()`, and `hideAll()` methods, so its methods are chainable. I'm happy to back out of this change if it's not consistent with the rest of the library.

### DIFF
--- a/docs/_components/accordions.md
+++ b/docs/_components/accordions.md
@@ -16,7 +16,7 @@ lead: Accordions are a list of headers that can be clicked to hide or reveal add
           aria-expanded="true" aria-controls="collapsible-0">
           First Amendment
         </button>
-        <div id="collapsible-0" aria-hidden="false" class="usa-accordion-content">
+        <div id="collapsible-0" class="usa-accordion-content">
           <p>
           Congress shall make no law respecting an establishment of religion, or prohibiting the free exercise thereof; or abridging the freedom of speech, or of the press; or the right of the people peaceably to assemble, and to petition the Government for a redress of grievances.
           </p>
@@ -24,10 +24,10 @@ lead: Accordions are a list of headers that can be clicked to hide or reveal add
       </li>
       <li>
         <button class="usa-button-unstyled"
-          aria-expanded="false" aria-controls="collapsible-1">
+          aria-controls="collapsible-1">
           Second Amendment
         </button>
-        <div id="collapsible-1" aria-hidden="true" class="usa-accordion-content">
+        <div id="collapsible-1" class="usa-accordion-content">
           <p>
           A well regulated Militia, being necessary to the security of a free State, the right of the people to keep and bear Arms, shall not be infringed.
           </p>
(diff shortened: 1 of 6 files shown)


docs/_components/accordions.md
```diff
@@ -135,23 +135,23 @@ lead: Accordions are a list of headers that can be clicked to hide or reveal add
...
       </li>
       <li>
-        Each button has a unique name <code>aria-controls=<wbr>’collapsible-#’</code> that associates the control to the appropriate region by referencing the controlled elements <code>id</code>.
+        Each button has a unique name <code>aria-controls=<wbr>"id"</code> that associates the control to the appropriate region by referencing the controlled element&rsquo;s <code>id</code>.
       </li>
       <li>
-        Each content area has an <code>aria-hidden</code> attribute set to either <code>true</code> or <code>false</code>. When <code>true</code>, the element (and all children) are neither visible or perceivable, and assistive technologies will skip this content.
+        Each content area will have its <code>aria-hidden</code> attribute set to either <code>true</code> or <code>false</code> by the component, depending on its corresponding button&rsquo;s <code>aria-expaded</code> attribute. To ensure that your content is accessible in the event that the JavaScript does not load or is disabled, you should not set <code>aria-hidden=<wbr>"true"</wbr></code> on any of your content areas.
```
joshbruce: "aria-expaded" shouldn't that be "aria-expanded"

kategarklavs:
@shawnbot I just looked over everything and have no proposed changes 👍

docs/_components/accordions.md
shawnbot: Yes, good catch!

PR #1252, opened 2016-06-10
[Form controls]: Fix input width in error state

maya:
## Description

Fixes input width in error state where the input was 4px too short bc it didn't account for the 4px border (on mobile).

Before:

screen shot 2016-06-10 at 2 39 25 pm

After:

screen shot 2016-06-10 at 2 39 41 pm


maya:
also cc'ing @ericadeahl just for visibility.

Issue #575, opened 2015-09-16
[Links] New tab link missing external link icon


angel:
Missing this icon (http://fortawesome.github.io/Font-Awesome/icon/external-link/) to the right of new tab.

carodew:
<img width="925" alt="screen shot 2015-09-16 at 4 52 42 pm" src="https://cloud.githubusercontent.com/assets/10144074/9918178/634be008-5c93-11e5-864f-d98cb0b84f67.png">

maya:
This never made it into development.

carodew:
The issue is that there is a WH initiative that links leaving .gov sites are required to warn you that you are leaving a government controlled site.

On the other hand, many agencies don't think this icon is enough to fulfill that requirement. 

Also, if this is why this is a high, the text needs to be different. But not before launch.

maryjom:
In addition, the code doesn't match the .pdf of the components_sticker_sheet_and_read_me_V.8 which shows the icon on the external link.

danamuses:
I would love to see guidance here about how to indicate exit links. For example, on Archives.gov we use an arrow icon which provides disclaimer language in the hover state or links to a full disclaimer page on our website.

PR #1253, opened 2016-06-11
[Links]: Add icon to external links

maya:
## Description

Adds icon to links that open externally via `usa-external_link` class.

Removes recommendation to open in a new tab as there's research from [Nielsen](http://libux.co/links-should-open-in-the-same-window/) and [others](https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/) that this is not a good recommendation, but I'd like if @ericadeahl and @bradnunnally could check this out first.

#### This is what it looks like:

screen shot 2016-06-10 at 4 41 50 pm

Resolves: #575


src/stylesheets/elements/_typography.scss
```diff
@@ -33,6 +33,27 @@ a {
...
+
+.usa-external_link {
+  &::after {
+    display: inline-block;
+    width: $base-font-size;
+    height: $base-font-size;
+    margin-bottom: -1px;
+    margin-left: 4px;
+    content: '';
+    background: url('../img/external-link.svg') no-repeat 0 0;
```
msecret: I thought most of our svg images had png fallbacks. Why is that not the case here?

src/stylesheets/elements/_typography.scss
```diff
@@ -33,6 +33,27 @@ a {
   }
 }
 
+// External link consider "effortless style approach":
+// [href^="http:"]:not([href*="my-domain.com"])
+// [href^="https:"]:not([href*="my-domain.com"])
+
+.usa-external_link {
+  &::after {
+    display: inline-block;
```
msecret: Can you alphabetize these?

src/stylesheets/elements/_typography.scss
```diff
@@ -33,6 +33,27 @@ a {
   }
 }
 
+// External link consider "effortless style approach":
```
msecret: Why not just do this now?

maya: It's a little bit more complicated of an approach. 

1.) It would mess up button-styled anchor links without some kind of reset on them.
2.) We'd have to figure out the best way to manage the user's domain in `[href^="http:"]:not([href*="my-domain.com"])` which I was not really sure what approach we wanted (e.g. a Sass variable, etc..). I was thinking we could always enhance-up to this option. What do you think?

src/stylesheets/elements/_typography.scss
maya: Fixed in 9e4becb.

src/stylesheets/elements/_typography.scss
msecret: Yea that sounds good.

src/stylesheets/elements/_typography.scss
maya: Fixed in c50fa2b.

maya:
@ericadeahl @bradnunnally: I wasn't sure if we wanted this style applied to external links throughout the site. Should we?

ericadeahl:
@maya @bradnunnally I could go either way on this, but I think it's probably good practice for us to use the styles we're recommending. So, yes. There may be some exceptions though, such as the links to the examples on the homepage (the icon makes sense for body copy links, but will look weird at header size.)

bradnunnally:
Agreed with @ericadeahl we should follow our own styles as much as we can. It's fine to break the rules when it makes the sense. /cc @maya

maya:
@bradnunnally @ericadeahl And what about the other point about the recommendation of opening new tab?

maya:
@ericadeahl @bradnunnally - slight issue, the icon is blue and the footer is currently blue as well, so they wouldn't appear. I realize we are changing the footer to white in the new designs so this wouldn't be an issue. We could create a white set of icons, but not sure it's worth the effort if it's going to change soon. I'm wondering what we should do in the meantime.

ericadeahl:
@maya This will likely be an issue for other folks as well, so it seems worth it to add a line here "This is a link to an external website on a dark background", with a light gray colored link/icon.

Re: new tabs, I was always a fan of external sites opening in new tabs, but I'm convinced by these arguments for linking in the same window as a best practice. Let's stick to that.

maya:
@kategarklavs will you be able to review some content changes. 

Changing:

> This is a link which opens in a new tab.

to something that shows it's opening an external website.

Currently using:

> This is a link which goes to an external website.

maya:
@ericadeahl the links are currently white, should we change them to light gray?

screen shot 2016-06-14 at 12 19 47 pm

maya:
Also, I want to flag to your attention @ericadeahl. We cannot change the color of the icon for `:visited` links. The reason is the `:visited` pseudo-class doesn't permit the use of pseudo-elements because of [privacy concerns](https://hacks.mozilla.org/2010/03/privacy-related-changes-coming-to-css-vistited/).

http://stackoverflow.com/questions/7371732/why-does-foo-alink-foo-avisited-selector-override-ahover-aactive-s

So visited external links will have the visited style but their icon will not.

screen shot 2016-06-14 at 12 48 02 pm

maya:
@ericadeahl should the "Learn more" link in the disclaimer bar have the icon?

screen shot 2016-06-14 at 2 04 53 pm

kategarklavs:
@maya Hello! My only change to the content you proposed is changing the _which_ to _that_: 

This is a link that opens in a new tab.

And that's it!

maya:
@kategarklavs done in 2bdc7f7. thanks!

ericadeahl:
@maya @bradnunnally I'm still not totally convinced this is the best practice for external links, and I've been reviewing some conflicting advice, so I'd like to mark this component for further review. However for the short term I think we can proceed as discussed. I recommend we:

- Use the blue icon for external links on a light background (despite the slight weirdness for visited links.) I took a look at using gray/black instead of blue for the icon color, but I think it gets confusing when the icon looks like it's more related to the surrounding text than to the link.

- Use #d6d7d9 for the external link and icon on a dark background (since that's what's specified in the Standards.)

- Remove the external link icon after the "learn more" link in the disclaimer

- One more thing: could we reduce the size of the icon relative to the text size? It would be less obtrusive if the height of the icon was closer to the x height rather than the cap height. So for body copy the height of the icon would be 11px: 
screen shot 2016-06-15 at 3 45 08 pm

bradnunnally:
Agreed with @ericadeahl's assessment. It's probably good enough for now, but given the research Erica has done it'll need more focused analysis at some point in the future.

maya:
@ericadeahl done! This is what it looks like:

screen shot 2016-06-16 at 2 57 46 pm

Also, added a white hover state for the links:

screen shot 2016-06-16 at 3 00 29 pm

ericadeahl:
@maya All looks good, thanks!!

rogeruiz:
Awesome work @maya. Thanks for getting this out the door @maya @ericadeahl @bradnunnally 👍

Issue #540, opened 2015-09-15
[Footer] Slim footer contact links incorrectly aligned in IE 10


colinpmacarthur:
**To reproduce**
1. Go to https://pages-staging.18f.gov/web-design-standards/footers/ in IE 10.
2. Scroll down to the "Slim" footer.
3. The phone number and email link are below the other links, not to their side. The email links is also not vertically aligned with the header link:
<img width="523" alt="screen shot 2015-09-15 at 3 27 21 pm" src="https://cloud.githubusercontent.com/assets/2092076/9887759/480548a4-5bbf-11e5-9a2d-4e5f6e0df326.png">

**Expected behavior**
From Chrome 44:
<img width="509" alt="screen shot 2015-09-15 at 3 29 39 pm" src="https://cloud.githubusercontent.com/assets/2092076/9887739/345c50b8-5bbf-11e5-9473-be5f728ffef5.png">

**Found on**
- SauceLabs, Windows 7, IE 10

PR #1255, opened 2016-06-13
[Slim footer]: Fix alignment

maya:
## Description

This fixes an issue in IE 10 where things were incorrectly aligned. The problem stemmed from IE 10 not respecting flexbox the same way other browsers do, so I've removed flexbox. It also removes the nesting of grids inside of grids bc this was a bit overcomplicated. Instead it uses a single level of 1/6 width columns, which is more consistent with the medium footer.

One difference is this will collapse differently in medium / tablet browsers (3 columns instead of 4) and have a border around the fifth and sixth column text and links on mobile.

#### This is what it looks like on medium screens:

screen shot 2016-06-10 at 10 33 59 pm

#### This is what it looks like on small screens:

screen shot 2016-06-10 at 10 34 13 pm

cc: @ericadeahl @bradnunnally 

Fixes: #540

### DIFF
--- a/docs/_components/footers.md
+++ b/docs/_components/footers.md
@@ -193,28 +193,28 @@ lead: Footers serve site visitors who arrive at the bottom of a page without fin
     </div>
     <div class="usa-footer-primary-section">
       <div class="usa-grid-full">
-        <nav class="usa-footer-nav usa-width-two-thirds">
+        <nav class="usa-footer-nav">
           <ul class="usa-unstyled-list">
-            <li class="usa-width-one-fourth usa-footer-primary-content">
+            <li class="usa-width-one-sixth usa-footer-primary-content">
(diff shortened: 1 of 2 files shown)


maya:
@ericadeahl @bradnunnally - could you review the changes above?

maya:
Looking at the sketch file, seems like we may want to remove the horizontal lines from the last two items on mobile.

ericadeahl:
@maya For the medium screen one: this isn't a big deal, but I think the contact info should probably appear in a single column to the right of the primary links. So on the same grid you're using, but in this order:

Primary link   Primary link   (800) CALL-GOV
Primary link   Primary link   info@agency.gov

On small screens, there shouldn't be a rule between the phone number and email address. It should match what's currently on the site:

screen shot 2016-06-13 at 3 21 21 pm

maya:
Superseded by #1259.

PR #1257, opened 2016-06-13
[Website]: Remove h1 from site header

maya:
## Description

Removes `h1` from header navbar as `h1`s should be used only for the title of a page and duplicate h1's are an accessibility problem.


rogeruiz:
@maya, would it make more sense to make the [`Getting Started`](https://standards.usa.gov/getting-started/) or the secondary heading on any of the other pages below the navbar the non-`<h1>` element and keep the navbar `<h1>` instead?

maya:
@rogeruiz no, page titles are correct and should be h1's as they are the primary headings of the page.

PR #1248, opened 2016-06-09
[Footer]: Fix footer social icon hit area

joshbruce:
## Description

Fixes #346

## Additional information

* This PR uses `display: inline-block` to allow setting height and
width to an inline element.
* Uses 44 pixels for both height and width per the Apple iOS HIG
* Concerns with `text-align: right` - could result in thumb hitting to
the right of the hit area - actually hitting the icon to the right of
the actual target. `text-align: center` used instead.
* Using `text-align: center` causes the right alignment of the icons to
be off when compared “Agency Contact Center” text. `position: relative;
left: 13px` used in order to compensate. Slight concern that this could
cause a horizontal scroll bar to appear on mobile - though, given the
overall alignment of the `social-links` container - concern may be moot.

Issue #706, opened 2015-09-23
[Footer] design update


mollieru:
v.8 release includes a Footer component. This is our second iteration of that design based on user feedback, which hasn't been updated in the v.8 code:

_Tip: click for full size_

![footers_ui_v2](https://cloud.githubusercontent.com/assets/3680494/10060537/6e6d06d8-621f-11e5-82ee-1cf643333151.png)

carodew:
As with the other footer issues, we'll be incorporating new feedback into the footer design so there are some forthcoming changes. Closing to prevent confusion.

PR #1262, opened 2016-06-16
[Footer]: Fix footer social icon hit area - flexible

### LINKED ISSUE #346: [Footer] Social media icons too small to respond consistently when tapped (multiple devices)
Social media icons are too small to respond consistently when tapped:
footer social link spacing

## Affected devices
- Moto E, Android 4.4, Chrome 44
- iPhone 6, iOS 8.4

joshbruce:
## Description

Alternative #1248
Fixes #346

## Additional information

#1248 updates the size of the social link icons, and adds a hit area
around the links to make them more mobile friendly. Visually, the links
are smaller; however, the active area of the link is larger (~45px
square).

Things I noticed from creating #1248:

- Changing icon sizes (#706) requires manually changing the height and
width per link - in two areas (once for the `svg` element and the other
for the `image` element).
- Adding a new icon (LinkedIn, for example), requires adding 6 new
lines to define the link.
- The CSS definition assumes all social links need to be held within a
container `.usa-social-links`, which has no definition outside of
defining each `a` element within the container.

This PR:

1. Re-engineers social icons as background images.
2. Removes dependency on `.usa-social-links` container.
3. Removes containing `div`.
4. Adds `.usa-social-link` class for defining global attributes (hit
area, for example); allowing any link to become a social link.
5. Adds `.facebook`, `.twitter`, `.youtube`, and `.rss` classes, which
must be used in conjunction with `.usa-social-link` to take effect.
6. Allows for introduction of more background for social media links by
extenders, without having to define dimensions and layouts.


joshbruce:
@maya and @ericadeahl - What SVG compiler are you using - I see the SVG generated for this is huge compared to what it was.

joshbruce:
screen shot 2016-06-16 at 1 08 05 pm

screen shot 2016-06-16 at 1 08 10 pm

src/stylesheets/components/_footer.scss
```diff
@@ -272,3 +272,43 @@ li.usa-footer-primary-content {
...
+
+  background-position: center center;
+  background-repeat: no-repeat;
+  display: inline-block;
+  height: $hit-area;
+  left: $left-align-adjust;
+  margin-left: $blank-space;
+  position: relative;
+  text-align: center;
+  width: $hit-area;
```
maya: I don't think we need variables. We only use variables if the value is repeated at least twice and the value is likely to be updated at least once.

shawnbot: I generally agree, but I think that the `$hit-area` variable is helpful here. @joshbruce maybe move the comments alongside the properties instead?

docs/_components/footers.md
```diff
@@ -73,32 +73,10 @@ lead: Footers serve site visitors who arrive at the bottom of a page without fin
...
-              <svg width="26" height="39" role="img" aria-label="RSS">
-                <title>RSS</title>
-                <image xlink:href="{{ site.baseurl }}/assets/img/social-icons/svg/rss25.svg" src="{{ site.baseurl }}/assets/img/social-icons/png/rss25.png" width="26" height="39" />
-              </svg>
-            </a>
-          </div>
+          <a class="usa-social-link facebook" href="#"><span class="usa-sr-only">Facebook</span></a>
+          <a class="usa-social-link twitter" href="#"><span class="usa-sr-only">Twitter</span></a>
+          <a class="usa-social-link youtube" href="#"><span class="usa-sr-only">YouTube</span></a>
+          <a class="usa-social-link rss" href="#"><span class="usa-sr-only">RSS</span></a>
```
maya: 1. Use the use the `sr-only()` mixin instead of `usa-sr-only` class so it's even cleaner?
2. Can you please indent these? Like this:
```
<a class="usa-social-link facebook" href="#">
  <span class="usa-sr-only">Facebook</span>
</a>
```
3. Update `usa-social-link` to `usa-social_link`
4. Update `facebook` (and others) to `usa-link-facebook`

/cc @msecret

src/stylesheets/components/_footer.scss
```diff
@@ -272,3 +272,43 @@ li.usa-footer-primary-content {
...
+
+  &.facebook {
+    background-image: url("../img/social-icons/png/facebook25.png");
+    background-image: url("../img/social-icons/svg/facebook25.svg");
+    background-size: auto 3rem;
```
maya: This can be moved into the rule above.

src/stylesheets/components/_footer.scss
```diff
@@ -272,3 +272,43 @@ li.usa-footer-primary-content {
     padding-left: 1em;
   }
 }
+
+.usa-social-link {
+  $hit-area: 4.5rem; // used for height and width
```
maya: Is there a reason we're using 4.5rem instead of 4.4rem - which is what Apple defines?

src/stylesheets/components/_footer.scss
joshbruce: Hmmmm...either way is fine for me. 

The initial reason - specifically for the hit area one is actually because, according to the CSS coding style, I couldn't put width and height next to each other alphabetically. The other two came about just to further separate the static values from the implementation - OOP-style.

Again, I'm fine for whichever.

docs/_components/footers.md
joshbruce: I'm not quite how this one lands for me...given that the current setup cannot be used without `usa-social_link`. This change would make it:

```
<a class="usa-social_link usa-link-facebook" href="#">
  <span>Facebook</span>
</a>
```

Having the `.usa-link-facebook` piece makes me think we are driving toward a Sass extension:

```
.usa-social_link {
  ...
  span {
    @include sr-only(); // At least I think that's what the comment about using the @mixin was driving at
  }
}

.usa-link-facebook {
  @extend .usa-social_link
  ...
}
```

This would give us HTMl like this:

```
<a class="usa-link-facebook" href="#">
  <span>Facebook</span>
</a>
```

Admittedly cleaner than the current PR, requires less global knowledge of the Sass, and easily extensible - even with a multiple class solution using `.usa-social_link` in conjunction with another - or extending it via Sass. The name does seem to imply there is a `.usa-link` somewhere - is there?

Thoughts?

src/stylesheets/components/_footer.scss
joshbruce: Not really. Will change.

src/stylesheets/components/_footer.scss
joshbruce: I'm not sure I'm following. Would this establish Facebook as being a "default" social link - then the others as add-ons?

I believe this is moot if we decide to go with the suggestion of using @extend from the comment in footers.md, because it would separate the rules entirely.

docs/_components/footers.md
maya: @joshbruce yeah I was talking with @msecret and actually think `.usa-social_link-facebook` is technically more correct, but he said it was implied / not necessary.

I don't have a problem with extends in this instance.

src/stylesheets/components/_footer.scss
maya: I'm OK with keeping `$hit-area` for that reason, but would remove the rest. 

Can we add a more descriptive comment on it:

>Link hit target is 44 x 44 pixels following Apple iOS Human Interface Guidelines

src/stylesheets/components/_footer.scss
maya: I mean `background-size: auto 3rem;` is repeated across all the social links so should go in `usa-social-link` class or in the extend if that's what we're using.

rogeruiz:
Hey @joshbruce. Looks like you'll need to rebase this branch onto the latest work on `staging`. 18F/web-design-standards#1261 has removed `screener` and that will get this PR to pass in `CircleCI`. If it's helpful, I can rebase these changes manually and submit a new pull request with that update. Thanks!

joshbruce:
@rogeruiz - Thanks. Saw the traffic and wasn't sure. Will push in two seconds and see what happens.

Might need some follow-on help, if you're up to it.

maya:
@joshbruce looking at the branch, this is what it looks like:
screen shot 2016-06-21 at 9 27 58 am

This is what it's supposed to look like:
screen shot 2016-06-21 at 9 29 35 am

@ericadeahl posted sizes for the icons in https://github.com/18F/web-design-standards/pull/1248#issuecomment-226310207 although they all appear to be the same 29px x 29px here.

src/stylesheets/components/_footer.scss
```diff
@@ -300,3 +300,50 @@ li.usa-footer-primary-content {
...
+  @extend .usa-social_link;
+  background-image: url("../img/social-icons/png/youtube15.png");
+  background-image: url("../img/social-icons/svg/youtube15.svg");
+}
+
+.usa-link-rss {
+  @extend .usa-social_link;
+  background-image: url("../img/social-icons/png/rss25.png");
+  background-image: url("../img/social-icons/svg/rss25.svg");
+}
```
maya: Add one line-break here. All of our files should have an extra space at the end of the file.

joshbruce: @maya - I'm not sure why the blank line is not showing in the diff...? I added periods to the comments in order to get this commit to work. My file shows 350 line long.

Thoughts?

joshbruce:
@maya - Odd. Here's what I see:

screen shot 2016-06-21 at 12 45 56 pm

The screenshot is from Safari 9.1.1 on the macOS 10.11.5.

The next one is from Chrome 51

screen shot 2016-06-21 at 12 52 14 pm

Thoughts?

maya:
@joshbruce was on the wrong branch, my bad! Looks like your screenshots :)

maya:
@joshbruce on mobile, the alignment looks like it's indented to the right, could we align this flush left?

screen shot 2016-06-21 at 10 06 35 am

Should be:

screen shot 2016-06-21 at 10 10 58 am

joshbruce:
@maya - Sorry - forgot the spacing between the icons on top and bottom - should be good now.

screen shot 2016-06-21 at 3 37 11 pm

src/stylesheets/components/_footer.scss
maya: It must be a github thing, just checked this out locally and looks correct. 👍

maya:
Looks good!

PR #1269, opened 2016-06-17
[ Website ]: Deployment Remove path from manifest & add to cf push

rogeruiz:
This patch adds the path parameter for the application to the `cf push`
command because the `path: _site` property in the main manifest was not
properly resolving since the folders moved into `config/cf`.


juliaelman:
🚀

Issue #595, opened 2015-09-16
[Search bar] clicking on 'search' reloads 'Search bar' page in Safari


mollieru:
*To recreate*
On 'search bar' page, click on any of the 'search' buttons. Page will reload. 

http://screencast-o-matic.com/watch/coQDFpfKVc

*Expected behavior*
No action should result from clicking on example search button. 

*Tested on*
OS 10.8
Safari 6.2

carodew:
Bug triage thoughts: 

*it's because there is a submit button that is trying to submit something
*short term solution would be to not make it a submit button, but that would make the code not relevant for people who download it
*long term solution might be to create a JS solution that would prevent the button from reloading the page

joshbruce:
Could disable the button as well - this would require adding a styleguide class to give the proper button color, but would make it so the page doesn't reload. Could also use JS prevent default.

PR #1275, opened 2016-06-17
[Website]: Prevent example search bars from reloading page when clicked

stephmarx:
## Description
Prevents default on objects with the class `.usa-search` which are inside of the `.preview-search-bar`. This prevents the example search bar forms from attempting to submit and reloading the page.

## Additional information

* Pull request also removes unnecessary whitespace on blank lines within the `styleguide.js` file.
* Resolves: #595


juliaelman:
LGTM - thank you @stephmarx!

Issue #990, opened 2016-02-06
"Return to top" link in footers


wslack:
I can't promise I know what I'm talking about, but the "return to top" links on https://playbook.cio.gov/designstandards/footers/ don't actually return me to the top of the page. However, the links on https://github.com/18F/web-design-standards/blob/18f-pages-staging/docs/_components/footers.md, where the code lives, do return me to the top. That's confusing.

Separately, in looking for "return to top" links in the wild, I didn't see any sites using them. Do we need to have these in the code if they aren't being used?

maya:
That's a good catch @wslack! We disabled the ability for clicking the hash `#` to return to the top of the page bc we're using it as a fake link on the footer links and didn't want someone to accidentally click a link and it go to the top. I didn't realize that by doing this, we removed the functionality of hitting the top of the page on the button it's meant to. That seems like it could be confusing!

To fix this we need to add a class to our JS to only target this function on `usa-footer-primary-section` class and anywhere else we'd like that behavior.

wslack:
Thanks @maya! If you do know of sites using that link, would like to see how that looks in practice.

carodew:
You know, we overrode that behavior without really considering whether or not people wanted to be able to test the functionality. It might be worth thinking about undo-ing that, or it might be something we think about including in page templates.

Drakenhart:
Return to top is sometimes used in FAQs if the list is long.
With Government websites as complex as they are sometimes, they often have massive FAQ lists.

Something to think about. ^.^

j-ta:
I am working on this.

bradnunnally:
@j-ta Awesome! Let me know how it goes and if you have any comments you'd like to share. I'm sitting over by the wall.

PR #1280, opened 2016-06-17
Remove unbind code because overriding default behavior is only useful…

j-ta:
## Description

Resolve "Return to top" link in footers Issue #990. There is code that unbinds the standard click event for primary links. This is only useful for the sample code page, and does not make sense for distributable code. I removed it because it is a feature that causes a bigger problem than it solves.

## Additional information

* Even though click event overrides have been removed, the "Return to Top" link still doesn't seem to work as expected. There might be other code that is overriding default click behavior.

### DIFF
--- a/src/js/start.js
+++ b/src/js/start.js
@@ -58,18 +58,14 @@ $(function () {
 
       $('.usa-footer-big nav ul').addClass('hidden');
 
-      $('.usa-footer-big nav .usa-footer-primary-link').unbind('click');
-
       $('.usa-footer-big nav .usa-footer-primary-link').bind('click', function () {
         $(this).parent().removeClass('hidden')
         .siblings().addClass('hidden');
       });
     } else {
-
       $('.usa-footer-big nav ul').removeClass('hidden');
-
-      $('.usa-footer-big nav .usa-footer-primary-link').unbind('click');
     }
+
   };
 
   footerAccordion();


maya:
Ouch, this should not have been in the `src` js, it should be in `docs/doc_assets/js/styleguide.js`. Thanks for discovering this! That being said, I think we want to move it to styleguide.js and the ideal solution would be to remove the behavior from "Return to Top" link without effecting the other footer links.

rogeruiz:
Awesome contribution, @j-ta. Thank you for your contribution. I'm going to be testing this locally to see if I can comment with a patch containing the necessary changes to get the `Return to Top` working. It looks like this can be merged in after I run local tests. Thanks. 👍

maya:
@rogeruiz just so you're aware as you work on this that the code is to prevent primary links from returning to the top which is the intended behavior. The goal is to get "return to top" links functioning without effecting the primary links non-functioning.

rogeruiz:
Alrighty, @j-ta. I looked into the issue a bit further and saw that the error was not tied to the `footerAccordion` function you modified in `src/js/start.js`. It's tied to the `handleDisabledLinks` function in the `docs/doc_assets/js/styleguide.js` file.

```diff
diff --git a/docs/doc_assets/js/styleguide.js b/docs/doc_assets/js/styleguide.js
index a79fb96..90a22c2 100644
--- a/docs/doc_assets/js/styleguide.js
+++ b/docs/doc_assets/js/styleguide.js
@@ -7,7 +7,11 @@ $(function(){
   function handleDisabledLinks() {
     $(document).on('click', 'a[href="#"]', function (event) {
       // Stop default browser action which would likely return to the top of the page
-      event.preventDefault();
+      // unless it's an actual return-to-top link
+      var isReturnToTop = 'return to top' === $(event.target).text().toLowerCase();
+      if (! isReturnToTop) {
+        event.preventDefault();
+      }
     });
   }
   handleDisabledLinks()
```

rogeruiz:
Closed in favor of #1285

PR #1289, opened 2016-06-21
Install autopilot for zero-downtime-push

rogeruiz:
With `autopilot` installed, a botched deployment won't kill our application. If the deployment fails, we should still be able to recover with no downtime.

https://github.com/concourse/autopilot#method


rogeruiz:
cc: @yozlet @msecret @maya

PR #1294, opened 2016-06-27
Prefer dashes over underscore in CSS class names

jessieay:
* Noticed one of these when using the standards because my CSS linter
  rules want all class names to have dashes, which is typical
* Consistency is easier, anyway!

### DIFF
--- a/docs/_components/footers.md
+++ b/docs/_components/footers.md
@@ -54,8 +54,8 @@ lead: Footers serve site visitors who arrive at the bottom of a page without fin
           </ul>
         </nav>
 
-        <div class="usa-sign_up-block usa-width-one-third">
-          <h3 class="usa-sign_up-header">Sign up</h3>
+        <div class="usa-sign-up-block usa-width-one-third">
+          <h3 class="usa-sign-up-header">Sign up</h3>
 
           <label class="" for="email" id="">Your email address</label>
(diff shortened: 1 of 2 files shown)


msecret:
@jessieay what linter and linting configuration are you using?

jessieay:
https://github.com/18F/micropurchase/blob/develop/.scss-lint.yml

maya:
We use 18F’s modified BEM naming methodology which uses underscores when block, elements, or modifiers have two words in them, and uses dashes in between. Also #5 in CSS architecture: https://standards.usa.gov/getting-started/

https://pages.18f.gov/frontend/#suggested-custom-methodology

Feel free to customize for your use, but we'll be sticking with the current methodology for now.

msecret:
@jessieay I'd also suggest to look at the [18F scss-lint configuration](https://pages.18f.gov/frontend/#css-linting). This will keep our whole organization consistent in CSS style and will be compliant with WDS standards. If you ever run into problems with that configuration, we can also change it.

jessieay:
@msecret @maya in that case, does this seem like a case where we *should* be using BEM? 

https://github.com/18F/web-design-standards/blob/29aac953002f1d295f9374bb6c3ea756c4c9175d/src/stylesheets/components/_footer.scss#L174

PR #1308, opened 2016-07-07
[ Website ]: Remap routes on deployments

rogeruiz:
## Description
This patch adds a map-route command to remap the routes for
`standards[-staging].usa.gov` when deploying `master` and `staging`.
The explicit remapping needs to happen because the `cf-plugin`
`zero-downtime-push` doesn't support anything but a `manifest` and
`path` flag. Because all of our manifests inherit the production
manifest and we rely on the default mapping for our shard domain and
custom mapping for our private domain, we must run `map-route` manually
after a successful `zero-downtime-push`.


juliaelman:
🚀

Issue #1313, opened 2016-07-08
Deadlink on https://standards.usa.gov/form-controls/#text-inputs


debracleaver:
steps to reproduce
1. navigate to https://standards.usa.gov/form-controls/#text-inputs
2. click on the links that says  "accessibility guidelines for all form controls"

expected results:
this is an actual link that takes you to an actual page

actual results:
negative.  it's just a fun thing to click on, doesn't actually link anywhere.

ericadeahl:
@debracleaver Thanks for the tip! Agreed that we probably don't want links that are just fun things to click on. After reviewing the content, it looks like there was never an external link, and the text is in fact referring to itself.

@maya @kategarklavs let's remove the link and change that sentence to:
"As you customize these templates, make sure they meet the accessibility guidelines in this introduction and as described for each control."

PR #1316, opened 2016-07-12
[Website]: Remove deadlink from form controls and form templates accessibility section

maya:
## Description

Removes deadlink from form controls and form templates section and updates the sentence text. 

Also fixes an incorrect link on form controls that was linking to itself when it should be linking to form templates.

### This is what it looks like:

screen shot 2016-07-12 at 10 57 00 am

Resolves: #1313.

### DIFF
--- a/docs/_components/form-controls/form-controls.md
+++ b/docs/_components/form-controls/form-controls.md
@@ -9,6 +9,6 @@ lead: Form controls allow users to enter information into a page.
 
 {% include accessibility.html %}
 
-<p>If you are a building a form with multiple controls, also consider the <a href="{{ site.baseurl }}/form-controls/">accessibility guidelines in the “Form Templates” section</a>.</p>
+<p>If you are a building a form with multiple controls, also consider the <a href="{{ site.baseurl }}/form-templates/">accessibility guidelines in the “Form Templates” section</a>.</p>
 
 {% include child-sections.html parent='Form controls' %}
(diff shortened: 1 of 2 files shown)


ericadeahl:
@maya That second sentence is also confusing. ("In addition, when creating forms with multiple controls or customizing these templates, be sure to:") I think we could just remove it. Thoughts @kategarklavs ?

kategarklavs:
@ericadeahl Agree that this could be clearer. How do you all feel about the following?

"When you're creating forms with multiple controls or customizing these templates, make sure the order of the form controls is the same in the HTML as it is on the screen. Please don't use CSS to rearrange the form controls. Screen readers narrate forms in the order they appear in the HTML, and rearranging that order can provide a compromised experience for some users."

If you all think the additional note at the end isn't necessary, feel free to cut(!)

maya:
@kategarklavs we're referring to the whole section here which includes many bullet points: https://standards.usa.gov/form-templates/ so I don't think we should change that.

ericadeahl:
@kategarklavs You can see the text in context here: https://standards.usa.gov/form-templates/ We're just trying to make the introduction text before the bullets clearer. It currently reads:

"As you customize these templates, make sure they meet the accessibility guidelines in this introduction and as described for each control.

In addition, when creating forms with multiple controls or customizing these templates, be sure to:"

The two sentences are referring to the same list of instructions, so "in addition" doesn't make sense, and "when customizing these templates" is redundant. I thought maybe we could just cut out the second sentence.

kategarklavs:
@ericadeahl Word -- thanks for clarifying (and +1 to removing the _in addition_). Would you all like me to take a more comprehensive look at the whole section, too, or is the removal of the _in addition_ good for now?

ericadeahl:
@kategarklavs Cool, I think we're good for now. @maya let's just remove that sentence starting with "in addition"

kategarklavs:
Sounds good!

maya:
@ericadeahl @kategarklavs done in 3a955dc.

PR #1323, opened 2016-07-14
[GitHub issue template]: Remove "Add issue type"

maya:
## Description

We recently added a `Type: Component` and `Type: Website` label so this should not be necessary in the template.

Thoughts?


juliaelman:
@maya agreed! However, this also removes the `Something else: Please add a descriptive title` bullet point and folks adding a `[Enter some info]` in the title. I am A+ for this improvement and for letting the labels do their job. Just want to make sure that you are cool with this prior to merging in.

maya:
I think it's OK bc in the text before this we explain how we want them to write a title.

PR #1329, opened 2016-07-15
Replace 'ol li' rule w/ 'ol > li' rule

### LINKED ISSUE #1328: Unordered lists nested inside ordered lists appear ordered
The following markup:

```html
<ol>
  <li>
    i am the first item in an ordered list.
    <ul>
      <li>I'm the first item in a nested <em>unordered</em> list.</li>
    </ul>
  </li>
</ol>
```

Looks like this:

> screen shot 2016-07-15 at 6 11 25 pm

The second list item should be bulleted rather than numbered.

I believe I know why this is happening and will issue a PR shortly.

toolness:
I think this fixes #1328 so that unordered lists nested within ordered lists work as expected.

### DIFF
--- a/src/stylesheets/elements/_list.scss
+++ b/src/stylesheets/elements/_list.scss
@@ -41,7 +41,7 @@ ul li {
   }
 }
 
-ol li {
+ol > li {
   counter-increment: table-ol;
   display: table-row;
 


maya:
LGTM. Thanks @toolness!

PR #1335, opened 2016-07-18
Fix header trigger for Android

rogeruiz:
## Description

This patch fixes the breaking change introduced in #1212.


maya:
LGTM 👍

PR #1359, opened 2016-07-22
[UI component] Accordion code refactoring

### LINKED ISSUE #1338: [Accordion]: Accordions should not treat every button as a toggle
## Description

Currently, the accordion.js file treats any button it finds inside of .usa-accordion as a toggle for that accordion. 

```
this.$('button').each(function () {...
```

This creates a few inconveniences:

1. All buttons, at any level, inside of the accordion will toggle the accordion and are treated as a drawer face (for lack of a better term). Thus using another button inside of the accordion that should not act as a toggle isn't possible. For example, if I have a form inside of an accordion the buttons associated with the form will toggle the accordion.
2. Using another element to toggle the accordion is not possible. If I want to use another HTML element to trigger the accordion I cannot. Or more to the point if I want to have an additional element trigger the accordion to close I'd need to write additional JS code to trigger the close.
3. If I nest an accordion, or more commonly a drawer of my own making inside an accordion, then again any button I use inside the drawer may inadvertently trigger the accordion.

Ideally, the JS code would look for a data attribute, or failing that a specific class name, that establishes the element as a toggle for the accordion. Perhaps we are abusing the accordion component by using it in a way that it wasn't intended to be used. In that case, we'd love a drawer component. :-)

As a side note, the ```.usa-unstyled-list``` class that is often added to the unordered list element that houses an accordion also applies it styles a bit too widely. ```.usa-unstyled-list li::before``` removes the bullet from all lists at any level that appear within rather than just first level of list items (i.e. ```.usa-unstyled-list > li::before```. Thus any unordered list that appears inside an accordion's content loses the list item bullets.

msecret:
## Description
Refactors the accordion to make the code easier to use, less specific, more concise and fixes #1338 while maintaining backward compatibility.

## Additional information

Does some of the following issues:
- Moves to using a class for the accordion button `.usa-accordion-button` to fix #1338. Also uses a specific child selector for the current button styles for backwards compatibility.
- Makes the accordion list a top level component rather having to next in another accordion div. This decreases complications, allows you to use any list type for an accordion, and decreases specificity.
- Is backwards compatible with existing accordions so existing code isn't broken.
- Adds a notice and documentation around the old, deprecated accordion to allow people to see how to update to the new accordion. Also includes the actual UI for the deprecated accordion so we can always ensure to test it moving forward.
- Fixes javascript to have the same behavior as the CSS concerning the button.

### DIFF
--- a/src/js/components/accordion.js
+++ b/src/js/components/accordion.js
@@ -70,7 +70,7 @@ Accordion.prototype.show = function ($button) {
  */
 Accordion.prototype.hideAll = function () {
   var self = this;
-  this.$('button').each(function () {
+  this.$('ul > li > button, .usa-accordion-button').each(function () {
     self.hide($(this));
   });
   return this;
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -5,7 +5,6 @@ $accordion-border: 3px solid $color-gray-lightest;
 // Accordion Styles
 
 @mixin accordion-list-styles {
-  // TODO copy
   @include unstyled-list;
   color: $color-base;
   margin: 0;
@@ -14,7 +13,6 @@ $accordion-border: 3px solid $color-gray-lightest;
 }
 
 @mixin accordion-list-item-styles {
-  // TODO copy
   background-color: $color-gray-lightest;
   font-family: $font-sans;
   list-style: none;
@@ -22,6 +20,7 @@ $accordion-border: 3px solid $color-gray-lightest;
   width: 100%;
 }
 
+// scss-lint:disable PropertyCount
 @mixin accordion-button-styles {
   background-color: $color-gray-lightest;
   background-image: url('../img/minus.png');
@@ -55,6 +54,7 @@ $accordion-border: 3px solid $color-gray-lightest;
     margin: 0;
   }
 }
+// scss-lint:enable PropertyCount
 
 .usa-accordion,
 .usa-accordion-bordered {
@@ -69,6 +69,7 @@ $accordion-border: 3px solid $color-gray-lightest;
   > ul {
     @include accordion-list-styles;
 
+    // TODO deprecated ruleset based on old accordion.
     > li {
       @include accordion-list-item-styles;
     }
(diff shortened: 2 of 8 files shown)


src/js/components/accordion.js
```diff
@@ -70,7 +70,7 @@ Accordion.prototype.show = function ($button) {
  */
 Accordion.prototype.hideAll = function () {
   var self = this;
-  this.$('button').each(function () {
+  this.$('ul > li > button, .usa-accordion-button').each(function () {
```
juliaelman: Why not just use the `.usa-accordion-button` as the target? Totally open if you think we need to cover both use cases, but seems like the class specificity will suffice.

src/stylesheets/components/_accordions.scss
```diff
@@ -4,70 +4,99 @@ $accordion-border: 3px solid $color-gray-lightest;
...
 .usa-accordion,
 .usa-accordion-bordered {
+  @include accordion-list-styles;
+
   + .usa-accordion,
   + .usa-accordion-bordered {
     margin-top: 1rem;
   }
 
+  // TODO deprecated ruleset based on old accordion.
```
juliaelman: Is this `TODO` still accurate?

src/stylesheets/components/_accordions.scss
```diff
@@ -4,70 +4,99 @@ $accordion-border: 3px solid $color-gray-lightest;
...
-      box-shadow: $focus-shadow;
+    // TODO deprecated ruleset based on old accordion.
+    button {
+      @include accordion-button-styles;
     }
 
-    &:hover {
-      background-color: $color-gray-lighter;
-      color: $color-base;
+    // TODO deprecated ruleset based on old accordion.
```
juliaelman: Is this `TODO` still accurate?

src/js/components/accordion.js
juliaelman: Ah! I see that you left this in for backwards compatibility. Totally okay to do this, but would like to know your thoughts on when we should remove this use case as we move towards `1.0`.

src/stylesheets/components/_accordions.scss
juliaelman: Dis-regard, as this is reference to the upgrades vs. newer class based specificity.

src/stylesheets/components/_accordions.scss
juliaelman: Dis-regard, as this is reference to the upgrades vs. newer class based specificity.

juliaelman:
Looks good! I added a few notes in regards to how we move forward with the deprecated code for the accordion button styles. Just want to get some thoughts around it, but I don't see it as something that blocks merging this in.

PR #1361, opened 2016-07-26
Change legend font size

jessieay:
Before:

screen shot 2016-07-26 at 2 36 12 pm

After:

screen shot 2016-07-26 at 2 38 24 pm

### DIFF
--- a/src/stylesheets/elements/_inputs.scss
+++ b/src/stylesheets/elements/_inputs.scss
@@ -126,7 +126,6 @@ select {
 }
 
 legend {
-  font-size: $h2-font-size;
   font-weight: $font-bold;
 }
 


msecret:
@ericadeahl pinging you to see what you think of the visual design implications of this change.

jessieay:
to give more context on visual design, this is what my form looked like with old legend size:

screen shot 2016-07-26 at 3 54 23 pm

jessieay:
and now
screen shot 2016-07-26 at 3 55 23 pm

ericadeahl:
@jessieay @msecret I'm not sure whether I understand the proposed change. Would this would reduce the size of the form template titles on https://standards.usa.gov/form-templates/ ("Name", "Mailing address", "Sign in")?

If so, I think we should hold off on making that change, because the appropriate title size depends on context. The smaller titles work on the Micro-purchase page, where there are several sections next to each other and the sections are split into separate boxes. In other contexts the large titles might be more appropriate.

jessieay:
@ericadeahl this is the change the legend style for fieldsets. we don't currently have any examples of this in the web design standards docs but the screenshots here show what we have now vs what it would look like with the proposed changes.

I believe this is a good change because fieldsets are usually used to group field together. When the legend is large, it is so much bigger than any other content in the form that it is overpowering. More on fieldsets: https://www.sitepoint.com/web-foundations/fieldset-html-element/

jessieay:
more on legends / fieldsets: http://www.htmlgoodies.com/tutorials/forms/article.php/3479111

jessieay:
Actually, I just inspected the code on the docs and realized that I am wrong: that is a fieldset! But it isn't being used like a normal fieldset. 

Perhaps there are also accessibility reasons for using a fieldset but in most apps I've worked with people only use them for visually grouping fields together.

In most forms I've worked with, he add a header to the top similar to the `Name` header in the form in the web design standards docs. With the fieldset legend being as large as it is, what would you recommend using as the main header for the entire form?

msecret:
I think I'm starting to see what you mean here.

- The WDS uses a fieldset and then a legend for the whole form. This means the legend has a large size applied to it.
- Micropurchase uses a fieldset and legend for groups or parts of a form, so the legend doesn't need such a large size.

Legend and fieldset can be used to describe a whole form, a group of elements or even one element, so both uses cases seem to make semantic sense. I would agree with @jessieay that applying a large font size on all legends doesn't necessarily seem correct, because there's no guarantee that the legends is the heading for the form. 

Maybe we could leverage the "only-child", "only-of-type" selector to only set the legend to a large font size if the fieldset is the only one on the form. If there are more then one, then we won't set the font size so large.

```css
form fieldset legend {
  font-size: $some-normal-font-size;
}

form fieldset:only-of-type legend {
  font-size: $h2-font-size;
}
```

jessieay:
that's a nice idea, @msecret.

juliaelman:
Reached out to @jessieay and we decided to close this pull request without merge. Let's plan on re-opening if the issue arises again, otherwise folks are welcome to override these styles with a custom stylesheet.

PR #1278, opened 2016-06-17
Searchbar text

PilgrimMemoirs:
## Description

Resized the search bar text for small and medium search bars from 1.7rem to 1.5rem

## Additional information

* In search-bar.md HTML file: Added new class to medium search bar form tag: usa-search-medium
* In _search.scss Stylesheet: Added text-size properties on .usa-search-medium and .usa-search-small inputs. Set properties to 1.5rem. 
* Final product:

screen shot 2016-06-17 at 3 04 59 pm

* Orignally looked like this: 

searchbox_og

PR #1362, opened 2016-07-27
[UI component]: Search bar text font sizing

donjo:
## Description

This is a continuation of the work started in #1278 where I cleaned up a few things

## Additional information

The medium and small search both use the same font size since the fields are the same height. I changed the size to reference the font size variable as requested and removed some extra code. 

screen shot 2016-07-27 at 4 38 53 pm


msecret:
Checking in with @ericadeahl on the visual design implications of this change. From a code perspective, everything looks good.

ericadeahl:
Looks good to me!

Issue #1353, opened 2016-07-21
[ Website ]: Refactor JavaScript dependency


rogeruiz:
## Description

Most of the of the JavaScript for the website should be coming either from functionality within the standards itself, or from the build system deploying the site. Little to no JavaScript code should exist on the website unless it fits into specific scopes of functionality. The reasoning behind this is that there should be less code that exists as a custom one-off for the website and instead this code should be available within the standards UI component themselves. Examples of this code is below:

```shell
├──  sticky-nav.js # used to stick the navigation on scroll. (experimental)
├──  scroll-to-top-for-hash.js # used to scroll when internal hash links are clicked.
├──  handle-disabled-links.js # used to disable any links with `href="#"`.
├──  generate-code-snippets.js # used to generate code accordions on the client-side.
└──  calculate-anchor-position.js # used by `scroll-to-top-for-hash` to anchor the scroll position properly.
```

## Explanation of scope for JavaScript on the website

- If the JS is used to gather user-data about the website, such as analytics, then it should remain on the site. 
- If the JS is used to test upcoming components, such as a sticky navigation currently on the website and not in the standards UI components, then it should remain on the site with a message in the `console` and in the code that mention it is in an experimental state.
    - If the JS is used to provide progressive enhancement to users of the documentation site, such as scrolling anchored text below a fixed navigation element, it should remain on the site with a similar message to the above ☝️ 

## Acceptance Criteria

- [x] Code snippets should be generated during the build step
- [x] Anchor tags `<a>` should be explicitly set to `href="javascript:void(0);"` to avoid jumping to the top of the page where appropriate
- [ ] Any JS code that still exists in the website should come with a `console` warning and documentation around it's reasoning along with information as to when it may or may not be making it back into the standards.

juliaelman:
Referencing #1343 so that we can track the work that happens between these two issues, as they possibly relate with one another.

maya:
> Anchor tags <a> should be explicitly set to href="javascript:void(0);" to avoid jumping to the top of the page where appropriate

This is what we had initially. So will we revert the decision made here: https://github.com/18F/web-design-standards/issues/743 ?

rogeruiz:
@maya yes I would say so. Because since using this `handle-disabled-links` JavaScript we've had an issue with the website and PR work associated with it.

Issue: #990
PR: #1285

francisrupert:
[...]

PR #1368, opened 2016-08-02
[ Website ]: Removing JavaScript from website

rogeruiz:
### Description

Removing the JavaScript from the website will help with maintaining the website going forward. Take a look at the AC from #1353.


rogeruiz:
Do we want to add some documentation for the reasoning behind #1378 in this PR? _We've gone back and forth on this issue of `href="#"` versus `href="javascript:void(0);"`_. I'd like to remove the ambiguity here and avoid discussions around what we suggest. The latter, `href="javascript:void(0);"` should be what our examples use so we don't need any additional javascript to prevent `href="#"` from going to the top, but we should have some documentation on the UI component for `links` maybe that discusses:

-  why we do this in the examples
-  why your _mileage may vary_ when setting `href="#"` when implementing the standards.

Thoughts on this @yozlet @maya @msecret @juliaelman @kategarklavs @jameshupp

maya:
@rogeruiz We use these in many components: footer, forms, etc so I don't think the reasoning for it would go in links since it's styleguide site specific thing. But I love the idea of explaining our reasoning, so we should do that on the issue that was originally opened for it and tag the people involved in it.

rogeruiz:
This does not touch the `sticky-nav.js` or any of the JavaScript related to the `side-navigation` UI component. Those will need to be worked on in another PR.

PR #1372, opened 2016-08-03
[Website] Removing descendent reference from unstyled list utility.

juliaelman:
## Description

This pull request cancels out the list styles from the websites sidenav that were introduced [here](https://github.com/18F/web-design-standards/commit/1bea7f69873bf753fe268508bbbdd2a5a3f549a2#diff-fced6c74fbf3140d02e92fd710e888eeL34). I am assuming that this does not work as expected b/c it is a mixin, but would need to do some more digging to verify.

### DIFF
--- a/src/stylesheets/core/_utilities.scss
+++ b/src/stylesheets/core/_utilities.scss
@@ -31,7 +31,7 @@
   margin: 0;
   padding: 0;
 
-  > li {
+  li {
     display: list-item;
     margin: 0;
 


msecret:
Is the addition of the descendent reference break something else somewhere on the site? Or are the accordions broken in some way?

juliaelman:
@msecret good question! As far as I can tell, this reference in the utility only breaks the sub navigation on the website. When debugging, I found that it is where the bug was introduced into the websites code base.

However, being specific here does make sense and am closing this in favor of #1374.

Issue #1091, opened 2016-03-30
[Lists]: ::before and ::after pseudo selectors should not be used for layout purposes


debracleaver:
Hey team,
My team is noticing that you're using pseudo selectors for layout purposes, namely to add margin to html elements.  This seems incorrect.  We'd love to know your reasoning for this, and/or for you to remove this wonky CSS and just add margin/padding to the element itself.  Here's an example from your CSS:

`ul li:after, ol li:after {
    content: "";
    display: block;
    margin-bottom: 0.5em;}`

That's ... bizarre.

justinelam:
+1

maya:
Hi there, thanks so much for the question. We get this a lot actually. The reason we do this is bc lists are actually difficult to align flush and this technique allows us to do so. 

We explain the reasoning here: https://github.com/18F/web-design-standards/issues/1027

Let us know if you have any questions!

debracleaver:
i see our favorite issue is back open :)

maya:
@debracleaver indeed! Closed it to soon, sorry, we typically wait around two weeks for a response from the issue opener before closing. Although we're discussing it internally as well. i.e. how important is standardizing alignment vs. working with the medium of the web.

maya:
Adding in Luke Keller's feedback from the blog post (https://18f.gsa.gov/2016/04/05/three-teams-using-the-draft-us-web-design-standards-talk-about-their-experiences/):

>I'd like to see the Standards [...] remove the pseudo elements from list items by default. I've never seen this approach to list items before, so it would be nice if this treatment was handled through a class, i.e. usa-list-item.

patrickcate:
I've run into this issue as well. After doing a fresh Drupal 7 install to look at the [Drupal 18f theme](https://github.com/18F/web-design-standards-drupal) project incorporating the standards, the list pseudo elements caused layout issues with the Drupal admin toolbar. 

I think this is going to cause major issues down the line as well because using the `ul` and `ol` tags for semantic html structures is very common. I foresee many sites trying to implement the WDS will have to regularly override these pseudo elements.

Perhaps a good middle ground would be to move the current pseudo element list styling to a utility class `usa-list`?

<img width="929" alt="screen shot 2016-04-21 at 9 58 15 pm" src="https://cloud.githubusercontent.com/assets/6277206/14740356/47f96b5c-085c-11e6-8fed-60ea17b77ef5.png">

zaneriley:
[...]

PR #1374, opened 2016-08-03
[Sidenav]: Add unstyled list mixin to inner list in sidenav

maya:
## Description

Since adding the direct selector to the unstyled list mixin, the styles don't carry down to a nested list in the sidenav. This adds the mixin to the sublist to fix that.

Supersedes #1372.

Before you hit Submit, make sure you’ve done whichever of these applies to you:

- [x] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [x] Run `npm test` and make sure the tests for the files you have changed have passed.
- [x] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [x] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #1375, opened 2016-08-03
[Lists]: Simplify lists to use browser defaults

maya:
## Description

Simplifies lists to use browser defaults, and not use on pseudo selectors.

Resolves: #1091.

Dependent on #1374 for sidenav to look correct.

Notable change is that we are switching styles, and bullets will look slightly larger depending on your browser default. Below is Chrome.

## Screenshots
**Before** (custom list and bullets):
screen shot 2016-08-03 at 1 02 13 pm

**After** (Chrome browser default):
screen shot 2016-08-03 at 1 39 50 pm

### DIFF
--- a/src/stylesheets/core/_utilities.scss
+++ b/src/stylesheets/core/_utilities.scss
@@ -26,9 +26,13 @@
 // Unstyled list helper
 
 @mixin unstyled-list() {
+  @include margin(0 null);
   list-style-type: none;
-  margin: 0;
-  padding: 0;
+  padding-left: 0;
+
+  > li {
+    margin-bottom: 0;
+  }
 }
 
 // Content size helpers
--- a/src/stylesheets/core/_utilities.scss
+++ b/src/stylesheets/core/_utilities.scss
@@ -26,23 +26,9 @@
 // Unstyled list helper
 
 @mixin unstyled-list() {
-  display: block;
   list-style-type: none;
   margin: 0;
   padding: 0;
-
-  > li {
-    display: list-item;
-    margin: 0;
-
-    &::before {
-      display: none;
-    }
-
-    &::after {
-      display: none;
-    }
-  }
 }
 
 // Content size helpers
--- a/src/stylesheets/elements/_list.scss
+++ b/src/stylesheets/elements/_list.scss
@@ -1,6 +1,6 @@
 ul,
 ol {
-  padding-left: 1.94em;
+  padding-left: 1.94em; // Approximately 15px left padding at default font size
 }
 
 li {
--- a/src/stylesheets/elements/_list.scss
+++ b/src/stylesheets/elements/_list.scss
@@ -1,5 +1,6 @@
 ul,
 ol {
+  @include margin(1em null);
   padding-left: 1.94em; // Approximately 15px left padding at default font size
 }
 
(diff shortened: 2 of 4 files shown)


src/stylesheets/core/_utilities.scss
```diff
@@ -26,22 +26,12 @@
 // Unstyled list helper
 
 @mixin unstyled-list() {
-  display: block;
+  @include margin(0 null);
```
msecret: What's the reason for this? I haven't seen this pattern before.

src/stylesheets/elements/_list.scss
```diff
@@ -1,20 +1,14 @@
 ul,
 ol {
-  @include margin(2em null 2em 0.9em);
-  list-style: none;
-  padding-left: 0;
-  display: table;
+  padding-left: 1.94em; // Approximately 15px left padding at default font size
```
msecret: Why would this be `em` rather then `rem`?

src/stylesheets/elements/_list.scss
```diff
@@ -1,20 +1,14 @@
...
+}
 
-  li {
-    line-height: $base-line-height;
-    margin-bottom: 0.75em;
-    margin-top: 0.75em;
+li {
+  line-height: $base-line-height;
+  margin-bottom: 0.5em;
```
msecret: Why would this be `em` rather then `rem`?

src/stylesheets/elements/_list.scss
maya: To keep it relative to the font size of the list: a bigger font = bigger padding.

src/stylesheets/elements/_list.scss
maya: Same as above: to keep it relative to the font size of the list: a bigger font = bigger padding.

src/stylesheets/core/_utilities.scss
maya: We had `margin: 0` before, when actually the only margins added in the list was top and bottom. We do not need to declare margin=0 on the left and right.

maya: Note: it's not undoing our margins (we did not add any margins for `ul`s), but webkits margins. You'll find:
```
-webkit-margin-before: 1em;
-webkit-margin-after: 1em;
```
applied to `ul` in the user agent styles. Which makes me think, should we add these in to be more explicit?

msecret: So this rule will be expanded to:

```
margin-bottom: 0;
margin-top: 0;
```

I definitely don't think it's a good idea to mess with the webkit specific margin before rules. Those should be overridden with regular margin rules.

maya: Right, I wasn't suggesting adding `-webkit-margin-before` ourselves. I meant should we add `margin-bottom: 1em;` ourselves, since webkit is already doing it + to make it more consistent across browsers. What do you think?

msecret: Agreed

maya: Done in 46a01a7. This will make it more consistent across browsers and match our paragraph text margins.

PR #1376, opened 2016-08-03
Rebuild dist folder.

toolness:
This is a follow-up to #1337 which rebuilds the `dist` directory in the cg-style branch of uswds. This seems to be a requirement for bringing the changes into cg-style itself, which (I think?) uses the built uswds files in `dist`.

@msecret am I doing the right thing here?


maya:
Curious: since we fixed the custom variables issue, is this still necessary?

rogeruiz:
interesting. what branch is this merging into? is the `dist/` directory still being tracked on that branch? @toolness @msecret

toolness:
Good questions!  We'd talked a bit about using the real USWDS 0.10.0 since the custom variables issue is fixed in that, but there is a bit of concern regarding how much work will be involved to make sure there aren't any visual regressions when we do this (for example, one we'll have to deal with in cg-style is the removal of auto-hiding elements with `aria-hidden="true"`).

@rogeruiz this is merging into the branch that cg-style is currently using to pull uswds into the cloud.gov style guide.  It's super out of date compared to the latest release of uswds, however, and I believe it needs the `dist/` directory because cg-style checks it out of git directly (instead of using npm) and seems to use the pre-built CSS instead of using the original SASS.

So right now I think the possible solutions are:

1. Just merge this PR now and we'll at least get the fix for #1328 into cg-style, which will be particularly helpful for the cloud.gov docs, as there are a number of places we want to have unordered lists nested within ordered ones. This is low-risk but still leaves cg-style using a really old fork of uswds; it's not a great long-term solution.

2. Leave this PR unmerged and change cg-style to use uswds 0.11.0 via NPM, since the reason that this branch needed to be used in the first place--something having to do with custom variables--has since been resolved in mainline uswds.  This is great for the long-term but does carry with it the risk of causing visual regressions for cg-style, since a lot has changed in uswds since cg-style's fork branched from it.

msecret:
I decided yesterday to move `cg-style` off this branch and track the npm package, so I don't think this is necessary anymore.

PR #1408, opened 2016-08-18
[Headers]: Fix code sample display

maya:
## Description

This fixes the code sample display by adding the partial code directly into the components as recommended by @rogeruiz. 

Previously, we had to show code sample html and the actual html for the partials depending on the use case or else it would render the partial inside the code sample. We tried doing this but were not able to get the indentation correctly displayed bc of how jekyll formats includes. It would reset all the indentation to the left for the partials.

Also, just noted some very strange display with the previous method on the headers page. This fix should resolve that. Although it would be very cool to use the mini-components that are consumed by bigger components, it's not possible at this point. 

cc: @yozlet @juliaelman @msecret


maya:
One question I have is: should we remove the partials (dropdown, mega menu, secondary nav) from the `src/html` directory?

Feel free to respond to this or emoji vote: 👍 or 👎

juliaelman:
@maya I 👍  this at first because I _think_ I know where you are heading, but am not entirely clear what your thoughts are as an alternative. Can you explain your thoughts a bit more?

maya:
@juliaelman sure, around what specifically?

juliaelman:
@maya in regards to this question:

> One question I have is: should we remove the partials (dropdown, mega menu, secondary nav) from the src/html directory?

maya:
@juliaelman @msecret partials deleted in 5bf151e. This is ready.

PR #1416, opened 2016-08-23
[WIP] [UI Component] Remove jQuery from add-headers branch

yozlet:
## Description

Remove all use of jQuery in this branch, and replace it with use of our growing utils collection, along with vanilla JS.

## Additional information

This work also contains new utils: `addClass` and `removeClass`, along with a couple of related bugfixes.

### DIFF
--- a/src/js/components/header/mobile.js
+++ b/src/js/components/header/mobile.js
@@ -1,8 +1,26 @@
-var $ = require('jquery');
+var select = require('../../utils/select');
+var addClass = require('../../utils/add-class');
+var removeClass = require('../../utils/remove-class');
+var dispatch = require('../../utils/dispatch');
 
-$('.usa-menu-btn, .usa-overlay, .usa-nav-close').on('click touchstart', function (e) {
-  $('.usa-overlay, .usa-nav').toggleClass('is-visible');
-  $('body').toggleClass('usa-mobile_nav-active');
-  $('.usa-nav-close').focus();
-  e.preventDefault();
+var navElements = select('.usa-menu-btn, .usa-overlay, .usa-nav-close');
+var toggleElements = select('.usa-overlay, .usa-nav');
+var navCloseElement = select('.usa-nav-close')[ 0 ];
+
+navElements.forEach(function (element) {
+  dispatch('click touchstart', function (e) {
+    toggleElements.forEach(function (element) {
+      toggleClass(element, 'is-visible');
+    });
+    toggleClass(document.body, 'usa-mobile_nav-active');
+    navCloseElement.focus();
+    return false;
+  });
 });
+
+function toggleClass (element, className) {
+  // TODO: Add IE8 support
+  if (element.classList) {
+    element.classList.toggle(className);
+  }
+}
--- a/src/js/components/header/mobile.js
+++ b/src/js/components/header/mobile.js
@@ -19,7 +19,6 @@ navElements.forEach(function (element) {
 });
 
 function toggleClass (element, className) {
-  // TODO: Add IE8 support
   if (element.classList) {
     element.classList.toggle(className);
   }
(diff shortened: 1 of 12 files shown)


yozlet:
Remaining work:
 - [ ] Make `dispatch()` accept an array or string list of event types

Question: Does `toggleClass()` in `mobile.js` need to be IE8 compatible? Right now it won't work but it won't error out either. And given that this code _presumably_ is only for mobile, I'm inclined to leave it.

maya:
hi @yozlet! Since we've merged the headers work into `staging` would it make sense to open this PR against `staging`?

maya:
@yozlet to answer your question, we officially support IE9+, so no we're not required to support IE8. A couple of times, if the fix is easy enough to support IE8 we've sometimes added it in.

maya:
The `mobile.js` is for small screens, so yes that's a fair assumption to leave it out -- I don't think many of our IE8 users will be on small screens.

yozlet:
OK, `dispatch()` is fixed, and we'll leave the integration test for a separate PR.

src/js/components/header/search.js
```diff
@@ -1,38 +1,52 @@
...
-    });
+    searchDispatcher = dispatch(document.body, 'click touchstart', searchOpenClickHandler);
+  }
+
+  return false;
+}
+
+function searchOpenClickHandler (event) {
+  var target = event.target;
+  if (! searchFormContains(target)) {
```
msecret: The space after `!` feels a little weird.

src/js/initializers/skip-nav.js
```diff
@@ -8,12 +8,15 @@ whenDOMReady(function () {
...
 
-  dispatch(elSkipnav, 'click', function () {
-    elMainContent.setAttribute('tabindex', '0');
-  });
-
-  dispatch(elMainContent, 'blur', function () {
-    elMainContent.setAttribute('tabindex', '-1');
-  });
+  if (elSkipnav) {
+    dispatch(elSkipnav, 'click', function () {
```
msecret: Should this be click and touch?

src/js/components/header/mobile.js
```diff
@@ -1,8 +1,25 @@
...
+    toggleElements.forEach(function (element) {
+      toggleClass(element, 'is-visible');
+    });
+    toggleClass(document.body, 'usa-mobile_nav-active');
+    navCloseElement.focus();
+    return false;
+  });
 });
+
+function toggleClass (element, className) {
```
msecret: Should this be a generic function rather then one just in this file?

msecret:
Reviewed everything looks good. I'd say this is ready to go after comments looked at.

src/js/components/header/search.js
yozlet: There for readability & obviousness, because (in my experience) people often miss the exclamation mark in `(!something)`.

That said, this is the only usage of that method. I was considering naming it `searchFormDoesNotContain()` but that seemed rather unwieldy. What do you think?

src/js/initializers/skip-nav.js
yozlet: No idea. I haven't changed the events that were there before, just rearranged them. It's very possible, but I don't have enough experience with event types and our target behaviours to make that decision. :(

src/js/components/header/mobile.js
yozlet: My attitude so far has been to leave potential util functions in their existing files if they're only used in that place. Once more than one component needs them, I pull them out (like `addClass` & `removeClass` elsewhere in this PR).

src/js/initializers/skip-nav.js
yozlet: OK, after discussion with people who understand this stuff, we've settled on just `click`.

PR #1418, opened 2016-08-24
[Gov banner]: Add gov banner

maya:
## Description

Adds banner addressing how the public can trust the government sites.

https://standards-staging-add-gov-banner.apps.cloud.gov/headers/basic/ 

Resolves: #1340.

### TODO:

- [x] Add markup to other headers and website #1464
- [x] JS PR #1462
- [x] Update screenshots to header + banner versions
- [x] Write some JavaScript to the effect of:
```
if #gov-banner aria-hidden=false 
then
add class .usa-banner-official-expanded to .usa-banner-official
```


src/img/angle-arrow-up-primary-hover.svg
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
+<!-- Generator: Adobe Illustrator 16.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" id="Capa_1" x="0px" y="0px" width="512px" height="512px" viewBox="0 0 284.929 284.929" style="enable-background:new 0 0 284.929 284.929;" xml:space="preserve">
+<g>
+	<path d="M282.082,195.285L149.028,62.24c-1.901-1.903-4.088-2.856-6.562-2.856s-4.665,0.953-6.567,2.856L2.856,195.285   C0.95,197.191,0,199.378,0,201.853c0,2.474,0.953,4.664,2.856,6.566l14.272,14.271c1.903,1.903,4.093,2.854,6.567,2.854   c2.474,0,4.664-0.951,6.567-2.854l112.204-112.202l112.208,112.209c1.902,1.903,4.093,2.848,6.563,2.848   c2.478,0,4.668-0.951,6.57-2.848l14.274-14.277c1.902-1.902,2.847-4.093,2.847-6.566   C284.929,199.378,283.984,197.188,282.082,195.285z" fill="#205493"/>
+</g>
```
msecret: I think this svg file can be optimized/minimized. I did a quick google search and found this: http://petercollingridge.appspot.com/svg-optimiser, but maybe there's something better.

src/stylesheets/components/_disclaimer.scss
```diff
@@ -1,10 +1,21 @@
 .usa-disclaimer {
   background-color: $color-gray-lightest;
+  padding-bottom: 7px;
 //  height: 2.3rem;
```
msecret: Can this be removed?

src/stylesheets/components/_disclaimer.scss
```diff
@@ -1,10 +1,21 @@
 .usa-disclaimer {
   background-color: $color-gray-lightest;
+  padding-bottom: 7px;
```
msecret: This value is technically more then 5px, can it be an `rem` unit?

src/js/components/banner.js
```diff
@@ -0,0 +1,16 @@
+var select = require('../utils/select');
+var addClass = require('../utils/add-class');
+var removeClass = require('../utils/remove-class');
+var dispatch = require('../utils/dispatch');
+
+var headers = select('.usa-banner-header');
+headers.forEach(function (header) {
+  var click = function (e) {
```
msecret: Does this require an `e.preventDefault()`?

ericadeahl:
Hi @maya! A few comments on the banner:

**On desktop and mobile:**

Please remove the space between "The gov means it's official" and the rest of the text, so it looks like this. The space was in the design, but the alignment is looking weird to me when I see it on the site.

screen shot 2016-08-29 at 3 47 13 pm

**On mobile:**

Ooh, this is looking better than the last time I looked at it!

I know I designed it this way, but I'm finding it jarring to have that "An official website" text jump around and get bold, and that dark line appear below. I think we should just treat this like the desktop version, so the top portion remains the same collapsed and expanded, like this:

mobile expanded copy 2

My only worry is that it might not be obvious enough how to close the banner. Wondering if we should add a big collapse button at the bottom. Thoughts? Is it weird to have duplicate ways to collapse the banner? @maya @bradnunnally @donjo 

mobile expanded copy

donjo:
@ericadeahl I like the two proposed changes, though I'm not sure how I feel about the collapse button at the bottom. If the entire top bar were a touch target (as opposed to just the hyperlink) could that work? Or some combination of the two with a lighter bottom-border and 1.5rem margin instead of 2rem? 

screen shot 2016-08-29 at 3 42 27 pm

(Note: that's a random light grey I chose via color picker, I didn't consult the style guide for something we should actually implement)

ericadeahl:
@donjo @maya Yeah I'm ok with that solution. The color of the bottom-border should match the rule that's under the top navigation on the site:

screen shot 2016-08-29 at 6 30 24 pm

@bradnunnally @lauraponce I know @donjo is out on vacation the rest of the week, so once this change is made, could one of y'all check this out on a mobile device and make sure it feels ok? One of the reasons I was thinking about removing the "X" was that when I looked at it on mobile I expected the close link to be in the same place as the open link; having to move my finger over to the X was cumbersome. But making the whole area the touch target could help that. It might be worth putting it in front of a few people to validate - I leave it to you!

bradnunnally:
@maya I'll take a look at this today.

bradnunnally:
My thoughts on the X is to leave it as is. It's a standard location for interactions like this and add it where the open was I think will make that area kind of crowded. Agreed making the whole upper banner the touch target though just to make it easy to tap. 

X for affordances, whole thing touch target for usability. 

/cc @maya @donjo @lauraponce

src/img/angle-arrow-up-primary-hover.svg
maya: Fixed in a1401288ab1a1457e8dd6974d15473507cbfba4c along with 2 other un-optimized SVGs.

src/stylesheets/components/_disclaimer.scss
maya: @msecret there's no `_disclaimer.scss` file anymore, this is gone.

src/stylesheets/components/_disclaimer.scss
maya: @msecret I think this was commented on an old version of the PR. The new work is in `_banner.scss` and this code doesn't exist.

src/js/components/banner.js
maya: @shawnbot @juliaelman @yozlet thoughts?

shawnbot: I don't think it _requires_ one, but it's probably a safe bet. The only reason I can think that it would help is if the banner lives in a `<form>` and you want to short-circuit submission when the button is clicked. You can solve this on the markup side with `type="button"`, but `e.preventDefault()` works too. Any other reasons, @msecret?

maya:
Does this mean we want the whole bar to be a click target on desktop too? @shawnbot @donjo @bradnunnally @lauraponce

src/js/components/banner.js
yozlet: I _think_ we're avoiding use of `Event.preventDefault()` because it's not fully supported in older IEs, but having said that, I'm not entirely sure of the story. I've tended to switch them to `return false;` (at the end of the function) even though that's not quite the same thing.

I'll look into it and see if I can get a more solid tactic here.

maya:
Made a few of @ericadeahl's change requests: 
- removed extra line break in 512dd46
- updated border color in c8fcbdd

maya:
@bradnunnally can you confirm from @donjo's screenshot in https://github.com/18F/web-design-standards/pull/1418#issuecomment-243249885. Are we keeping the increase in font size on mobile, but just un-bolding it?

src/js/components/banner.js
yozlet: OK, here's how to avoid breakage on IE8 (which I think is worthwhile):

    (event.preventDefault) ? event.preventDefault() : event.returnValue = false;

... which can be anywhere in the function, _if_ we _do_ want `preventDefault()` behaviour.

maya:
Just spoke with @bradnunnally - ideally, the whole banner would be a touch target on mobile and _not_ on desktop. However, for MVP it's OK if it's not the whole banner a target on mobile. We are still meeting Apple HID's recommendation of 44px touch target on the close button.

I will investigate how heavy the lift is to make sure.

maya:
Added documentation reviewed by @kategarklavs in #1444.

maya:
Full touch target on mobile is added! cc: @bradnunnally

src/js/initializers/banner.js
```diff
@@ -0,0 +1,9 @@
+var whenDOMReady = require('../utils/when-dom-ready');
+
+whenDOMReady(function initBanner () {
+
+  // Banner
+  require('../components/banner');
```
yozlet: I didn't catch this earlier, but `banner.js` should really be exporting an initialisation function which this initialiser code calls, similar to what the Accordion, Footer and other components do. (Apart from `header.js`, which we should probably fix as well. Sorry, I didn't think about this properly before!)

src/js/components/banner.js
yozlet: OK, given Shawn's comment, I guess we do want this behaviour. @maya, can you add the line above to the top of the click function?

src/js/utils/dispatch.js
```diff
@@ -6,15 +6,15 @@
  * @param {Function} listener - the function to be executed
  * @returns {Object} - containing a <tt>trigger()</tt> method for executing the listener, and an <tt>off()</tt> method for detaching it
  */
-module.exports = function dispatch (element, eventTypes, listener) {
+module.exports = function dispatch (element, eventTypes, listener, options) {
```
yozlet: The new `options` arg looks potentially useful, but what's using it?

src/js/components/banner.js
maya: Added in 2e5da47. @yozlet @shawnbot

src/js/utils/dispatch.js
maya: @shawnbot would you happen to know?

shawnbot: When I was looking at using `dispatch()` for the banner I originally thought that being able to pass `true` (or `{capture: true}`) to catch clicks during the "capture" phase would be helpful. The changes we made should probably stay in so that we can pass listener options at some later point if need be.

src/js/initializers/banner.js
shawnbot: I defer to you on this one, @yozlet. IIRC, we cribbed the banner initializer from one of the other ones, and (at that time, at least) whichever one we copied didn't export an initialization function.

src/js/utils/dispatch.js
yozlet: If it's not being used then I strongly recommend taking it out. Unused code infrastructure is great for confusing people, but not much else. Especially this kind of change, which is very easy to add when we need it.

src/stylesheets/components/_disclaimer.scss
```diff
@@ -1,7 +1,6 @@
 .usa-disclaimer {
   background-color: $color-gray-lightest;
-  font-size: 1.2rem;
-  height: 2.3rem;
+//  height: 2.3rem;
```
juliaelman: Can we just remove this?

juliaelman: Ah! I see now that we are removing this entirely 👍

src/html/header-basic.html
```diff
@@ -1,6 +1,40 @@
 
   <a class="usa-skipnav" href="#main-content">Skip to main content</a>
   <header class="usa-header" role="banner">
+    <!-- Gov banner BEGIN -->
+    <div class="usa-banner">
+      <div class="usa-accordion">
```
juliaelman: Is there a reason why we need to wrap this with `usa-accordion`?

maya: In order for the accordion behavior to function, it needs this.

src/html/header-basic.html
```diff
@@ -1,6 +1,40 @@
 
   <a class="usa-skipnav" href="#main-content">Skip to main content</a>
   <header class="usa-header" role="banner">
+    <!-- Gov banner BEGIN -->
```
juliaelman: This gets added to our code snippets. I worry that it might be possibly be confusing to users when looking at the snipper. They might wonder if the skipnav above this line is also part of our example code.

Also, are we doing this anywhere else in our other code snippets?

src/html/header-basic.html
```diff
@@ -1,6 +1,40 @@
...
+          <div class="usa-banner-guidance-ssl usa-width-one-half">
+            <img class="usa-banner-icon usa-media_block-img" src="{{ site.baseurl }}/assets/img/icon-https.svg" alt="SSL">
+            <div class="usa-media_block-text">
+              <p>This site is also protected by an SSL (Secure Sockets Layer) certificate that’s been signed by the U.S. government. The <strong>https://</strong> means all transmitted data is encrypted  — in other words, any information or browsing history that you provide is transmitted securely.</p>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+    <!-- Gov banner END -->
```
juliaelman: Same comment as [above](https://github.com/18F/web-design-standards/pull/1418/files#r77702485).

src/stylesheets/components/_banner.scss
```diff
@@ -0,0 +1,198 @@
+.usa-banner {
+  background-color: $color-gray-lightest;
+  padding-bottom: 0.7rem;
+
+  @include media($small-screen) {
+    font-size: $h6-font-size;
+    padding-bottom: 0;
+  }
+
+  .usa-accordion-content {
```
juliaelman: Is there a reason why we are using `usa-accordion-content` and not `usa-banner-content`?

src/html/header-basic.html
maya: Do you mean are we adding a comment anywhere else in our examples?
No, we're not, but it might be a good practice to do so for complex components made up of many components. It could help add more clarity to the code. I appreciate how Google Material Design Lite does this: https://getmdl.io/components/index.html#layout-section

> They might wonder if the skipnav above this line is also part of our example code.
Since it all gets added to the code sample, I think they'll be able to see the entire snippet and know it's all together. 

Additionally, we've not heard that concern from users before. We've actually heard the opposite, that it's hard to read/confusing when the snippets are too large (at least that's what I remember hearing from the listening tour).

src/stylesheets/components/_banner.scss
maya: No reason, updated in 48683f0.

juliaelman:
@maya left some notes and took a quick look at the HTML/CSS and looks GTG. 🚀

yozlet:
@maya JS refactor is merged!

maya:
Quick check - there's a flash of content on page load. Is that OK?

donjo:
Oh, a thing I missed earlier when vertical aligning was that the down arrow needs to be aligned to match the text now that the text moved.

```
.usa-banner-button[aria-expanded=false]:after {
   vertical-align: middle;
}
```

maya:
@donjo `.usa-banner-button[aria-expanded=false]:after` and `.usa-banner-button[aria-expanded=true]:after` are actually relatively in the same location to each other and the icon is relative to the button so it moved together with the button text when we added `vertical-align: middle;` to the button.

maya:
@donjo Since this may be a background position issue rather than a vertical align issue, I moved it down 1px here: 167928e and it seemed to look right.

juliaelman:
@maya in reference to https://github.com/18F/web-design-standards/pull/1418#issuecomment-245671798

This flickering occurs do to the `aria-hidden` attr being added dynamically with JS and the "hiding" styles being added only when that attribute is attached to the given selector. Upon further review, this seems to be happening across various components that utilize this method of implementation.

I'll be opening an issue for solving in the Draft U.S. Web Design Standards overall. This should not block merging this PR into staging.

Issue #1343, opened 2016-07-19
Separation of WDS website and WDS components into different GH repos


juliaelman:
# Description

The core team has decided to move forward with the separation of the website/docs codebase and the components code into different repos. Based off of [user research and feedback](https://github.com/18F/web-design-standards-ux/tree/master/June_Chicago_User_Observation), our goal is to create a better understanding of how the two work together both for the users of the Standards and our contributor base.

## Tasks
- [x] [Comms strategy for announcing split to user base](https://github.com/18F/web-design-standards/issues/1393) 
- [x] Create new repo with website code separated out
- [x] Create `je-split_repo` branch (FKA `split-repo`)
- [x] [Remove website dependencies from `web-design-standards`](https://github.com/18F/web-design-standards/issues/1394)
- [x] [Remove website code from `web-design-standards`](https://github.com/18F/web-design-standards/issues/1395)
- [x] [Test Standards as stand alone package](https://github.com/18F/web-design-standards/issues/1396)
- [x] [Update documentation](https://github.com/18F/web-design-standards/issues/1397)
- [x] [Documentation review](https://github.com/18F/web-design-standards/issues/1398)
- [x] [QA code base](https://github.com/18F/web-design-standards/issues/1399)

## Additional information

We'll need to be clear in our comms and outreach to our users prior to the finalization of this work. See https://github.com/18F/web-design-standards/issues/1341 for the long term comms plan for scheduling this messaging.

kategarklavs:
@juliaelman Hey! Thanks for tagging me in this. I'll block off some time on your calendar (and/or on other folks' calendars) to pick your brains about the necessary documentation.

joshbruce:
Tagging my team so they are aware: @BrianGilmanGSA, @ChristyHermansen, @jenbesergsa

juliaelman:
The groundwork has been completed for the new docs/website repo here: https://github.com/18F/web-design-standards-docs

I have added a disclaimer in the [README.md](https://github.com/18F/web-design-standards-docs/blob/master/README.md) stating referencing back to this issue.

yozlet:
The new [web-design-standards-docs repo](https://github.com/18F/web-design-standards-docs) is now auto-deploying to https://wds-docs.apps.cloud.gov/ (`master` branch) and https://wds-docs-staging.apps.cloud.gov/ (`staging` branch). Right now we're all just working straight into `master` but before launch we'll resume the PR -> staging -> master workflow.

[...]

PR #1419, opened 2016-08-24
Repo split

shawnbot:
This is a place to track some of the specific work we're doing on the repo split in #1343. We can flesh this out more, but here's what's done already and what remains, as far as I can tell:

- [x] Merge the `integrate-stylelint` branch so we can ditch `scss-lint` and Ruby altogether.
- [x] Remove references to `scss-lint` and replace with `stylelint`.
- [x] Remove all Jekyll and deployment stuff from `circle.yml`.
- [x] Remove a bunch of unused dependencies from `package.json`.
- [x] Resolve all issues in #1343.
- [x] Resolve merge conflicts
- [x] Review

/cc @juliaelman @yozlet @Journerdism


juliaelman:
Merge conflicts are now resolved. It looks like the `docs/` folder was not explicitly removed, which is what was causing the plethora of conflicts. Simple fix.

Also, removed a few more tasks and references to docs. 🎉

juliaelman:
@shawnbot no tears! Merge conflicts solved.

juliaelman:
Adding a note here that we will want to remove `prism.js` as part of the Standards themselves, as it is not something that supports the package.

See https://github.com/18F/web-design-standards/blob/je-repo_split/package.json#L39

Issue #1398, opened 2016-08-17
[Split Repo] Documentation Review


Journerdism:
## Description 
Update documentation

## Acceptance Criteria
- [x] Review documentation and merge to publish
- [x] Coincide with the [Comms strategy for announcing to current user base](https://github.com/18F/web-design-standards/issues/1393)

## Additional information [optional]
* Related to this epic: https://github.com/18F/web-design-standards/issues/1343#issue-166414812

kategarklavs:
Howdy! Is this documentation ready for review? (If so, I'm happy to start reviewing.)

juliaelman:
@jameshupp adding you to this issue too, as you have already started this work here: https://github.com/18F/web-design-standards/pull/1420 🎉

PR #1420, opened 2016-08-25
[Repo split] Updating docs and removing directions for running the website.

juliaelman:
## Description

This pull request updates the technical documentation in our contribution guidelines. The commands outlined now focus on the `npm` commands that can be used.

I have also added our CSS and JavaScript coding guidelines for folks to follow and reference for their contributions, similar to what we are doing [here](https://github.com/18F/web-design-standards-docs/blob/master/CONTRIBUTING.md#coding-guidelines).

This is also downstream from the [`je_repo_split` branch work](https://github.com/18F/web-design-standards/tree/je-repo_split).

Ref #1419, #1343 + #1398


juliaelman:
@jameshupp here is an initial take on the technical documentation we discussed earlier this week. Please let me know if you'd like to go over IRL too!

jameshupp:
Only one comment as-is: the style guide reference has a justification for the CSS portion of the style guide but none for the JS portion. Options:

* remove the CSS-specific references from the first sentence after `###CSS` and put it before that header, since the idea of consistent code across 18F seems valuable regardless of where it's happening.
* add a justification after `###JS` that explains why a JS style guide is important
* do neither, because what people pay attention to about CSS and JS are qualitatively different enough

If it's the first bullet, I'm happy to take a shot at it myself.

Nothing else stands out to me that's an issue for this file on its own. I'm inclined to host the license info in either just this file or just the README, but I'm still working through the README and am not totally sure which yet. Might be a few more things at that point, but I'll include those in my own PR.

juliaelman:
@jameshupp I made some updates to the copy and consolidated the coding conventions section. Let me know what you think and if that makes sense to you.

As for the licensing info, that is something I would recommend circling back with the comms team about. I was out on maternity leave when this got added, so there might be reasons why it is included in the README and CONTRIBUTING files, rather than just in a standalone LICENSE file. //cc @awfrancisco

maya:
@jameshupp Chatted with @brittag and she recommended:

>I’d suggest having the long version in LICENSE, a clear note in CONTRIBUTING that contributions are CC0 with a link to LICENSE, and a brief note in README

>it doesn’t need to be the long version in more than one place in my opinion

>the readme should basically point to the LICENSE and CONTRIBUTING files so that people don’t miss that they exist while reading the README

jameshupp:
@juliaelman Looks perfect then!

@brittag's recommendation sounds perfect to me on licensing, unless @awfrancisco can shed light on a need to do otherwise. I'm going to make that the plan for the README and will PR into this branch to account for it both there and in CONTRIBUTING.

awfrancisco:
I'm still waiting to hear back from @vzvenyach about his talks with OGC on the contributing docs. I'm not sure where that conversation stands.

jameshupp:
Okay I'll follow up on that point, but in the meantime, this particular PR shouldn't be held up for that. It doesn't really implicate the issue.

juliaelman:
The updates mentioned per Britta's note above were made 451d47527b33132a4127e1467b26797c839eecd1

Merging this in and thanks all for your help and feedback ^5

PR #1421, opened 2016-08-25
[Repo split]: Add missing images to src

maya:
## Description

Adds missing images that were in `docs` but should have been in `src`.

cc: @yozlet @juliaelman


juliaelman:
🚀

PR #1441, opened 2016-08-31
[Politespace + jQuery removal] Fixing Politespace in combination with jQuery removal.

juliaelman:
This pull request fixes a few issues with  [`Politespace`](https://github.com/filamentgroup/politespace/) that occur due to the removal of `jQuery` + updates the `Politespace` API:

- `jQuery` was not being added to the `docs` correctly and was failing silently 😭 
- `Politespace` has recently (within the past 9 days) updated their package. I have explicitly stated in the `package.json` the usage of v0.1.4, which resolves the handling of this error. 

Since we are [removing `Politespace` and moving towards `Fieldkit`](https://github.com/18F/web-design-standards/issues/1385#issuecomment-239886877), my recommendation is for the team to *not* spend the time upgrading.

### DIFF
--- a/docs/doc_assets/js/start.js
+++ b/docs/doc_assets/js/start.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var $ = require('jquery');
+window.$ = window.jQuery = require('../../../node_modules/jquery/dist/jquery');
 var calculateAnchorPosition = require('./components/calculate-anchor-position');
 var stickyNav = require('./components/sticky-nav');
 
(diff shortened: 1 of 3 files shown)


docs/doc_assets/js/start.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var $ = require('jquery');
+window.$ = window.jQuery = require('../../../node_modules/jquery/dist/jquery');
```
rogeruiz: Should we just reference `require('jquery');` here rather than the whole path? Otherwise it looks good. :100:

juliaelman: Yes! Much cleaner and fixed here caca489

yozlet:
Yay! 👍

PR #1445, opened 2016-08-31
Removing gov banner mention

kategarklavs:
Julia let me know that we don't include the "Official U.S. government site" banner in our headers -- updated the copy to reflect this.


juliaelman:
@kategarklavs perfect- thank you!

PR #1449, opened 2016-08-31
Merging header component lead copy into release branch

juliaelman:


kategarklavs:
This also looks good — merging! 🐭

PR #1462, opened 2016-09-06
Refactor banner & header JS to export init functions

yozlet:
This mainly wraps the banner & header (search & mobile) initialisation code in exported functions, but also does a little tidying up to ensure some DOM selections happen _after_ the DOM is ready.

@shawnbot: Can you review?

@maya: I'm not sure how to QA all of this. Could you give the three changed files (gov banner, search header, mobile header) a quick pass before you merge?

### DIFF
--- a/src/js/components/banner.js
+++ b/src/js/components/banner.js
@@ -3,15 +3,22 @@ var addClass = require('../utils/add-class');
 var removeClass = require('../utils/remove-class');
 var dispatch = require('../utils/dispatch');
 
-var headers = select('.usa-banner-header');
-headers.forEach(function (header) {
-  var click = function (e) {
-    (event.preventDefault) ? event.preventDefault() : event.returnValue = false;
-    var expanded = this.getAttribute('aria-expanded') === 'true';
-    var toggleClass = expanded ? addClass : removeClass;
-    toggleClass(header, 'usa-banner-header-expanded');
-  };
-  select('[aria-controls]').forEach(function (button) {
-    dispatch(button, 'click', click.bind(button));
+function bannerClickHandler (event) {
+  (event.preventDefault) ? event.preventDefault() : event.returnValue = false;
+  
+  var expanded = this.getAttribute('aria-expanded') === 'true';
+  var toggleClass = expanded ? addClass : removeClass;
+  toggleClass(header, 'usa-banner-header-expanded');
+}
+
+function bannerInit () {
+  var headers = select('.usa-banner-header');
+
+  headers.forEach(function (header) {
+    select('[aria-controls]').forEach(function (button) {
+      dispatch(button, 'click', headerClickHandler.bind(button));
+    });
   });
-});
+}
+
+module.exports = bannerInit;
--- a/src/js/components/banner.js
+++ b/src/js/components/banner.js
@@ -3,20 +3,21 @@ var addClass = require('../utils/add-class');
 var removeClass = require('../utils/remove-class');
 var dispatch = require('../utils/dispatch');
 
-function bannerClickHandler (event) {
+function headerClickHandler (event) {
   (event.preventDefault) ? event.preventDefault() : event.returnValue = false;
   
-  var expanded = this.getAttribute('aria-expanded') === 'true';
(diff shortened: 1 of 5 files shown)


src/js/components/banner.js
```diff
@@ -3,15 +3,22 @@ var addClass = require('../utils/add-class');
...
-headers.forEach(function (header) {
-  var click = function (e) {
-    (event.preventDefault) ? event.preventDefault() : event.returnValue = false;
-    var expanded = this.getAttribute('aria-expanded') === 'true';
-    var toggleClass = expanded ? addClass : removeClass;
-    toggleClass(header, 'usa-banner-header-expanded');
-  };
-  select('[aria-controls]').forEach(function (button) {
-    dispatch(button, 'click', click.bind(button));
+function bannerClickHandler (event) {
```
shawnbot: This should either be called `headerClickHandler` or the reference below should be updated.

maya:
@yozlet while I can't speak to the JS files themselves, I just checked the web UI and those components (gov banner, search header, mobile header) are all functioning correctly.

src/js/components/banner.js
```diff
@@ -3,15 +3,22 @@ var addClass = require('../utils/add-class');
...
+  (event.preventDefault) ? event.preventDefault() : event.returnValue = false;
+  
+  var expanded = this.getAttribute('aria-expanded') === 'true';
+  var toggleClass = expanded ? addClass : removeClass;
+  toggleClass(header, 'usa-banner-header-expanded');
```
shawnbot: The problem with moving this function out of the headers loop is that the `header` variable is out of scope now. One solution might be to bind the click handler to the header and reference the button through `event.target`, a la:

```js
var headerClickHandler = function (event) {
  // ...
  var expanded = event.target.getAttribute('aria-expanded') === 'true';
  // ...
  toggleClass(this, 'usa-banner-header-expanded');
};

headers.forEach(function (header) {
  var headerClick = headerClickHandler.bind(header);
  select('[aria-controls]').forEach(function (button) {
    dispatch(button, 'click', headerClick);
  });
});
```

src/js/initializers/header.js
```diff
@@ -1,12 +1,14 @@
...
 
 whenDOMReady(function initHeaders () {
 
   // Search Toggle
-  require('../components/header/search');
+  searchInit();
 
   // Mobile Navigation
-  require('../components/header/mobile');
+  mobileInit();
```
shawnbot: I think there's some trailing whitespace here?

src/js/components/banner.js
yozlet: Ouch, good catch, thanks!

yozlet:
Argh, yes, what @shawnbot said. @maya, can you double-check? Because at the very least, the banner click _shouldn't_ work (though I'll fix it in another commit)

maya:
Ah yes, the click works bc it's using the accordion component but the class that needed to be added does not :(

src/js/components/banner.js
yozlet: Great idea, thank you!

src/js/initializers/header.js
yozlet: Good catch!

yozlet:
OK, fixed. @maya, can you give it another go?

maya:
@yozlet it works!

yozlet:
@shawnbot, can you give it one more look?

src/js/components/header/search.js
```diff
@@ -3,10 +3,7 @@ var addClass = require('../../utils/add-class');
 var removeClass = require('../../utils/remove-class');
 var dispatch = require('../../utils/dispatch');
 
-var searchForm = select('.js-search-form')[ 0 ];
-var searchButton = select('.js-search-button')[ 0 ];
-var searchButtonContainer = select('.js-search-button-container')[ 0 ];
-var searchDispatcher;
+var searchForm, searchButton, searchButtonContainer, searchDispatcher;
 
 if (searchButton && searchForm) {
```
shawnbot: This whole `if` statement can go away, right?

yozlet: Doh! Yes, thank you.

Issue #1478, opened 2016-09-13
[Split Repo] [Documentation]: how to make updates given split repos


jpyuda:
## Description

Per our sprint-planning meeting of 9/13, we'll create a set of instructions for how to make updates/run the site locally given the repo split. Here's a general outline of the process:

- [x] Generate a list of issues folks might encounter/questions they might have, and write responses to said questions
- [x] Technical review of info
- [x] Add to relevant part of README
- [x] Also, make sure everyone on the team knows where the draft version of the doc lives

maya:
Documentation here (tested by team): https://github.com/18F/web-design-standards-docs#using-the-latest-version-of-the-uswds-package

kategarklavs:
@maya Ah, you read my mind! (I literally just tagged Julia in a comment in a Google doc asking whether we should spell out the process/whether we should provide only high-level guidance). 

I'll update the README and then ping you for review — sound good?

kategarklavs:
Update: The web-design-standards-docs README has a new section titled "Using the latest version of the uswds package," which has already been reviewed/approved by the team. I talked to @maya, and she confirmed that no updates to the web-design-standards README are necessary. All in all, I think we're in good shape. That said, if you all feel like we need additional documentation, let me know and we can reopen this.

PR #1485, opened 2016-09-15
Using the latest version info!

kategarklavs:
Per #1478 , here's info about using the latest version of the Standards.

@maya — would you be willing to review & let me know whether this looks correct (or whether it needs more work)? Many thanks in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Previously, the website and documentation for the Draft U.S. Web Design Standard
   * [Download](#download)
   * [Install using npm](#install-using-npm)
   * [Use another framework or package manager](#use-another-framework-or-package-manager)
+  * [Use the latest version of the `uswds` package](#use-the-latest-version-of-the-uswds-package)
 * [Our use of branches](#our-use-of-branches)
 * [Need installation help?](#need-installation-help)
 * [Contributing to the code base](#contributing-to-the-codebase)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -143,6 +144,22 @@ If you’re using another framework or package manager that doesn’t support NP
 
 If you’re interested in maintaining a package that helps us distribute the Draft U.S. Web Design Standards, the project’s build system can help you create distribution bundles to use in your project. Please read our [contributing guidelines](CONTRIBUTING.md#building-the-project-locally-with--gulp-) to locally build distributions for your framework or package manager.
 
+### Use the latest version of the `uswds` package
+
+If you'd like to run the latest version of the Standards, follow these steps:
+
+1. Clone the latest version of the [web-design-standards repo](https://github.com/18F/web-design-standards/tree/staging).
```
maya: Since this is the `web-design-standards` repo, you shouldn't need a link to it (you're here!). We should probably link to the `web-design-standards-docs` repo instead.

README.md
```diff
@@ -17,6 +17,7 @@ Previously, the website and documentation for the Draft U.S. Web Design Standard
   * [Download](#download)
   * [Install using npm](#install-using-npm)
   * [Use another framework or package manager](#use-another-framework-or-package-manager)
+  * [Use the latest version of the `uswds` package](#use-the-latest-version-of-the-uswds-package)
```
maya: So I'm not sure if this is needed, or if it is maybe it should be stated differently bc it deals with utilizing the docs repo. Linking to that repo by stating something along the lines of "run the docs site" and linking to that repo may be enough.

kategarklavs: Ahh, I see — I may have been confused about the nature of the request (I thought both READMEs needed to be updated). But, being that the docs README is updated, I'll go ahead and cancel these changes ...

PR #1498, opened 2016-09-22
Fixing mobile.js from having duplicative events from firing

### LINKED ISSUE #1496: JavaScript error for basic header
## Problem with JavaScript for basic header

Website: https://standards.usa.gov/headers/basic/#

## Description

When you view the header in a mobil screen size, clicking the "menu" button opens the mobile menu properly. But clicking the "X" button to close does not work. The menu closes and then re-opens immediately.

1. Visit https://standards.usa.gov/headers/basic/#
2. Adjust screen size to be mobile
3. Click "menu"
4. Click "x" to close out of menu

## Additional information

Gif of what I am seeing:

sep-21-2016 12-54-07

juliaelman:
This pull request fixes the issue around the mobile sidenav touch events not *actually* closing the sidenav. The error occurred due to the fact that both the `click` and `touchstart` would fire at the same time, this never actually closing the the mobile "touch" version of the sidenav.

Fixes #1496

### DIFF
--- a/src/js/components/header/mobile.js
+++ b/src/js/components/header/mobile.js
@@ -3,6 +3,8 @@ var addClass = require('../../utils/add-class');
 var removeClass = require('../../utils/remove-class');
 var dispatch = require('../../utils/dispatch');
 
+var clickHandler = ('ontouchstart' in document.documentElement ? 'touchstart' : 'click');
+
 function toggleClass (element, className) {
   if (element.classList) {
     element.classList.toggle(className);
@@ -25,7 +27,6 @@ function handleNavElements (e) {
 
 function mobileInit () {
   var navElements = select('.usa-menu-btn, .usa-overlay, .usa-nav-close');
-  var clickHandler = ('ontouchstart' in document.documentElement ? 'touchstart' : 'click');
 
   navElements.forEach(function (element) {
     dispatch(element, clickHandler, handleNavElements);
--- a/src/js/components/header/mobile.js
+++ b/src/js/components/header/mobile.js
@@ -9,20 +9,26 @@ function toggleClass (element, className) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/js/components/header/mobile.js
```diff
@@ -9,20 +9,26 @@ function toggleClass (element, className) {
   }
 }
 
-function mobileInit () {
-  var navElements = select('.usa-menu-btn, .usa-overlay, .usa-nav-close');
+function handleNavElements (e) {
+
   var toggleElements = select('.usa-overlay, .usa-nav');
   var navCloseElement = select('.usa-nav-close')[ 0 ];
```
shawnbot: Should these selections be scoped to `this` or `e.target`? IOW, is there any chance that there might be multiple nav elements on the page?

rogeruiz: That's a good point @shawnbot. But, I can't be sure what the `e.target` is going to be here right? I don't think there will be multiple navigations on the page. I think other parts of the codebase also rely on the idea of including elements globally rather than contextually.

shawnbot: Yep, @juliaelman and I agreed that `this` is what we want, not the event target.

juliaelman: @shawnbot took a bit of time to take a closer look at this approach from holistic stand point. I agree with @rogeruiz on the multiple navs and also that we probably don't want to scope the code base to encourage that type of behavior. We also don't focus the scope in this way throughout the current JS code base and think it makes sense for us to be consistent. 

However, I did make the `clickHandler` available globally in 9754434. 😄

shawnbot: Sounds good to me! We can always revisit it in the future if need be.

PR #1500, opened 2016-09-22
Set default value for `$asset-pipeline`

jessieay:
* If you look at http://bourbon.io/docs/#font-face, the `asset-pipeline`
  argument sent to the bourbon `font-face` mixin is what is used
* This update enables Rails apps to override `$asset-pipeline` and use
  font mixin provided by the design standards
* See https://github.com/18F/uswds-rails-gem/issues/14

### DIFF
--- a/src/stylesheets/core/_fonts.scss
+++ b/src/stylesheets/core/_fonts.scss
@@ -3,7 +3,7 @@
   '#{$font-path}/sourcesanspro-light-webfont',
   300,
   normal,
-  $asset-pipeline: false,
+  $asset-pipeline: $asset-pipeline,
   $file-formats: eot woff2 woff ttf
 );
 
@@ -12,7 +12,7 @@
   '#{$font-path}/sourcesanspro-regular-webfont',
   400,
   normal,
-  $asset-pipeline: false,
+  $asset-pipeline: $asset-pipeline,
   $file-formats: eot woff2 woff ttf
 );
 
@@ -21,7 +21,7 @@
   '#{$font-path}/sourcesanspro-italic-webfont',
   400,
   italic,
-  $asset-pipeline: false,
+  $asset-pipeline: $asset-pipeline,
   $file-formats: eot woff2 woff ttf
 );
 
@@ -30,7 +30,7 @@
   '#{$font-path}/sourcesanspro-bold-webfont',
   700,
   normal,
-  $asset-pipeline: false,
+  $asset-pipeline: $asset-pipeline,
   $file-formats: eot woff2 woff ttf
 );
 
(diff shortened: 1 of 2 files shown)


jessieay:
Another thought on this: maybe we shouldn't even pass the arg, and just declare the scss variable instead?

See https://github.com/thoughtbot/bourbon/blob/v4-stable/app/assets/stylesheets/css3/_font-face.scss 

The font-face mixin itself uses `$asset-pipeline: $asset-pipeline`

looks like the bourbon default for `$asset-pipeline` is  false:

https://github.com/thoughtbot/bourbon/blob/a50249ad24521bda3ac60f0c5cd3febeb7ff2fe0/core/bourbon/settings/_settings.scss#L36

but you can set it to true and I'd assume that would cause an override anywhere that we are calling the `font-face` mixin as long as we don't pass the `$asset-pipeline` arg?

src/stylesheets/core/_fonts.scss
```diff
@@ -3,7 +3,7 @@
   '#{$font-path}/sourcesanspro-light-webfont',
   300,
   normal,
-  $asset-pipeline: false,
+  $asset-pipeline: $asset-pipeline,
```
tysongach: No need for this, as the mixin defines this argument to this value by default.

tysongach:
Hi @jessieay!

> Another thought on this: maybe we shouldn't even pass the arg, and just declare the scss variable instead?

Yes, this is correct. I would recommend setting the global variable (`$asset-pipeline`) once instead of per mixin include.

> looks like the bourbon default for $asset-pipeline is false

Yes, it’s `false` by default. One note: The settings you linked to is for our new v5.0 settings, which are defined differently than in v4.x, which I believe we’re using here (v5 is currently in beta).

In Bourbon 5.0 we are moving away from Sass variables for global library settings and instead wrapping them in a Sass map.

So, v4.x setting definitions look like this: 

```scss
$asset-pipeline: true;
$em-base: 14px;
// etc.
```

And in v5.x, they’ll be defined like this:

```scss
$bourbon: (
  "contrast-switch-dark-color": #000,
  "contrast-switch-light-color": #fff,
  "global-font-file-formats": ("ttf", "woff2", "woff"),
  "modular-scale-base": 1em,
  "modular-scale-ratio": $major-third,
  "rails-asset-pipeline": false,
);
```

(note that some of the setting names have been updated, as well, to provide more clarity)

> but you can set it to true and I'd assume that would cause an override anywhere that we are calling the font-face mixin as long as we don't pass the $asset-pipeline arg?

Yep. Any mixin that deals with assets will look at the global variable, but you can also override that when you call the mixin.

Hope this helps!

jessieay:
@tysongach Thank you so much for weighing in! I just updated this PR to remove the `asset-pipeline`

juliaelman:
@jessieay @tysongach thank you for your patience! This has been reviewed and is slated for release `0.13.0`.

PR #1514, opened 2016-10-04
Tidy gulp

shawnbot:
- Don't infer the task name from `__filename` in fonts, images, and javascript gulp configs.
- Reference `gutil.colors` as [chalk](https://www.npmjs.com/package/chalk) in doc-utils so it's clearer which terminal color API we're using.

### DIFF
--- a/config/gulp/doc-util.js
+++ b/config/gulp/doc-util.js
@@ -1,5 +1,6 @@
 var pkg = require('../../package.json');
 var gutil = require('gulp-util');
+var chalk = gutil.colors;
 var notifier = require('node-notifier');
 
 var shellPrefix = '$';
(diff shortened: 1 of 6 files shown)


juliaelman:
Looks good! Is there a reason why we want to keep `var task` in each gulp config file? Looks like we aren't using that variable anywhere other than in the top level task assignment (e.g. `sass`). Might make sense to remove that variable and just name the task accordingly.

juliaelman:
@shawnbot yup! Let's do it. ^5

juliaelman:
Going to go ahead and give the 👍 on this and merge in. We can make that update in another PR, but it makes sense to have the var in there.

PR #1519, opened 2016-10-06
Merge master -> staging

shawnbot:
Staging is behind right now because we skipped the staging step for v0.13.1. This gets it back up to speed.


juliaelman:
Looks like Circle CI didn't update here once the build passed - weird. Rebooted build and all is well!

🚀

PR #1520, opened 2016-10-06
Tidy up and lint unit tests

### LINKED ISSUE #1510: Add unit tests for the header component
We have JS unit tests for every component except the new header. Let's add some!

shawnbot:
This PR tidies up the unit test suite and adds all of the test JavaScript files to our eslint glob list so that our tests are held to the same standards as the rest of the code.

Changes proposed:
- [x] Refactor the DOM tests to use [jsdom-global](https://github.com/rstacruz/jsdom-global) via mocha options, eliminating the `setup.js` require that just exports jQuery.
- [x] Move the eslint configuration to YAML files, so that it's easier to update, and add a test-specific "subconfiguration" in `test/.eslintrc.yml`.


juliaelman:
@shawnbot I think turning this pull request into an initial refactor is ok. I checked out some of the work you've done to get started (e.g. whitespace cleanup) here.  What else would you like to do?

We probably want some basic tests in for headers as soon as possible, but I don't see the harm in some initial refactor tasks.

spec/unit/toggle-fields/template.js
```diff
@@ -1,11 +1,11 @@
 var template = [
   '<form>',
-  '<input id="password" name="password" type="password"/>',
+  '<input id="password" name="password" type="password">',
   '<input id="confirmPassword" name="confirmPassword" type="password">',
   '<p class="usa-form-note">',
   '<a title="Show my typing" href="#" class="usa-show_multipassword" aria-controls="password confirmPassword">Show my typing</a>',
   '</p>',
-  '</form>'
-].join(',');
```
shawnbot: Joining with `,` was producing some weird (but perfectly valid) HTML.

spec/unit/validator/template.js
```diff
@@ -1,13 +1,17 @@
+/* eslint-disable indent */
 var template = [
-'<div>',
-  '<ul class="usa-checklist" id="validation_list">',
-    '<li data-validator="length">Be at least 8 characters</li>',
-    '<li data-validator="uppercase">Have at least 1 uppercase character</li>',
-    '<li data-validator="numerical">Have at least 1 numerical character</li>',
-  '</ul>',
-  // JSDOM doesnt properly support multiline attributes :(
```
shawnbot: @rogeruiz this actually ended up being a problem with the `.join(',')` below, which was producing invalid HTML.

PR #1522, opened 2016-10-07
Reorganize header components

### LINKED ISSUE #1510: Add unit tests for the header component

shawnbot:
This is a reorganization of the headers "component" JS into two discrete components:

1. The **navigation** component manages the expanding and collapsing mobile nav and overlay
1. The **search** component manages the showing and hiding of the search input in the header

In this PR:
- [x] Rename `header/mobile.js` to `navigation.js`
- [x] Rename `header/search.js` to `search.js`
- [x] Reorganize initializers to require both of the above and order components alphabetically, for easier scanning. (This assumes there are no interdependencies between components, which appears to be the case.)
- [x] Add detection of touch events to the search components so that they work on mobile browsers (**untested**)
- [x] Write tests for the search component
  - [x] Introduce a polyfill for [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden), a la [this one](https://github.com/valtlai/HTMLElement.prototype.hidden/blob/master/elem-hidden.js)
  - [x] ~~Introduce a `stylize()` utility to support the hidden polyfill, with its own tests~~
- [x] Write (cursory) tests for the navigation component, fixes #1510
- [x] This PR includes the work in #1520 _and_ #1521, so we'll need to rebase this branch after those get merged

When all this is done, we'll need to update [the docs](https://github.com/18F/web-design-standards-docs/tree/staging/_components/headers) to:

1. Include separate pages for the "new" search and navigation components (and migrate any related content from the headers page to these pages)
1. Rewrite the [headers documentation](https://standards.usa.gov/headers/) to reflect the new component structure
1. Update the standalone header example pages?


spec/unit/navigation/template.html
```diff
@@ -0,0 +1,84 @@
+
+<div class="usa-navbar">
+  <button class="usa-menu-btn">Menu</button>
+</div>
+
+<div class="usa-overlay"></div>
+
+<nav role="navigation" class="usa-nav">
+  <div class="usa-nav-inner">
+    <button class="usa-nav-close">Close</button>
```
shawnbot: Not in the test suite, no. But if you want the test template to be more representative of our recommended markup, then yes!

spec/unit/navigation/navigation.spec.js
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it('hides the nav when the overlay is clicked', function () {
+    menuButton.click();
+    overlay.click();
+    assert.equal(isVisible(nav), false);
+    assert.equal(isVisible(overlay), false);
+  });
+
+  describe('off()', function () {
```
shawnbot: Not necessarily, but I fleshed out the tests anyway. Thanks for pointing it out!

nickbristow:
@shawnbot i think i understand. Its :100 percent fine to hide from everyone, even if JS shows it later. 

Did i get that right?

juliaelman:
@shawnbot just to clarify, are you wanting to have both `hidden` and `aria-hidden` attributes on certain elements? I am not sure we should/or need to do both.

juliaelman:
@shawnbot ah! Thank you for clarifying the path you are wanting to take in regards to hiding/showing in general. Just want to make sure we are over thinking here and that we are not using two "hiding" type of values on elements. 

Mergin' 🚀

Issue #1529, opened 2016-10-19
Identify version in CSS comments


shawnbot:
Currently there are no comments in the CSS identifying which version of the system the file was installed from. This would be a really helpful thing both for humans and potential future robots to determine an upgrade path.

shawnbot:
This was done in #1530.

PR #1530, opened 2016-10-19
Rename all.scss to uswds.scss; add version comment

### LINKED ISSUE #1417: [Sass] Easier use with node-sass include paths
Currently, both of options for including the Sass sources are kind of ugly. You have to do one of the following:

1. Reference the `all.scss` via its full path, or path relative to the including Sass file, as in [cg-style](https://github.com/18F/cg-style/blob/d9406d994af686a8acac80849e2e08087a939771/src/css/base.scss#L4):

  ```scss
  @import '../../node_modules/uswds/src/stylesheets/all.scss';
  ```

1. Or you can add a path to node-sass's [`includePaths` option](https://github.com/sass/node-sass/#includepaths), a la:

  ```js
  sass.render(..., {
    includePaths: 'node_modules/uswds/src/stylesheets'
  });
  ```

  Which is nicer, but it means that you have to ambiguously refer to the `all` file like this:

  ```scss
  @import 'all'; // via uswds
  ```

  Which will not work as expected if you have your own `all.scss` locally.

## Description

I really like the way that [Bourbon](https://github.com/thoughtbot/bourbon#installing-with-npm-and-using-a-node-based-asset-pipeline) and [Neat](https://github.com/thoughtbot/neat#installing-with-npm-and-using-a-node-based-asset-pipeline) do it, which is to export an `includePaths` array from the top-level JS module, and then you can load them in Sass via their formal names:

```js
// sass config
{
  includePaths: [/* your local include paths here */]
    .concat(require('bourbon').includePaths))
    .concat(require('bourbon-neat').includePaths))
}
```

```scss
@import 'bourbon';
@import 'neat';
```

It would be great if WDS followed this pattern, so users could do the same:

```js
// sass config
{
  includePaths: [/* your local include paths here */]
    .concat(require('uswds').includePaths))
    .concat(require('bourbon').includePaths))
    .concat(require('bourbon-neat').includePaths))
}
```

```scss
@import 'uswds';
@import 'bourbon';
@import 'neat';
```

Alternatively, as Bourbon and Neat's docs suggest, we could consider offering some examples for using WDS with [Eyeglass](http://eyeglass.rocks/).

shawnbot:
### Sass entrypoint: `uswds`
This PR renames `src/stylesheets/all.scss` to `uswds.scss`, which fixes #1417 by making it possible to import in Sass by adding the `src/stylesheets` directory to their include path and importing it with:

```scss
@import 'uswds';
```

For users of the existing `all.scss`, I've included a fallback `_all.scss` that prevents breakage for users of the old Sass entrypoint via `@import 'all';`.

### CSS version comment
I've also included a fix for #1529 in the form of a CSS comment at the top of both the un-minified and minified CSS files that looks like this:

```css
/*! uswds v0.13.1 */
```

where `0.13.1` will be whatever our `version` field specifies in `package.json`. I've placed this comment at the top of our `uswds.scss` file, so any users who import this file will get the comment in their CSS, which can help us debug future issues in user-generated CSS.

> ℹ️ The `!` is a convention that suggests minifiers preserve the comment, and ensures that it isn't stripped in `uswds.min.css`.

### Tests!
There are tests for all of this:

1. The "sass/build" tests confirm that running `gulp build` generates the expected `uswds.css` and `uswds.min.css` files, and that both of them contain the expected version string.

  > ⚠️ This slows down the tests because we're shelling out to `gulp build` (which is "wasting" time also building the JavaScript), but I think that it's worth it. We may need to play with the 15-second timeout a bit if the build ends up taking longer on CI than on my computer.

1. The "sass/include" tests confirm that rendering raw Sass that imports either `uswds` or `all` from our include path doesn't generate an error.


juliaelman:
@shawnbot a few things:

1) We need to update our documentation to state that we are making this change: https://github.com/18F/web-design-standards/blob/staging/README.md
2) Let's open a PR in https://github.com/18F/web-design-standards-docs that correlates with this change so that we don't run into any breakage on the site (e.g. `gulp` tasks)

spec/sass/build.spec.js
```diff
@@ -0,0 +1,69 @@
...
+};
+
+before(function (done) {
+  this.timeout(15000);
+  build(done);
+});
+
+describe('build output', function () {
+
+  it('generates CSS at dist/css/uwsds.css', function () {
```
juliaelman: s `uwsds.css/uswds.css`

spec/sass/build.spec.js
```diff
@@ -0,0 +1,69 @@
...
+    var distFilename = path.join(distPath, 'uswds.css');
+    assert.ok(
+      fs.existsSync(distFilename),
+      'the file does not exist: ' + distFilename
+    );
+  });
+
+  it('generates minified CSS at dist/css/uwsds.min.css', function () {
```
juliaelman: s `uwsds.css/uswds.css`

shawnbot: DERP

juliaelman:
🚀

Issue #1535, opened 2016-10-24
Eliminate SCSS variable duplication


shawnbot:
Currently, our "core" SCSS variables are duplicated in [_defaults.scss](https://github.com/18F/web-design-standards/blob/master/src/stylesheets/core/_defaults.scss) and [_variables.scss](https://github.com/18F/web-design-standards/blob/master/src/stylesheets/core/_variables.scss), and our grid settings are duplicated _again_ in [_grid-settings.scss](https://github.com/18F/web-design-standards/blob/master/src/stylesheets/core/_grid-settings.scss), as noted in #1532.

This is a problem: It's confusing for us to maintain, and it's confusing for users of the Sass source who want to customize the system. I've noted most of the documentation-related improvements on customization guidance in #1534, but this issue is about the code itself, and how we can make our SCSS variables easier for us to maintain and for our users to selectively override.

In [the docs](https://standards.usa.gov/getting-started/developers/#css-architecture), we specifically suggest that users copy `_variables.scss` and import the modified copy before the rest of the uswds source. The problem with this advice is that copying all of the variables locks that user into all of the default values at the time they copied that file. **The solution is to only set variables that you intend on overriding.** Mechanically, here's how it should work:
1. Remove [_defaults.scss](https://github.com/18F/web-design-standards/blob/master/src/stylesheets/core/_defaults.scss) (or deprecate them so as not to introduce issues for users, as I've done in #1530)
2. Add `!default` to all of our variable declarations in [_variables.scss](https://github.com/18F/web-design-standards/blob/master/src/stylesheets/core/_variables.scss) so that they can all be overridden
3. Update the [reference](https://github.com/18F/web-design-standards/blob/8f14a0c93ba6cb199b07d39e64a2dba4e0431a35/src/stylesheets/all.scss#L10) in our SCSS entrypoint from `core/defaults` to `core/variables`

Then, instead of maintaining a copy of `_variables.scss`, users would remove all of the variables they _haven't_ changed, which prevents them from being "locked in" to those values even if we change them in a later release.

shawnbot:
Cc'ing @joshbruce on this. Josh: Am I on the right track here? Does this capture #1532, and can we consolidate the issues if so?

shawnbot:
I'd like to merge #1530 before getting started on this, and work in parallel on release notes that document how users should update their Sass to accommodate the next release. My hope would be that we can get users to refactor their imports down to something like:

```scss
// customizations
@import 'variables';
[...]

Issue #1532, opened 2016-10-22
[Sass]: Grid-settings breakpoints are duplicative


joshbruce:
## Description

`/stylesheets/core/_grid-settings.scss` defines mobile-first breakpoints
`/stylesheets/core/_variables.scss` defines the same mobile-first breakpoints
## Steps to reproduce the issue
1. Open `/stylesheets/core/_grid-settings.scss` and `/stylesheets/core/_variables.scss`
2. See commented block for "Mobile first breakpoints"
## Additional information [optional]

What was the rationale behind the duplication?

shawnbot:
Thanks, @joshbruce! I'm going to take a look at this as part of #1534.

shawnbot:
I'm fairly certain that this was resolved in #1536, where we deprecated `core/grid-settings` altogether. 🎉

PR #1536, opened 2016-10-24
Refactor all variables as defaults

shawnbot:
**⚠️ As configured, this will merge into #1530. If we intend to follow through with this, we should merge that PR first, then set the base branch for this one to `staging` and rebase. ⚠️**

This is a proposed fix for #1535 and #1532. It deprecates `core/defaults` and `core/grid-settings` in favor of `core/variables` with `!default` variable declarations that allow any variable to be overridden if it's set before being imported.

I'm documenting some of my findings from searching through github for usage examples in #1535, but in the meantime I would love a gut check.


src/stylesheets/core/_variables.scss
```diff
@@ -1,96 +1,121 @@
...
-$lead-line-height:    1.7;
-
-$font-sans:           'Source Sans Pro', $helvetica;
-$font-serif:          'Merriweather', $georgia;
-
-$font-normal:         400;
-$font-bold:           700;
+// These variables drive almost every aspect of the U.S. Web Design
+// system. Every variable is declared as !default so that it can be
+// overridden by a previous import.
```
juliaelman: I've got an example of this in the new `README.md` changes [here](https://github.com/18F/web-design-standards/blob/28a608222e79780184c3bb7e0d39f75848a195df/README.md#sass). Once this gets merged in, I'll update here to point to that example. ^5

PR #1539, opened 2016-10-26
Update README to better accommodate Windows implementation

juliaelman:
This pull request includes some updates in our `README.md` to better accommodate Windows users and also users of varying degrees of experience. A few things to note:

- Adding more definition around varying acronyms to be more explicit in what they mean or refer to.
- Added better examples around how folder and file structures work together
- Be more explicit on how to import and/or compile assets when installing with `npm`

Also, this PR is merging into https://github.com/18F/web-design-standards/pull/1530 (!)

Refs #1533

PR #1540, opened 2016-10-27
Fix installation step formatting

shawnbot:
This reformats a few of the procedural steps in the README to move the block-level elements between steps "into" the preceding step, which fixes the issue of numeric ordering being reset.

Props to @acolter for noticing this in #1539; sorry, I had an itchy trigger finger and merged before fixing this!


juliaelman:
@shawnbot I think that that build error was due to caching on Circle CI. Re-ran the build without caching and all is well now. 😺 

Mergin'.

Issue #1524, opened 2016-10-11
[Buttons]: Outline button hover issue on type="button"


maya:
## Description

When using a button `type="button"` with the `usa-button-outline` style, the hover state has a filled background.
### This is what it looks like:

<img width="100" alt="screen shot 2016-10-11 at 11 55 43 am" src="https://cloud.githubusercontent.com/assets/5249443/19284340/65c6c032-8faa-11e6-8050-38074d04f001.png">

shawnbot:
@maya I'm having a hard time reproducing this. Can you post a test case?

maya:
Just checked Tock, where this was initially happening and it doesn't seem to be an issue anymore!

Issue #1523, opened 2016-10-11
[Form buttons]: Submit button display issues


maya:
## Description

When a submit button is in a form it gets additional styles added to it. I'd like it to display a secondary button side-by-side to it but instead it drops below and has a wider width than the other button. So I have to add custom styles to remove these styles.

<img width="233" alt="screen shot 2016-10-11 at 11 27 43 am" src="https://cloud.githubusercontent.com/assets/5249443/19283271/1230bb16-8fa6-11e6-85e3-5e733d47a1f8.png">

shawnbot:
Update: the fix for this is ready in #1544.

shawnbot:
Fixed in #1544! 🎉

PR #1544, opened 2016-10-31
Button style fixes

shawnbot:
This PR addresses ~some issues~ an issue related to button styling:

- [x] The CSS previously targeted `[type=submit]:disabled`, rather than `.usa-button:disabled` #1523
- [x] ~The `usa-button-outline` styles produce illegible button text in #1524~


src/stylesheets/elements/_buttons.scss
```diff
@@ -152,7 +152,7 @@ button,
 }
 // scss-lint:enable PropertyCount
 
-[type="submit"]:disabled,
+.usa-button:disabled,
```
msecret: Could there be any `[type="submit"]` code around that relies on this?

shawnbot: I think that's possible, yeah. Aside: should we track (potentially) breaking changes in a changelog that gets committed, so that we can add them as we go?

msecret: I'm also fine with marking it as a breaking change, either is fine.

Issue #1434, opened 2016-08-30
[Refactor CSS] Split header component into separate files


juliaelman:
The JS and testing work was done in #1522. We still need to reorganize the CSS, then transfer this work over to Marigold (18F/marigold#22).

juliaelman:
Completed in #1595

PR #1595, opened 2016-11-30
Refactor header CSS

shawnbot:
This is a fix for #1434 that refactors the CSS to match the JS component refactor in #1522. Changes:

1. Move _most_ of the navigation-related CSS from the `header` partial into a new `navigation` partial.
1. Alphabetize the component partial import order in `uswds.css` so that it's easier to read, and _suggests_ that our components shouldn't depend on one another. This lead me to...
1. Move the sidenav list mixins into `core/utilities` so that they're available in `navigation` regardless of the component import order.

It looks like it works for me locally, but I'd love another pair of eyes on it. The bigger issue, though, is:

The `header` partial is much lighter now (hooray!), but it does quite a bit of styling its child navigation and search "components".  Honestly, I'm not sure what the best way is to handle these: should the header-qualified selector rules (`.usa-header .usa-search`, as opposed to `.usa-search`) stay in the `header` partial, or should "variants" of each contained component live with the rest of the component's own styles (e.g. `.usa-header .usa-search` moves to the `search` partial)?


juliaelman:
@shawnbot this LGTM! 

To answer your question, this is something we should think about taking to our users. AFAIK, most folks are downloading the compiled files at this point. My thought is to first inquire who is using (or wanting to use the Sass) and determine the best path forward from that user set.

//cc @bradnunnally would love to know your thoughts too!

PR #1606, opened 2016-12-06
Remove politespace, tidy up

### LINKED ISSUE #1385: [Package]: Politespace caused build to fail on customized standards
## Description

We are creating an extension of the USWDS; whereby, we leverage the good work here and the modular approach to extend or override. 
1. It seems Node.js 3.x uses a flat folder structure for dependencies:  https://docs.npmjs.com/how-npm-works/npm3 - therefore, we had some problems with includes and requires.
2. We tried copying just the `uswds.min.js` and just requiring it into our build. However, ran into issues with JS compiling due to Politespace not being found. Therefore, we copied over `/node_modules/uswds/src/js` to our `/src/js/vendor` (following the copy vendor files methodology. Same issue.
3. We modified the copy of the `uswds.min.js` to go one folder higher to find the Politespace stored in our `/node_modules`. The had it be ignored when we do `copy-vendor-javascript`. This does not seem like a permanent solution.

Thoughts?
1. Copy the `/node_modules/uswds/src/js` directory to `/src/js/vendor`
2. Create `/src/js/start.js` with `require('./vendor/start.js');`
3. Run `$ gulp release`
4. **Should fail**
5. Modify the `/src/js/vendor/start.js`:

From:

`var Politespace = require('../../node_modules/politespace/src/politespace').Politespace;`

To:

`var Politespace = require('../../../node_modules/politespace/src/politespace').Politespace;`
1. Run `$ gulp release`
2. Should build.
## Additional information

gsa/sam-web-design-standards#45

gsa/sam-web-design-standards#13

screen shot 2016-08-11 at 10 15 24 am

`$ gulp copy-vendor-javascript`

```
gulp.task('copy-vendor-javascript', function (done) {

  dutil.logMessage('copy-vendor-javascript', 'Compiling vendor JavaScript');

  var stream = gulp.src([ // TODO: Should we copy the USWDS Sass, as it is a vendor?
      // './node_modules/uswds/dist/js/uswds.min.js'
      './node_modules/uswds/src/js/**/*.js',
      '!./node_modules/uswds/src/js/start.js'
    ])
    .on('error', function (error) {
      dutil.logError('copy-vendor-javascript', error);
    })
    .pipe(gulp.dest('src/js/vendor/'))
    .pipe(gulp.dest('_site-assets/js/vendor/'))

  return stream;
});
```

shawnbot:
Fixes #1385 by removing the [Politespace] dependency. There's some general tidying in here, too.

- [x] Remove `politespace` from npm `dependencies`
- [x] Update 3rd-party licensing information in `LICENSE.md`
- [x] Remove the `verify-jquery` JS helper because we don't need it
- [x] Add JS sourcemap comments to `uswds(.min).js` to ease debugging
- [x] Fix (unreported, unrelated) issue of improper seach initialization in f307643
- [x] Remove `data-politespace` and other Politespace-specific attributes from the [address form template](https://github.com/18F/web-design-standards/blob/remove-politespace/src/html/address-form.html#L77)
- [x] Remove the unused `prismjs` dependency
- [x] Move `jquery` to dev dependencies

### ⚠️  Breaking change alert!
**This is a breaking change for users of the [address form template](https://standards.usa.gov/form-templates/#address-form)!** When upgrading to the next version, users of the address form template (as-is) will need to [download Politespace](https://github.com/filamentgroup/politespace#download-politespacejs-and-politespacecss) separately and include the [jQuery snippet](https://github.com/18F/web-design-standards/blob/6d36bc09cb3625f3d119e179c3c60e1f95388c76/src/js/initializers/politespace.js) that we've removed—and which, to be 100% transparent, **does not work now** for users who don't also include jQuery alongside USWDS.

I think we should ask the mailing list if anyone is actually using this form template as-is so that we don't waste time writing too much deprecation documentation for something that nobody uses. What do you think, @juliaelman @msecret?

[Politespace]: https://github.com/filamentgroup/politespace


src/js/components/search.js
```diff
@@ -57,9 +57,8 @@ function searchInit () {
   searchButton = select('.js-search-button')[ 0 ];
   searchButtonContainer = select('.js-search-button-container')[ 0 ];
 
-  closeSearch();
-
   if (searchButton && searchForm) {
+    closeSearch();
```
msecret: Whats the reason for this change?

msecret: Oh I see

package.json
```diff
@@ -38,7 +38,6 @@
     "cross-spawn": "^2.1.5",
     "jquery": "^2.2.0",
     "normalize.css": "^3.0.3",
-    "politespace": "0.1.4",
     "prismjs": "^1.5.1"
```
shawnbot: Nope, not in this repo. Just [in the docs](https://github.com/18F/web-design-standards-docs/blob/add-politespace/package.json#L83)!

config/gulp/javascript.js
```diff
@@ -58,7 +58,7 @@ gulp.task(task, [ 'eslint' ], function (done) {
         basename: dutil.pkg.name,
         suffix: '.min',
       }))
-    .pipe(sourcemaps.write('.', { addComment: false }))
+    .pipe(sourcemaps.write('.', { addComment: true }))
```
shawnbot: Good call; will do.

package.json
```diff
@@ -38,7 +38,6 @@
     "cross-spawn": "^2.1.5",
     "jquery": "^2.2.0",
```
shawnbot: Yes, it's still used in the tests, e.g. [here](https://github.com/18F/web-design-standards/blob/remove-politespace/spec/unit/accordion/accordion.spec.js#L3).

package.json
juliaelman: @shawnbot sorry, I meant `prismjs` 😁

package.json
juliaelman: Cool, that is what I thought. Probably makes more sense to move to `devDependencies`.

juliaelman:
@shawnbot just added specific html for the address form on the docs site 🎉  should resolve the issue you outlined above and help this get merged in.

juliaelman:
Just making a note that @shawnbot and I talked about a few more changes and then I think this is good to go.

package.json
shawnbot: done in 6747ef5 👌

juliaelman:
LGTM :shipit:

PR #1610, opened 2016-12-08
Stylelint fixes

shawnbot:
Fixes our `gulp stylelint` task so that we can add it back in as a dependency to `gulp sass` and ensure that we're not breaking our own Sass style rules.

### DIFF
--- a/src/stylesheets/core/_fonts.scss
+++ b/src/stylesheets/core/_fonts.scss
@@ -1,3 +1,4 @@
+/* stylelint-disable at-rule-empty-line-before */
 @include font-face(
   'Source Sans Pro',
   '#{$font-path}/sourcesanspro-light-webfont',
@@ -61,3 +62,4 @@
   normal,
   $file-formats: eot woff2 woff ttf
 );
+/* stylelint-enable */
--- a/src/stylesheets/core/_utilities.scss
+++ b/src/stylesheets/core/_utilities.scss
@@ -102,11 +102,13 @@
     height: $size;
     width: $size;
 
+    /* stylelint-disable block-closing-brace-newline-after, at-rule-empty-line-before */
     @if $direction == 'after' {
       margin-left: $margin;
     } @else {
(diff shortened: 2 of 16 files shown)


src/stylesheets/core/_fonts.scss
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable at-rule-empty-line-before */
```
msecret: Why are we disabling this rule here? Couldn't we just add a newline on line 1?

src/stylesheets/core/_utilities.scss
```diff
@@ -150,7 +152,7 @@
       z-index: 1;
     }
 
-    &.usa-current {
+    &.usa-current { /* stylelint-disable-line selector-no-qualifying-type */
```
msecret: This rule is not re-enabled.

But I'd also potentially prefer to just move `.usa-current` outside of `a` because I'm not sure the extra specificity is needed?

msecret: Actually, it might be required because things like `a:visited` might override it.

src/stylesheets/core/_utilities.scss
```diff
@@ -175,7 +176,7 @@
     line-height: $heading-line-height;
 
     &:hover,
-    &.usa-current {
+    &.usa-current { /* stylelint-disable-line selector-no-qualifying-type */
```
msecret: Is this rule ever re-enabled? Or does it need to be?

shawnbot: `stylelint-disable-line` doesn't need to be re-enabled; it only applies to the line 😀

src/stylesheets/core/_fonts.scss
shawnbot: IIRC, stylelint was throwing errors for every `@include` rule in the file.

PR #1626, opened 2016-12-14
Add npm publish to Circle CI when merged into master.

juliaelman:


circle.yml
```diff
@@ -7,6 +7,15 @@ test:
...
-    - npm run build:package # Run the release process
+    - npm run build
   post:
-    - ls -agolf dist/ # Ensure that build:package worked
\ No newline at end of file
+    - ls -agolf dist/ # Ensure that build:package worked
+
+deployment:
+  npm:
+    branch: master
```
msecret: Do we use git tags for this at all? So we publish a git tag (which shows up on the github releases page) and then publish to npm?

circle.yml
```diff
@@ -7,6 +7,15 @@ test:
...
+    owner: 18F
+    commands:
+      - npm run build:package # Build distributed package files
+      - echo -e "$NPM_USERNAME\n$NPM_PASSWORD\n$NPM_EMAIL" | npm login
+      - npm run publish
```
msecret: We had good success using a npm module called `check-publish` for this, as it will check if the current package.json version is already on npm, and only attempt to publish if it's not.

See [cg-style build](https://github.com/18F/cg-style/blob/master/circle.yml#L33)

msecret: Yea great idea @adborden

circle.yml
juliaelman: We do use git tags for each of our releases. At the moment that only happens when we publish our release notes, hence pointing the at `master`. I'd like to change this up in the future where we remove the `release` branch and target releases with tags instead. 

Hope that makes sense!

juliaelman:
Okay! @msecret @adborden thank you for your feedback. I've made some updates around your suggestions and replaced the npm login vars with an API token route instead - much smarter! I also added the `check-publish` command.

msecret:
@juliaelman @adborden I'm getting the follow arrow when I pull this down and try and run `npm install` something.

```
Error: Failed to replace env in config: ${NPM_API_TOKEN}
```

juliaelman:
@msecret Per the npm docs, I think you need to the following to your `.profile`:

> You will also need to add this to your environment variables on your development machine. In OSX or Linux, you would add this line to your ~/.profile:
>
> `export NPM_TOKEN="00000000-0000-0000-0000-000000000000"`
>
> and then refresh your environment variables:
>
> `source ~/.profile`

juliaelman:
Ah! Also, noting that you should s `NPM_TOKEN/NPM_API_TOKEN`, since that is the var we are using here.

adborden:
@juliaelman sorry for the churn on this, but the discussion has been useful. We just reverted the npm token change on cg-style because it would require everyone touching the code to have a token set, and therefore an npm account. We _could_ work around this by only adding the `.npmrc` file before deploy (not committed to the project) by adding a step in circleci:

```
- "echo '//registry.npmjs.org/:_authToken=${NPM_API_TOKEN}' > .npmrc"
- npm install
...
```

But I haven't tested this yet. I would go ahead with the user/email setup, and then when I have a chance to make sure this works well, I'll submit a PR on wds.

juliaelman:
@adborden agreed! Glad we are hashing this out and figuring out best approaches.

I'd like to continue down the path of using api tokens rather than the username/password approach. I've reached out directly to the folks at Circle CI and will also reach out to our infrastructure folks. I'll update here once I hear back!

adborden:
Sounds good. FWIW, travis's approach is to create the `.npmrc` file at
deploy time, similar to what I have above.
https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/npm.rb


--
Aaron D Borden
Human and Hacker

juliaelman:
@adborden okay! I think that we've got a final solve here. I pushed up a few changes:

1) Removed `.npmrc` file 517bd25
2) Setting the `NPM_TOKEN` in the `circle.yml` file instead! 5876345

I wonder if we should set that inside of the `deployment` part of the config, since we don't need to set the token outside of it for other tasks - thoughts?

adborden:
Yeah, setting it for deployment only is a great idea.

juliaelman:
@adborden @msecret okay! I *think* this is good to go. Ship it?

juliaelman:
@adborden ping! Just wanted to let you know that the build was fixed when I re-built without the cache. 🎉

juliaelman:
FWIW I think that is a Circle CI config that is completely un-related to this PR. See https://github.com/18F/web-design-standards/issues/1642

adborden:
LGTM, thanks!

PR #1661, opened 2016-12-22
Move lodash.debounce to dependencies

shawnbot:
This is a simple fix for errors related to the unmet `lodash.debounce` dependency when the package isn't npm installed with the `--dev` option. There's no reason that it should be a dev dependency since we're requiring it directly in our code [here](https://github.com/18F/web-design-standards/blob/8f130eb3384aaf33fa4c7782a8717ac2ba0df5af/src/js/initializers/footer.js#L1).

This will allow our module to be required directly in JS, and to be built with [browserify](http://browserify.org/) (and automatically with the [browserify CDN](https://wzrd.in/)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "bourbon-neat": "github:thoughtbot/neat#neat-1.8.0-node-sass",
     "classlist-polyfill": "^1.0.3",
     "cross-spawn": "^2.1.5",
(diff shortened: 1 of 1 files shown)


juliaelman:
@shawnbot truth and good call! Mergin' 🚀

PR #1667, opened 2017-01-09
Add US National Library of Medicine

shawnbot:
The [NIH's US National Library of Medicine support site](https://support.nlm.nih.gov/) uses the Standards. I'm not sure what our convention for labeling these is, so let me know if the "NIH" bit is either unnecessary or should be un-abbreviated, removed, or otherwise changed.


juliaelman:
@shawnbot thanks for adding! I think we can leave `NIH` out for this.

Issue #1642, opened 2016-12-20
CircleCI: generates minified CSS test fails


bruffridge:
Occasionally the CircleCI build fails because it doesn't find the minified css file.

https://circleci.com/gh/bruffridge/web-design-standards/8

Looks like it also happens occasionally in your repo:
https://circleci.com/gh/18F/web-design-standards/1145

Any ideas what is causing this?

shawnbot:
I'm scratching my head over this one. My first hunch is that the async gulp call isn't finishing before the tests run, but I'm not 100% sure if that's even possible.

juliaelman:
Following up here per a conversation I just had with @shawnbot about this. 

My gut is telling me that we need to add some cache busting as part of our Circle CI builds. This error goes away if you [rebuild without cache](https://circleci.com/docs/how-cache-works/#clearing-the-cache). We're going to circle back with some of our infrastructure folks to determine a solve. Stay tuned!

shawnbot:
@bruffridge can you confirm that rebuilding without the cache makes the error go away?

shawnbot:
So [this](https://github.com/18F/web-design-standards/blob/staging/spec/sass/build.spec.js#L15) is the place to look if anyone gets around to this before I get back from vacation. I think what might be happening here is that `npm` or `gulp` is forking and the`child_process.exec()` callback is firing when before some deeper "ancestor" process exits. That's what I would _expect_ to be the issue with builds failing with no obvious pattern, anyway.

@msecret @rogeruiz Do you guys have any idea what might be going on here?

shawnbot:
See #1680 for what I think is a robust fix on top of #1664.

juliaelman:
@bruffridge we've merged in your PR, as well as refined some parts of the build process for our Sass gulp task and testing. After testing locally, as well as on a test branch, these solves seem to have resolved this issue. 🎉 

Closing this issue for now. If we see the problem arise again, we can open a new issue.

PR #1680, opened 2017-01-19
Fix mysterious minified CSS build failures

shawnbot:
This is, I think, a more robust fix for #1642. TL;DR:

* Refactor `sass` gulp task to use [cssnano] and use a single stream with multiple outputs ("expanded" and minified CSS w/sourcemap)
* Similarly refactor the `javascript` gulp task to reuse the same stream for minification
* Nix the `merge-stream` dependency
* Speed up `mocha spec/sass/build.spec.js` by having it only run `gulp sass` instead of `npm run build`, which does a bunch of other, unrelated work

After merging #1664 (thanks, @bruffridge!), I ran the tests a bunch of times locally and discovered that gulp was _still_ doing something funky with the "default" and "minified" streams in our `sass` task—but only under certain and rather mysterious conditions. As far as I can tell, either [merge-stream](https://www.npmjs.com/package/merge-stream) (which we used to "merge" the two streams so that gulp would get an `end` stream event only when both streams were complete) was not detecting completion of both streams or ignoring the minified one altogether, or gulp was for some reason failing to sync the minified virtual file to disk. My script to test this was:

```sh
rm -r dist/css; gulp sass && ls -la dist/css
```

I ran this a bunch, and at some point, without any local modifications, I was able to get this to reliably produce _only_ `dist/css/uswds.css`, and not the minified version. So I refactored, and ended up with a gulp task that recycles the "default" stream and passes it through [cssnano] with an additional `rename()` and sourcemap transform. After refactoring, I consistently ended up with all three `dist/css` files (the expanded CSS, minified CSS, and sourcemap).

As a bonus, I applied the same refactoring to our JavaScript minification, which should also be much faster now that we're not spawning two separate (but identical) browserify bundles in parallel. This allowed us to nix the `merge-stream` dependency altogether. 🎉 

[cssnano]: http://cssnano.co/

### DIFF
--- a/config/gulp/javascript.js
+++ b/config/gulp/javascript.js
@@ -37,30 +37,22 @@ gulp.task(task, [ 'eslint' ], function (done) {
     debug: true,
   });
 
-  defaultStream = defaultStream.bundle()
+  var stream = defaultStream.bundle()
     .pipe(source('components.js'))
     .pipe(buffer())
     .pipe(rename({ basename: dutil.pkg.name }))
     .pipe(gulp.dest('dist/js'));
 
-  var minifiedStream = browserify({
-    entries: 'src/js/start.js',
-    debug: true,
-  });
-
-  minifiedStream = minifiedStream.bundle()
-    .pipe(source('components.js'))
-    .pipe(buffer())
+  stream
     .pipe(sourcemaps.init({ loadMaps: true }))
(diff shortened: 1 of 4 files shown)


juliaelman:
@shawnbot this looks good to me 🚀

PR #1702, opened 2017-02-07
Add code coverage report

shawnbot:
#### TODO

- [x] Replace codecov with [Code Climate](https://docs.codeclimate.com/docs/javascript)
- [x] Review from @juliaelman 

Original PR description:

> This adds automated code coverage reports generated by [istanbul] and summarized with [codecov.io]. The [initial summary](https://codecov.io/gh/18F/web-design-standards/list/add-coverage/src/js) is actually great: **our test suite covers 91.83% of our code**. (See below for more info.)
> 
> We'll need to merge this all the way up to `master` to get the full benefits of [codecov's] support for detecting coverage diffs between branches, but for now the branch-specific reports are very useful.
> 
> As for the coverage misses, most appear to be `attachEvent()` detection conditionals for IE8 and below. It should be safe to remove these since we don't formally support anything below IE9, but that will only improve our coverage. Other misses are errors thrown in expected failure cases, some of which don't appear to have corresponding tests, and all of which should be easy to backfill.

[codecov.io]: https://codecov.io
[istanbul]: https://github.com/gotwarlost/istanbul


juliaelman:
Adding a note here that we are going to move over to [Code Climate](https://docs.codeclimate.com/docs/setting-up-test-coverage) instead of Codecov.

juliaelman:
WFM 🎸

PR #1721, opened 2017-02-14
Add grid offset classes

colinalford:
## Description

I'm working on the sam.gov project with GSA. We needed offset classes that work with the grid. They should work like bootstrap offsets, so that when applied, the content within this class will be shifted x columns to the right. Neat provides the shift mixin to easily accomplish this.

This is a quick implementation that solves the particular needs of our project.

## Additional information

[Offset concept as implemented by Bootstrap](http://getbootstrap.com/css/#grid-offsetting)

### DIFF
--- a/src/stylesheets/core/_grid.scss
+++ b/src/stylesheets/core/_grid.scss
@@ -152,3 +152,48 @@
 .usa-end-row {
   @include omega();
 }
+
+// Grid Offsets
+.usa-offset-one-twelfth {
+  @include shift(1 of 12);
+}
+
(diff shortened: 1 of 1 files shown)


maya:
I've also seen classes written as `.usa-shift-one-twelfth` in line with the include in case you'd like to consider an alternate naming convention.

line47:
This is great! Id love to see this type of flexibility with the grid.

donjo:
Going to merge this in and document in #2008.

Thanks!

Issue #1725, opened 2017-02-21
Publish release v1.0


shawnbot:
From our [release docs](https://github.com/18F/web-design-standards/blob/release-1.0.0/RELEASE.md#release-process), here is what we need to do:

- [x] run `npm version major --no-tag` on the `release-1.0.0` branch (or just update the "version" field in [package.json](https://github.com/18F/web-design-standards/blob/release-1.0.0/package.json#L3) manually)
- [x] [create a PR](https://github.com/18F/web-design-standards/compare/master...release-1.0.0) to merge `release-1.0.0` to `master`: #1726
- [x] draft the release notes for the [v1.0.0 release](https://github.com/18F/web-design-standards/releases/new?tag=v1.0.0) and get them reviewed
- [x] create a pre-release to test our automatic npm publish logic (see: #1731, #1732)
- [x] merge the v1.0.0 PR
- [x] upgrade the docs to use the new release version
- [x] [create a PR](https://github.com/18F/web-design-standards-docs/compare/master...staging) to publish the new docs site: 18F/web-design-standards-docs#176
- [x] merge the docs PR
- [x] publish the [v1.0.0 release on GitHub](https://github.com/18F/web-design-standards/releases/tag/v1.0.0) 
- [x] merge master back to staging: #1734
- [x] 🌮 🎉

juliaelman:
Noting here that we are splitting release tasks for 1.0. I'll be tackling the website release and @shawnbot the Standards themselves. ^5

shawnbot:
DONE!

PR #1726, opened 2017-02-21
Release 1.0.0

shawnbot:
This is the PR that will merge v1.0.0 to master, where we can tag and publish it to npm. cc #1725


juliaelman:
@shawnbot I think we are missing some new additions from `staging` in here? We might need to merge into staging first, see https://github.com/18F/web-design-standards/compare/release-1.0.0...staging

PR #1737, opened 2017-02-24
Add comment about Bourbon and Neat Installation

manavm1990:
As indicated in the basic instructions for both of these great libraries, doing a 'normal' setup *should* result in subdirectories being created, and, we would need to indicate this in our @import(s).

<!-- Please feel free to remove whatever sections/lines in this aren’t relevant.

Use the title line as the title of your pull request, then delete these lines.

Bourbon and Neat are usually installed to their own subdirectories, and the current import statements imply that they are just all thrown in together with lib/

This relatively minor fix just addresses the import directory structure.

### DIFF
--- a/src/stylesheets/uswds.scss
+++ b/src/stylesheets/uswds.scss
@@ -1,8 +1,9 @@
 /*! uswds @version */
 
+// As the notes below indicate, doing a 'standard' 'bourbon install' and 'neat install' well put the actual SCSS in it's own subdirectory.
 // Vendor -------------- //
-@import 'lib/bourbon';
-@import 'lib/neat';
+@import 'lib/bourbon'; // Most likely this would say: 'bourbon/bourbon'
+@import 'lib/neat'; // Most likely, this would be: 'neat/neat'
 @import 'lib/normalize';
 
 // Core -------------- //


juliaelman:
@manavm1990 thank you for your interest and contributing to the U.S. Web Design Standards. 

Since Bourbon and Neat are both [part of the the Standards](https://github.com/18F/web-design-standards/blob/staging/package.json#L36), adding this note to the code base might make folks think that they need to be added manually. 🤔 

It might make more sense to add something to [our developer documentation](https://standards.usa.gov/getting-started/developers/) that explicitly states that these libraries are part of our code base. Thoughts?

manavm1990:
Yes, probably for the docs. NP. Sounds good. It's really a very minor thing
any way.


---
Manav Misra
Founder & Lead Visioneer
618-281-7252 <16182817252>



var greatResults = ourCode(yourVision);
<https://twitter.com/VisionWebSoft>

juliaelman:
@manavm1990 I think you hit upon something that @evetsreklaw brought up in our public Slack channel. Just to clarify, are you also using Dreamweaver or have any blockers for installing Node?

manavm1990:
😂 Never DreamWeaver - well not for like 4 or 5 years!

I have no blockers.

It's just that if you do: bourbon install in your 'stylesheets' directory,
then, what happens is it becomes: stylesheets/bourbon - and in here is the
bourbon.scss and that imports from all other folders.

So, if someone was using Bourbon (or Neat), they would need to put: @import
borbon/bourbon (else it will not find bourbon) - assuming they are not
manipulating the 'default' way of installing Bourbon as provided on
bourbon.io


---
Manav Misra
Founder & Lead Visioneer
618-281-7252 <16182817252>



var greatResults = ourCode(yourVision);
<https://twitter.com/VisionWebSoft>

juliaelman:
Ok cool - thanks for clarifying! I am going to close this pull request and recommend we take look a closer look at our documentation around Standards Sass + Bourbon + Neat information. It sounds like it might be confusing how all of these live together.

Issue #1746, opened 2017-02-28
Output HTML suitable for performance testing


shawnbot:
Per #815, one thing that we _have_ to do is output HTML pages suitable for performance testing. More specifically, we need a suite of HTML documents that properly reference all of the assets (CSS, images, and JavaScript) in a way that they can be served up publicly to testing services such as Google PageSpeed Insights. These pages might also prove useful for visual regression testing (#1698).

This dovetails with efforts to refactor our HTML "templates" (which are really just Jekyll templates that get copied to the docs site) so that they're more component-based, e.g. so that we can define the banner as a component that's included by the header in each of its variations, rather than hard-coded into four different Jekyll templates. That work is under discussion in #1596.

At first I was thinking that these might just be our two [page templates](https://standards.usa.gov/page-templates/), but we might want to take this a step further and create some more "representative" pages, or even test some of the more complicated content on the docs site (without, erm, testing the docs site's customizations).

shawnbot:
There's also some discussion of this in #1342.

shawnbot:
This is basically done in #1749. Our build process now generates standalone HTML files from Nunjucks templates, and Federalist builds everything in our `dist` directory. You can see our [docs](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/develop/html/page-docs.html) and [landing](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/develop/html/page-landing.html) pages live there.

Issue #1738, opened 2017-02-24
gov banner text on websites and HTTPS is inaccurate


konklone:
## Description

This text contains some inaccuracies (screenshot from https://tech.gsa.gov) -

<img width="1105" alt="screen shot 2017-02-24 at 1 57 06 pm" src="https://cloud.githubusercontent.com/assets/4592/23321359/2876cff6-faad-11e6-9df2-e69c6922e632.png">

The two big issues are -

* Not every government websites ends in `.gov` or `.mil`. Besides `.fed.us` also being a valid "official" suffix, there are government websites that use .com and .org. For example, [here's a list DHS keeps of theirs](https://www.dhs.gov/dhs-sites-outside-gov-or-mil-domains), and [here's a (stale) dataset of other URLs kept by GSA](https://github.com/gsa/govt-urls). While OMB wants people to use .gov and .mil only, we can't tell users that all government websites use them.

* The HTTPS certificate information says that the US government signed the certificate, which is not true. Commercial certificate authorities do that, and this will differ per-agency. In the case of tech.gsa.gov, Let's Encrypt signed the certificate. In addition, `SSL` is an outdated marketing term and not an accurate description of the certificate or protocol in use. I am also not sure what the reference to `browsing history` is for. I would like to rewrite this language entirely and refocus it.

## Steps to reproduce the issue

1. Visit a site with the full banner.
2. Click the `Here's how you know` link in the banner.

## Additional information [optional]

The relevant places with text seem to be:

* https://github.com/18F/web-design-standards/blob/master/src/html/header-basic.html
* https://github.com/18F/web-design-standards/blob/master/src/html/header-basic-mega.html
* https://github.com/18F/web-design-standards/blob/master/src/html/header-extended.html
* https://github.com/18F/web-design-standards/blob/master/src/html/header-extended-mega.html

shawnbot:
Thank you for this, @konklone! For me this signals a need to refactor our header code and promote the banner as a discrete component. See #1596 for more info.

juliaelman:
@konklone thanks for opening this issue! While @shawnbot works on the header refactor, let's schedule some time to pair and talk about how we can update the language.

juliaelman:
Talked with the team and @konklone today about this. We will be removing the copy in the banner until we can come up with a more general solution for this guidance.

shawnbot:
[...]

Issue #815, opened 2015-10-09
Tracking performance of web design standards


msecret:
As the standards grow, I think it's important to measure the performance impacts of using the library to ensure as change come in performance doesn't degrade down to unacceptable levels.

I think the best way to do this would be to add performance tracking to example template pages we've discussed creating: https://github.com/18F/web-design-standards/issues/759. In terms of metrics, I'd suggest something generic like speed index as the main metric. I'd image this would be part of the CI process OR something similar to speedcurve.com, where we'd just track over time.

I don't think we need any hard and fast rules on what would be acceptable performance, because people could leverage additional techniques (critical CSS, etc) to improve performance on their sites. I think it's just good to have an eye on performance as development continues

mollieru:
Cool, good idea @msecret -- we'll consider this in our future roadmap.

shawnbot:
I'm upgrading the priority for this one because we need to be able to measure performance before we start trying to improve it.

yowill:
@lauraponce Check this out!

shawnbot:
Okay, so after a bit of research, I think there are a bunch of tasks to stub out in separate issues for this one. As I write this, I'm thinking that we probably won't get all of this done in a single sprint.

- [ ] Decide which metrics to measure
- [ ] Decide which tool(s) to use
- [ ] Get this repo to generate testable HTML files, and possibly to publish them somewhere public (Federalist) that they can be tested by services such as [Google PageSpeed Insights]
- [ ] Get reporting working on our CI service with the chosen tool(s)
- [ ] Figure out how to track variance (e.g. between scores on a given branch and `master`, which is how our code coverage tracking works), where reports "live" that they can be compared (CI build artifacts?), etc.
- [ ] Decide if, and under which conditions, a variance fails a CI build (e.g. if the [Google PageSpeed Insights] score goes down by 5 points or more, or if the [YSlow] grade drops)
- [ ] Run through the process a couple of times to test and validate it

Before I go and start filing issues, does anyone have any thoughts? Are these tasks too granular, or not granular enough? @msecret @lauraponce @juliaelman @yowill 

Also, I have a lot more written down about tools, but I didn't want to dump it all here.

[Google PageSpeed Insights]: https://developers.google.com/speed/pagespeed/insights/
[YSlow]: http://yslow.org/

shawnbot:
Okay, and with #1791 all of the sub-tasks from the above comment have been broken out. We're going to close this issue and move forward with the more granular ones. ✌️

PR #1749, opened 2017-03-01
Refactor component and page templates

shawnbot:
The major thrust of this PR is to refactor our component and page templates, which currently exist as Jekyll-specific HTML templates that we use on the docs site, as more generally useful [Nunjucks] templates and standalone, static HTML that will eventually drive our performance tests (see: #815).

⚠️  **I haven't fully considered whether this should be a breaking change**, because I don't know whether any of our users are actually _using_ the existing `src/html` directory. But to my knowledge they're not, and I doubt that they're particularly useful to anyone because they're Jekyll templates that really only make sense in the context of the docs site. So I would vote for this being a new feature, which means that this will bump the **minor** version (e.g. from `1.0.0` to `1.1.0`).

Anyway, here's the deal:

- [x] Move the templates from `src/html` to `src/template`.
- [x] Refactor all asset paths from Jekyll-specific `{{site.baseurl}}/assets/...` to `{{uswds.path}}/...`.
- [x] Add a new (?) `html` gulp task that builds static HTML from the templates using [gulp-nunjucks] with the `uswds.path` context to `..` so that the files generated in `dist/html` point to the assets in the `dist` directory.
- [x] Run the generated HTML through  [gulp-html-beautify] to produce consistent indentation, even though component partials aren't indented.
- [x] Factor out common, non-component HTML snippets from the page templates into `meta.html`, `styles.html`, and `scripts.html`.
- [x] Refactor page templates to include partials for their respective header and footer.
- [x] Rename page templates from `template-{name}.html` to `page-{name}.html`.
- [x] Break `footers.html` into separate files for each variant (`footer-slim`, `-medium`, `-big`).
- [x] Move the gov banner out of the header templates and into `banner.html`, and remove the expandable content per #1738.
- [x] Move the skipnav link out of header partials and into the page templates, so that it lives in the same file as its corresponding `<main>` element.
- [x] Fix JavaScript whitespace issues in `gulpfile.js`.
- [x] Make it build on Federalist: [landing](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/feature-refactor-templates/html/page-landing.html), [docs](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/feature-refactor-templates/html/page-docs.html) per #1746
- [x] Add tests
- [x] Review from @juliaelman.
- [x] Update relevant documentation.

[Nunjucks]: https://mozilla.github.io/nunjucks/
[gulp-nunjucks]: https://www.npmjs.com/package/gulp-nunjucks
[gulp-html-beautify]: https://www.npmjs.com/package/gulp-html-beautify


juliaelman:
Excellent! Quick note: we will also need to test that the code samples on standards.usa.gov are still compiling and displaying correctly.

juliaelman:
@shawnbot something else to think about is name consistency. We use both `html` and `templates` to refer to these files. We also have `src/html` and `dist/templates`. This is confusing and we should pick one or the other. I am leaning towards html so we are more explicit and template might infer that these files contain a full page layout.

Thoughts?

juliaelman:
Thanks for clarifying. While *we* understand the compilation process of going from a template to raw html, not all of our users might be used to this process. My thoughts are based off of the user research that we did on [Marigold](https://github.com/18F/marigold) for usability testing on the Fractal interface. Some users were confused what the difference between the `View` and `HTML` tabs were or how they related to each. This may also be true for users when looking at the `template` to `html` folders when referring to or using this code base.

That said, I am of the mindset to implement, test and iterate! Going to add some documentation now to better outline what all of these things mean and will submit a pull request against the `feature-refactor-templates` branch.

juliaelman:
Submitted a pull request to add some docs in our README about the new templates folder. 

Also, I think `/src/template/figure.html` still has a reference to `{{ site.baseurl }}`.

juliaelman:
Ok! So since we determined that the figure template is not currently "published" we can either remove or leave as is.

Issue #1754, opened 2017-03-02
Basic header mobile nav fills the screen


shawnbot:
Our friend @mgwalker appears to have stumbled on a nasty bug with our CSS that, for some reason, doesn't reveal itself in our page templates. He showed us a bare-bones page with a stock header nested in a couple of divs that mysteriously extended to the height of the entire page:

![image](https://cloud.githubusercontent.com/assets/113896/23488277/758d5232-fea0-11e6-8fc3-cc6b55d7d5ef.png)

I thought it might have something to do with nesting and selector specificity, but after messing around in the inspector I ruled that out and discovered that [this combination of position and size](https://github.com/18F/web-design-standards/blob/06d34219096071d859ed78fe3d0ea0812e9dc67e/src/stylesheets/components/_navigation.scss#L42-L43) produces CSS that does the wrong thing:

```css
.usa-nav {
  position: fixed;
  top: 0;
  right: 0;
  bottom: 0;
  left: auto;
  height: 100%;
  /* ... */
}
```

Specifically, the `height: 100%` is redundant because we've already specified `top` and `bottom`. At large screen widths we [revert the `position` and `width`](https://github.com/18F/web-design-standards/blob/06d34219096071d859ed78fe3d0ea0812e9dc67e/src/stylesheets/components/_navigation.scss#L63-L64), but not `height`. For some reason, the 100% height doesn't expand to fill the page in our page templates, but in the test case it does. Disabling that rule in the inspector fixes it, and doesn't affect the fixed-position mobile layout because the height is implied by setting `top` and `bottom`.

(Aside: This is a _really_ good example of an issue that would be easier to detect and debug if we had a standalone test page for our header. [Fractal], anyone?)

I haven't seen anyone else report this, but it's reasonable to assume that other users have run into this issue, and I would like to categorize it as a major bug because it tripped me up, and I'm a maintainer!

[Fractal]: http://fractal.build

shawnbot:
One thing I'd like to suggest we do here is cook up a standalone test case so that we can be 100% sure both that we're able to reproduce this and that our proposed fix resolves it. Here's one way that we could get there:

[...]

PR #1761, opened 2017-03-02
Fix errant .usa-nav height: 100%

shawnbot:
This is a fix for #1754 that replaces the call to our Sass `size()` mixin for our navigation at mobile screen widths (which sets both the width and height) with setting only `width`.

cc @mgwalker for visibility, and help testing with his use case


mgwalker:
FWIW, I applied this patch in my project and it's looking good!  Let me know if there's more formal testing I can do to help.  😄

juliaelman:
🚀  this solve will be part of the next release! Nice work!

PR #1775, opened 2017-03-09
[Typography]: Move heading mixins from typography file to utilities file

### LINKED ISSUE #1763: Move mixins to their own files
Currently, both mixins and rulesets sometimes live within the same Sass file. For example, [`elements/_typography`](https://github.com/18F/web-design-standards/blob/develop/src/stylesheets/elements/_typography.scss#L81-L116) includes rulesets for base HTML elements, but also includes Sass mixins. This prevents someone from being able to import _just_ the mixins into their Sass project and would be much more usable if they were moved to a separate, mixins-only, file.

I'd propose either:

1. Create a `core/_mixins.scss` file, similar to how you currently have a `core/_variables.scss` file
2. Or, preferably, follow a pattern like [ITCSS](http://thomasbyttebier.be/blog/less-css-mess) where there's a `tools` and `settings` layer for things like mixins and variables.

This would allow an existing Sass project to do:

```scss
@import '~uswds/core/mixins';
@import 'my-custom-mixins';

.my-heading {
  @include h1;
}
```

I'd be happy to make this pull request, but would like to get your thoughts before moving things around 😁 

cc #1342

sawyerh:
This resolves #1763 by moving the heading mixins to the utilities file with the other mixins.

### DIFF
--- a/src/stylesheets/core/_utilities.scss
+++ b/src/stylesheets/core/_utilities.scss
@@ -1,3 +1,41 @@
+// Heading mixins
+@mixin title {
+  font-size: $title-font-size;
+  font-weight: $font-bold;
+}
+
+@mixin h1 {
+  font-size: $h1-font-size;
+  font-weight: $font-bold;
+}
+
+@mixin h2 {
+  font-size: $h2-font-size;
+  font-weight: $font-bold;
+}
+
+@mixin h3 {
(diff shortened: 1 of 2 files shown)


juliaelman:
LGTM - thanks for the pull request @sawyerh! 🚀

PR #1792, opened 2017-03-14
Removing :first-child and :last-child margin for .usa-grid-full.

juliaelman:
Tested this on our docs site locally and everything looks good on the site, page templates demo. Would be good to take another visual walk through to double check and make sure this works 💯  before merging in. 

On another note, we'll need to be explicit about this change in whatever release this goes out with. 🚀


juliaelman:
Ping @shawnbot

PR #1806, opened 2017-03-21
Refactor accordion logic; fix nested button events

### LINKED ISSUE #1762: Buttons within accordions get assigned .usa-button-accordion event listeners
## Description

When a button is placed within an accordion content div, the button gets assigned event listeners as if it were the usa-accordion-button.  Clicking the button will trigger the accordion to close when it shouldn't.

In the changes since 0.9.x, it seems that care was taken to avoid some of the problems that were present in 0.8.x.  However, it seems one line of code was missed.

1. Use this code to test:
````
<ul class="usa-accordion">
  <li>
    <button class="usa-accordion-button" aria-expanded="false" aria-controls="formContent">View form</button>

    <div id="formContent" class="usa-accordion-content">
      <form>
        <label for="my-input">Number between 1 and 5:</label>
        <input id="my-input" type="number" min="1" max="5" />
        <button type="submit">Submit</button>
      </form>
    </div>
  </li>
</ul>
````

2. Type in a number greater than 5
3. Click the submit button
4. The accordion closes, and a JS error occurs because it cannot render the validation message

## Fixing the problem
In uswds.js line 656 does not have the same CSS selector as line 712:
````
var buttons = select('button', this.root); // line 656
var buttons = this.$('ul > li > button, .usa-accordion-button'); // line 712
````

Changing line 656 to the following solves the problem (tested locally):
`var buttons = select('ul > li > button, .usa-accordion-button');`

### PR DESCRIPTION
**Pull request by shawnbot**
This is a slight refactoring of our accordion JavaScript intended to fix #1762. Specifically:

* Introduce common variables for the CSS selectors used to query for accordion button elements.
* Make DRY the accordion's `show()` and `hide()` methods, which both now rely on a single function to get the "target" of a clicked button via the `aria-controls` attribute.
* Add a `select()` method, the equivalent of `querySelector()`, to make instances of `this.$(selector)[ 0 ]` more readable.

### DIFF
--- a/src/js/components/accordion.js
+++ b/src/js/components/accordion.js
@@ -24,13 +24,21 @@ function showPanelListener (el, ev) {
   return false;
 }
 
-function getTargetOf (button) {
-  var id = button.getAttribute('aria-controls');
+/**
+ * @name getTargetOf
+ * @desc Get the target of an element, according to the id listed in its
+ * `aria-controls` attribute.
+ * @param {HTMLElement} source the source element
+ * @return {Element} the target element
+ * @throws {Error} an error is thrown if no such target exists
+ */
+function getTargetOf (source) {
+  var id = source.getAttribute('aria-controls');
   var target = document.getElementById(id);
   if (target) {
     return target;
   } else {
-    throw new Error('No accordion button target with id "' + id + '" exists');
+    throw new Error('No accordion target with id "' + id + '" exists');
   }
 }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/js/components/accordion.js [hunk 1]
```diff
@@ -1,5 +1,14 @@
+'use strict';
 var select = require('../utils/select');
 
+var SELECTOR_BUTTON = 'ul > li > button, .usa-accordion-button';
```
juliaelman: Let's see if we can be more specific on this and add a class name on `button`

Issue #1596, opened 2016-12-01
Further refactor header component


shawnbot:
Here are some further thoughts on the "contextual" (search and nav within header) styles that I started to consider in #1595, and what we might do about them:

1. The _header-specific_ search component needs its own unique class name, rather than a qualifying selector (`.usa-nav-secondary .usa-search`) so that it can be used anywhere. See also: #1435; I think one good solution is to make the current `.usa-search` appear as the focused one does now,  provide a new class that styles it like the current button when it's not focused, then do away with the button. (v1!)
1. None of the primary nav's styles should be scoped to the header, since we don't "allow" it to live anywhere else. This should be easy to tease apart, though I think there are some selectors (the mobile-only `.usa-navbar`, specifically) that might be trickier to place.
1. The "basic" and "extended" header variants each change very specific things about contained elements. `.usa-header-basic`, for instance, [appears to](https://github.com/18F/web-design-standards/blob/24bebbd61fdae89fd89f9cf2d0034eb063113dba/src/stylesheets/components/_header.scss#L120-L134) just tweak the logo positioning on mobile and modifies the width of the navbar (?), and the extended header has [a lot more going on](https://github.com/18F/web-design-standards/blob/24bebbd61fdae89fd89f9cf2d0034eb063113dba/src/stylesheets/components/_header.scss#L138-L221). We should take a closer look at this and see if we can eliminate some of this complexity so that the header is easier to test and maintain.

shawnbot:
Here are a couple of thoughts on refactoring after spending some more quality time with the header:

* Trying to generalize a template for the header (see [this](https://github.com/18F/web-design-standards/blob/b900cb0c389cf36277ce169f0fd3735a7eb40ea0/components/header/header.njk) abandoned example) proved really difficult, mostly because the differences in HTML from one to the other weren't obvious.

    For instance, the basic header wraps all of its contents in a `.usa-nav-container`, but the extended one doesn't. This makes for a hard-to-read diff and an awkward workaround in the generalized template that shouldn't be necessary. **More importantly**, though, it makes for a potentially difficult and/or frustrating experience when a user decides to add or remove the secondary nav. One solution for this would be to always include the container div, but negate its sizing in the `usa-header-extended` styles.

[...]

Issue #1594, opened 2016-11-30
The header should support a logo image


shawnbot:
Currently our header doesn't offer any guidance for where to place a logo, how big it should be, or anything like that. I imagine that many government agencies would like to be able to put a logo next to their name, and it would be nice if they could do that without having to twiddle with `width` and `height` attributes in the HTML—in other words, if we sized it reasonably/appropriately in CSS.

guitarzan:
I'm glad you're addressing this! We've recently wrestled with adding a logo to USWDS to comply with our agency co-branding standards. Here's what we came up with. It's working well for us, but I'd appreciate your input.

SCSS: 

```scss
.usa-header-basic {
    .gaepd-logo {
    @include media($nav-width) {
      bottom: 0;
      position: relative;
    }
  }
  .gaepd-navbar {
    @include media($nav-width) {
      width: auto;
    }
  }
}
.gaepd-logo-link {
    display: flex;
    align-items: center;
}
.gaepd-logo-link img {
    width: 3rem;
    margin-right: 1rem;

    @include media($nav-width) {
        width: 8rem; 
        margin: 1rem 2rem 1rem 0;
    }
}
```

HTML: 

```html
    <header class="usa-header usa-header-basic" role="banner">
        <div class="usa-nav-container">
            <div class="usa-navbar gaepd-navbar">
                <button class="usa-menu-btn">Menu</button>
                <div class="usa-logo gaepd-logo" id="logo">
                    <em class="usa-logo-text">
                        <a accesskey="1" title="Home" aria-label="Home" class="gaepd-logo-link" href="/">
                            <img src="/assets/logos/epd-icon.svg" alt="" /> Environmental Protection Division
                        </a>
                    </em>
                </div>
            </div>
            <nav role="navigation" class="usa-nav">
                <button class="usa-nav-close">
                    <img src="/assets/img/close.svg" alt="close">
                </button>
                <ul class="usa-nav-primary usa-accordion">
                    <li>
                        <a title="Home" aria-label="Home" href="/">Home</a>
                    </li>
                    <li class="js-search-button-container">
                        <a class="usa-header-search-button js-search-button" href="/Search">Search</a>
                    </li>
                </ul>
            </nav>
        </div>
    </header>
```

And here is how it looks on desktop/mobile:
![image](https://cloud.githubusercontent.com/assets/737326/25093699/662a11d2-2361-11e7-8986-6dc2959f25b2.png)
[...]

PR #1823, opened 2017-03-30
Refactor header templates

shawnbot:
This is an attempt to better understand (and document) how the four variants of our header component work and differ from one another, with an eye toward considering how best to refactor them in #1596. This PR doesn't touch our CSS or JavaScript, but does pretty drastically change our Nunjucks templates and the HTML output:

1. Templatize the header title and navigation elements so that they can accept dynamic data, rather than maintaining duplicative static HTML.
1. Simplify the four header configurations or variants (`basic`, `basic-mega`, `extended`, and `extended-meta`) so that they rely primarily on common includes for our "masthead" (site title), primary, and secondary navigation.
1. Introduce [a single page](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/refactor-header/html/page-headers.html) where we can test and compare all four headers in one place.
1. Only output HTML for our `page-*` templates: [docs](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/refactor-header/html/page-docs.html), [landing](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/refactor-header/html/page-landing.html), and [headers](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/refactor-header/html/page-headers.html).

See the comments below for observations based on this work.


line47:
I like the idea of showing the search textbox and search button instead of using JS to show/hide the search functionality. One other search related item I wanted to add was that when on a small screen and viewing the navigation the search is at the bottom of the list of navigation options. I would expect the search to be at the top in the navigation.

donjo:
I think your breakdown of the 3 main visual styles makes sense. I have some comments on suggested CSS changes based on how this is broken down so should we save that for another thread or are they still relevant here? 

One example would be how the overall height of the navbar is set through a mix of heigh on `usa-navbar` and padding-top on `usa-nav` both of which are applied to different elements and have to stay in sync. 

❤️ the header preview page!

line47:
@shawnbot I really really like the idea of modularizing these components and making a tighter 1:1 relationship! ➕

config/gulp/template-data.js
```diff
@@ -0,0 +1,61 @@
...
+
+var fakeLink = function (text) {
+  return {
+    text: text,
+    href: 'javascript:void(0)',
+  };
+};
+
+module.exports = {
+  package: require('../../package.json'),
```
msecret: I'm a little confused why this is needed.

src/templates/header-basic-mega.html
```diff
@@ -1,173 +1,11 @@
...
-      <button class="usa-menu-btn">Menu</button>
-      <div class="usa-logo" id="logo">
-        <em class="usa-logo-text">
-          <a href="#" title="Home" aria-label="Home">Department of <br>Web Standards</a>
-        </em>
-      </div>
-    </div>
+    {% include 'masthead.html' %}
+
     <nav role="navigation" class="usa-nav">
```
msecret: I would kinda expect the nav component to include a `nav` element, but it also sounds like it would be hard to split them up with it? What were your thoughts on this?

src/templates/masthead.html
```diff
@@ -0,0 +1,12 @@
+<div class="usa-navbar">
+  <button class="usa-menu-btn">Menu</button>
```
msecret: Just as a secondary note, not relevant to this PR, the `usa-menu-btn` class seems a little out of place. It doesn't follow the usual naming convention for buttons of being `button` rather then shortened `btn` and it's confusing where it's supposed to be as a component ('menu' or 'button').

src/templates/masthead.html
```diff
@@ -0,0 +1,12 @@
...
+  <div class="usa-logo" id="logo">
```
msecret: Is there reason for `logo` to be split into own component, or is that not necessary?

src/templates/nav-primary.html
```diff
@@ -0,0 +1,31 @@
+<button class="usa-nav-close">
```
msecret: This sounds like it fills a similar space as `usa-menu-btn`? Or maybe not at all, but unsure where this fits in.

src/templates/nav-primary.html
```diff
@@ -0,0 +1,31 @@
...
+  <img src="{{ uswds.path }}/img/close.svg" alt="close">
+</button>
+<ul class="usa-nav-primary usa-accordion">
+  {% for link in nav.links %}
+  <li>
+    {% if link.links %}
+    <button class="usa-accordion-button usa-nav-link" aria-expanded="false" aria-controls="{{ link.id }}">
+      <span>{{ link.text }}</span>
+    </button>
+    <ul id="{{ link.id }}" class="usa-nav-submenu{% if nav_mega %} usa-megamenu usa-grid-full{% endif %}">
```
msecret: So another note that best not addressed now:

If you need to use the mega menu here, you have to add both `usa-megamenu` and `usa-grid-full`. I wonder if we could use sass to mixin `usa-grid-full` into the megamenu class so only that's needed. Might improve the dev experience to only require one extra class.

config/gulp/template-data.js
shawnbot: It's used in [id.html](https://github.com/18F/web-design-standards/blob/0ef5f8a8ea4ce4d9bdb47db8dd091cfd87794821/src/templates/id.html) to output the package name and version in a comment on each page.

src/templates/header-basic-mega.html
shawnbot: The problem here is that different variants of the header nest the navigation elements differently: the extended headers wrap the nav contents in `.usa-nav-inner`, but the basic ones don't. If we think of this as a documentation task, I think that it succeeds in highlighting issues like this. 😄

src/templates/masthead.html
shawnbot: Probably not right now, but it's worth considering in #1594.

src/templates/nav-primary.html
shawnbot: I'm honestly not sure either.

src/templates/nav-primary.html
shawnbot: I agree 100%.

PR #1835, opened 2017-04-06
Video embed classes/mobile menu fix

donjo:
Adds embed container classes that can be used to make youtube/vimeo embeds responsive and ensures the mobile menu works on pages with embeds that would normally be wider than the viewport. 

screen shot 2017-04-06 at 4 14 26 pm


src/stylesheets/elements/_embed.scss
```diff
@@ -0,0 +1,20 @@
+.embed-container {
```
shawnbot: Let's call this `usa-embed-container`.

src/stylesheets/elements/_embed.scss
```diff
@@ -0,0 +1,20 @@
...
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+  }
+}
+
+.embed-container {
+  $ratio-widescreen: 56.25%; // Assumes 16:9 aspect ratio for video content – the default on most services.
```
shawnbot: I think this might make more sense as a mixin with a configurable aspect ratio, then use the mixin to generate the styles for our `.usa-embed-container`:

```scss
@mixin embed-container($aspect) {
  // ...
}

.usa-embed-container {
  @include embed-container(16/9);
}
```

PR #1836, opened 2017-04-07
Implement event delegation; refactor JS and tests

### LINKED ISSUE #1423: Use event delegation to preserve interactive behaviors in Angular, React, etc.
The current component + initializer setup is kind of a pain to maintain, and has a pretty significant flaw: if you add a new component  to the DOM (or remove one from it) at runtime, it won't work unless you call the initializer(s) again—which will, in turn, add duplicate event listeners to the previously initialized elements. 

[Event delegation](https://davidwalsh.name/event-delegate) can be used to listen for events at the top (or some other "high") level of the document and act only on those that match certain criteria, such as having a target that matches a given CSS selector. For instance, instead of adding individual event listeners to each button in an accordion, you can add a single `click` event listener to the accordion root element that toggles whichever button is clicked.
   
For instance, with [receptor](/shawnbot/receptor) we could dynamically handle any accordion in the document like so:
   
   ``` js
   var delegate = require('receptor').delegate;
   document.body.addEventListener('click', delegate(
     '.usa-accordion [role=tab][aria-controls]',
     function clickTab(e) {
       toggleTab(this);
     }
   );
   
   function toggleTab(tab) {
     var accordion = tab.closest('.usa-accordion');
     // tab toggle logic here
   }
   ```

[Custom events](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events) are a potentially nice way to add "hooks" to components for users to customize their behavior. For instance, @maya's new banner component might listen for a more specific `toggle`  (or `usa-toggle`?) event from its button to toggle the header class, rather than having to know which type of element to listen for more general `click` events on. (Event delegation would help in cases like this, too.)

shawnbot:
### [:eyes: Preview on Federalist](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/feature-event-delegation/)

⚠️  **ES5 users who reference submodules directly will need to transpile their JavaScript** when this is released. For instance, if you `require('uswds/src/js/components/accordion')` and bundle your JavaScript with browserify or webpack, you'll need to incorporate the respective Babel plugin ([babelify](https://github.com/babel/babelify) or [babel-loader](https://github.com/babel/babel-loader), respectively). Users of React or Angular 2 **do not** need to do anything special.

This PR fixes #1423 by replacing all of our "direct" event listeners (added to individual elements once, on DOM ready) with delegated listeners on the `<body>` that _should_ work regardless of what happens in the rest of the DOM. I've created standalone [React](https://federalist-proxy.app.cloud.gov/site/shawnbot/uswds-react-test/) and [Angular 1](http://codepen.io/shawnbot/pen/jmNNOy) test cases that reference the [JS built by this branch](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/feature-event-delegation/dist/js/uswds.min.js), and both appear to work well. Here's the lowdown:

* All components are implemented as [behaviors](https://github.com/shawnbot/receptor#behavior) with additional `on()` and `off()` methods that add and remove all of the declared listeners to the `<body>` on DOM ready (in the respective [initializers](https://github.com/18F/web-design-standards/tree/feature-event-delegation/src/js/initializers)). Our [wrapped behavior](https://github.com/18F/web-design-standards/tree/feature-event-delegation/src/js/utils/behavior.js) provides a couple of additional niceties:
* If a behavior has an `init()` method, that will be run by the behavior's `on()` method automatically (before listeners are added) to initialize any relevant elements. **XXX** maybe this should be called `setup()`?
* If a behavior has a `teardown()` method, that will be run by the behavior's `off()` method automatically, to clean up any temporary state.

The upside of this is that if you (the user of the Standards) render parts, or even the entirety, of your page's content from an AJAX request or with another framework, you don't have to worry about event listeners being blown away and having to add them again.

Along the way, I've basically refactored all of the JS from the ground up. Here's what else happened:

* The "main" script has been renamed from `start.js` to `uswds.js`, to better match the destination filename.
* The main script now exports a `uswds` object, which currently has two properties:
    * `prefix` is the `usa` prefix used to derive CSS selectors for each component behavior.
    * `components` is an object of named component behaviors, each of which can be added (`add()`/`on()`) and removed (`remove()`/`off()`) to any element root at runtime.
* The `components/x.js` + `initializers/x.js` duo of files for each component has been replaced by a single "behavior" script in `components/x.js`. In other words, component definitions _are_ behaviors.
* Some functions that lived in the `components` directory have been converted into stateless functions (`toggleFieldMask()`, etc.) and moved to `utils`.
* Some ambiguously named component scripts (such as [forms.js](https://github.com/18F/web-design-standards/pull/1836/files#diff-982273f4d8a323739b20cf02c05a3712)) have been replaced with more specific ones (such as [password.js](https://github.com/18F/web-design-standards/pull/1836/files#diff-bcba32846ef13a95106c559f74cdfee6)).
* The "password" component (used in the [password reset](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/feature-event-delegation/components/detail/password-reset-form.html) and [sign in](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/feature-event-delegation/components/detail/sign-in-form.html) form templates) now allows you to supply a `data-hide-text` attribute to specify what's displayed when you toggle it, and derives a sensible default by replacing "Show" with "Hide" in the initial text content (respecting case) if there is none. Previously, [these values were hard-coded](https://github.com/18F/web-design-standards/blob/06d34219096071d859ed78fe3d0ea0812e9dc67e/src/js/initializers/forms.js#L11-L12) in JavaScript.
* All of the ES5 shimming is now done with third-party npm modules, such as [array-foreach] and [array-filter].
* There's a new, better [ponyfill](https://github.com/18F/web-design-standards/blob/483361b6d125a0a87f87078e959d3328f30bea31/src/js/polyfills/dataset.js) for [HTML5 dataset].
* All of the scripts in `src/js/vendor` are gone now. Those were a relic of the site, AFAICT.

### Unit test improvements
Along the way, I've also made some improvements to our unit test suite and removed some unnecessary dependencies:

* There are now unit tests for the poorly documented (big) footer faux-accordion and skip-nav tabindex switching, which fixes https://github.com/18F/web-design-standards-docs/issues/238.
* Our unit test suite now uses the Node.js built-in [`assert` module][node-assert] instead of [should].
* We have officially removed jQuery from all of our unit tests. 🎉 
* Unit tests that previously imported HTML templates as JS now read [actual](https://github.com/18F/web-design-standards/blob/acf4a6d9a93f7cf8c00c7bcc53740245df8cbe8e/spec/unit/accordion/template.html) [HTML](https://github.com/18F/web-design-standards/blob/acf4a6d9a93f7cf8c00c7bcc53740245df8cbe8e/spec/unit/banner/template.html) [files](https://github.com/18F/web-design-standards/blob/acf4a6d9a93f7cf8c00c7bcc53740245df8cbe8e/spec/unit/footer/template.html) as strings. Eventually, these files may be generated from the Fractal templates, or tests may be run on Fractal's HTML output.

### ES2015
I've also taken the liberty of upgrading the component JS to ES2015 (ES6) and incorporating [Babel] into our builds. This makes our source JS a **lot** more concise, and at a fairly insignificant size penalty:

File | Before | After
:--- | ---: | ---:
`uswds.js` | 104K | 152K
`uswds.min.js` | 16K | 20K
`uswds.min.js.map` | 100K | 128K

[array-filter]: https://www.npmjs.com/package/array-filter
[array-foreach]: https://www.npmjs.com/package/array-foreach
[node-assert]: https://nodejs.org/dist/latest-v6.x/docs/api/assert.html
[should]: https://shouldjs.github.io/
[Babel]: http://babeljs.io/
[HTML5 dataset]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset


toolness:
Cool!

I'm not very familiar with how the USWDS JS components currently work, so I guess one question I have is, how do the virtual DOM interop concerns you have for this PR not *also* apply to the current state of the JS, outside of this PR?

I've used third-party code with React a bit (see the [lots of legacy](https://gist.github.com/toolness/b2e63cc6df4420bee557e932e7e5c941#lots-of-legacy) section of the CALC React migration notes) but the main way I've evaluated whether to rewrite a dynamic component in React vs. use a native DOM implementation has largely centered around how "encapsulated" the component is in the DOM.

For example, with CALC, the d3 histogram was an easy decision to make because it's easy to just take that one DOM node and tell React "let me manage this DOM node, don't mess with it".  It's just a black box to React, and we can just subscribe to React's component lifecycle methods to make the native DOM changes required when the component's props are changed, remove any event listeners when the component leaves the DOM, and so on.

On the other hand, things might get more complicated if there's lots of interleave between DOM and virtual DOM components.  An example of this might be a native DOM-based accordion component, where the "frame" of the accordion is managed by native DOM code, but you want content *inside* the accordion panels that's managed by a virtual DOM.  This is something I don't have much experience with, so it might be possible if you're careful, but it's also where I'd start seriously considering just re-implementing the native DOM code in React, to avoid any possibility of things exploding unexpectedly. Again, though, I don't have a lot of experience with this kind of situation, so I could be wrong.

joshbruce:
Wow! @shawnbot you've been busy. Not sure why I just now got the email on this. Having issues with local development and things...gonna tag in some teammates to see if they can help out.

src/js/utils/behavior.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+const assign = require('object-assign');
+const forEach = require('array-foreach');
+const Behavior = require('receptor/behavior');
+
+const sequence = function () {
```
jseppi: I think you could use the [spread operator](https://babeljs.io/learn-es2015/#ecmascript-2015-features-default-rest-spread) to write this line as

```js
const sequence = function(...seq) {
```

and remove the following line (`const seq = [].slice.call(arguments);`

src/js/utils/behavior.js
```diff
@@ -0,0 +1,31 @@
...
+  const seq = [].slice.call(arguments);
+  return function (target) {
```
jseppi: Similarly, you could use ES2015 default args to write this as

```js
return function (target=document.body) {
```

src/js/utils/behavior.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+const assign = require('object-assign');
```
jseppi: Consider using `import` in place of all `require`s. While it wouldn't really have any immediate effect, it would pave the way for tree-shaking in your bundler (either webpack2 or rollup).

jseppi: Could also obviously be handled later in a separate PR.

src/js/components/accordion.js
```diff
@@ -1,122 +1,76 @@
...
-    self.hide(button);
+const getAccordionButtons = accordion => {
+  return filter(select(BUTTON, accordion), button => {
+    return button.closest(ACCORDION) === accordion;
   });
-  return this;
 };
 
-module.exports = Accordion;
+module.exports = behavior({
```
jseppi: Along the same vein of using `import`, consider using `export`/`export default`. Perhaps in a separate PR.

src/js/components/banner.js
```diff
@@ -1,21 +1,23 @@
...
-  (event.preventDefault) ? event.preventDefault() : event.returnValue = false;
+const PREFIX = require('../config').prefix;
+const HEADER = `.${PREFIX}-banner-header`;
+const EXPANDED_CLASS = `${PREFIX}-banner-header-expanded`;
 
-  this.classList.toggle('usa-banner-header-expanded');
-}
+const toggleBanner = function (event) {
+  event.preventDefault();
+  this.closest(HEADER).classList.toggle(EXPANDED_CLASS);
```
jseppi: Is `closest` polyfilled? I think it needs to be for IE. http://caniuse.com/element-closest

src/js/components/search.js
```diff
@@ -1,76 +1,74 @@
...
-
-function searchOff () {
-  if (activateDispatcher) {
-    activateDispatcher.off();
-  }
-  if (deactivateDispatcher) {
-    deactivateDispatcher.off();
-  }
-}
+const CLICK = ('ontouchstart' in document.documentElement)
```
jseppi: consider extracting `CLICK` since it is used in multiple places

src/js/components/search.js
```diff
@@ -1,76 +1,74 @@
...
-  searchForm.classList.remove(VISUALLY_HIDDEN);
-  var input = searchForm.querySelector('[type=search]');
-  if (input) {
-    input.focus();
+  if (active) {
+    const input = form.querySelector(INPUT);
+    if (input) {
+      input.focus();
+    }
+    const listener = ignore(
```
jseppi: explanatory comment would be good here

src/js/components/skipnav.js
```diff
@@ -0,0 +1,26 @@
...
+  // href already begins with '#'
+  const id = this.getAttribute('href').slice(1);
+  const target = document.getElementById(id);
+  if (target) {
+    target.setAttribute('tabindex', 0);
+    target.addEventListener('blur', once(event => {
+      target.setAttribute('tabindex', -1);
+    }));
+  } else {
+    // throw an error?
```
jseppi: Hrm, might want a single strategy for dealing cases where an expected element is missing. I see some other cases where an error is thrown: https://github.com/18F/web-design-standards/pull/1836/files#diff-6449513e1e61be79d366e960d8578ffbR16

I don't have a good feeling for what is "correct." On one hand, you would help developers by making problems obvious. On the other, it might break otherwise functional sites, which could be nasty for users.

src/js/components/skipnav.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+};
+
+module.exports = behavior({
+  'click': {
```
jseppi: Why not same `touchstart` || `click` as used in other components?

src/js/utils/behavior.js
shawnbot: Yeah, we're still using Node 6, so I was sticking to the subset of ES2015 supported natively. Let's tackle this in a future PR. :)

src/js/utils/behavior.js
shawnbot: Same, re: ES2015 and Node v6.

src/js/utils/behavior.js
shawnbot: Same, re: ES2015 and Node v6.

src/js/components/banner.js
shawnbot: Yeah, it's actually [polyfilled by receptor](https://github.com/shawnbot/receptor/blob/170bfca7877305e3d5dae38dc3f997101a4f195b/delegate/index.js#L2) via `element-closest`, which _also_ polyfills `matches()`. (Which I guess means that we're not necessarily playing nice in that respect...)

src/js/utils/behavior.js
jseppi: But the `es2015` babel config you're using with `babelify` will take care of this :)

src/js/components/skipnav.js
shawnbot: Yeah, I'm on the fence about this one. In other cases (e.g. with `aria-controls`, it doesn't feel like an error if no elements resolve because it's _theoretically_ possible to have a button that toggles just its own state. But in this case, the link should resolve, and if it doesn't, that's at least an accessibility error.

src/js/components/skipnav.js
shawnbot: Very good question :)

src/js/utils/behavior.js
shawnbot: That's true, but our mocha/jsdom tests `require()` the modules directly. ☹️

jseppi: 😬 Perhaps explore using `jest` and its babel runner: https://github.com/facebook/jest#using-babel

src/js/components/skipnav.js
shawnbot: After looking at this a bit more closely, I'm going to leave it as-is for now. Throwing an error seems like the right thing to do, but that would effectively stop the link from working, and I'm a little weary that just doing `link.getAttribute('href').slice(1)` is naive. I'd rather users just don't get the `tabindex` behavior because their link is malformed (and the link still works, possibly) than preventing the link from being followed altogether.

PR #1863, opened 2017-04-17
Add Fractal and embed container class to v1.1.0

shawnbot:
This will:

1. Replace the HTML templates listed in #1796 with Fractal
1. Add the `.usa-embed-container` class, introduced in #1835


.nvmrc
```diff
@@ -0,0 +1 @@
+6
```
shawnbot: ⚠️ The commit message says "v7", but this is clearly v6.

PR #1867, opened 2017-04-18
Update text max-width to ch (character) units

### LINKED ISSUE #1866: [Typography]: Use ch (character) units for text max-width
## Description

The current max-width for text is 53rem or 530px. This works for the default font Source Sans Pro size 17px but not for any other typeface someone may use, for example the alternate font recommendation of Merriweather size 15px, or any other style or size of font.

### Expected
53rem is a fixed width and not relative to the font used. Since the 53rem is a translation of the [66 character-per-line ideal](https://webtypography.net/2.1.2), the Standards should just use character units to declare this. It's more in line with what someone might expect and is more versatile for different font families and sizes. 

### Support
screen shot 2017-04-18 at 9 42 31 am
Support is pretty good: IE 9 and up supports it but is slightly different - it supports the width of the glyph, not its surrounding space, so might be slightly shorter. Since the actual recommendation is a range 50-75 characters, I think as long as it's within that range, readability should be OK.

maya:
## Description

Uses character units for text max-width. Removes unneeded `p:not(.usa-font-lead)` declaration. 

Fixes #1866.

### DIFF
--- a/src/stylesheets/core/_variables.scss
+++ b/src/stylesheets/core/_variables.scss
@@ -8,7 +8,7 @@
 // own partials, then @import '{path/to/uswds/}core/variables'.
 
 // Typography
-// Removing the !default from $em-base so we are not inheriting that 
+// Removing the !default from $em-base so we are not inheriting that
 // value from Bourbon.
 $em-base:             10px;
 $base-font-size:      rem(17px) !default;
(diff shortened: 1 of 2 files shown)


maya:
Bc the `variables.scss` file changed so much, this now conflicts. Will be easier to close this PR and open a new one later.

PR #1908, opened 2017-04-28
Add API shims for private 1.1 JS APIs

### LINKED ISSUE #1885: Decide how to communicate JS refactoring in our next release
I tried running a tool called [cracks](https://github.com/semantic-release/cracks) to see how our develop branch fares against the 1.0.0 test suite, which essentially tells us whether we're following semantic versioning. Currently, it fails right out of the box because several of our internal modules have moved (`toggle-field-mask` and `toggle-form-input`), and some have been removed entirely (such as the `dispatch` helper).

I'd like to make 1.2 backwards compatible so that we can release it ASAP for our users who need the event delegation fix, but I want to lay out all of the options for the team first. To be honest, I probably went overboard in refactoring _all_ of our JavaScript. Even though our users aren't (as far as I can tell, anyway) referencing many of our internal modules, there are some who use our component initializers directly, and those users will be hosed if they upgrade. **However**, the fixes are straightforward, and in most cases will involve removing code that's no longer relevant, such as for components that no longer need to be re-initialized when React modifies the DOM.

There are a few ways that we can go about this, sorted roughly in order of level of effort:

1. **Bump to 2.0.0**. This is by far the easiest because it means that all we'd really end up doing is bumping the version and documenting upgrade instructions from v1. We could also bundle bigger changes into this release, including CSS refactors... Or we could just get people used to the idea of semantic versioning by following 2.x with 3.0.0 in the not-too-distant future.

1. **Disregard private APIs in our commitment to semantic versioning**. This would be kind of rude, but if we can be clearer about what constitutes the public JavaScript API then we can tell users not to rely on anything outside of it being stable. Basically, we'd pledge only to support using _all_ the JS via `require('uswds')` until the internal API is "stable". This is easy, but requires the most communication because we'd still need to publish a thorough upgrade guide.

1. **Don't use `cracks`—at least for now**. For any module that's changed, rename the refactored `.js` file and introduce a v1 API "shim" that uses the new behavior under the hood. This adds a bit of bloat to our JS, but gives us better backwards compatibility. We can introduce `cracks` when we bump to v2.0.0.

1. **Use `cracks` to block breaking changes**. This means resurrecting modules that were removed, adding references to ones that have been renamed, and re-implementing some old APIs on top of the new behaviors. This will require the most work, but make it easier to tell if we're making breaking changes in the future.

shawnbot:
Fixes #1885. Here's what's changed:

* Some old "component" definitions that have been refactored into utility functions in v1.2 have been rewritten as backwards compatibility shims. For instance, the `components/toggle-field-mask` shim uses the `utils/toggle-field-mask` function to implement the 1.x component behavior.
* There is now a directory of 1.x API tests from the 1.0.0 release branch that are run in addition to the current version tests. The ones that rely on HTML templates have been updated to reference the current version templates, leaving a couple of cracks that we deal with separately...
* We're now using [cracks](https://github.com/semantic-release/cracks) to run the current code against the most recent release's unit tests (based on the semantic version). This means that any 1.x release will run the code against the most recent release's (`^1`) test suite, and will produce an error if any tests fail. When we bump to 2.0.0 we can remove all of the backwards compatibility shims and 1.x tests, and keep the `cracks` test in place to ensure that future 2.x releases don't introduce breaking API changes.

    `npm run cracks` performs this test.
* The `cracks` test is run after `npm test` on Circle CI.
* There is now a `mocha` npm script that makes it easier to run mocha with the shared options, e.g.

    ```sh
    npm run mocha -- spec/unit/**/*.spec.js
    npm run mocha -- spec/1.x/*.spec.js
    ```

* There's also a `spec` npm script that runs only the mocha specs, which should be the equivalent of running `gulp test`:

    ```sh
    npm run spec
    ```


src/js/components/accordion.js
```diff
@@ -43,18 +42,30 @@ const toggleButton = (button, expanded) => {
 };
 
 /**
+ * @param {HTMLButtonElement} button
+ * @return {boolean} true
+ */
+const showButton = button => toggleButton(button, true);
```
msecret: Are we OK using arrow functions for our browser compatibility?

src/js/components/accordion.js
```diff
@@ -63,15 +74,39 @@ module.exports = behavior({
...
+ *
+ * module.exports = behavior({...});
+ */
+const Accordion = function (root) {
+  this.root = root;
+  accordion.on(this.root);
+};
+
+// copy all of the behavior methods and props to Accordion
+const assign = require('object-assign');
```
msecret: There's also this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

But I assume it's not compatible enough with other browsers?

src/js/components/toggle-form-input.js
```diff
@@ -0,0 +1,17 @@
+// TODO: remove this file in 2.0
+const toggleFormInput = require('../utils/toggle-form-input');
+const resolveIdRefs = require('resolve-id-refs');
+
+module.exports = function (control, showText, hideText) {
+  /*
+  if (showText && hideText) {
```
msecret: Should this code be taken out?

spec/1.x/dispatch.spec.js
```diff
@@ -0,0 +1,81 @@
...
+  after(function () {
+    global.document = oldDocument;
+    global.window = document.defaultView;
+  });
+  */
+
+  it('attaches an event to an element', function () {
+    var element = document.body;
+    var flag = false;
+    var listener = function (e) { flag = true; };
```
msecret: This could probably be done a little cleaner with a library like sinon.js, where you can use spys. But it's fine for now.

src/js/components/accordion.js
shawnbot: Yep, we're compiling to ES5 with Babel.

src/js/components/accordion.js
shawnbot: Yeah, `object-assign` is a ponyfill.

shawnbot: Addendum: [browser support](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility) is pretty good, but doesn't cover IE and some mobile browsers.

src/js/components/toggle-form-input.js
shawnbot: Nope, this is the v1.x shim :)

spec/1.x/dispatch.spec.js
shawnbot: Agreed, though we'll be jettisoning the whole `dispatch` utility in v2.0.0.

Issue #1879, opened 2017-04-24
Update JS bundling guidance post-refactor


shawnbot:
Now that #1836 is merged, we need to update our documentation for how to incorporate our JavaScript into browser bundles running up to the next release. The first thing we need to do:

- [x] #1885 **Decide whether the JS refactor bumps to v1.2.0 or 2.0.0**

Then, we need to:

- [x] Explain what will need to change **if** users are requiring specific submodules, e.g. `require('uswds/src/js/components/navigation')`.
- [x] Point people at resources for how to use Babel, specifically tools like [webpack's babel-engine-plugin](https://github.com/SamVerschueren/babel-engine-plugin), which will do the Right Thing with `uswds` imports.
- [x] ~Explain the new `main` and `browser` fields in `package.json`, and what they mean for people using different tools, namely browserify and webpack.~ We don't provide a `browser` field, so this is moot.

shawnbot:
This will need to be done as part of #1915.

shawnbot:
One thing we could here might be to show diffs between v1.1 and v1.2 bundling examples, maybe as part of #1923?

shawnbot:
Support for [webpack babel-engine-plugin](https://github.com/SamVerschueren/babel-engine-plugin) is done in #1940.

shawnbot:
This is basically done. Remaining documentation tasks moved to #1951.

PR #1923, opened 2017-05-12
Bundling/customization examples

shawnbot:
This is part of the fix for #1879, and includes standalone "project" directories that demonstrate how to bundle the Standards JS and CSS with different tools:

- [x] JS: Browserify
- [x] JS: Webpack 2
- [x] CSS: node-sass
- [ ] ~CSS: Ruby Sass~


donjo:
@shawnbot I think the directory should probably have a brief readme that says how to use the examples. I tried my best to put on my beginner's mind hat and came out with a few questions that a readme in the example directory could probably answer. I don't know how much we want to assume about js/sass knowledge but I'd probably lean towards "this is my first project getting third party code from a package manager" as the assumed experience level. 

Open questions/comments/suggestions: 
- What do I do with these examples? do I copy them? create them in the root of my project? 
- Where do I run `npm install`? Does it matter which directory I'm in? 
- What else goes in `main.js`? (maybe some commented out examples? suggesting structure for including custom JS you write?) 
- A short explainer on how `package.json` works, specifically the versioning component.

donjo:
It's also possible that a more involved how-to could be part of a training we develop or instructional content we generate while doing work for an agency partner. So maybe these READMEs don't need to be more detailed than they already are.

PR #1925, opened 2017-05-15
updating lists and alerts when used in content body

### LINKED ISSUE #1924: Clean up performance examples
In order to have clear information for examples in the performance section, the design of the examples should be changed.

- [ ] The width of the example sections matches up with the rest of the content
- [ ] There's no title to the examples, just the content

donjo:
### [:eyes: Preview on Federalist](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/list-alerts-formatting/components/detail/alerts.html)

fixes #1924 

These were mostly done to fix things on the new performance pages in the docs. The added alert class can be used for alerts that show up in the middle of paragraphs (as opposed to spanning the top of a page).

This change could be considered breaking for users who intend for lists in `usa-content` to span the entire page so we may want to move it to the docs codebase until it can be packaged with a release.

### DIFF
--- a/src/components/test/kitchen-sink.njk
+++ b/src/components/test/kitchen-sink.njk
@@ -16,6 +16,22 @@
 
       <p class="usa-font-lead">{{ page.lead }}</p>
 
+      <p class="usa-content">
+        {{ page.body }}
+        
+          <ul>
+          {% for list in page.lists %}
+            <li>{{ list }}</li>
+          {% endfor %}
+          </ul>
+
+          <ol>
+          {% for list in page.lists %}
+            <li>{{ list }}</li>
+          {% endfor %}
+        </ol>
+      </p>
+
       {% for spec in inner %}
       <section id="{{ spec.id }}">
         <h2>{{ spec.title }}</h2>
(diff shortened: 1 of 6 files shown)


donjo:
@shawnbot yep. There is another issue related to changing the text width var from `rem` to `ch` but I think we can make and test that decision separately

donjo:
@shawnbot added some previews for lists within `usa-content` on the [Kitchen Sink](https://federalist.fr.cloud.gov/preview/18f/web-design-standards/list-alerts-formatting/components/detail/kitchen-sink.html) page. 

Not sure if there are any other cases (beyond accordions) I'm not thinking of where a list would exist inside the class for paragraph text. If there are, maybe there is a better approach than adding 2 more exceptions to that rule.

src/components/test/kitchen-sink.njk
```diff
@@ -16,6 +16,22 @@
 
       <p class="usa-font-lead">{{ page.lead }}</p>
 
+      <p class="usa-content">
+        {{ page.body }}
+        
+          <ul>
```
shawnbot: The indentation is one level too deep from here until the `</ol>`.

PR #1935, opened 2017-05-22
Merge develop => release-1.2.0

shawnbot:
This PR merges work that we've committed to `develop` since first cutting our v1.2 release branch. ~It looks like there are some merge conflicts, so we'll need to handle those first.~


amm365dev:
M m

PR #1953, opened 2017-05-30
Alerts: Fix color contrast A11y of links

sawyerh:
This addresses https://github.com/18F/web-design-standards/issues/1907, changing the link color within Alerts to the darker variation of the `primary` color, which meets WCAG2.0 AA.

image


donjo:
@sawyerh thanks for catching this! Could you change the link href to `href="javascript:void(0);"` as that's the pattern we're trying to use for blank links across the standards site?

sawyerh:
@donjo Ah got it — updated! 👍

PR #1988, opened 2017-06-22
Fixing search role div around search forms

### LINKED ISSUE #1654: Missing ARIA Landmark role="search"
**Issue**:
The page is missing an ARIA landmark role of "search" for the Search area(s) of the page. Landmark roles assist screen reader users by conveying where they are on the page and can be used as an alternate way to navigate the page.

**Solution**:
Add ARIA role="search" for the Search regions of the page. The Search role must surround the visible label, the search field, and the search button.

Refer to: http://www.w3.org/TR/wai-aria/roles#search

Example:
```
<form action="…">
  <div role="search">
    <input type="search" aria-label="Search"> <button type="submit">Search</button>
  </div>
</form>
```

donjo:
fixes #1654 

The div that gives a search field `role="search"` was inside the `<form>` element but it should wrap the form as referenced in #1654

### DIFF
--- a/spec/unit/navigation/template.html
+++ b/spec/unit/navigation/template.html
@@ -44,15 +44,15 @@
         </li>
       </ul>
       <div class="usa-nav-secondary">
-        <form class="usa-search usa-search-small js-search-form">
-          <div role="search">
+        <div role="search">
+          <form class="usa-search usa-search-small js-search-form">
             <label class="usa-sr-only" for="search-field-small">Search small</label>
             <input id="search-field-small" type="search" name="search">
             <button type="submit">
               <span class="usa-sr-only">Search</span>
             </button>
-          </div>
-        </form>
+          </form>
+        </div>
         <ul class="usa-unstyled-list usa-nav-secondary-links">
           <li class="js-search-button-container">
             <button class="usa-header-search-button js-search-button">Search</button>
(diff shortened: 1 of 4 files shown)


toolness:
Hmm, actually, I'm confused... The description for #1654 doesn't mention anything about wrapping the `<form>` in a `<div role="search">`, and in fact the example in the description has the opposite:

```html
<form action="…">
  <div role="search">
    <input type="search" aria-label="Search"> <button type="submit">Search</button>
  </div>
</form>
```

The [ARIA specification](https://www.w3.org/TR/wai-aria/roles#search) doesn't seem to mention anything that contradicts this, either.

You seem to have mentioned that everything looks OK to you in https://github.com/18F/web-design-standards/issues/1654#issuecomment-309852945 so I'm confused as to what prompted this change. Are there specific ATs that don't work properly with things the way they are?

donjo:
Ahh sorry I should have updated the issue. I had reached out to @davidstenger on slack and he provided further clarification. The example given in the original issue was an example of an area that needed to be changed _not_ the correct way to do it. 

Here's what he had to say:

> No problem at all. Looks like the role=“search” has been added but in the wrong spot. It should come before the `<form>` example: 
```
<div role=“search”>
  <form method=“get” class=“search-form” action=“https://www.search.com“>
    <label>
        <span class=“screen-reader-text”>Search for:</span>
        <input type=“search” class=“form-control” placeholder=“Search…” value=“” name=“s” title=“Search for:“>
    </label>
    <input type=“submit” class=“btn btn-primary” value=“Search”>
    <!-- debug: searchform.php -->
  </form>
</div>
```

toolness:
Ohh got it! Yeah, in that case this looks great!

Issue #1432, opened 2016-08-30
Document "validator" component behavior


juliaelman:
Refs https://github.com/18F/web-design-standards/issues/1342

shawnbot:
This is _mostly_ done in #1836. The [`validator` "component"](https://github.com/18F/web-design-standards/blob/develop/src/js/components/validator.js) needs to be described and documented, though.

shawnbot:
Here's a narrative description of how the [validator component](https://github.com/18F/web-design-standards/blob/develop/src/js/components/validator.js) works:

### Behavior
Given an `<input>` element with a `data-validation-element` attribute:

* Whenever the input is changed (on `keyup` or `change`), [validate the input](https://github.com/18F/web-design-standards/blob/develop/src/js/utils/validate-input.js) using the following steps:
    * Determine the validation "checklist" by getting the `data-validation-element` attribute, which is interpreted as either the `id` of another element in the document or a CSS selector if it begins with `#`.
    * If the specified validation checklist element doesn't exist, throw an error. ✖️ 
    * Iterate over the keys of the input element's [dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset), and for each that begins with `'validate'`:
        * Lowercase the key to get everything after the `data-validate-` prefix of the attribute name to get the validator name.
        * Form a CSS selector from the validator name, e.g. `[data-validator="numeric"]`, and query the checklist for the corresponding element.
        * If the corresponding checklist element doesn't exist, throw an error. ✖️ 
        * Test the `data-validate-*` attribute value as a regular expression against the input's value.
            * If the regular expression matches the input value, set `aria-checked="true"` and add the `usa-checklist-checked` class.
            * Otherwise, set `aria-checked="false"` and remove the `usa-checklist-checked` class.

### Examples
See the [unit test template](https://github.com/18F/web-design-standards/blob/develop/spec/unit/validator/template.html) for a reference implementation. On the site, the [password reset form](https://standards.usa.gov/components/form-templates/#password-reset-form) demonstrates it interactively.

### Styles
The CSS that controls the display of the checklist [lives in `src/stylesheets/components/_forms.scss`](https://github.com/18F/web-design-standards/blob/f6029a6d233a54f28382d6fc0ef79bde9d46176f/src/stylesheets/components/_forms.scss#L133-L164).

donjo:
Tagging @toolness for guidance on where to put this documentation

toolness:
[...]

PR #1992, opened 2017-06-24
Actually use validator component.

### LINKED ISSUE #1990: Reset password example fails w/ "Cannot read property 'validationElement' of undefined"
When investigating #1432 and trying out the [password reset form](https://standards.usa.gov/components/form-templates/#password-reset-form), I noticed that it wasn't working, and the following was logged to my console whenever I typed a character in the "new password" field:

```
Uncaught TypeError: Cannot read property 'validationElement' of undefined
```

toolness:
This is an attempt to fix #1990.

Because I'm very unfamiliar with the USWDS codebase, though, I'm _probably_ doing something wrong here.

It seems #1990 was occurring because `components/password.js` was attaching the `validate` function to a key event handler; it was being passed a keyboard event, but expected an HTML element, which is why the exception was being thrown.

Tests didn't catch this because the tests were hooking up the validator component to the sample template, though. And that was working fine.

So, this changes things so that the validation aspect of password fields is handled by the validator component, *not* the password component. The password component *is* still needed for toggling password visibility on forms, though, so I didn't get rid of it.

## However...

However, I *think* this means that the [`js-validate_password`](https://github.com/18F/web-design-standards/search?utf8=%E2%9C%93&q=js-validate_password&type=Code) class may no longer be needed, since JS code no longer needs it.  (The validation logic is now handled by the validator component, which is triggered by the `input[data-validation-element]` selector instead.)

If that's the case, it would require also updating the example(s) that use this component in the documentation, and maybe incrementing the minor version of USWDS.

## But...

What confuses me is why the validator component wasn't already "enabled" in the first place.  Because it wasn't included in `components/index.js`, its event listener wasn't being added to `document.body` in `start.js`, which meant that _documenting_ the validator component as per #1432 would be futile, since we'd effectively be documenting dead code. This PR fixes that, but I'm puzzled as to why the code was like that in the first place.

Perhaps it's something that should have been done as part of #1836, but might have been overlooked due to the size and complexity of that PR?

### DIFF
--- a/src/js/components/index.js
+++ b/src/js/components/index.js
@@ -6,5 +6,6 @@ module.exports = {
   password:   require('./password'),
   search:     require('./search'),
   skipnav:    require('./skipnav'),
+  validator:  require('./validator'),
 };
 
--- a/src/js/components/password.js
+++ b/src/js/components/password.js
@@ -1,6 +1,5 @@
 'use strict';
 const behavior = require('../utils/behavior');
-const validate = require('../utils/validate-input');
 const toggleFormInput = require('../utils/toggle-form-input');
 
 const CLICK = require('../events').CLICK;
@@ -16,8 +15,5 @@ const toggle = function (event) {
 module.exports = behavior({
   [ CLICK ]: {
     [ LINK ]: toggle,
-  },
-  'keyup change': {
-    '.js-validate_password': validate,
   },
 });


toolness:
Ok, @donjo messaged me on Slack and said he couldn't find any mention of the validator's disabled-ness being intentional, so we're going to go ahead and merge it.  I'll file a separate issue to deal with removing the `js-validate_password` class.

Issue #1929, opened 2017-05-16
Radio button and checkbox issues


maya:
1. When I placed the radio buttons and checkboxes in a three column layout, I experienced some of the form elements overlapping with other form elements in the adjacent column. This was not good bc I couldn't select the intended form element.
2. There is not a pointer style cursor (hand) when I hover over the elements, so it's not obvious where someone can click.
3. The input does not have the proper dimensions, since the label is stacked above the input and is clickable, it's generally not an issue, but this caused some issues (#1).
<img width="520" alt="screen shot 2017-05-15 at 5 18 05 pm" src="https://cloud.githubusercontent.com/assets/5249443/26084653/81ef7faa-3992-11e7-954e-a1869ccdf0fa.png">

I was able to fix these issues with the following code: https://github.com/18F/clinical-trials-prototype/blob/clean/_sass/_custom.scss#L183-L194

toolness:
@maya I think the link you provided to your custom fixes is now out-of-date because that file has been changed since you created the link... Is this the commit that fixed things on your end?

https://github.com/18F/clinical-trials-prototype/commit/9bc2d20a697968eacdd94a1d12f09c38f453910e

donjo:
Fixed in #2002

PR #2002, opened 2017-06-29
Fix multi-column checkbox/radio issues

toolness:
This is an attempt at fixing #1929.

I've created a fractal page under "Test", adjacent to "Kitchen sink", which attempts to reproduce the issue.

### DIFF
--- a/src/components/test/multi-column-checkboxes.njk
+++ b/src/components/test/multi-column-checkboxes.njk
@@ -1,0 +1,11 @@
+<main class="usa-grid usa-section">
+  <div class="usa-width-one-third">
+    {% include '07-form/controls/checkboxes.njk' %}
+  </div>
+  <div class="usa-width-one-third">
+    {% include '07-form/controls/checkboxes.njk' %}
+  </div>
+  <div class="usa-width-one-third">
+    {% include '07-form/controls/checkboxes.njk' %}
+  </div>
+</main>
--- a/src/components/test/multi-column-checkboxes.njk
+++ b/src/components/test/multi-column-checkboxes.njk
@@ -1,11 +1,17 @@
 <main class="usa-grid usa-section">
-  <div class="usa-width-one-third">
-    {% include '07-form/controls/checkboxes.njk' %}
-  </div>
-  <div class="usa-width-one-third">
-    {% include '07-form/controls/checkboxes.njk' %}
-  </div>
-  <div class="usa-width-one-third">
-    {% include '07-form/controls/checkboxes.njk' %}
-  </div>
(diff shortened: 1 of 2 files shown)


toolness:
@maya, can you take a look at the [fractal example](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/multi-column-checkboxes/components/detail/multi-column-checkboxes.html) I added in this PR?  Does it reproduce the same behavior you were seeing?

Also, does the HTML template in this PR seem OK to you?  I wanted to use a list for the options, and I notice your [fix](https://github.com/18F/clinical-trials-prototype/commit/9bc2d20a697968eacdd94a1d12f09c38f453910e) for your own code used `li` in its CSS, but I'm unclear on how to create a multi-column list with USWDS, since a `<ul>` can only contain `<li>` elements as children, i.e. individual `<li>`s can't be wrapped in `<div class="usa-width-one-third">`. Nor does using `<li class="usa-width-one-third">` seem to do the trick either...

toolness:
Wait, I am really confused. At first I assumed this bug only occurred when the individual checkboxes of the same `<fieldset>` were in multiple columns, but I am having a hard time getting it to work.

toolness:
Is #1929 even specific to a three-column layout?  I seem to be able to produce the same behavior with just a one-column layout... Though that said, the only "bug" I've been able to produce on my end is that the bounding box of the very last list item extends kind of far vertically, e.g. I can click almost a full line below the last option and it will get toggled.  But the other list elements seem to be triggered fine...

toolness:
Oy I am having a hard time reproducing the bug.  @maya can you provide an HTML snippet that reproduces it?

toolness:
Um wait, I think I may have actually reproduced it and fixed it.

I decided to take an approach of using the `sr-only()` mixin on checkboxes and radios, as it reduces the complexity of our CSS and also makes sense semantically, since (as I understand it) all actual visual styling of our checkboxes and radios is coming from our rules on the `<label>` .

The only remaining thing I don't understand from @maya's fix in https://github.com/18F/clinical-trials-prototype/commit/9bc2d20a697968eacdd94a1d12f09c38f453910e is that she included the following rule:

```sass
.usa-fieldset-inputs {
  li {
    position: relative;
  }
}
```

I'm not sure I understand the purpose of this part, so I haven't included it in my fix... @maya do you remember why you included it in your fix?

maya:
@toolness yeah it looks like it was an issue with the `left` value causing this issue. Sorry this issue was confusing 😬.

toolness:
Hehe no worries! I guess this PR is ready for review...

PR #2006, opened 2017-06-30
Remove deprecated Bourbon mixins and remove warnings

maya:
This removes any instance of the code that uses deprecated bourbon mixins. Seeing all those warnings is not a great experience. 

Fixes: https://github.com/18F/web-design-standards/issues/1904
Related: https://github.com/18F/web-design-standards/pull/2003

cc @toolness

### DIFF
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -38,7 +38,9 @@
 
 .usa-nav {
   $sliding-panel-width: 26rem;
+
   @include position(fixed, 0 0 0 auto);
+
   background: $color-white;
   border-left: 1px solid $color-gray-light;
   border-right: 0;
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -38,10 +38,7 @@
 
 .usa-nav {
   $sliding-panel-width: 26rem;
-
   @include position(fixed, 0 0 0 auto);
-  @include transform(translateX($sliding-panel-width));
-
   background: $color-white;
   border-left: 1px solid $color-gray-light;
   border-right: 0;
@@ -49,24 +46,24 @@
   flex-direction: column;
   overflow-y: auto;
   padding: 2rem;
+  transform: translateX($sliding-panel-width);
   width: $sliding-panel-width;
   z-index: $z-index-nav;
 
   @include media($nav-width) {
     @include padding(5rem 0 0 null);
-    @include transform(translateX(0));
-
     border-left: none;
     display: block;
     float: right;
     overflow-y: visible;
     position: relative;
+    transform: translateX(0);
     width: auto;
   }
 
   &.is-visible {
-    @include transform(translateX(0));
-    @include transition(all 0.3s ease-in-out);
+    transform: translateX(0);
+    transition: all 0.3s ease-in-out;
   }
 
   nav {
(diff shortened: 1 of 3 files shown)


maya:
Does the USWDS use autoprefixer? If it does, browser prefixes will be added in and no need for some of these Bourbon mixins. If not, you may want to check the browsers you support and add the prefixes in manually.

toolness:
Woot thanks @maya!!!  This is awesome, sorry we haven't looked at it yet... will soon!!

toolness:
According to our [gulp sass task](https://github.com/18F/web-design-standards/blob/5cf8e96504036d479fcc5c71e1ac0b11d22b1662/config/gulp/sass.js), it looks like we do indeed use autoprefixer!

thisisdano:
I have one general comment on this that I'd raised in Slack a while back but will recapitulate here:

>However, if we don't use the `rem()` function and write in straight rem, we'll have an issue if the base font changes from 10px. But if a site has a different base font size (say 15px), we _should_ be able to just adjust the base font size variable in our default overrides to the site's base (in this case, 15px) and the USWDS rem-sized rules should just update and the site should display the same. (A long way of saying that the base font size shouldn't matter to the display of the USWDS styles.) But, if our own variables aren't calculated using a function like `rem()` we're gonna see the problem you point out. So, if we don't use calculated variables for our sizing, we really should.

The main point is that removing a function like `rem()` means that sizes previously calculated with that function are now hard-coded and more susceptible to being damaged by a change to the base font size.

Perhaps this is OK as a temporary fix, to address what I consider a mistake that the Bourbon team is making. But I think that we want to use calculated values _more broadly_ in the long term, and not be writing hardcoded rem.

maya:
@thisisdano issues around base font size came up in Design Systems Slack: https://design-systems.slack.com/archives/C0ENULNQZ/p1498750113039260

This PR doesn't remove the rem function itself as we're still using an old version of Bourbon, but just the uses of it that trigger the warnings, and don't have an effect on the system as the function just ends up converting it to the rems that this changes it to anyway. 

Not saying we should do this, but GDS made a case to move from rem to px for browsers consistency (rem does some weird things in different browsers) which would also resolve the base font size issues.

thisisdano:
Thanks @maya — that's interesting context. rem to px feels like a big philosophical shift, and one that feels good to me. It would make lots of things seem (be?) more consistent, understandable, and modular (i.e. less dependent). I still have all ye olde concerns about hardcoded rem, but I'm confident that this is a temporary thing.

toolness:
Woot!

@maya it looks like there's [one linter error](https://circleci.com/gh/18F/web-design-standards/1545?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) that's preventing CircleCI from being OK with this PR:

> src/stylesheets/components/_navigation.scss
> 41:3  ✖  Expected empty line before at-rule   at-rule-empty-line-before

One that's fixed, I think this PR is ready to merge!

src/stylesheets/components/_navigation.scss
```diff
@@ -38,7 +38,9 @@
 
 .usa-nav {
   $sliding-panel-width: 26rem;
+
   @include position(fixed, 0 0 0 auto);
+
```
maya: There should not be spaces here per the 18F front end guide which this project follows: https://frontend.18f.gov/#order @msecret @toolness

Issue #2013, opened 2017-07-06
Accordion buttons can't be used as surfaces for touch-based scrolling


toolness:
I noticed when browsing the Standards site on my phone that, while accordion buttons are extremely responsive to touch and open instantly when touched, they can't be used as scrolling surfaces.  This could cause anxiety or frustration for users on pages with lots of accordions, since the screen could consist of mostly collapsed accordion buttons at any given time, thus providing very little room for the user to actually scroll *past* the accordions.

If this behavior is undesirable, I think changing it might require a bit of a trade-off though, in that the accordions may need to become a bit less responsive to touch, e.g. waiting a couple hundred ms to see if the touch turns into a scroll before collapsing/expanding the accordion. Personally I think that trade-off is worth it, though, as intending to scroll but accidentally toggling an accordion instead can be quite frustrating, but I'd like to hear other opinions.

toolness:
Fixed by #2015!

PR #2015, opened 2017-07-06
Bind only to click events, not touchstart.

### LINKED ISSUE #1993: Accordions don't work via mouse on devices that support both touch and mouse
https://standards.usa.gov/components/form-templates/

The code accordions aren't expanding anymore - I know there were some new releases, this seems to have broken since

toolness:
This fixes #1993 and #2013 by binding event listeners to `click` instead of `touchstart`, even on devices that support touch.

A full rationale for the change is at https://github.com/18F/web-design-standards/issues/1993#issuecomment-313410624, though I also included a synopsis of it in the comments in this PR.

I've tested this out on multiple iOS devices, Firefox and Chrome on Android, and multiple desktop browsers and it seems to work fine.

### DIFF
--- a/src/js/events.js
+++ b/src/js/events.js
@@ -1,5 +1,15 @@
 module.exports = {
-  CLICK: ('ontouchstart' in document.documentElement)
-    ? 'touchstart'
-    : 'click',
+  // This used to be conditionally dependent on whether the
+  // browser supported touch events; if it did, `CLICK` was set to
+  // `touchstart`.  However, this had downsides:
+  //
+  // * It pre-empted mobile browsers' default behavior of detecting
+  //   whether a touch turned into a scroll, thereby preventing
+  //   users from using some of our components as scroll surfaces.
+  //
+  // * Some devices, such as the Microsoft Surface Pro, support *both*
+  //   touch and clicks. This meant the conditional effectively dropped
+  //   support for the user's mouse, frustrating users who preferred
+  //   it on those systems.
+  CLICK: 'click',
 };


toolness:
@paultag and @mjoyce91, can you try the website at the following URL and let me know if the accordions work on it?

http://atul-uswds-test.s3-website-us-east-1.amazonaws.com/

Note that this URL might go down sometime in the next few days because it's in a temporary location that gets cleared once per week, but if that happens I'll put it back up monday morning.

JJFranko:
@toolness - Tested in Chrome V59 on Surface Pro 4 : **Appears to be a success**. 
Accordions and Menu Button (@ width < 950px) work for Mouse click and Trackpad tap and Screen Touch events.

mjoyce91:
@toolness - I'll have access to my computer this evening and will provide an update.

toolness:
Awesome, thanks for taking a look at this everyone!

@msecret I think this is ready for review.

toolness:
I'm wary of binding to touchstart, though, because of #2013--the accordion buttons have a tiny bit of latency on touchscreens when they bind to `click`, but the advantage is that they can be used as scrolling surfaces, which is important when the page consists mostly of accordions.

That said, I may be wrong, but I *think* that `touchstart` pre-empts `click` events from ever being synthesized as a result of touch-based actions, so binding to both events *should* be OK, if we want to fix #1993 without addressing #2013.

Since you've approved this PR, though, I'm going to merge this for now since we've verified with the affected users that this fix works well, and we can re-address in a separate issue/PR if we want to change things further!

PR #2017, opened 2017-07-07
Collapse navigation on nav link click

### LINKED ISSUE #1962: Dropdown menus with internal links don't collapse after link is selected
Hello! @line47 and I were working on [methods.18f.gov](https://methods.18f.gov/) yesterday and trying to solve a problem whereby selecting a link in the header drop-down doesn't collapse the drop-down after selection. This isn't a problem when the header link is on a different page, since the page reload effectively "collapses" the drop-down, but when the link is to an internal link on the same page (e.g. `#design-studio`), the location is navigated-to by the browser but the dropdown doesn't un-collapse.

For the short term, we've implemented a hacky fix in https://github.com/18F/methods/pull/241 whereby we manually toggle `.usa-nav-primary [aria-expanded=true]` and `.usa-nav-primary [aria-hidden=false]` to re-collapse the menu on `hashchange`. However, this may be fragile going forward, so we'd rather use a more well-supported/documented solution.

Is there already a recommended fix for this that we missed?  Or can I somehow help contribute docs and/or code that would help?

toolness:
This attempts to fix #1962.

## To do

- [x] Ensure that focus works properly when a link is clicked.
- [x] Consider adding another example to the fractal site which includes anchor links, and ensure it works properly.
- [x] Add more tests.
  Tests can be run with `mocha --opts spec/mocha.opts spec/unit/navigation/navigation.spec.js`.

### DIFF
--- a/src/js/components/navigation.js
+++ b/src/js/components/navigation.js
@@ -2,12 +2,14 @@
 const behavior = require('../utils/behavior');
 const forEach = require('array-foreach');
 const select = require('../utils/select');
+const accordion = require('./accordion');
 
 const CLICK = require('../events').CLICK;
 const PREFIX = require('../config').prefix;
 
 const CONTEXT = 'header';
 const NAV = `.${PREFIX}-nav`;
+const NAV_LINKS = `${NAV} a`;
 const OPENERS = `.${PREFIX}-menu-btn`;
 const CLOSE_BUTTON = `.${PREFIX}-nav-close`;
 const OVERLAY = `.${PREFIX}-overlay`;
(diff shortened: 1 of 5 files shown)


src/js/components/navigation.js
```diff
@@ -42,6 +44,23 @@ const navigation = behavior({
...
+    [ NAV_LINKS ]: function() {
+      // A navigation link has been clicked! We want to collapse any
+      // hierarchical navigation UI it's a part of, so that the user
+      // can focus on whatever they've just selected.
+
+      // Some navigation links are inside accordions; when they're
+      // clicked, we want to collapse those accordions.
+      const acc = this.closest(accordion.ACCORDION);
+      if (acc) {
+        accordion.getButtons(acc).forEach(btn => accordion.hide(btn));
```
toolness: Is it actually possible for a navigation to *not* contain an accordion?  I thought it was, but looking through the examples in our fractal site, I'm actually not sure anymore... If it's not possible, then I guess we might as well get rid of this conditional, maybe.

toolness:
Ok, I think this is ready for feedback!

src/js/components/navigation.js
msecret: I'd keep the conditional for the future, it seems like it would make sense for me that it would come up.

toolness:
Thanks! Yeah, I was originally looking for some kind of `NAV_CLICK` event but I couldn't find one, so I just used the most straightforward approach I could think of. At least the tests are written in such a way that if we decide to refactor later to use such an event, we hopefully won't need to rewrite the tests.

That said, I do like the idea of adding such events, and documenting them nicely, the way [Bootstrap documents its JS events](http://getbootstrap.com/javascript/) (except sans jQuery).

PR #2020, opened 2017-07-11
Mobile nav: allow for a mobile-specific navigation by hiding it on desktop

brockfanning:
## Description

The standards.usa.gov site is using a class called "sidenav-mobile" to allow a mobile-specific navigation that gets hidden on desktop. This approach would be useful to other sites using the standards, but the CSS that hides the menu on desktop is not part of the library (it's specific to standards.usa.gov). If that CSS could be added to the library, then sites would be able to use this approach simply by using a certain class in their markup, instead of needing to add custom CSS.

In this PR I used the class: usa-nav-mobile

EDIT after 2nd commit: I realized that in order to be useful, another class would be needed to target the desktop-only menu. So I've added the corresponding code for a usa-nav-desktop class.

## Additional information

If you think this is worth adding to the library, there should also probably be some documentation of the use of this class, so that it doesn't end up as a hidden feature that nobody knows about. I could use some suggestions about the best place for that, and am happy to add to this PR as needed.

### DIFF
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -103,6 +103,12 @@
     display: none;
   }
 }
+.usa-nav-desktop {
+  display: none;
+  @include media($nav-width) {
+    display: block;
+  }
+}
 
 // Primary navigation ------------- //
 
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -96,6 +96,14 @@
   }
 }
 
+// Mobile-only navigation --------- //
+
+.usa-nav-mobile {
+  @include media($nav-width) {
+    display: none;
+  }
+}
+
 // Primary navigation ------------- //
 
 .usa-nav-primary {


AndreaSigz:
Thanks @brockfanning. @toolness or @donjo any thoughts?

toolness:
Thanks @brockfanning! Sorry for the long response time on this one, and thanks for explaining the context behind this on the CMS call last week.

It strikes me that this could potentially be used beyond the concept of pure navigation, to be something that a site can generically use to make *any* content mobile-only or desktop-only. What do you think of renaming the classes to e.g. `usa-desktop-only` and `usa-mobile-only`? Hmm, I guess the only strange thing about that is that the breakpoint depends on a variable called `$nav-width`, which does suggest the concept of *navigation* specifically... Hmmm.

Actually, so others have context of how you're using this, can you post a screenshot or two of how you're using this class in your Drupal theme/module?

brockfanning:
@toolness @AndreaSigz Sorry for the long delay. I'm not particular about the variable names, though the use of $nav-width does make it seem pretty navigation-specific.

Here are a few screenshots that hopefully help illustrate the use-case. In each of these screenshots you can see there are 2 divs that contains menus: "usa-nav-mobile" and "usa-nav-desktop". This allows for the 2 menus to have distinct content. The point of this PR is to make it easy that the appropriate menu displays at the appropriate device width.

First, here's a shot with a narrower width. In this one, in the dev tools I'm highlighting the usa-nav-mobile div.
vfsc--pr-2020--usa-nav-mobile-1

Next here's the same width, but in the dev tools I'm highlighting the usa-nav-desktop div. Note it has a "display:none" rule.
vfsc--pr-2020--usa-nav-mobile-2

Next here's a shot with a wider width. In this one, in the dev tools I'm still highlighting the usa-nav-desktop, which now has a "display:block" rule.
vfsc--pr-2020--usa-nav-desktop-1

Finally here's a shot with the same wider width, but in the dev tools I'm highlighting the usa-nav-mobile div, which now has a "display:none" rule.
vfsc--pr-2020--usa-nav-desktop-2

AndreaSigz:
No worries @brockfanning for the delay! Thanks for getting back to us

maya:
Hi @brockfanning, thanks for trying to take a stab at this idea. 

You're right that the standards site is using a mobile-specific navigation, but that's because when the site first launched, it was before we developed the navigation component. Also, the Standards site's mobile navigation is a little different from what we're currently offering bc the contents of the page appear in a side navigation rather than a dropdown from the header navigation. So this will need to be changed at some point to either add this functionality into the USWDS package or use the provided headers.

The way our CSS is written is mobile-first, so we don't add mobile-specific classes to our codebase, rather start from mobile sizes and expand out to desktop sizes when it's needed. Taking your suggested code, I think it can be simplified to:

```
// Navigation display helper classes --------- //

.usa-nav-mobile {
  display: none;
}

.usa-nav-desktop {
  @include media($nav-width) {
    display: block;
  }
}
```

*Note: we'll need to discuss with the team around adding this helper classes in, probably when we take another look at the headers. Will let you know when it's on our roadmap.

brockfanning:
@maya Unless I'm misunderstanding, I don't think my PR could be simplified to that snippet. The point of it is to allow for completely different markup to display on either side of the $nav-width breakpoint. To give an example, given this markup:
```
<ul class="usa-nav-mobile">
  <li>Menu A</li>
</ul>
<ul class="usa-nav-desktop">
  <li>Menu B</li>
</ul>
```
This PR was intended to allow menu A (only) to display up to $nav-width, and menu B (only) to display beyond $nav-width.

Maybe you intended this?
```
.usa-nav-desktop {
  display: none;
  @include media($nav-width) {
    display: block;
  }
}
```

This would allow for displaying menu A (only) up to $nav-width, and menus A and B (both) beyond $nav-width. That's some improvement, but unfortunately it's just short of what I was trying to allow for in this PR.

maya:
@brockfanning sorry, I just rechecked and your code is 👍.  

I think the team still needs to do a bit of research on this to better understand different use cases and if this is a desired feature by others. The example on the USWDS site is very specific to that site, it's a fixed-width sidenav and previously dropped above the footer on mobile before being hidden as it is now. Ideally, we'd want to use the components provided in the library so would want to know if we should add this pattern/ability into the library or use an existing pattern.

So we can better understand your use case, in your example, what items are you trying to only show in mobile/desktop nav widths?

maya:
Thanks again for opening this @brockfanning. At the moment, the team is more focused on fixing bugs, but may re-open this when we have a headers-focused sprint and see that something like this would be beneficial.

PR #2024, opened 2017-07-11
fixes #2023 (search box above nav on IE11 bug)

### LINKED ISSUE #2023: Search box is above top nav in IE11
In IE 11, search box is above top navigation in usa-header-basic instead of on same level. [view bug](https://standards.usa.gov/components/headers/basic-mega/)

screen shot 2017-07-11 at 12 04 26 pm

bruffridge:
IE 11 doesn't support display: initial. Since initial sets display to the elements default, I just explicitly set display to block which is the default display property of div.

### DIFF
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -419,7 +419,7 @@
     flex-direction: column;
 
     @include media($nav-width) {
-      display: initial;
+      display: block;
       float: right;
       margin-top: -4.8rem;
     }


src/stylesheets/components/_navigation.scss
```diff
@@ -419,7 +419,7 @@
     flex-direction: column;
 
     @include media($nav-width) {
-      display: initial;
+      display: block;
```
toolness: I wonder if it'd be useful to write:

```css
  /* For IE11 */
  display: block;
  /* For newer browsers */
  display: initial;
```

That way the `display: initial` would still be set for all browsers that actually support that rule.

However, I'm unfamiliar with the CSS in question, so I don't actually know if `display: block` is all that's needed for things to look just like they do with `display: initial`... If it is, then I agree that *only* using `display: block` is the way to go!

toolness: Er wait, I'm being silly. I think just replacing `initial` with `block` is fine!

toolness:
Thanks for the contribution, @bruffridge !

PR #2026, opened 2017-07-12
Short header tweaks

msecret:
## Description

This fixes some of the header issues as identified in https://github.com/18F/web-design-standards/issues/1753. It fixes:

- The nav link alignment so the expanded menu items fill their vertical space

## Additional information

* An unrelated fix is for just the header fractal component to use the class structure in the same way the USWDS site does.

### DIFF
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -193,7 +193,7 @@ $z-index-nav:     9000;
       &[aria-expanded=false],
       &[aria-expanded=true] { /* stylelint-disable-line selector-no-qualifying-type */
         @include media($nav-width) { /* stylelint-disable-line selector-no-qualifying-type */
-          background-position: right 1.5rem top 50%;
+          background-position: right 1.5rem top 44%;
         }
       }
     }
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -138,6 +138,7 @@
   }
 
   button {
+    $button-vertical-offset: 36%;
     @include button-unstyled;
     @include font-smoothing;
     font-weight: $font-normal;
@@ -175,7 +176,7 @@
       @include media($nav-width) {
         background-image: url('#{$image-path}/angle-arrow-down.png');
         background-image: url('#{$image-path}/angle-arrow-down.svg');
-        background-position: right 1.5rem top 36%;
+        background-position: right 1.5rem top $button-vertical-offset;
       }
     }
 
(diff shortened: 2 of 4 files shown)


toolness:
@donjo can you review this one?

src/stylesheets/components/_header.scss
```diff
@@ -193,7 +193,7 @@ $z-index-nav:     9000;
       &[aria-expanded=false],
       &[aria-expanded=true] { /* stylelint-disable-line selector-no-qualifying-type */
         @include media($nav-width) { /* stylelint-disable-line selector-no-qualifying-type */
-          background-position: right 1.5rem top 50%;
+          background-position: right 1.5rem top 44%;
```
donjo: This should be 54% (in conjunction with the other change I'm requesting)

src/stylesheets/components/_navigation.scss
```diff
@@ -138,13 +138,15 @@
...
+    $button-vertical-offset: 36%;
+
     @include button-unstyled;
     @include font-smoothing;
     font-weight: $font-normal;
     padding: 1.4rem 1.5rem 1rem 1.8rem;
 
     @include media($nav-width) {
-      @include padding(null 3rem 1.9rem 1.5rem);
+      @include padding(null 3rem 3rem 1.5rem);
```
donjo: I think 1.9rem was correct otherwise the text isn't vertically centered

src/stylesheets/components/_header.scss
msecret: I thought based on the conversation, @maya was saying the nav wasn't meant to be vertically centered liked that, so I actually purposely didn't do it.

donjo: Looking back at the issue I see where that was coming from now. However, we do have a new visual design lead so maybe @rtwell will care to chime in on how we should approach things going forward. 

Ryan, my proposed style is included above where the text is vertically centered, the dropdown starts below the nav, and the hover highlight is aligned with the bottom of the nav bar while remaining inside that container.

donjo: screen shot 2017-07-17 at 5 38 59 pm

fwiw we can also adjust the bottom border on the span so it doesn't line up with the bottom of the nav if we want to fit the style but keep things centered.

maya: @msecret if you're thinking of this comment https://github.com/18F/web-design-standards/issues/1753#issuecomment-306972426 it wasn't about that it shouldn't be vertically centered, it was about not aligning the hover bar to the baseline. I often use a short header with the nav vertically centered. I've seen some designers align it to the top of the logo tho, different designers may want different treatments, so it may be good to account for this and allow a bit of flexibility.

donjo:
going to approve and merge this since it fixes the issue and the small things I was talking about will get addressed in @rtwell design audit

PR #2027, opened 2017-07-12
Updating Gov Banner

### LINKED ISSUE #1748: GovBanner 2.0
With the recent changes proposed #1738, we should revisit the GovBanner at a later date to ensure that our intent behind the GovBanner isn't lost. 

Intent:
- Educate users (general public) on what they should look for when using a federal government website

Why?
- One of the goals of the WDS is to instill trust in people when using a federal website, part of that is educating them on what makes something a federal website vs a non-federal website. It's often too easy to make a normal site look/feel/sound like it's part of the government.

donjo:
fixes 18f/web-design-standards-docs#315
fixes #1748 

This updates the gov banner text with more accurate information (derived from USA.gov).

This may still need to be updated on other sites but should show the updated version in Fractal.

Thanks @konklone!

### DIFF
--- a/spec/unit/banner/template.html
+++ b/spec/unit/banner/template.html
@@ -15,13 +15,17 @@
           <p>
             <strong>The .gov means it’s official.</strong>
             <br>
-            Federal government websites always use a .gov or .mil domain. Before sharing sensitive information online, make sure you’re on a .gov or .mil site by inspecting your browser’s address (or “location”) bar.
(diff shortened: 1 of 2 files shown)


konklone:
:+1: from me. What are the best instructions to users of the USWDS to help them migrate?

donjo:
We'll make sure to drop some instructions in the release notes but it should be as simple as copy/pasting some HTML from https://components.standards.usa.gov (will link to the more specific page) into the appropriate part of their app.

PR #2028, opened 2017-07-12
Include sass files in the release zip file

### LINKED ISSUE #1739: Add documentation for Sass users who do not have access to Node
The Standards documentation currently only outlines compilation of the Standards using Node and gulp. This solution does not include an option for those who do not have access or cannot install Node. 

And while we do provide the [compiled assets](https://standards.usa.gov/getting-started/download/), Sass is not included in these files. We should create an option for those who are wanting to use Sass, but do not have access to Node.

Thanks to @evetsreklaw who brought up this issue in our public Slack channel!

toolness:
This fixes #1739 by including our SCSS files, along with vendor dependencies like Neat and Bourbon, in the release zip file.

To do:

- [x] Verify with @evetsreklaw that the bundled SCSS works well with Dreamweaver CC.
- [ ] Add documentation so folks who want to use the SCSS without node can do so (we may need to do this as a separate issue/PR in the docs repository).

### DIFF
--- a/config/gulp/sass.js
+++ b/config/gulp/sass.js
@@ -43,6 +43,15 @@ gulp.task('copy-vendor-sass', function () {
   return stream;
 });
 
+gulp.task('copy-dist-sass', function () {
+  dutil.logMessage('copy-dist-sass', 'Copying all SASS to dist dir');
+
+  var stream = gulp.src('src/stylesheets/**/*.scss')
+    .pipe(gulp.dest('dist/scss'));
+
+  return stream;
+});
+
 gulp.task(task, [ 'copy-vendor-sass' ], function () {
 
   dutil.logMessage(task, 'Compiling Sass');
--- a/config/gulp/build.js
+++ b/config/gulp/build.js
@@ -51,6 +51,9 @@ gulp.task('build', function (done) {
       'images',
       'fonts',
     ],
+    // We need to copy the SASS to dist *after* the sass task, to ensure
+    // that vendor libraries have been copied to the SASS directory first.
+    'copy-dist-sass',
     done
   );
 


toolness:
@evetsreklaw I'm including a "prototype" bundled ZIP file below. The SCSS is in the `uswds-1.2.1/scss` directory. Can you try using it with Dreamweaver CC to see if it fits your needs?

[uswds-1.2.1-with-scss.zip](https://github.com/18F/web-design-standards/files/1143567/uswds-1.2.1-with-scss.zip)

config/gulp/sass.js
```diff
@@ -43,6 +43,15 @@ gulp.task('copy-vendor-sass', function () {
   return stream;
 });
 
+gulp.task('copy-dist-sass', function () {
```
toolness: I wanted to have `copy-vendor-sass` be a dependency here, but gulp seems to *always* run task dependencies even if they've already been run during the gulp process' lifetime, so I left it out, because I didn't want to execute `copy-vendor-sass` multiple times during the doggone build process.

I guess another option might be to have `copy-dist-sass` mention `copy-vendor-sass` as a dependency, and then have `sass` depend of `copy-dist-sass` instead of `copy-vendor-sass`, but I don't like that either because copying the SASS to the dist directory is really orthogonal to compiling the SASS in general. Blerg.

evetsreklaw:
@toolness It works perfectly in Dreamweaver CC 2017 - uswds.css generated without error.  This would have saved me so much time. Thank you!
dwtest

toolness:
That's awesome @evetsreklaw!

Now I just have to write docs... Since those will happen over on the docs repo, i wonder if we should just review + merge this PR now.  Thoughts @msecret ?

msecret:
@toolness yea it's fine to merge the code first then write docs

PR #2030, opened 2017-07-13
Fix navigation accessibility issues

### LINKED ISSUE #1649: User is not informed about the Menu has been collapsed
**Issue**:
In Responsive view,  when a user selects a Close button, once the action is performed and Menu is closed, a screen reader is silent and immediate feedback of what was performed is not notified for screen readers.

**Solution**:
After screen reader users perform an action on the page they should hear that about the changes on the page. This can be done by setting focus to the back to the Menu button. See related focus issue 31F-3.

### LINKED ISSUE #1653: Focus indicator is lost in Responsive view.
**Issue**:
In Responsive view, the visual focus is lost on the page after the menu button. Even the Menu is closed the menu items are still included in the tab order. 
See image [ResponsiveViewFocus_ExtendedHeader.png](https://drive.google.com/open?id=0ByQeU_iOqP_iWTZ6eG9wZkhVODg)

**Solution**:
After the Menu is closed, set the visual focus to back to the Menu button.

### LINKED ISSUE #2029: Collapsible nav menu is not very accessible for keyboards and ATs
This has been mentioned in a number of separate issues but I wanted to create an issue that documents everything that needs to be done with this component.

* When desktop users have a small viewport, the navigation gets collapsed behind the "menu" button, but its close button and links are still in the tab order. I think this is because they're all actually off-screen, waiting to "slide into" the screen from the right once the menu button is selected. Regardless, this means that when users press <kbd>Tab</kbd>, their focus disappears into a wormhole for a bit as they select those off-screen links and buttons. (Update: this is basically #1653.)
* Similarly, when the "menu" button is shown, the screen reader announces the close button and all the nav links, even though they're not visible on the page. It's actually debatable as to how problematic this is: if the user is *only* experiencing the site with a screen reader, we might actually be able to get away with applying `aria-hidden="true"` to the menu and the close button, thereby having the navigation always be present to screen-reader users regardless of viewport size. However, I think some people use screen readers in concert with sighted browsing, and such an experience would likely be confusing.
* ~~There's no ARIA correlation between the "menu" button and the navigation that gets collapsed/un-collapsed, so screen-reader users won't really know what's happening when they activate it.~~ Ehh, I've tried this with NVDA and I don't think it's that big a deal.  I think maybe adding an `aria-label` of "navigation menu" might make things a bit more understandable, since it takes them to a navigation landmark, but that's something that site authors can add to their HTML if they want...

toolness:
This fixes a number of accessibility-related bugs with navigation:

* Fixes #2029 
* Fixes #1653 
* Fixes #1649

It also addresses a few other focus issues that are documented in the GitHub/code comments below.

To do:
- [x] Add tests.
- [x] ~~Consider ensuring that keyboard focus can't leave the navigation menu once it's activated. The menu essentially seems to function like a dialog box, and behaves much like one from a mouse/touch standpoint, so we might want to make sure it behaves like one from a keyboard standpoint as well.~~ Filed separately as #2031.

### DIFF
--- a/spec/unit/navigation/navigation.spec.js
+++ b/spec/unit/navigation/navigation.spec.js
@@ -2,6 +2,7 @@
 const assert = require('assert');
 const fs = require('fs');
 const path = require('path');
+const sinon = require('sinon');
 const navigation = require('../../../src/js/components/navigation');
 const accordion = require('../../../src/js/components/accordion');
 
@@ -64,6 +65,39 @@ describe('navigation toggle', function () {
     menuButton.click();
     navLink.click();
     assert.equal(isVisible(nav), false);
+  });
+
+  it('focuses the close button when the menu button is clicked', function () {
+    menuButton.click();
+    assert.equal(document.activeElement, closeButton);
+  });
+
+  it('focuses the menu button when the close button is clicked', function () {
+    menuButton.click();
+    closeButton.click();
+    assert.equal(document.activeElement, menuButton);
+  });
+
+  it('collapses nav if needed on window resize', function () {
+    menuButton.click();
+    sinon.stub(closeButton, 'getBoundingClientRect').returns({ width: 0 });
+    try {
+      window.dispatchEvent(new CustomEvent('resize'));
+    } finally {
+      closeButton.getBoundingClientRect.restore();
+    }
+    assert.equal(isVisible(nav), false);
+  });
+
+  it('does not collapse nav if not needed on window resize', function () {
+    menuButton.click();
+    sinon.stub(closeButton, 'getBoundingClientRect').returns({ width: 100 });
+    try {
+      window.dispatchEvent(new CustomEvent('resize'));
+    } finally {
+      closeButton.getBoundingClientRect.restore();
+    }
+    assert.equal(isVisible(nav), true);
   });
 
   it('does not show the nav when a nav link is clicked', function () {
--- a/spec/unit/navigation/navigation.spec.js
+++ b/spec/unit/navigation/navigation.spec.js
@@ -11,6 +11,7 @@ const TEMPLATE = fs.readFileSync(path.join(__dirname, 'template.html'));
 describe('navigation toggle', function () {
   const body = document.body;
 
+  let sandbox;
   let nav;
   let overlay;
   let closeButton;
@@ -32,6 +33,7 @@ describe('navigation toggle', function () {
     menuButton = body.querySelector('.usa-menu-btn');
     accordionButton = nav.querySelector('.usa-accordion-button');
     navLink = nav.querySelector('a');
+    sandbox = sinon.sandbox.create();
   });
 
   afterEach(function () {
@@ -39,6 +41,7 @@ describe('navigation toggle', function () {
     body.className = '';
     navigation.off();
     accordion.off();
+    sandbox.restore();
   });
 
   it('shows the nav when the menu button is clicked', function () {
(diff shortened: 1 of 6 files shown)


src/stylesheets/components/_navigation.scss
```diff
@@ -62,11 +63,14 @@
     overflow-y: visible;
     position: relative;
     width: auto;
+    visibility: visible;
   }
 
   &.is-visible {
     @include transform(translateX(0));
-    @include transition(all 0.3s ease-in-out);
+    @include transition(transform 0.3s ease-in-out);
```
toolness: Changing the transition from `all` to just `transform` was crucial for ensuring that we could still focus on the close button immediately after the user clicked on the menu button: with transition set to `all`, even `visibility` was being transitioned, which meant that when `focus()` was called on the close button, nothing happened because the button wasn't actually visible yet!

src/stylesheets/components/_navigation.scss
```diff
@@ -51,6 +51,7 @@
   padding: 2rem;
   width: $sliding-panel-width;
   z-index: $z-index-nav;
+  visibility: hidden;
```
toolness: This ensures that keyboard focus can't go into the nav menu unless it's visible, either because desktop mode is active, or because the mobile nav is active.

src/js/components/navigation.js
```diff
@@ -7,7 +7,6 @@ const accordion = require('./accordion');
 const CLICK = require('../events').CLICK;
 const PREFIX = require('../config').prefix;
 
-const CONTEXT = 'header';
```
toolness: I got rid of this and the focus logic that required the close button be inside it because the close button *isn't* inside a `<header>` on standards.usa.gov, which meant that the close button wasn't even being focused on that site after the user activated the menu. 😞

toolness:
Ok, I think this is ready for review. I've manually tested it on Chrome, Firefox, and IE11 and it seems to work fine on them.

spec/unit/navigation/navigation.spec.js
```diff
@@ -66,6 +67,39 @@ describe('navigation toggle', function () {
...
+    assert.equal(document.activeElement, menuButton);
+  });
+
+  it('collapses nav if needed on window resize', function () {
+    menuButton.click();
+    sinon.stub(closeButton, 'getBoundingClientRect').returns({ width: 0 });
+    try {
+      window.dispatchEvent(new CustomEvent('resize'));
+    } finally {
+      closeButton.getBoundingClientRect.restore();
```
msecret: Rather then doing this in a finally, you can use the sinon sandbox and restore in in an afterEach. Then you wouldn't need both of these. Or is there something else about the `try` that's beneficial?

src/js/components/navigation.js
```diff
@@ -26,20 +25,43 @@ const toggleNav = function (active) {
...
+    // visible (this may have been what the user was just focused on,
+    // if they triggered the mobile nav by mistake).
+    menuButton.focus();
   }
+
   return active;
 };
 
+const resize = () => {
+  const isActive = document.body.classList.contains(ACTIVE_CLASS);
```
msecret: Shouldn't we have a util for this?

msecret: Or I guess we're not, seems pretty simple so maybe it's fine

spec/unit/navigation/navigation.spec.js
toolness: Ah nice, I didn't know about [sinon sandboxes](http://sinonjs.org/releases/v1.17.7/sandbox/)!  Thanks, I'll convert this to use them.

toolness:
Ok, I just resolved your concerns and also fixed a merge conflict due to #2006.  I also added some comments around some of the CSS changes I've made, because I think they're particularly non-obvious, in that they don't change the behavior of the site for *sighted* users, but do affect its usability for keyboard and AT users.

PR #2033, opened 2017-07-14
Update Banner Component - US Flag

tbolt:
## Description

I noticed upon zooming in on the US Flag that there were only 11 stripes and 9 stars. I swapped it for a public domain [SVG](https://en.wikipedia.org/wiki/File:Flag_of_the_United_States.svg) and now it will appear correctly, albeit a bit small. 

**Normal**
screen shot 2017-07-14 at 4 25 58 pm

**Zoomed in**
screen shot 2017-07-14 at 4 14 23 pm

I think it would be a good idea to migrate all of the favicons over as well. Preserving the accuracy of flags is important.

## Additional information

[SVG Browser Support](http://caniuse.com/#feat=svg)

Something I noticed is that there were rules for the US Flag img that were being over-written by specificity. Due to this I simply put a padding instead of margin-top (which would always be overwritten by a previous, but more-specific class)
notice

### DIFF
--- a/src/components/banner/banner.njk
+++ b/src/components/banner/banner.njk
@@ -2,7 +2,7 @@
   <div class="usa-accordion">
     <header class="usa-banner-header">
       <div class="usa-grid usa-banner-inner">
-      <img src="{{ uswds.path }}/img/favicons/favicon-57.png" alt="U.S. flag">
+      <img src="{{ uswds.path }}/img/usa-flag.svg" alt="U.S. flag">
       <p>An official website of the United States government</p>
       <button class="usa-accordion-button usa-banner-button"
         aria-expanded="false" aria-controls="gov-banner">
--- a/src/img/usa-flag.svg
+++ b/src/img/usa-flag.svg
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="1235" height="650" viewBox="0 0 7410 3900">
+<rect width="7410" height="3900" fill="#b22234"/>
+<path d="M0,450H7410m0,600H0m0,600H7410m0,600H0m0,600H7410m0,600H0" stroke="#fff" stroke-width="300"/>
+<rect width="2964" height="2100" fill="#3c3b6e"/>
+<g fill="#fff">
+<g id="s18">
+<g id="s9">
+<g id="s5">
+<g id="s4">
+<path id="s" d="M247,90 317.534230,307.082039 132.873218,172.917961H361.126782L176.465770,307.082039z"/>
+<use xlink:href="#s" y="420"/>
+<use xlink:href="#s" y="840"/>
+<use xlink:href="#s" y="1260"/>
+</g>
+<use xlink:href="#s" y="1680"/>
+</g>
+<use xlink:href="#s4" x="247" y="210"/>
+</g>
+<use xlink:href="#s9" x="494"/>
+</g>
+<use xlink:href="#s18" x="988"/>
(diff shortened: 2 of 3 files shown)


AndreaSigz:
Thanks @teebolt. @toolness @donjo

maya:
I think the current US flag icon was chosen bc it displays well at that small favicon size, not to be an accurate representation of the actual US flag. It's pretty common for icons to look different at different sizes and I would recommend to keep the icon as is bc the actual flag blurs everything together at that small size.

jondandois:
What about using the Star Spangled Banner version of the flag?  
screen shot 2017-07-20 at 8 00 51 am
https://commons.wikimedia.org/wiki/File:Star-Spangled_Banner_flag.svg
It's a real version of the flag, has fewer stars, and there is also a public domain SVG available.

maya:
@thisisdano is our visual design point person, and hoping he can provide some direction here. Any thoughts?

thisisdano:
I think there are good arguments to make about how we display our flag. How do we best preserve the legibility of the flag at small sizes? It's impossible to reproduce a fine level of detail at certain levels of scale, and the art of iconography is in capturing the essential and removing everything else. 

**Success criteria**
- Crisp and clean at the favicon size of 16x16 pixels (or 32x32px displayed as 16x16px)
- Use the same image for the official banner that we use for the favicon
- Clearly reads as the United States flag with no magnification


**Constraints**
- The aspect ratio of the official US flag is essentially 2:1 ([actually 1.9:1](http://www.ushistory.org/betsy/flagetiq3.html)) 
- 7 red bars and 6 white bars, alternating, starting and ending with red
- 50 white stars in a blue rectangle (the union): aspect ratio 1.4:1
- The union height is 7 bars (red to red)


**Solutions**

Scaling down a standard US flag results in a muddy image without clear distinction between the bars, and complete dissolution of the stars. (FYI, it's hard to display this with pixel fidelity in the following screenshots, and every screen has different display density.)



While this is an "accurate" flag, it doesn't read as accurate. The stars and bars both are reduced to soup. How might we rebuild the flag file to make it read better? Assigning each bar to a single pixel gives a clear distinction between bars, but requires changing the number of bars. Getting closest to the accurate 2:1 ratio results in a flag with 9 bars:




To my eye, this reads as too few bars, and the resulting union is too small to display many stars. Moving to 11 bars results in a flag that's closer to my mental image of the flag:




Additionally, an accurate flag reads as too dark and unsaturated at small size. You could be forgiven for thinking the flags below were reproduced in grayscale, particularly at a glance, and if the brightness on your machine is turned down:



This brighter version reads more clearly as red at small size. (And I should be clear that the changes are explicitly for very small sizes. They don't hold up at even double the size.):



- - -

I know that facts matter, and the flag is the flag is the flag, but in a specialized display environment there are compromises, and as a visual designer I'm focussed on the experiential, and most often hew to the "adhere to the spirit of the law rather than the letter of the law" dictum. And in the mind's eye, the flag looks something like this:



Clear red and white bars, a blue squarish rectangle in the upper-left corner from red bar to red bar with some number of distinct stars set in an offset pattern inset from the boundary, all inside a flag that's more rectangular than an old tv, but less rectangular than a new one.

And all this leads me to suggest a small flag very similar to the flag we currently use:



But what I might suggest is that we build a cleaner, sharper version (or versions) and do a better job of controlling its display (by, say, limiting its width in the banner to 16px). 

And while I'm being something of an apostate, I may as well say that I'd be open to using a high-contrast black and white version of the flag icon in certain situations. I just want to work to honor the spirit of the flag and present it in the best possible light over multiple display environments. 

:us:

thisisdano:
I'm closing this PR to address this issue via #2190. There's a lot to think about and I appreciate all the ideas and input.

PR #2035, opened 2017-07-17
[WIP] Improve the way we import Sass and make it available to client code

### LINKED ISSUE #2032: Sass can't easily be used when installing via npm + git
I was trying to create a sample project that used USWDS via the bleeding-edge version on GitHub, rather than the latest official release on `npm`.  However, when I tried to use USWDS's SASS, it failed to compile because the vendor SASS files (e.g. Bourbon) weren't in their expected locations at `lib/bourbon`.

It would be nice if this could "just work" so that folks could easily have projects that use a branch of the standards from git, rather than being required to either use an npm version or use `npm link`.

I'm not sure what the best way to do this is. I noticed that Blendid just adds `node_modules` to the SASS path; if we do that too, we could potentially import our dependencies via their path from `node_modules`.  For example, Bourbon could be loaded via:

```sass
@import "bourbon/app/assets/stylesheets/bourbon"
```

Then folks who wanted to use USWDS via git could install bourbon as a peer dependency, at least... As things are now, they would somehow have to get `@import "lib/bourbon"` to work, which would likely mean manually setting up some kind of gulp-ish task.

Another alternative might be to add all SASS dependencies to `dependencies` (they're currently in `devDependencies`, and then add a `postinstall` task that basically does what `gulp copy-vendor-sass` does.

toolness:
This is an attempt to fix #2032 while also (hopefully) simplifying the way we pull in SASS dependencies.

Instead of copying dependencies to the `src/stylesheets/lib` directory, we use them directly by changing our SASS include paths.  By not hard-coding the paths to our dependencies, this allows dependent modules to more easily reuse our code directly from git repositories, or even swap one version of our dependencies out for another if they really want to.

To do:

- [x] Add tests to make sure the SASS files in the release zip file (added in #2028) can still be imported without requiring any dependencies.
- [ ] Bourbon and Neat both have [`index.js`](https://github.com/thoughtbot/bourbon/blob/master/index.js) [files](https://github.com/thoughtbot/neat/blob/master/index.js) that allow client code to import them and discover their include paths (which we use in this PR).  It might be nice for us to provide similar functionality to clients of our own code, but I'm not really sure what best practices around this are. Some possibilities might be:
    * Move our CSS/SASS dependencies from `devDependencies` to `dependencies` so that anything importing `uswds` will also get the required CSS/SASS dependencies.
    * Document our CSS/SASS dependencies and leave it up to the importer to include them in their project.
- [ ] Ensure that the docs repo can still import and use `uswds` once this is done.

### DIFF
--- a/config/gulp/sass.js
+++ b/config/gulp/sass.js
@@ -1,3 +1,4 @@
+var path = require('path');
 var gulp = require('gulp');
 var dutil = require('./doc-util');
 var sass = require('gulp-sass');
@@ -7,14 +8,12 @@ var sourcemaps = require('gulp-sourcemaps');
 var rename = require('gulp-rename');
 var linter = require('@18f/stylelint-rules');
 var pkg = require('../../package.json');
+var sassIncludePaths = require('../../sass-include-paths');
 var filter = require('gulp-filter');
 var replace = require('gulp-replace');
 var runSequence = require('run-sequence');
 var del = require('del');
 var task = 'sass';
-
-var entryFileFilter = filter('uswds.scss', { restore: true });
-var normalizeCssFilter = filter('normalize.css', { restore: true });
 
 gulp.task('stylelint',
   linter('./src/stylesheets/{,core/,components/,elements/}*.scss',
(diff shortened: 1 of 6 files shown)


toolness:
Hmm, we might actually want to save this for 2.0 because I think it will require anyone that imports `uswds` via npm to change their code.

toolness:
I've copied some of the stuff in this PR, specifically 0f7a6b96a6051bb1cc934b5bf5e05a600ff078b3, over to #2036, so eventually this PR will be a bit smaller.

msecret:
How would it change for people import uswds? So do we publish an `includePaths` property now similar to bourbon / neat?

toolness:
Yeah, I think we'd publish an `includePaths`, and then folks who want to use our SCSS would either need to:

* Take our `includePaths` into account when using our SCSS at `src/stylesheets`, *or*
* Supply their own versions of our dependencies (bourbon / neat / normalize) and make sure they're on the SASS path so that the `@import 'bourbon'` etc. used by our SCSS works, *or*
* Use our pre-collected, standalone SCSS stored in `dist/scss` (this would only work if they checked out a version directly from `npm`, which has gone through our build step; it couldn't be used when installing uswds w/ a GitHub URL b/c `dist` wouldn't exist)

Blerg.

maya:
Hi @toolness, I noticed the `WIP` tag on this PR and wanted to check in: is this something you'll try to get to a ready state or should we close it?

toolness:
Nah, not really--let's close it!

Issue #2046, opened 2017-07-17
Add unit tests for scrolling expanded accordions into view 


toolness:
We merged #2025 but I hadn't written any tests for that logic, so I should add some tests!

Looks like fixing this might take some stubbing, so we might want to wait until #2030 (which adds sinon as a dependency) is merged.

maya:
Your suggestion is appreciated, however it is not on the current product roadmap.

PR #2047, opened 2017-07-18
Add tests for is-in-viewport.js

toolness:
This addresses part of #2046 but doesn't resolve all of it.

### DIFF
--- a/src/js/utils/is-in-viewport.js
+++ b/src/js/utils/is-in-viewport.js
@@ -1,13 +1,13 @@
 // https://stackoverflow.com/a/7557433
 function isElementInViewport (el, win=window,
-                              htmlEl=document.documentElement) {
+                              docEl=document.documentElement) {
   var rect = el.getBoundingClientRect();
 
   return (
     rect.top >= 0 &&
     rect.left >= 0 &&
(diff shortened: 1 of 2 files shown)


src/js/utils/is-in-viewport.js
```diff
@@ -1,11 +1,14 @@
 // https://stackoverflow.com/a/7557433
-module.exports = function isElementInViewport (el) {
+function isElementInViewport (el, win=window,
+                              htmlEl=document.documentElement) {
```
msecret: When I see `htmlEl`, I don't really differentiate it from `el`. Should it be `docEl` or something?

msecret: Would it make sense to have `win` and `htmlEl` be optional, defaulting to `window` and `documentElement` if they don't exist?

toolness: Good idea, I will switch it to `docEl`... For your second comment, I think that is what I'm doing by using [ES6 default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters)?

msecret: Ha oh duh, I don't know why I didn't read it as that.

PR #2048, opened 2017-07-18
Add basic type-checking via TypeScript.

### LINKED ISSUE #2005: Add linting step that runs TypeScript in JS checking mode
TypeScript 2.3 added a new option called [`--checkJs`](https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files) which allows us to get a lot of the benefits of strong typing without having to actually write code in TypeScript, or integrate the TypeScript compiler into our build system.

The mode works primarily by (A) bringing in type annotations for all the third-party libraries we use, and (B) automatically parsing any JSDoc-style comments in our code to infer type annotations. The really nice thing about this is that we just *document* our code and TypeScript effectively helps us verify that our documentation is correct, and that all our code integrates properly. So we get both better internal documentation *and* better code quality.

As an example, if we used this, it would have prevented #1990 from ever occurring, as that was caused by a type error that would've been caught by TypeScript's `--checkJs`.

toolness:
This fixes #2005 by adding a simple TypeScript task that checks a whitelist of files, defined in `tsconfig.json`, for type safety. Right now we just test [`src/js/utils/select.js`](https://github.com/18F/web-design-standards/blob/develop/src/js/utils/select.js) because it was fairly simple and already had JSDoc comments in it, which is what [TypeScript's `--checkJs` mode](https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files) uses to infer typings.

To make sure it works, and/or to see what TypeScript's type feedback is like, try tinkering with `select.js` in weird ways. If you remove the `context = window.document` line, for instance, you should get a `Object is possibly 'null'.` error because we've enabled TypeScript's [strict null checking](https://ariya.io/2016/10/typescript-2-and-strict-null-checking).

### DIFF
--- a/src/js/utils/select.js
+++ b/src/js/utils/select.js
@@ -1,5 +1,11 @@
 'use strict';
 
+/**
+ * @name isElement
+ * @desc returns whether or not the given argument is a DOM element.
+ * @param {any} value
+ * @return {boolean}
+ */
 const isElement = value => {
   return value && typeof value === 'object' && value.nodeType === 1;
 };
@@ -8,8 +14,9 @@ const isElement = value => {
  * @name select
  * @desc selects elements from the DOM by class selector or ID selector.
  * @param {string} selector - The selector to traverse the DOM with.
- * @param {HTMLElement} context - The context to traverse the DOM in.
- * @return {Array.HTMLElement} - An array of DOM nodes or an empty array.
+ * @param {Document|HTMLElement?} context - The context to traverse the DOM
+ *   in. If not provided, it defaults to the document.
+ * @return {HTMLElement[]} - An array of DOM nodes or an empty array.
  */
 module.exports = function select (selector, context) {
 
@@ -17,7 +24,7 @@ module.exports = function select (selector, context) {
     return [];
   }
 
-  if (context === undefined || !isElement(context)) {
+  if (!context || !isElement(context)) {
     context = window.document;
   }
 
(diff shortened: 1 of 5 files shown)


src/js/utils/select.js
```diff
@@ -8,16 +14,17 @@ const isElement = value => {
...
+ * @return {HTMLElement[]} - An array of DOM nodes or an empty array.
  */
 module.exports = function select (selector, context) {
 
   if (typeof selector !== 'string') {
     return [];
   }
 
-  if (context === undefined || !isElement(context)) {
+  if (!context || !isElement(context)) {
```
toolness: The check for `context === undefined` is actually unneeded here because `isElement` takes care of that case.

However, TypeScript's [strict null checking](https://ariya.io/2016/10/typescript-2-and-strict-null-checking) is not smart enough to dig into our `isElement()` function and infer its return value if `context` happens to be null or undefined, so we'll explicitly check for null and other falsy values here, ensuring that TypeScript knows exactly what's going to happen if `context` has one of those values.

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "noEmit": true,
+    "allowJs": true,
+    "checkJs": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitAny": true,
+    "strictNullChecks": true
```
toolness: These are all documented in [TypeScript's compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

toolness:
Woot!

PR #2050, opened 2017-07-20
Prepare fractal components for inclusion in documentation site

toolness:
This makes various changes to the fractal components so they can be used directly as examples in the documentation site. It's being developed in parallel with https://github.com/18F/web-design-standards-docs/pull/357.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -3,11 +3,9 @@
 !src/
 
 build/
-config/
 coverage/
 spec/
 .github/
 .editorconfig
 .eslintrc.yml
-gulpfile.js
 circle.yml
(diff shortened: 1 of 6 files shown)


toolness:
Erm, so in 3774c5ed0459f32778c8573151af55607f1f5078, I removed the gulpfile and its dependencies in `config` from the `.npmignore` file, which at least allows dependencies to install the package from a `git` URL and build the SASS/federalist site/etc on its end if it really wants to. I'm using this right now in 18F/web-design-standards-docs#357 to make CircleCI pull the version of uswds in *this* PR and use it when running tests.

msecret:
I think this part is ready to go.

PR #2054, opened 2017-07-21
Fix Contribuiting link in Table of Contents

fpoumian:
## Description

I noticed the link for the "Contributing to the code base" was broken in the Table of Contents and was pointing to #contributing-to-the-codebase instead of "contributing-to-the-code-base, so I just fixed it.

Maybe we could start using something like [Doctoc](https://www.npmjs.com/package/doctoc) so that we can more easily generate the Table of Contents? Let me know what you think and I can take care of that.


AndreaSigz:
Thanks @fpoumian. What do you think @donjo @toolness?

PR #2055, opened 2017-07-25
Run aXe on (almost) all fractal components.

toolness:
This is a bit like https://github.com/18F/web-design-standards-docs/pull/344, but for the core standards rather than the docs site.  It also uses the Fractal API to dynamically get a list of all available components, and then runs aXe on each of them (with the exception of a few).

It also embeds the tests in a mocha test suite, which allows us to defer all test running logic and options (e.g., running aXe only on one fractal component) to mocha instead of implementing it ourselves.

These tests also run aXe on the page at two different resolutions: 412x732 and 1280x732 (dubbed "small desktop" and "large desktop" respectively).

To do:
- [ ] Resolve as many TODOs for failing aXe tests as we can, file issues for the rest.
- [x] Ensure this works on CircleCI.
- [x] Add docs to readme.


tsconfig.json
```diff
@@ -6,7 +6,10 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noImplicitAny": true,
-    "strictNullChecks": true
+    "strictNullChecks": true,
+    "types": [
+      "node"
+    ]
```
toolness: Well, this is annoying. Basically, our inclusion of `chrome-launcher` adds `@types/core-js` to our dependencies, which TypeScript then gets fussy over because I guess it's an ancient version of that library or something.  In any case, we're not actually type-checking any of our code that uses `chrome-launcher` or `core-js`.  There's no way to black-list packages under `@types` in TypeScript, so instead we'll have to take the whitelisting approach and explicitly list every `@types` package we want to enable here. Blerg.

toolness:
Note: I've changed CircleCI to use [Ubuntu 14.04 (Trusty)](https://circleci.com/docs/1.0/build-image-trusty/) so installing Chrome works.  (The docs repository was already configured to use this image, which is why this problem didn't crop up when I made https://github.com/18F/web-design-standards-docs/pull/344.)

toolness:
Hmm, that's interesting, we're getting an `Ensures buttons have discernible text` error on the `search` component on CircleCI that isn't showing up on my local machine.  That said, my local machine has a slightly different version of Chrome than the one that CircleCI is using, so that might have something to do with it... Odd.

spec/delayed-root-suite.js
```diff
@@ -0,0 +1,14 @@
+const fractal = require('../fractal');
+
+exports.fractalLoad = fractal.components.load();
+
+Promise.all([
+  exports.fractalLoad,
+]).then(() => {
+  // https://mochajs.org/#delayed-root-suite
+  run();
```
msecret: Where is run defined?

spec/axe.spec.js
```diff
@@ -0,0 +1,162 @@
...
+    // terminate the process when it's done running tests.
+    before('start fractal server', () => server.start());
+
+    before('initialize chrome', done => {
+      getChrome().then(newChrome => {
+        chrome = newChrome;
+        chromeHost = chrome.host || 'localhost';
+        serverUrl = `http://${HOSTNAME}:${server.port}`;
+        done();
+      }).catch(done);
```
msecret: Should this be `done.fail` like Jasmine? Or will Mocha be smart enough to know that this failed and exit with `1` rather then `0`?

toolness: Nah, `done()` in mocha is like a node callback function in that it gets called with a falsy first argument to represent "no errors" or an exception as the first argument to represent an error.

spec/delayed-root-suite.js
toolness: I'll add a bit more context in this file to explain this better!

toolness:
Ok, I think this PR is big/complex enough that it should probably just get merged now, and then I'll file separate issues to fix the individual fractal components that are failing the aXe tests and resolve those later.

@msecret I noticed you approved these changes, but I did make a few relatively small changes after your review--should I go ahead and merge or do you want to take a second look?

PR #2057, opened 2017-07-26
Fix aXe violations for document-title, html-has-lang, frame-title.

toolness:
This fixes a few of the quick fixes from #2055.


src/components/layouts/_base.njk
```diff
@@ -1,11 +1,11 @@
 <!doctype html>
-<html lang="{{ lang }}">
+<html lang="{% if lang %}{{ lang }}{% else %}en{% endif %}">
   <!-- generated by {{ package.name }}@{{ package.version }} -->
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <title>{{ meta.title }}</title>
+    <title>{% if meta.title %}{{ meta.title }}{% else %}USWDS Fractal example document{% endif %}</title>
```
toolness: I was tempted to just remove the `{% if %}` statement from both of these modified lines entirely, as I doubt we're actually defining `lang` or `meta.title` anywhere, but I left them in just in case...

src/components/utilities/embed-container/embed-container.njk
```diff
@@ -1,3 +1,3 @@
 <div class="usa-embed-container">
-  <iframe src="https://www.youtube.com/embed/iLD4Bu6I2I8" frameborder="0" allowfullscreen></iframe>
+  <iframe src="https://www.youtube.com/embed/iLD4Bu6I2I8" frameborder="0" allowfullscreen title="Introducing the US Web Design Standards (YouTube)"></iframe>
```
toolness: Related: #2058

PR #2062, opened 2017-07-26
Tables decrease specificity

msecret:
## Description

In order to make it easier for users to develop with the standards, CSS specificity in elements and components should be kept as low as possible so things are easy to override.

Fixes: https://github.com/18F/web-design-standards/issues/1428

## Additional information

This could be a breaking change in the following case:
- A user is using the standards and some other CSS library on one project.
- The other library, imported after the standards, makes some changes to `td`, `th` default styling.
Before: The standards would have overridden that default styling due to being nested in the `table` tag.
Now: The styling will not be overridden if declared after the un-nested standards tags.

This could also apply when not using a library, but that would mean the user actively has CSS code that doesn't do anything (which is definitely something that could potentially happen). 

### DIFF
--- a/src/stylesheets/elements/_table.scss
+++ b/src/stylesheets/elements/_table.scss
@@ -2,30 +2,29 @@ table {
   border-spacing: 0;
   margin: 2em 0;
   min-width: 100%;
+}
 
-  thead {
-    th,
-    td {
-      background-color: $color-gray-lightest;
-    }
-  }
-
+thead {
   th {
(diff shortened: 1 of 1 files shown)


msecret:
Because this is a breaking change, it needs to go into a major release, probably not develop. Is there a way to do that? Should a branch be created for version 2?

maya:
@msecret looking through this and was wondering how is this a breaking change?

PR #2064, opened 2017-07-26
Fix fractal buttons contrast

### LINKED ISSUE #2059: Fractal "buttons" component raises aXe violations
Currently running the "buttons" component through the aXe tests in #2055 fails, which is why they're being skipped at the moment. Based on the aXe output, it looks like aXe has problems with buttons that have the following `class` attributes:

* `usa-button-outline-inverse usa-button-hover`
* `usa-button-outline-inverse usa-button-active`

Given that these represent hover and active states, there might be an exception to WCAG AA's color contrast requirements that effectively make these violations false positives. If that's the case, I think there's some aXe configuration options we can pass in to ignore anything with these classes, maybe.

toolness:
Fixes #2059.

As mentioned in https://github.com/18F/web-design-standards/issues/2059#issuecomment-318204822, it seems like the last row of secondary buttons was meant to have a dark background. It's even given a class called `button_wrapper-dark` but there's no CSS for it defined anywhere, so this row of buttons has a white background.

~~This removes that class and just adds an inline `style="background: black"` to fix things:~~

This replaces that class with `usa-background-dark`:

> buttons-with-usa-background-dark

It's kind of weird that the background runs all the way to the end of the page. But it's also weird that none of the buttons in any of the rows have any margin separating them from the left edge of the page. I suspect some styling was once provided for both `button_wrapper` and `button_wrapper-dark` but it disappeared at some point and now we have kind of weird styling.

### DIFF
--- a/spec/axe.spec.js
+++ b/spec/axe.spec.js
@@ -22,7 +22,6 @@ const AXE_OPTIONS = JSON.stringify({
   },
 });
 const SKIP_COMPONENTS = [
-  'buttons',         // TODO: Resolve color contrast issues and remove.
   'search',          // TODO: Resolve discernible text issues and remove.
 ];
 const SMALL_DESKTOP = {
--- a/src/components/02-buttons/buttons.njk
+++ b/src/components/02-buttons/buttons.njk
@@ -30,7 +30,8 @@
     <button class="usa-button-outline usa-button-hover">Hover</button>
   </div>
 
-  <div class="button_wrapper button_wrapper-dark">
+  <div class="button_wrapper" style="background: black">
+    <!-- These buttons should only appear on a dark background to meet WCAG AA color contrast requirements. -->
     <button class="usa-button-outline-inverse" type="button">Default</button>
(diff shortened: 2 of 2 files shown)


toolness:
@donjo any idea what the `button_wrapper` / `button_wrapper-dark` CSS used to be?  If we could dig through the git history to find it, maybe we can get this component to look a bit less broken...

toolness:
Oh, I didn't know about `usa-background-dark`!  Using this class now, instead of `background: black`.

msecret:
Pretty sure I know what happened:

`button_wrapper-dark` is a styleguide class, meaning it's not in the standards at all, it's only in the additional SCSS file on the docs site to help with things like this: styling needed just for the documentation. So when we moved the HTML from there to here, we didn't move the class, so it didn't work. Your solution sounds fine.

toolness:
Ahhh so that's what it is! Thanks for the explanation @msecret!

I guess we'll be talking about this kind of thing during the fractal meeting in a few minutes...

PR #2065, opened 2017-07-27
Fix search button a11y

### LINKED ISSUE #2061: Fractal "search" component raises aXe violations on small desktops
Currently the "search" component raises aXe violations in its small desktop test introduced in #2055, so it's disabled for now. It looks like this is a legitimate bug: basically, if the screen size is small enough, only the magnifying glass icon is shown for the search button, and its text is given `display: none`, which means that screen readers won't be able to communicate anything about the button.

toolness:
Fixes #2061.

### DIFF
--- a/src/stylesheets/components/_search.scss
+++ b/src/stylesheets/components/_search.scss
@@ -50,10 +50,10 @@ $usa-btn-big-width:     11.6rem;
   }
 
   .usa-search-submit-text {
-    display: none;
+    @include sr-only();
 
     @include media($small-screen) {
-      display: block;
+      @include not-sr-only();
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/stylesheets/components/_search.scss
```diff
@@ -50,10 +50,10 @@ $usa-btn-big-width:     11.6rem;
   }
 
   .usa-search-submit-text {
-    display: none;
+    @include sr-only();
```
msecret: Yea I found this early too, glad it was caught by axe

src/stylesheets/core/_utilities.scss
```diff
@@ -49,6 +49,15 @@
   left: -999em;
 }
 
+// This "negates" the sr-only() helper; useful if
+// we want the default state of something to be
+// for screen readers only, but we want to force
+// it to show for sighted users under more specific
+// conditions.
+@mixin not-sr-only() {
```
msecret: Ha I was thinking of exactly this too.

PR #2066, opened 2017-07-27
Use a data URL for the embed example

### LINKED ISSUE #2058: Consider using a local video (or image?) for the embed example
We currently have a fractal component representing a responsive embed container, and it embeds a YouTube video:

https://components.standards.usa.gov/components/detail/embed-container.html

Something I noticed while working on #2055 is that since our test suite now loads every fractal component in headless Chrome, this particular fractal component increases our test time significantly--by about 500ms per screen size on my current internet connection, but I assume it'd be even slower if I were on a spotty internet connection.  The test would probably fail if I was working offline, too.

Is there a local resource we could use instead of this YouTube video?  For instance, could we just use a local image that has a similar aspect ratio as the video?

toolness:
This fixes #2058 by using a local `data:` URL for the embed example.

I also added some text below the embed in the example, to make sure that text displays under the embed properly, rather than e.g. on top of it or something weird.

### DIFF
--- a/spec/axe.spec.js
+++ b/spec/axe.spec.js
@@ -20,6 +20,9 @@ const AXE_OPTIONS = JSON.stringify({
     // ignore that rule.
     'bypass': { enabled: false },
   },
+  exclude: [
+    [ 'iframe[src^="data:"]' ],
+  ],
 });
 const SKIP_COMPONENTS = [
   'buttons',         // TODO: Resolve color contrast issues and remove.
--- a/spec/axe.spec.js
+++ b/spec/axe.spec.js
@@ -14,15 +14,21 @@ const { fractalLoad } = require('./delayed-root-suite');
 const HOSTNAME = os.hostname().toLowerCase();
 const REMOTE_CHROME_URL = process.env[ 'REMOTE_CHROME_URL' ];
 const AXE_JS = fs.readFileSync(__dirname + '/../node_modules/axe-core/axe.js');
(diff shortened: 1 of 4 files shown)


spec/axe.spec.js
```diff
@@ -20,6 +20,9 @@ const AXE_OPTIONS = JSON.stringify({
     // ignore that rule.
     'bypass': { enabled: false },
   },
+  exclude: [
+    [ 'iframe[src^="data:"]' ],
```
toolness: I'm not sure why, but aXe takes a lot longer to run if it needs to check iframes, so I'm explicitly telling it not to dive into iframes with a `data:` URI source here.

spec/axe.spec.js
```diff
@@ -14,6 +14,15 @@ const { fractalLoad } = require('./delayed-root-suite');
...
 const AXE_JS = fs.readFileSync(__dirname + '/../node_modules/axe-core/axe.js');
+const AXE_CONTEXT = JSON.stringify({
+  exclude: [
+    // For some reason aXe takes a lot longer if it needs to dive into
+    // iframes with data: URIs. The content of these iframes is just for
+    // non-USWDS example content anyways, so just skip them to speed things
+    // up.
+    [ 'iframe[src^="data:"]' ],
+  ],
+});
```
toolness: For more details on this object, see https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#context-parameter.

Issue #1698, opened 2017-02-03
Add visual regression tests


shawnbot:
From #1673:

> Visual regressions are a category of bug in which changes to some part of the design implementation (HTML, CSS, or JavaScript) inadvertently affect the appearance of a component or page. Regressions are sometimes) intentional, but most often they are caused by overly broad bug fixes that cause problems in other contexts. (For instance, if somebody were to change the way that ul elements were styled, they might forget that accordions are implemented as unordered lists and forget to update the corresponding styles.) There are lots of different tools and techniques to perform visual regression tests, but they generally do some variation of:
> 
> 1. Given a specific state of each page or component, take a "baseline" screenshot from one or more browsers
> 1. Each time the test suite is run, take new screenshots of each component or page, in each state, and diff the pixels of the new and "baseline" images
> 1. If the number of pixels that differ exceeds some threshold (e.g. 1%), the changes must be manually approved
> 
> The simplest and most lightweight [approach](https://medium.com/@fabien.rassinier/non-visual-regression-in-github-pull-requests-85e469f7a010) to this that I've seen is to commit the screenshots to git and do manual approvals for images that differ via GitHub pull requests. One issue [mentioned by CM on Slack](https://gsa-tts.slack.com/archives/g-frontend/p1485194580000788) was that this approach (assuming it regularly generates slightly different images) might cause the git repo to balloon in size and quickly become unwieldy to sync. I think we need to do more research before committing to an approach, or a specific solution such as [BackstopJS] or [Screenr].

To do:

- [ ] Check in on #1021, which is a WIP PR to incorporate [Screenr]
- [ ] Evaluate tools and/or techniques using a subset of Standards components and report back on which one(s) are worth considering, including:
    * [Screenr]
    * [BackstopJS]
    * [The GitHub method](https://medium.com/@fabien.rassinier/non-visual-regression-in-github-pull-requests-85e469f7a010)
- [ ] Integrate the tool or technique
- [ ] Document our process for using it, assuming manual approvals for regressions are required

[Screenr]: https://screener.io/
[BackstopJS]: https://garris.github.io/BackstopJS/

toolness:
Oh hey, this is basically done as of #2067, #2100, and #2129.

PR #2067, opened 2017-07-27
[WIP] Add experimental support for local visual regression testing

toolness:
This adds experimental support for *local* visual regression testing (#1698).

By "local" I mean that the reference images that detect whether visual regressions have occurred are stored locally on a developer's system, and not committed to version control.

Here's an example of what the output looks like, when I add `outline: 4px solid black` to the radio button label CSS:

visual-regression-tool-screenshot-2

The visual diff is courtesy [Resemble.js](https://huddle.github.io/Resemble.js/).

Implementation-wise, it's essentially a kludge that sits atop the aXe test suite introduced in #2055. This is because the aXe test suite already takes care of loading every fractal component in headless Chrome, so it was easy to add another test that does a bit of screenshotting.

## Development workflow

It's easiest to explain how this works using an example.

1. A developer runs the following command:

    ```
    ENABLE_SCREENSHOTS=yup mocha --opts spec/mocha.opts spec/axe.spec.js
    ```

    For every Fractal component, the test runner detects if a screenshot for it exists in the `screenshots` directory. If not, it takes a screenshot of the component and saves it there. This is called "update" mode, and the screenshot generated is called the **golden** screenshot.

    On the other hand, if the screenshot *does* exist, it takes a screenshot of the component and compares it to the existing screenshot, raising an error if they do not match. This is called "compare" mode. For reference, the screenshot is saved to a different file, called the **fail** screenshot. Some metadata is also saved, so that an HTML5 front-end can be used to visually compare the two.

2. The developer makes refactoring changes to any SCSS files.

3. The developer re-runs the command from (1) to make sure no visual regressions have occurred. If they have occurred, a helpful failure message is shown, which points the user to an HTML5 file where they can visually compare the differences.

This can also be used by reviewers; for instance, if someone is reviewing a PR for a CSS refactoring, they can check out `develop`, run the screenshotter to generate the golden screenshots, then switch to the PR branch and re-run the screenshotter in compare mode.

## Why local?

1. I consider local visual regression testing to be a prerequisite for a service/CI-based approach. Developers should be able to make changes and get quick feedback about visual regressions--i.e., they shouldn't have to push to GitHub and wait for a service to tell them whether there's something amiss.

2. Visual regression testing is new to us, and it *seems* like something that's liable to produce false positives. If we integrate it into CI as something that *must* pass at all times, we run the risk of eventually ignoring it, if the false positive rate is too high. So I'd rather start out with it being an opt-in that can be used to suggest *potential* visual regressions, rather than being used as an undisputed source of truth.

3. To compare screenshots in a non-local way, we might have to guarantee that screenshots generated on different developers' systems will be identical, but that's very hard to do. For example, some aspects of Fractal components may be delegated to user-agent stylesheets, which may then be delegated to OS-specific stylesheets. We're using the Chrome DevTool Protocol's [`Emulation` domain](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/) to smooth out these differences, but some may still persist. And that's not accounting for the ways different operating systems might [antialias text](https://www.joelonsoftware.com/2007/06/12/font-smoothing-anti-aliasing-and-sub-pixel-rendering/), which could introduce more complications.

## To do

Some of these can be punted to later PRs.

- [ ] Determine whether this is actually useful.
- [ ] Take screenshots at multiple sizes.
- [ ] Document the mechanism (especially the `ENABLE_SCREENSHOTS` environment variable).

### DIFF
--- a/spec/axe.spec.js
+++ b/spec/axe.spec.js
@@ -198,6 +198,51 @@ fractalLoad.then(() => {
           return cdp.Emulation.setDeviceMetricsOverride(SMALL_DESKTOP)
             .then(() => runAxe(cdp));
         });
+
+        if (process.env.ENABLE_SCREENSHOTS) {
+          const ROOT_DIR = `${__dirname}/..`;
+          const REL_DIR = `screenshots`;
+          const ABS_DIR = `${ROOT_DIR}/${REL_DIR}`;
+          const REL_PATH = `${REL_DIR}/${item.handle}.png`;
+          const ABS_PATH = `${ROOT_DIR}/${REL_PATH}`;
+          let doCompare = fs.existsSync(ABS_PATH);
+
+          it(doCompare ? 'is identical to previous screenshot'
+                       : 'saves screenshot for comparison later', () => {
+            const metrics = JSON.parse(JSON.stringify(SMALL_DESKTOP));
+            const { Page, Emulation } = cdp;
+
+            metrics.fitWindow = true;
+
+            return Emulation.setDeviceMetricsOverride(metrics)
+              .then(() => Page.getLayoutMetrics())
+              .then((result) => {
+                metrics.height = result.contentSize.height;
+                return Emulation.setDeviceMetricsOverride(metrics);
+              })
+              .then(() => Page.captureScreenshot({ format: 'png' }))
+              .then(result => {
+                const data = Buffer.from(result.data, 'base64');
+
+                if (doCompare) {
+                  const goldenData = fs.readFileSync(ABS_PATH);
+                  if (!goldenData.equals(data)) {
+                    return Promise.reject(new Error(
+                      `Screenshot of "${item.handle}" does not match ` +
+                      `${REL_PATH}! If ${REL_PATH} represents an old ` +
+                      `screenshot that is no longer valid, please delete it.`
+                    ));
+                  }
+                } else {
+                  if (!fs.existsSync(ABS_DIR)) {
+                    fs.mkdirSync(ABS_DIR);
+                  }
+
+                  fs.writeFileSync(ABS_PATH, data);
+                }
+              });
+          });
+        }
       });
     }
   });
--- a/spec/axe.spec.js
+++ b/spec/axe.spec.js
@@ -135,6 +135,8 @@ const RUN_AXE_FUNC_JS = function runAxe (options) {
 let getChrome = REMOTE_CHROME_URL ? getRemoteChrome : launchChromeLocally;
 
 fractalLoad.then(() => {
+  const handles = Array.from(fractal.components.flatten().map(c => c.handle));
+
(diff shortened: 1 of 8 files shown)


spec/axe.spec.js
```diff
@@ -198,6 +198,51 @@ fractalLoad.then(() => {
...
+              .then((result) => {
+                metrics.height = result.contentSize.height;
+                return Emulation.setDeviceMetricsOverride(metrics);
+              })
+              .then(() => Page.captureScreenshot({ format: 'png' }))
+              .then(result => {
+                const data = Buffer.from(result.data, 'base64');
+
+                if (doCompare) {
+                  const goldenData = fs.readFileSync(ABS_PATH);
```
msecret: Why is this called `goldenData`?

toolness: Oh, I've sometimes heard of this notion of "the expected file contents" as being called a "golden file". For example, it looks like Go's test suite for `gofmt` uses files with a `.golden` extension to contain the expected output of various calls and stuff:

https://golang.org/src/cmd/gofmt/gofmt_test.go

That said, it's confusing out of context, especially since I don't use the word "golden" anywhere else in this code, so I'll change it to something else like `previousData`.

toolness:
Good point @msecret -- I'm currently working on a very basic front-end that can show a visual diff. Hope to have something presentable by demo time today!

Also, I thought a bit more about the word "golden" and I think it's actually appropriate (or at least, I can't think of a better word)--I just have to use the word more,  not less, so readers can better understand what it means.

toolness:
@msecret does this seem OK to you for a first pass?  I realized you approved it a few days ago, but I've added the whole HTML5 front-end with visual diffing since then... I also improved the front-end since I last demo'd it to y'all on Friday.

If this looks good I will add docs and punt the other stuff to issues; I'm full-time on PCLOB this week so I won't be able to work on this PR and other core USWDS stuff as much as I've been doing.

spec/visual-regression-tester.js
```diff
@@ -32,13 +32,24 @@ class VisualRegressionTester {
...
-        metrics.height = result.contentSize.height;
-        return Emulation.setDeviceMetricsOverride(metrics);
+        // This is weird, I'm not really sure what I'm doing, but after
+        // a bunch of experimentation, this seems to do what we want, i.e.
+        // capture a full-size screenshot of the entire page.
+        //
+        // At the time of this writing, though, Emulation.setVisibleSize()
+        // is documented as being both experimental *and* deprecated, so
+        // this code may not work for long. - AV 7/28/2017
+        return Emulation.setVisibleSize({
```
msecret: I think the most successful thing would be to take a screenshot of each component in isolation. That would mean if you change the height of something at the top of the page, it wouldn't invalidate everything below it. That being said, I think that's out of scope for this work.

PR #2095, opened 2017-08-03
Alert visual changes

msecret:
## Description

Change visual design and refactor code for alerts component. Fixes https://github.com/18F/web-design-standards/issues/2080.

Preview: [Alerts on Federalist](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/ms-alert_visual_changes/components/detail/alerts--default.html)

Still TODO:
- [x] Figure out how to get correct colors for the left-hand bars in a clean way. Currently it's using a color function to generate the colors from the base alert color, but it's not accurate to the spec.

- [x] The slim variant needs to have the smaller icon.

## Additional information

- I changed things up quite a bit by having spacing editable variables which get used to size things in this component. I'm still mulling over whether this is the best idea, and what should the intervals of these variables be. I do generally think it'd be good for consistency to have something like this.

- I made the default alert work with just one line, no header, which I think is an important piece of functionality that was missing before.

- I've broken up the code samples into variants as it's the new pattern we're using.

### DIFF
--- a/src/components/05-alerts/alerts.njk
+++ b/src/components/05-alerts/alerts.njk
@@ -1,42 +1,6 @@
-
-  <div class="usa-alert usa-alert-success">
-    <div class="usa-alert-body">
-      <h3 class="usa-alert-heading">Success Status</h3>
-      <p class="usa-alert-text">Lorem ipsum dolor sit amet, <a href="javascript:void(0);">consectetur adipiscing</a> elit, sed do eiusmod.</p>
-    </div>
+<div class="usa-alert {{ alert.classes }}" {% if alert.role %} role="{{alert.role}}"{% endif %}>
+  <div class="usa-alert-body">
+    <h3 class="usa-alert-heading">{{ alert.title }}</h3>
+    <p class="usa-alert-text">Lorem ipsum dolor sit amet, <a href="javascript:void(0);">consectetur adipiscing</a> elit, sed do eiusmod.</p>
   </div>
-
-  <div class="usa-alert usa-alert-warning">
-    <div class="usa-alert-body">
-      <h3 class="usa-alert-heading">Warning Status</h3>
-      <p class="usa-alert-text">Lorem ipsum dolor sit amet, <a href="javascript:void(0);">consectetur adipiscing</a> elit, sed do eiusmod.</p>
-    </div>
-  </div>
-
-  <div class="usa-alert usa-alert-error" role="alert">
-    <div class="usa-alert-body">
-      <h3 class="usa-alert-heading">Error Status</h3>
-      <p class="usa-alert-text">Lorem ipsum dolor sit amet, <a href="javascript:void(0);">consectetur adipiscing</a> elit, sed do eiusmod.</p>
-    </div>
-  </div>
-
-  <div class="usa-alert usa-alert-info">
-    <div class="usa-alert-body">
-      <h3 class="usa-alert-heading">Information Status</h3>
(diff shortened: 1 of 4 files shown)


rtwell:
this looks fantastic! 

can we adjust the color of the left border to be #00a6d2?

msecret:
Yea I'm figuring out the best way to get the colors now

rtwell:
i noticed that the icon is scaling when it reaches a smaller width—is that default behavior? doea our type have a scale for mobile? This may be a broader conversation best had somewhere else?

msecret:
@rtwell yea that is the existing mobile behavior. Because I didn't see a mobile mockup, I kept that behavior, but we can definitely revisit it. It would be easer (and get rid of code) to keep the icon size the same on mobile.

rtwell:
seems like `no-header` and `slim` could be one and the same (the slim design), we just need to shrink that icon a wee bit

src/components/05-alerts/alerts.njk
```diff
@@ -1,42 +1,8 @@
...
-      <h3 class="usa-alert-heading">Information Status - Paragraph Width</h3>
-      <p class="usa-alert-text">Multi line. Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui atione voluptatem sequi nesciunt. Neque porro quisquam est, qui doloremipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo.</p>
-    </div>
-  </div>
+<div class="usa-alert {{ alert.classes }}" {% if alert.role %}role="{{alert.role}}"{% endif %}>
+  <div class="usa-alert-body">
+    {% if alert.title %}
+    <h3 class="usa-alert-heading">{{ alert.title }}</h3>
+    {% endif %}
+    <p class="usa-alert-text">{{ alert.content }}</p>
```
toolness: Since these fractal components are becoming more likely to be included in the docs (see e.g. https://github.com/18F/web-design-standards-docs/pull/357), I'm curious if we should start using `{%-` and `-%}` to [remove excess blank lines](https://mozilla.github.io/nunjucks/templating.html#whitespace-control) from the resulting HTML code?  Using it on the above code, for instance, would get rid of the blank lines in the HTML rendering:

```html
<div class="usa-alert usa-alert-info">
    <div class="usa-alert-body">

        <h3 class="usa-alert-heading">Information Status</h3>

        <p class="usa-alert-text">Lorem ipsum dolor sit amet, <a href="javascript:void(0);">consectetur adipiscing</a> elit, sed do eiusmod.</p>
    </div>
</div>
```

It also makes the code snippet nicer to read on the fractal site too, of course, but it's really up to you.

msecret: I agree, would definitely be good to do that.

donjo: Oh nice! Didn't realize that was a thing. Those random whitespaces were really bugging me.

donjo:
I think the `margin-bottom` on the slim version is too large. There is a `padding-top: .5rem;` and bottom should be the same amount. Also, should it be padding on both top and bottom or is there a reason to have one side being positioned by margins and the other padding?

msecret:
@donjo I'll take a look at the spacing. I think the issue with margins is more to do with when there's a single line vs the header + line. In the former case, I add more bottom margin to the line to make up for the difference in height, and ensure the height stays consistent. This mean's in code, you can have either one line or line + header and not have to add another class.

msecret:
@donjo fixed the spacing issue, was due to a bad selector that I didn't realize.

PR #2098, opened 2017-08-08
Button refresh and Color Variables

donjo:
### 👀  [Preview implementation in Fractal](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/button-refresh/components/detail/buttons--default.html)

screen shot 2017-08-08 at 4 06 41 pm

Color variables are being reworked so colors are defined separately from their function within the system. Instead of the design system's primary color being set to a random hex value, they will now be set to a named color that is defined elsewhere in the `_variables.scss`
 
**This**
```
$color-primary:  #0071bc;
```
**Becomes This**
```
$color-blue:  #0071bc;
$color-primary:  $color-blue;
```

### DIFF
--- a/src/components/02-buttons/buttons.njk
+++ b/src/components/02-buttons/buttons.njk
@@ -1,6 +1,8 @@
 
 <button class="{{ classes }}">Default</button>
 {% if (category != 'disabled') and (category != 'big') %}
-<button class="{{ classes }} usa-button-active">Active</button>
 <button class="{{ classes }} usa-button-hover">Hover</button>
+<button class="{{ classes }} usa-button-active">Active</button>
+<button class="{{ classes }} usa-button-focus">Focus</button>
+<button class="{{ classes }} usa-button-disabled">Disabled</button>
 {% endif %}
--- a/src/components/02-buttons/buttons.njk
+++ b/src/components/02-buttons/buttons.njk
@@ -5,9 +5,9 @@
   <button class="{{ classes }} usa-button-focus">Focus</button>
   {% if (category == 'secondary') -%}
     <button class="usa-button-secondary-disabled">Disabled</button>
-  {% elif (category == 'secondary-inverse') %}
+  {% elif (category == 'secondary-inverse') -%}
     <button class="usa-button-secondary-inverse-disabled">Disabled</button>
-  {% else %}
+  {% else -%}
     <button class="usa-button-disabled">Disabled</button>
   {%- endif %}
 {%- endif %}
--- a/src/components/02-buttons/buttons.njk
+++ b/src/components/02-buttons/buttons.njk
@@ -1,8 +1,21 @@
-
 <button class="{{ classes }}">Default</button>
 {% if (category != 'disabled') and (category != 'big') %}
-<button class="{{ classes }} usa-button-hover">Hover</button>
-<button class="{{ classes }} usa-button-active">Active</button>
-<button class="{{ classes }} usa-button-focus">Focus</button>
-<button class="{{ classes }} usa-button-disabled">Disabled</button>
+  <button class="{{ classes }} usa-button-hover">Hover</button>
+  <button class="{{ classes }} usa-button-active">Active</button>
+  <button class="{{ classes }} usa-button-focus">Focus</button>
+  {% if (category == 'secondary') %}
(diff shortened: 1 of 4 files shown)


src/stylesheets/core/_variables.scss
```diff
@@ -32,54 +32,74 @@ $font-normal:         400 !default;
...
-$color-primary-alt-darkest:  #046b99 !default;
-$color-primary-alt-light:    #9bdaf1 !default; // lighten($color-primary-alt, 60%)
-$color-primary-alt-lightest: #e1f3f8 !default; // lighten($color-primary-alt, 90%)
-
-$color-secondary:            #e31c3d !default;
-$color-secondary-dark:       #cd2026 !default;
-$color-secondary-darkest:    #981b1e !default;
-$color-secondary-light:      #e59393 !default; // lighten($color-secondary, 60%)
-$color-secondary-lightest:   #f9dede !default; // lighten($color-secondary, 90%)
+$color-blue:                 #0071bc !default;
```
msecret: I like this naming system.

src/stylesheets/elements/_buttons.scss
```diff
@@ -122,7 +94,7 @@ button,
     }
   }
 
-  &.usa-button-outline-inverse {
+  &.usa-button-secondary-inverse {
```
msecret: This looks like a breaking change right? Because if somebody has a button with outline-inverse, then the button won't be targeted for those rules anymore? Did you have thoughts on whether it's OK to break this?

If we feel that it's necessary, we can keep the code the way you have it, and add a "Deprecated" section at the bottom of the file that has the rules to ensure nothing will break. Then we can remove the deprecated section once we upgrade to 2.0?

src/stylesheets/elements/_buttons.scss
```diff
@@ -155,18 +155,53 @@ button,
...
+  &.usa-button-focus {
     background-color: $color-gray-lighter;
     border: 0;
     box-shadow: none;
-    color: $color-gray-dark;
+  }
+}
+
+.usa-button-secondary:disabled,
+.usa-button-secondary-disabled {
```
msecret: Based on my understanding of the WDS CSS, I'd expect there not to be another `usa-button-secondary-disabled` class, but that if you add the `usa-button-disabled` to a button with `usa-button-secondary` it would correctly apply the right disabled styles. I definitely think `.usa-button-secondary:disabled` makes sense though

src/stylesheets/elements/_buttons.scss
```diff
@@ -155,18 +155,53 @@ button,
...
+  &.usa-button-active,
+  &:focus,
+  &.usa-button-focus {
+    background-color: $color-white;
+    border: 0;
+  }
+}
+
+.usa-button-secondary-inverse:disabled,
+.usa-button-secondary-inverse-disabled {
```
msecret: Same as 173

src/components/02-buttons/buttons.njk
```diff
@@ -1,6 +1,21 @@
-
 <button class="{{ classes }}">Default</button>
 {% if (category != 'disabled') and (category != 'big') %}
-<button class="{{ classes }} usa-button-active">Active</button>
-<button class="{{ classes }} usa-button-hover">Hover</button>
+  <button class="{{ classes }} usa-button-hover">Hover</button>
+  <button class="{{ classes }} usa-button-active">Active</button>
+  <button class="{{ classes }} usa-button-focus">Focus</button>
+  {% if (category == 'secondary') %}
```
msecret: I think we're supposed to use `{%-` here to remove the whitespace? I might have that wrong though, I haven't done it yet on my PR.

donjo: took care of this one

src/stylesheets/elements/_buttons.scss
```diff
@@ -152,24 +156,48 @@ button,
...
+  &:active,
+  &.usa-button-active,
+  &:focus,
+  &.usa-button-focus {
+    background-color: $color-white;
+    border: 0;
+  }
+}
+
+.usa-button-secondar:disabled {
```
sawyerh: 👋  Just flagging that this one has a typo

AndreaSigz:
Thanks @sawyerh! @toolness @donjo

PR #2100, opened 2017-08-10
Better headless chrome tests

toolness:
This adds some neat stuff to our aXe and visual regression tests like:

* Running the visual regression tests on both small and large desktop sizes (until now, we ran them only on small desktops)
* Factoring out some of the code into separate modules to make everything more understandable and less spaghetti-like
* Making `spec/visual-regression-tester.js` an executable script that can be used to directly execute visual regression tests, rather than requiring developers to run `mocha` with special environment variables
* Actually documented how to use the tester in `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,6 +88,36 @@ bin)/gulp` instead of `gulp`.)
 Note that running the tests also requires an installation of
 Chrome v59 or higher (v60 if you're on Windows).
 
+#### Visual regression testing
+
+The Standards come with optional tooling for detecting visual regressions,
+which can be especially useful if you're refactoring CSS.
+
+These tests work by comparing current screenshots of the Standards' Fractal
+components to "golden" screenshots that represent what the components are
+supposed to look like.
+
+Golden screenshots are stored on your local development system *only*;
+they're not version controlled.
+
+To generate the golden screenshots, run:
+
+```
+node spec/visual-regression-tester.js --updateGolden
+```
+
+Then, make any CSS refactorings (or switch to a branch that has them).
+
+To compare the current state of your CSS to the golden screenshots, run:
+
+```
+node spec/visual-regression-tester.js
+```
+
+If the current screenshots don't match their golden counterparts, you will
+be directed to an HTML file that visually shows the differences between
+any conflicting screenshots.
+
 ### Building
 
 To build the `uswds` package in preparation for releases, run:
--- a/spec/axe-tester.js
+++ b/spec/axe-tester.js
@@ -1,0 +1,72 @@
+const fs = require('fs');
+const assert = require('assert');
+
+const AXE_JS = fs.readFileSync(__dirname + '/../node_modules/axe-core/axe.js');
+
+const AXE_CONTEXT = JSON.stringify({
+  exclude: [
+    // For some reason aXe takes a lot longer if it needs to dive into
+    // iframes with data: URIs. The content of these iframes is just for
+    // non-USWDS example content anyways, so just skip them to speed things
+    // up.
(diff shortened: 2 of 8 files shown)


toolness:
Erg, so this diff is kinda big, but it's mostly due to the refactorings... @msecret I'm happy to walk you through these on a video call, let me know if that'd be helpful!

spec/axe-tester.js
```diff
@@ -0,0 +1,72 @@
...
+    if (viols.length > 0) {
+      return Promise.reject(new Error(
+        `Found ${viols.length} aXe violations: ` +
+        JSON.stringify(viols, null, 2) +
+        `\nTo debug these violations, install aXe at:\n\n` +
+        `  https://www.deque.com/products/axe/\n`
+      ));
+    }
+    return Promise.resolve();
+  });
```
msecret: Does this whole promise block need a catch?

spec/chrome-fractal-tester.js
```diff
@@ -0,0 +1,114 @@
...
+      headless ? '--headless' : '',
+    ],
+  });
+}
+
+function getRemoteChrome () {
+  const info = urlParse(REMOTE_CHROME_URL);
+  if (info.protocol !== 'http:')
+    throw new Error(`Unsupported protocol: ${info.protocol}`);
+  return new Promise(resolve => {
```
msecret: Can this just return a `Promise.resolve`?

CONTRIBUTING.md
```diff
@@ -88,6 +88,36 @@ bin)/gulp` instead of `gulp`.)
...
+```
+node spec/visual-regression-tester.js --updateGolden
+```
+
+Then, make any CSS refactorings (or switch to a branch that has them).
+
+To compare the current state of your CSS to the golden screenshots, run:
+
+```
+node spec/visual-regression-tester.js
```
msecret: Should we add npm script commands for these just got consistent

toolness: Hmm, my only concerns about that are that folks are likely going to want to run this with arguments. They can run the tester with `--help` to get CLI help, and as with mocha, they can run it with `-g <pattern>` to only run tests that match a certain pattern, which is particularly useful because these tests take longer to run than pure unit tests.

While passing extra arguments to npm/yarn *can* be done with  `--`, it adds extra typing and isn't very discoverable. With npm, updating the golden screenshots would become `npm run visual-regression-tester -- --updateGolden` which is more of a mouthful. Also, unless we come up with a shorter name, `visual-regression-tester` takes long to type and can't be easily autocompleted with <kbd>Tab</kbd> like `spec/visual-regression-tester.js` can.

Another annoying thing with npm < 5 (but fixed in yarn and npm 5) is the huge amount of debug output that appears when something fails, which this command is frequently expected to.

So, I guess I just don't feel like this is a very good fit as an `npm` script, but if you feel strongly about it, we can convert it over.

spec/axe-tester.js
toolness: This is just the [`runAxe`](https://github.com/18F/web-design-standards/pull/2100/files#diff-50474fe1a3bbc30931f2806609b64f13L84) function from the old code, and it's expected that the caller catch anything that is thrown.  Right now, the promise is simply [passed on to Mocha](https://github.com/18F/web-design-standards/pull/2100/files#diff-97c5e1389ae6185cf3b249aa29a07828R91) so Mocha will nicely display any exceptions that are thrown.

spec/chrome-fractal-tester.js
toolness: Hmm, this is just old code that's been moved to a new file, but that's a good point, I'm not sure why this isn't just a `Promise.resolve`! I'll change it.

PR #2105, opened 2017-08-15
WIP Change the focus state color

msecret:
## Description

## Additional information

* Screenshot images
* Notes


msecret:
So I experimented with the dashed outline focus and here's what I found:

We can apply positive padding and negative margin to inline elements to add some space between the element and the outline while keeping the element as a whole from moving:

screen shot 2017-08-16 at 3 03 20 pm
screen shot 2017-08-16 at 3 03 30 pm

For block level elements, like buttons, this becomes a problem, because they're whole structure is already defined by their large padding. So if if set the custom padding they become too small:

screen shot 2017-08-16 at 3 07 23 pm

Here is what they look like without the padding:

screen shot 2017-08-16 at 3 08 29 pm

msecret:
There is also the `outline-offset` CSS3 property, which adds exactly what we want: space between the outline and the element.

screen shot 2017-08-16 at 3 16 22 pm

The `outline-offset` property does not have any support in IE, but has support in the rest of the browsers. Using it in IE would probably mean there's not space between the outline and the element.

msecret:
Closing this in favor of new way of doing it

PR #2112, opened 2017-08-17
Change to single focus state

msecret:
In order to start from the blank single focus state on all
elements.

## Description

## Additional information

* Screenshot images
* Notes

### DIFF
--- a/src/stylesheets/elements/_buttons.scss
+++ b/src/stylesheets/elements/_buttons.scss
@@ -146,7 +146,7 @@ button,
   &:active,
   &.usa-button-active,
   &:focus,
-  &.usa-button-focus {
+  &.usa-focus {
     background-color: $color-gray-lighter;
     border: 0;
     box-shadow: none;
@@ -165,7 +165,7 @@ button,
   &:active,
   &.usa-button-active,
   &:focus,
-  &.usa-button-focus {
+  &.usa-focus {
     background-color: $color-white;
     border: 0;
   }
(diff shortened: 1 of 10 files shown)


src/stylesheets/elements/_buttons.scss
```diff
@@ -45,12 +44,6 @@ button,
     text-decoration: none;
   }
 
-  &:focus,
-  &.usa-button-focus {
```
donjo: The removal of some of these classes makes it so the focus state isn't being show on the fractal example like it should (example of how it should look below)

screen shot 2017-08-21 at 2 13 14 pm

Doesn't have to be done in the same fashion as the past but some `njk` files may need to be adjusted otherwise.

msecret: What if we just did a `usa-focus` class that could work for focus state on any element? Or should there but a `button-focus` class specifically?

donjo:
Looks great 👍

sawyerh:
Curious to hear the thoughts behind this switch to a light gray focus state. Was there research showing the default blue focus state was being missed? Is there evidence that a dashed gray outline is better? Not saying I disagree with the choice, but would be great to hear more about why the change?

rtwell:
@sawyerh here's one example that came to mind: https://github.com/18F/web-design-standards/issues/1655 

Blue-on-blue for the focus state isn't the easiest thing to see, especially when the focus has no padding to help separate it from the element. Our choice for this iteration accomplished a few things:
1) the dotted styling helped it feel different from the rest of the elements and functionalities on the page
2) the middle gray color was the most flexible choice to sit on either a dark background or a light background
3) offsetting the focus from the elements made it easier to see, as well

sawyerh:
@rtwell Got it! Thanks for the context 👍

PR #2116, opened 2017-08-21
Update form templates to new visual design

msecret:
## Description

Changes the form templates to include "Optional" tags on form fields rather then "Required" tags. 

I added a input-optional class to be used in similar fashion as the input-required class. I think keeping the two consistent like this is better.

## Additional information

Requires https://github.com/18F/web-design-standards/pull/2115 to be complete as that PR modifies the dropdown menu which is part of this issue.

screen shot 2017-08-22 at 11 54 55 am
screen shot 2017-08-22 at 11 54 45 am

### DIFF
--- a/src/components/07-form/templates/address-form.njk
+++ b/src/components/07-form/templates/address-form.njk
@@ -5,7 +5,8 @@
       <label for="mailing-address-1">Street address 1</label>
       <input id="mailing-address-1" name="mailing-address-1" type="text">
 
-      <label for="mailing-address-2">Street address 2 <span class="usa-additional_text">(Optional)</span></label>
+      <label for="mailing-address-2" class="usa-input-optional">
+        Street address 2</label>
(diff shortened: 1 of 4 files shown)


src/components/07-form/templates/address-form.njk
```diff
@@ -17,7 +18,7 @@
         <div class="usa-input-grid usa-input-grid-small">
           <label for="state">State</label>
           <select id="state" name="state">
-            <option value></option>
+            <option value>- Select -</option>
```
donjo: does this need to be set to `option value=""`? I saw that's what it was in a different PR for the `dropdown.njk` file and I'm not sure which is right.

msecret: I realized, I set it to `value=""` in the other file, but this already existed as `value`. So I think this one is correct and I have to change the other one.

PR #2113, opened 2017-08-18
Disable form submission on fractal site.

toolness:
Fixes #2101.

PR #2119, opened 2017-08-22
Release 1.3.1

toolness:
This is a candidate for a patch release that just cherry-picks #2092 and #2113.


toolness:
Ok, @donjo told me that CircleCI's caching system might be responsible for the build failing until I committed fcd799533a79c374adeee0bb475bc9750d9c88c7 here.  It looks like that commit fixed everything, though, so that's cool.

PR #2122, opened 2017-08-22
Add variants for labels to be consistent

msecret:
All components should make use of variants if they have them to
stay consistent.


msecret:
I wonder if there's some sort of better njk solution. Or adding the empty space in the config file would technically work.

PR #2123, opened 2017-08-22
[DO NOT MERGE YET] Merge master back into develop

toolness:
Hmm, because this just cherry-picked `develop` changes to begin with, this should be a straightforward merge with just the version change to `package.json`. I'm unclear on why github thinks that the cherry-picked changes aren't already in `develop`, so until I figure that out, um, please don't merge this I guess.


donjo:
@atul sounds good

PR #2131, opened 2017-08-24
Visual design header

msecret:
Update the header to new visual design.

Fixes https://github.com/18F/web-design-standards/issues/2086


msecret:
@rtwell Please take a look. https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/ms-visual_design_header/components/detail/header--basic.html

I've found a lot of inconsistencies with whats even in the "Old"/original mockups and what's actually part of the header, so I wasn't really sure anymore what I should update. Especially the mobile view, it seems like a lot of elements are missing. Should I add things on to get it as close to your new mockup as possible or should we re-access?

msecret:
I fixed part of the problem where the nav actually blocks the logo. In terms of the nav moving down, I'm not sure much can be done there because there's just physically not enough room for both the nav and the logo to fit.

src/components/08-navigation/nav-primary.njk
```diff
@@ -9,11 +9,25 @@
       <span>{{ link.text }}</span>
     </button>
     <ul id="{{ nav.id_prefix }}{{ link.id }}" class="usa-nav-submenu{% if nav.mega %} usa-megamenu usa-grid-full{% endif %}">
-      {% for child in link.links %}
-      <li>
-        <a href="{{ child.href }}">{{ child.text }}</a>
-      </li>
-      {% endfor %}
+      {% if nav.id_prefix == "basic-" or nav.id_prefix == "extended-" %}
```
donjo: could you do the `-%}` close style on these if/else statements in the njk files so the generated html samples in Fractal don't have a bunch of extra blank lines. 

There's an example in the button PR of how it works here https://github.com/18F/web-design-standards/pull/2098/files#diff-c95f7975ef5ea8e3cfc3fb942baa9055R2

src/components/08-navigation/nav-primary.njk
```diff
@@ -1,5 +1,5 @@
 <button class="usa-nav-close">
-  <img src="{{ uswds.path }}/img/close.svg" alt="close">
+  <img src="{{ uswds.path }}/img/close-primary.svg" alt="close">
```
donjo: This doesn't seem to be working 
screen shot 2017-08-31 at 11 31 05 am

msecret: On which one of the nav pages are you seeing the button not appearing? I had this problem locally at first, and I ended up having to run `npm run build` every time I changed an image, otherwise they wouldn't be copied over to the dist folders correctly. I'm seeing this on the Federalist for the basic nav:

screen shot 2017-08-31 at 11 15 34 am

msecret:
The current nav on [components site](https://components.standards.usa.gov/components/detail/header--basic-mega.html) doesn't actually have any visual indicator when it's active, so I'm not really sure what that's supposed to look like. Maybe it had a visual indicator that broke at some point? I can add one if I see what it looks like.

donjo:
@msecret 
header - mobile - expand

This is the screenshot I meant to attach for that one. I don't think we ever had this style in the past and it was something new that @rtwell added.

msecret:
@rtwell when we discussed the header, we didn't talk about adding that visual indicator. Should we add it at this point?

src/components/08-navigation/nav-primary.njk
donjo: hmmm it's magically working now

PR #2151, opened 2017-09-28
Release version 1.4.0

toolness:
## Accessibility improvements

* The search button's submit text is now visible to screen readers on all viewport sizes (#2065).
* A number of accessibility improvements have been made to the fractal site (#2057, #2064).

## Visual changes

* Alerts now have a high-contrast left border to help differentiate them from each other and surrounding content (#2095).

  image

* Buttons' corners are more round to help them feel more button-y and to differentiate them from other elements, like labels (#2098). The spacing of the type inside the buttons is now balanced.

  image

* The focus state has been updated to address concerns about contrast (#2112). Users told us the blue focus state on the common blue background color was hard to see. Our update does the following:

  1. Chooses a neutral gray that has higher contrast that works well on dark backgrounds and light (big caveat here: while there is no perfect focus state color for every site, one goal we had here is to help designers _think critically about what focus state works best for the site they are designing_).

  2. We offset the focus state a bit in order for the type to breathe a wee bit and help make it more noticeable, especially on objects that are flush the edge of a browser window.

  image

* Dropdown menus have been updated (#2115). We changed the following:

  1. We changed the icon from a down-caret to a selector. 

  2. The default state for a dropdown menu should indicate that the menu contains a list of options, rather than the first item on the list (see the example below for an illustration of this).

  image

* A new `usa-input-optional` class has been added in favor of `usa-input-required` (#2116). The latter is now deprecated and will be removed in a future release of the Standards, due to evidence that labeling optional fields over required ones is a [best](http://uxmovement.com/forms/why-users-fill-out-less-if-you-mark-required-fields/) [practice](https://www.nngroup.com/articles/web-form-design/). **If you feel strongly about this issue, however, please contact the Standards team, as we may un-deprecate `usa-input-required` if a compelling argument exists.**

* Checkboxes have been updated (#2117). Specifically:

  1. Spacing adjustments have been made for a cleaner look.
  2. The widget has slightly less rounded corners.
  3. The disabled state looks more disabled (note that accessibility contrast requirements do not apply to disabled elements).

  image

* Labels have been updated (#2118). Specifically:

  1. They have less rounded corners to look less button-y.
  2. We've changed the labels from primary-blue to gray to differentiate them from buttons.
  3. We've adjusted the spacing to feel more comfy.

  image

* The footer has been visually revamped (#2130). Specifically:

  1. Primary blue links didn't have enough contrast on the light gray background.
  2. The agency logo has been scaled down to a medium size for the medium footer.
[...]

PR #2152, opened 2017-09-29
Fix disabled buttons

toolness:
**Note: This is a PR against #2151, not `develop`!**

[ 😎 PREVIEW ](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/fix-disabled-buttons/components/detail/buttons--primary.html)

This attempts to fix the problems with disabled buttons explained in https://github.com/18F/web-design-standards/pull/2151#issuecomment-332958442.

There's multiple ways we can go about fixing this:

1. Bring back the `usa-button-disabled` class and all its variants, which seem to have been removed in #2098 (where this regression seems to have originated).  I think this may be necessary to classify our release as a minor version update, since otherwise it's a breaking change, as clients will have to change the way they mark up disabled buttons.

2. Don't bring back any of the classes that were removed, and modify our CSS/fractal component markup as needed to make them work with the `disabled` attribute and the button's base class instead. This will result in a cleaner markup API but it's a breaking change, so I recommend not doing it unless we upgrade our major version (i.e., release 2.0).

3. Attempt to have the best of both worlds by introducing the new markup API described in (2), but keeping the old `usa-button-disabled` class and its variants mentioned in (1) and documenting them as deprecated.

@donjo @msecret thoughts on this?

### DIFF
--- a/src/components/02-buttons/buttons.njk
+++ b/src/components/02-buttons/buttons.njk
@@ -4,11 +4,18 @@
 {%- endif %}
 
 <button class="{{ classes }}">Default</button>
-{% if (category != 'big') -%}
+{% if (category != 'disabled') and (category != 'big') -%}
   <button class="{{ classes }} usa-button-hover">Hover</button>
   <button class="{{ classes }} usa-button-active">Active</button>
   <button class="{{ classes }} usa-focus">Focus</button>
   <button class="{{ classes }}" disabled>Disabled</button>
+  {% if (category == 'primary') -%}
+    <button class="usa-button-disabled" disabled>Disabled (deprecated)</button>
+  {%- elif (category == 'secondary') -%}
+    <button class="usa-button-secondary-disabled" disabled>Disabled (deprecated)</button>
+  {%- elif (category == 'secondary-inverse') -%}
+    <button class="usa-button-secondary-inverse-disabled" disabled>Disabled (deprecated)</button>
+  {%- endif %}
 {%- endif %}
 
 {% if (classes == 'usa-button-secondary-inverse') %}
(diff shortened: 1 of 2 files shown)


toolness:
Also, as far as I can tell, we don't currently have `:disabled` styles for the following button styles:

* Default (not *primary* but default, i.e. buttons w/o any `usa-button` class)
* Secondary inverse
* Aqua
* Gray
* Red

donjo:
@toolness if I recall the discussions @rtwell and I had when we were doing this originally, those button types should all have the same disabled button style (the solid light gray) as the primary button so they'll need whatever css changes are necessary to make that style show up for those button types

toolness:
Oh ok. Should I just assign that style to all `:disabled` buttons, as well as `usa-button-disabled` and its variants, to keep backwards compatibility with old markup for now? (i.e., option 3 in the original description of this PR)

toolness:
Ok I think I am almost done with implementing option 3. The only thing left seems to be the disabled variant for the secondary inverse style:

> screen shot 2017-09-29 at 11 58 03 am

Unfortunately, the disabled style looks very similar to (perhaps identical to) the hover style.  Any suggestions @donjo @rtwell ?

donjo:
@toolness for that one I think we decided just using a fully solid grey disabled button with black text (even though all the others are outline)

rtwell:
disabled should look disabled, and color contrast isnt an issue here. imo the button outline and text should be dark, like so:
image

toolness:
Ok, how's this?

>

rtwell:
@toolness love it!

PR #2160, opened 2017-10-04
update password guidance

donjo:
Updates the password guidance as referenced in 18F/web-design-standards-docs#371

Because the previous example did not align with NIST password guidelines we decided that the Standards should not take an opinion on what the password reset guidance should be, just where you could potentially provide that information. 

We could link to additional resources for formulating password guidelines on the docs site page for password reset.

### DIFF
--- a/src/components/07-form/templates/password-reset-form.njk
+++ b/src/components/07-form/templates/password-reset-form.njk
@@ -15,12 +15,7 @@
     </div>
 
     <label for="password">New password</label>
-    <input id="password" name="password" type="password"
-      aria-describedby="validation_list"
-      data-validate-length=".{8,}"
-      data-validate-uppercase="[A-Z]"
-      data-validate-numerical="\d"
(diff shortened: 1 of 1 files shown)


toolness:
Hmm, the new example is funky to me b/c the example is still implying that people *should* provide validation feedback, but because the example doesn't use _real_ validation rules, we don't use the validator component in our example... honestly i don't think it's a great idea to remove the validator component entirely then, b/c it's kind of like us just removing a tool that we still encourage people to use, but that we refuse to use in our examples.

So one alternative might be to use the new copy provided in this PR, but also *keep* the validator component (it's [documented](https://standards.usa.gov/components/form-controls/#validation) in a different section of the docs), and link to it from the "update password" example, mentioning that it might be useful in providing live feedback for whatever their password requirements happen to be. That way devs can "connect the dots" and use the validator component in their own concrete password validation implementations, even though we don't use it in our example.

Does that make sense?

PR #2167, opened 2017-10-06
Update PULL_REQUEST_TEMPLATE.md [WIP]

maya:
Since we run `npm test` and aXe (for accessibility) as part of our CI, I wonder if these manual tests are irrelevant, since they'll flag here automatically. Thoughts @donjo @thisisdano?


donjo:
I'm up for removing them if it makes things easier and a bit less intimidating. One thing to note is that the CI doesn't build on PRs from folks outside the organization so maybe that's why they were put there in the first place (see [this PR](https://github.com/18F/web-design-standards/pull/2020) for an example).

maya:
Closing this for now since public contributors don't have the tests run. We can come back to this at a later point.

Issue #2171, opened 2017-10-08
Button changes, no documentation


stphnwlkr:
It appears in 1.4 there were some significant changes to the buttons and I cannot find any reference. It appears usa-button-outline became usa-button-secondary, and usa-button-secondary became usa-button-red. I am not a fan of the change as it moves away from the standardized color naming, but I need to know if this is going to be permanent so that I can make the appropriate updates in several projects.

joshbruce:
I actually appreciate the change with primary and secondary (was already using them that way. The specific change I appreciate is the separation between intent (function) and straight color (see `variables`).

Though I am considering a button class in my extension:

- `usa-button-destructive`
- (and possibly, though not likely) `usa-button-destructive-outline`

Then the style and naming make sense from a UX perspective. 

- Primary action.
- Secondary action.
- Destructive action.

Could translate to:

- Save.
- Cancel.
- Delete.

Respectively.

Didn’t even know about `usa-button-red` - thanks for that.

evetsreklaw:
I am of the school that CSS is about styling and not about user interaction. This way you simplify and minimize they styling you need (write once use in several places). Since I am not building it myself, I just need to know what the road ahead will be so I can make the necessary changes. In my opinion, moving from 1.3 to 1.4 should not be breaking basic styling and certainly not without documentation.

joshbruce:
That's fair. Think I come from the school that colors should be named colors. :)

Given the changes to the `variables.scss` separating the functional names from the color names - and having the functional names reference the color names I wouldn't be surprised if 2.0 saw the functional color name become deprecated entirely; thereby, having the palette be divided between primary, secondary (and possibly tertiary).

evetsreklaw:
I guess it would have helped if I looked at variables.scss in detail. I think it answers my question about the road ahead. I am definitely not opposed to colors being named colors. I agree with your assessment for 2.0.

maya:
Thanks for bringing this to our attention. I could see how this could be confusing without any documentation. We'll take a look at this issue and get back to you.

evetsreklaw:
I just went through my code and made the updates. I would like to make a recommendation that the terms "primary" and "secondary" should only relate to hierarchy and never be related to color. The button-primary and button secondary are not logical names (IMO). Since default and primary are the same, do away with primary and just use usa-button and usa-button-inverse. Secondary-inverse should just be use-button-black-outlined.

maya:
From @thisisdano in #2237:

[...]

PR #2175, opened 2017-10-11
[Hero component]: Change button to use primary button

maya:
Changes the hero template from using `usa-button-secondary` to just `usa-button` (primary). This button used to be red but some changes were made to the buttons that made  `usa-button-secondary` into the outline button.

Preview: https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/fix-landing-button/components/detail/layout--landing.html

### DIFF
--- a/src/components/hero/hero.njk
+++ b/src/components/hero/hero.njk
@@ -13,7 +13,7 @@
       {% endif %}
 
       {% if hero.button %}
-      <a class="usa-button usa-button-big usa-button-secondary" href="{{ hero.button.href }}">{{ hero.button.text }}</a>
+      <a class="usa-button usa-button-big" href="{{ hero.button.href }}">{{ hero.button.text }}</a>
       {% endif %}
     </div>
   </div>


maya:
@donjo some questions came up around the buttons that @thisisdano and I wanted to square away first in #2171. We went with the primary button bc it was more stable.

thisisdano:
@donjo @maya So, this change here is a result of the confusion over the way we changed buttons in `1.4.0` — the confusion is mostly outlined in #2171, but I will recap, mostly for my own benefit, because... it's complicated 🤷‍♀️ . So, as of `1.4.0`:

- `button-primary` is a non-outlined button and `button-secondary` is an outlined button with a white background. 
- `primary` and `secondary` no longer refer to colors (at least for buttons)
- we use named colors for button backgrounds, of which we seem to have three: `aqua`, `red`, and `gray`... and a blue color (not named, but `$color-primary`) as a default, if no color is specified
- there seems to be no class for changing outline colors of `button-secondary`
- we seem to still use `$color-primary` and `$color-secondary` as functional color variables, _including_ for styling the buttons themselves

This is confusing. And not documented.

It feels like a problem to introduce what could be considered a breaking change in a point release.

All of that said, what do we do about the issue at hand? We need to fix this button so it's actually legible. And it would make sense to fix it in a way that's less dependent on the `1.4.0` changes, just so we can address those changes without introducing more tech debt into our own code.

Keeping the button styled with `.button-primary.button-big` seems like something that's safe and future-forward, even if we temporarily lose the red background.

maya:
Merged this in so we can get the fix landed.

PR #2179, opened 2017-10-12
Select component: Pixel-fit arrow-both.svg

sawyerh:
This is super picky 😅 but makes the icon look less blurry, especially on retina screens.

image


maya:
@sawyerh not picky at all! I'm working on a related issue here #2178 and uploaded a 2x png. Does that resolve it or does it need more work?

maya:
Ah, just realized you shared an SVG, not a PNG. Wonder how it compares to the new SVG I uploaded: https://github.com/18F/web-design-standards/pull/2178/files#diff-c92e431a8d1091d484058af08dee8196

sawyerh:
@maya ah I missed that one. It looks like the SVG in #2178 is still a little misaligned with the pixel grid:

image

maya:
Got it, I pulled in your code to my PR too: https://github.com/18F/web-design-standards/pull/2178/files#diff-c92e431a8d1091d484058af08dee8196

PR #2191, opened 2017-10-25
Fix focus state on hover

### LINKED ISSUE #2184: [Links]: Focus disappears on hover
Focus disappears on hover and it shouldn't.

maya:
Fixes focus state by adding it after a link i.e. `a:focus` bc `*:focus` was causing it to disappear on hover.

Moves focus code from `base.scss` to `typography.scss`. 
- Is there a reason it should be in `base.scss` that I'm not aware of?

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/fix-focus)

Fixes #2184.

## Ways to improve
We can make focus a mixin so the 2 CSS declarations are not in two separate places or add `a:focus` into the other block, but I also like for link behavior to be together in the link code block.

### Question 
Is `*:focus` working on any other component?

### DIFF
--- a/src/stylesheets/core/_base.scss
+++ b/src/stylesheets/core/_base.scss
@@ -28,3 +28,9 @@ body {
 [hidden] {
   display: none !important; /* stylelint-disable-line declaration-no-important */
 }
+
+*:focus,
+.usa-focus {
+  outline: $focus-outline;
+  outline-offset: $focus-spacing;
+}
--- a/src/stylesheets/core/_base.scss
+++ b/src/stylesheets/core/_base.scss
@@ -31,6 +31,5 @@ body {
 
 *:focus,
 .usa-focus {
-  outline: $focus-outline;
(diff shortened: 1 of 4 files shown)


donjo:
My guess is that it was in `_base` because it's applying to input fields, buttons, nav items, etc. so I'm not sure if `_typography` is the place for it (though makes sense for the `a:focus` rule). I'm not sure `_base` is the most descriptive or expected place to find the rule but I'm not sure there is a better one for that.

maya:
@donjo that makes sense. Moved it back to `base.scss` in e22bc97 and leaving the `a:focus` rule in typography. Do you think it should be a mixin since it's in two places?

donjo:
@maya I like that idea!

maya:
@donjo done in bb70ee2. Also, had to move `utilities.scss` import ahead of `base.scss` in the import order in fe75f10 since it was referencing it.

PR #2192, opened 2017-10-25
[Tables]: Fix table style and accessibility

### LINKED ISSUE #1495: Tables - captions
https://standards.usa.gov/tables/

In the accessibility section you talk about using `<caption>` to add a title to a table.
modeling in the examples would be useful as the `<h6>` text above the tables are the "titles" of the tables. 
Other elements (e.g. th) and attributes (e.g. scope) are modeled. So should `<caption>`
`<caption>Bordered Table</caption>`
`<caption>Borderless Table</caption>`

also consider making the titles of the tables an `<h2>` or `<h3>` rather than `<h6>`, 
`<caption><h2>Bordered Table</h2></caption>`
`<caption><h2>Borderless Table</h2></caption>`

### LINKED ISSUE #1621: Table element: min-width should not be set
Adding a table to a page using the Web Design Standards should not take up the entire available width. There are many tables that present only a small amount of tabular information. I'm not sure how readable the tables are when they take up the entire width.

table-standards

Tables should expand to fit the width; no further.

Would it be possible to drop the `min-width: 100%` styling?

maya:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/fix-tables)

## Visual design
- Reduces top/bottom spacing on table cells to match the latest visual designs. 
- Removes the gray background on borderless table.
- Removes `min-width: 100%` from tables so it doesn't take up the whole width for smaller data. Fixes #1621. Smaller table example:
screen shot 2017-10-25 at 5 02 06 pm

## Accessibility
- Fixes accessibility issue where the table title should be a caption (Fixes #1495). 
- Left aligns `<caption>` and adds a bottom margin

**Question**: How do we style the `<caption>`? 

I was going back and forth whether style the caption to look like the h6 style, add an h6 into the caption to give it that style (but I don't want to promote people to copy and paste this). I ended up just floating it left and adding a reasonable bottom margin and leaving the basic text styling. Does the left aligned caption make sense or should we leave the browser default centered? @thisisdano

### DIFF
--- a/src/components/04-tables/tables.njk
+++ b/src/components/04-tables/tables.njk
@@ -56,7 +56,7 @@
       </tr>
       <tr>
         <th scope="row">Declaration of Sentiments</th>
-        <td>MadeA document written during the Seneca Falls Convention outlining the rights that American women should be entitled to as citizens.</td>
+        <td>A document written during the Seneca Falls Convention outlining the rights that American women should be entitled to as citizens.</td>
         <td>1848</td>
       </tr>
       <tr>
--- a/src/components/04-tables/tables.njk
+++ b/src/components/04-tables/tables.njk
@@ -1,7 +1,5 @@
-
-  <h6>Bordered Table</h6>
-
   <table>
+    <caption>Bordered table</caption>
     <thead>
       <tr>
         <th scope="col">Document title</th>
@@ -33,9 +31,8 @@
     </tbody>
   </table>
 
-  <h6>Borderless Table</h6>
-
   <table class="usa-table-borderless">
+    <caption>Borderless table</caption>
     <thead>
       <tr>
         <th scope="col">Document Title</th>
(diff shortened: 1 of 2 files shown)


thisisdano:
In general, I don't like mixing centered heds with left-justified heds, but the "caption" here is maybe an exception, because the caption falls directly above the table heds, and the cell inset makes left alignment tricky. I think I'd go for something like your `h6` suggestion — something allcap and letter-spaced (to something like `0.05em`), and also centered.

screen shot 2017-10-30 at 11 50 19 am

thisisdano:
Good changes in general, tho! 👏

maya:
Going with h5 styles per @thisisdano's recommendation. Also adds a flush left alignment for the first heading item in the borderless table.

joshbruce:
That `caption` element is also slightly different than the `thead` element in that it mimics the purpose of a figure caption, only for tables. Of course, we could debate on the justification of text for a figure caption as well. ;)

PR #2194, opened 2017-10-26
Reduce ambiguity in footer-big secondary links

thisisdano:
This reduces the line height of secondary links in the footer, and compensates vertical rhythm with with padding so multi-line links have less leading than the space between adjacent links.

**Note: this ALSO addresses an issue where changing the import order in the `uswds.scss` broke the grid — and maybe other stuff too.**

line-height

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -357,8 +357,8 @@ li.usa-footer-primary-content {
 
     li:not(.usa-footer-primary-link) {
       line-height: 1.35;
-      padding-bottom: rem(6px);
-      padding-top: rem(6px);
+      padding-bottom: .35em;
+      padding-top: .35em;
     }
 
     .usa-footer-primary-link {
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -341,7 +341,7 @@ li.usa-footer-primary-content {
   }
 
   ul {
-    padding-bottom: 2.5rem;
+    padding-bottom: rem(24px);
 
     @include media($medium-screen) {
       padding-bottom: 0;
@@ -355,8 +355,10 @@ li.usa-footer-primary-content {
       }
     }
 
-    li {
-      line-height: 2em;
+    li:not(.usa-footer-primary-link) {
+      line-height: 1.35;
+      padding-bottom: rem(6px);
+      padding-top: rem(6px);
     }
 
     .usa-footer-primary-link {
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -357,8 +357,8 @@ li.usa-footer-primary-content {
 
     li:not(.usa-footer-primary-link) {
       line-height: $heading-line-height;
-      padding-bottom: .35em;
-      padding-top: .35em;
+      padding-bottom: 0.35em;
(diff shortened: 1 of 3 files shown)


src/stylesheets/components/_footer.scss
```diff
@@ -355,8 +355,10 @@ li.usa-footer-primary-content {
       }
     }
 
-    li {
-      line-height: 2em;
+    li:not(.usa-footer-primary-link) {
+      line-height: 1.35;
```
maya: Can we use one of our line height variables for this? https://github.com/18F/web-design-standards/blob/develop/src/stylesheets/core/_variables.scss#L24-L26

src/stylesheets/components/_footer.scss
```diff
@@ -341,7 +341,7 @@ li.usa-footer-primary-content {
   }
 
   ul {
-    padding-bottom: 2.5rem;
+    padding-bottom: rem(24px);
```
maya: This `rem(24px)` mixin technique was removed from use in https://github.com/18F/web-design-standards/pull/2006 due to Bourbon deprecation and loss of support in Bourbon 5.0, so I think we shouldn't use this.

maya:
Adding a preview link: 

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/dw-footer-leading)

maya:
>Note: this ALSO addresses an issue where changing the import order in the uswds.scss broke the grid — and maybe other stuff too.

Not sure I totally understand the problem: can you describe how the grid order was not working and how it relates to this PR. I'm also moving the `utilities.scss` up in #2192 so I want to make sure we're in sync.

src/stylesheets/components/_footer.scss
maya: When I use `$heading-line-height` or `1.3`, it seems to have the same effect / doesn't change the size, so maybe we can stick with that?

src/stylesheets/components/_footer.scss
thisisdano: I remember when this happened, and I'm comfortable removing this function, _but_ I still think the Bourbon deprecation is a mistake and writing in raw `rem` makes the system too dependent on its base size declaration (10px). That is, if the System has to play nice with an existing codebase that's already defined its base size, we're SOL. Adding our own function to replace the soon-to-be-deprecated Bourbon function would make the whole system more flexible. But perhaps this is an argument for another day!

src/stylesheets/components/_footer.scss
thisisdano: I'll use `$heading-line-height` here. Forgot those existed.

src/stylesheets/components/_footer.scss
```diff
@@ -355,8 +355,10 @@ li.usa-footer-primary-content {
       }
     }
 
-    li {
-      line-height: 2em;
+    li:not(.usa-footer-primary-link) {
+      line-height: $heading-line-height;
+      padding-bottom: .35em;
+      padding-top: .35em;
```
maya: @thisisdano also, can you add a zero before the decimal, so it's `0.35em`?

thisisdano:
@maya I just addressed that last review comment

maya:
LGTM, good work @thisisdano! 🎉

PR #2198, opened 2017-10-27
Consistent form control labels

### LINKED ISSUE #2183: [Form controls]: Required and optional labels are inconsistent
We're using "Label - _optional_" and "Label (*Required)" and should use one or the other to be consistent. 

Need to decide:
1. Dash or parentheses
2. _italic_ or normal font
3. lowercase or Sentence case

Note: there was some discussion around italics being difficult for people with dyslexia here: https://github.com/18F/web-design-standards/pull/1504

donjo:
Fixes #2183 

Applying the knowledge from #1504 I adjusted the content supplied by the `usa-input-required` and `usa-input-optional` to be as similar as possible.

- I went with lowercase for both words because an uppercase `O` next to a `(` is hard to read.
- Kept the `*` in required for the accessibility reasons
- Colors stayed the same. 
- Removed italics from optional

### DIFF
--- a/src/stylesheets/elements/_inputs.scss
+++ b/src/stylesheets/elements/_inputs.scss
@@ -88,13 +88,12 @@ select {
 
 .usa-input-required:after {
   color: $color-secondary-darkest;
-  content: ' (*Required)';
+  content: ' (*required)';
 }
 
 .usa-input-optional:after {
   color: $color-gray-medium;
-  content: ' - optional';
-  font-style: italic;
+  content: ' (optional)';
 }
 
 label {


thisisdano:
lgtm. this is all in line with gov.uk guidance, which makes sense to me. http://govuk-elements.herokuapp.com/form-elements/#form-optional-fields

couple Qs:

- Should we include the `(optional)` text as pseudocontent? There's perhaps a modest accessibility gain to hardcoding it in [link](https://tink.uk/accessibility-support-for-css-generated-content/) — and I have _slight_ queasiness about pseudocontent in general.
- Does the standard field have to be `usa-input-required`? Might `usa-input` be just as clear and more in line with our guidance to consider required the default?

maya:
>- Does the standard field have to be `usa-input-required`? Might `usa-input` be just as clear and more in line with our guidance to consider required the default?

The standard field is just `<input>`, `.usa-input-required` if for the required styling.

thisisdano:
Do we have better reasons than gov.uk for why we explicitly mark both required fields and optional fields?

> - only ask for the information you absolutely need
> - if you do ask for optional information, mark the labels of optional fields with ‘(optional)’
> - don’t mark mandatory fields with asterisks

maya:
No, we only mark optional and that's what we should in the docs site. The required is just left in the CSS so not to break anything for folks that were using it before. Also, there's a use case where some people will use the required state if 10 optional fields and 1 required field. So we have it there but don't promote it.

maya:
Re: hardcoding vs. CSS content, that may be a question for @nickbristow. It was hardcoded when I was on the project so assume it was done for developer convenience and bc screen readers read content, but I haven't looked into the minutia of this.

maya:
Merging as visual design and front-end is 👍 and we can continue the convo here or in another issue.

thisisdano:
Re: required — I saw this

> - Kept the * in required for the accessibility reasons

and assumed that meant that we were marking required fields with an asterisk, which is counter to the gov.uk guidance:

> don’t mark mandatory fields with asterisks

I'm definitely not saying we need to follow their guidance by default, but it points to a difference of opinion around which I don't know the reasoning.

thisisdano:
Re: pseudocontent
https://tink.uk/accessibility-support-for-css-generated-content/ is my reference for the accessibility implications of pseudocontent, but those aren't the only issues I have with pseudocontent. That is: is content _content_ (markup) or _styling_ (css)? Is this a distinction that's too old-school?

PR #2199, opened 2017-10-27
Finishing big button polish

### LINKED ISSUE #2144: Fix big button

donjo:
Fixes #2144 

When 1.4 was created we missed a few changes on the big button class necessary for it to match the designs below. 

screen shot 2017-10-27 at 4 04 31 pm

Changes: 
- Font size increased to 24pt
- Border radius doubled from default
- Horizontal padding increased (in the Sketch file the big button is 30px wider on each side than the standard button so I applied that here. However, our current standard buttons are not as wide in code as they are in Sketch. I'm inclined to default to the narrower presentation we're currently running with)

cc/ @thisisdano

### DIFF
--- a/src/components/hero/hero.config.yml
+++ b/src/components/hero/hero.config.yml
@@ -5,10 +5,7 @@ context:
   hero:
     callout: 'Hero callout:'
     title: Call attention to a current priority
-    link:
-      text: Link to more about that priority
-      href: 'javascript:void(0)'
+    paragraph: Support the callout with some short explanatory text. You don't need more than a couple of sentences.
     button:
       text: Learn about what we do
       href: 'javascript:void(0)'
-
--- a/src/components/hero/hero.njk
+++ b/src/components/hero/hero.njk
@@ -8,8 +8,8 @@
         {{ hero.title }}
       </h2>
 
-      {% if hero.link %}
-      <a class="usa-hero-link" href="{{ hero.link.href }}">{{ hero.link.text }}</a>
+      {% if hero.paragraph %}
+      <p class="usa-hero-paragraph">{{ hero.paragraph }}</p>
       {% endif %}
 
       {% if hero.button %}
--- a/src/components/hero/hero.njk
+++ b/src/components/hero/hero.njk
@@ -9,7 +9,7 @@
       </h2>
 
       {% if hero.paragraph %}
-      <p class="usa-hero-paragraph">{{ hero.paragraph }}</p>
+      <p>{{ hero.paragraph }}</p>
       {% endif %}
 
       {% if hero.button %}
--- a/src/components/hero/hero.njk
+++ b/src/components/hero/hero.njk
@@ -13,7 +13,7 @@
       {% endif %}
 
       {% if hero.button %}
-      <a class="usa-button usa-button-big" href="{{ hero.button.href }}">{{ hero.button.text }}</a>
+      <a class="usa-button" href="{{ hero.button.href }}">{{ hero.button.text }}</a>
       {% endif %}
     </div>
   </div>
--- a/src/stylesheets/components/_hero.scss
+++ b/src/stylesheets/components/_hero.scss
@@ -29,20 +29,9 @@
   > *:first-child {
     @include margin(0 null $spacing-medium null);
   }
-
-  .usa-button {
-    padding: 1.5rem 3rem;
-    margin-top: $spacing-medium;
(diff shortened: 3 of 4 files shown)


maya:
@donjo the big button is used on the landing page and now looks a little funky when it drops to two lines: https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/jd-big-button/components/detail/layout--landing.html

## New big button:
screen shot 2017-10-30 at 11 40 30 am

## Current big button:
screen shot 2017-10-30 at 11 41 48 am

donjo:
It looks like the hero image button was using font size overrides and the `usa-big-button` class was being applied just to add the vertical padding. Since the button more closely resembles a standard button I decided to take the class off and add the padding to the `.usa-button` style.

Another alternative would be to just use the `usa-button` class with no customizations but it would be making the font larger and the vertical padding less and no longer look like the old version.

thisisdano:
👀

thisisdano:
Well, that sure is a big button. Gotta say, I think it looks a little horsey and off-kilter.

- - -
Original: 

screen shot 2017-10-31 at 9 59 44 am

- - -

Here it is with 8px border radius and 3em x-padding:

screen shot 2017-10-31 at 10 00 23 am

This seems a bit tighter to me. I'd have to go in and modify the Sketch file if we made this change. What do you all think?

And one super-froofy pedantic point: When we say the font size has increased to `24pt` (points) it's actually increased to `24px` (CSS pixels; as distinct from hardware pixels). This is a very small point, and I apologize for bringing it up, but I can't help myself!

maya:
@thisisdano also wanted to make sure you gave some feedback around how the button should be applied in the hero component. Could you take a look?

donjo:
I agree, I like the 8px and 3rem padding one more so I swapped in that change.

thisisdano:
So, re: the hero button — I don't really like that it has a lot of custom styling, and I think I'd prefer if it were a little closer to the default style. I also don't really understand why our hero includes both a text link and a call to action. Is it necessary to have two actions in the hero? Can we just stick with the button?

And perhaps, instead of a link, we might want to include something like a couple sentences of followup text.

_Aaaaand_ it seems like there's more custom spacing under the `h2` than strictly necessary. `2rem` seems like plenty. 

If we do all that, we get an example like the following:

screen shot 2017-11-01 at 3 07 51 pm

Now, is all that crucial to addressing `usa-button-big`? Ha obviously not. But there's a benefit to using fewer customizations in our examples, and trying to work closely as possible to our default styles. If the default styles don't do what we want, why not? What can we do to either improve the defaults, provide sensible variants, or use some other technique (like utility classes) to avoid proliferating custom styles and edge cases? I want us to be real about the strengths and capabilities of our system, and try not to hide the weaknesses with customization.

thisisdano:
Oh, and one more thing (Columbo-style): Do we need to assign links `$color-gray-lighter` in `section-dark`? If links were regular white, like text, we wouldn't need an override for our friend the button here.

maya:
💯 to you hero change suggestions. In terms of _this_ PR, I think if we use the regular button as @donjo has done already and remove its custom styling we should be good. And we can :raccoon: the rest of the hero improvements in another issue/PR. Regarding the link color on dark background, that was from a previous designer that thought light gray links look nice on dark backgrounds. You can see a similar treatment here: https://labs.usa.gov/. But you have the power to change that.

thisisdano:
@maya agreed. And I also agree that gray links look good on blue, _but_ perhaps not good enough to outweigh the consequences. :)

As Thoreau was so thoughtfully edited once: 
> Simplify, ~~simplify~~.

donjo:
@thisisdano ready for re-review. I ended up making the hero changes too.

thisisdano:
If we want to go ahead and make the hero changes, I'd suggest

- remove the custom styling from the hero button, that is the `.usa-hero-callout .usa-button` rules
- remove custom class from the `p.usa-hero-paragraph `

- - -

before:
screen shot 2017-11-08 at 11 00 42 am

- - -

after:
screen shot 2017-11-08 at 11 00 56 am

- - -

Is this an aesthetic improvement? Arguably not (perhaps arguably so?), but it simplifies the code, makes it a bit more flexible, and helps us work closer to the defaults, which I think is a useful way to keep us aware of the expressive gamut of the default system.

donjo:
@thisisdano sounds good. done!

PR #2202, opened 2017-10-30
Add vertical space around sidenav at small widths

thisisdano:
👀 Preview: [link](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/dw-sidenav-separation/components/preview/layout--docs.html)

I wanted to add separation around the sidenav at small widths, but it's tricky because our grid imposes a margin of `0` on the first-child. I didn't want to get into fixing the grid, but I did want to make the sidenav work, so I 
- used padding instead of margin to add 24px at the top
- added a rule on the top and bottom to differentiate from the page content
- used negative margin to push the sidenav out to the viewport margin at narrow widths
- adjusted the sidenav `li` padding to align `li`s to the page margin. (page margin is 15px and the li padding was 18px, and it was just close enough to be annoying).

screen shot 2017-10-30 at 1 51 02 pm

- - -

It still looks kinda stupid at medium width, but that seems to be more a limitation of our grid and its insistence that the `one-quarter` grid appear at 50% at medium width — and I didn't want to add some kind of high-specificity overrides.

screen shot 2017-10-30 at 1 53 17 pm

- - -

Everything should be back to normal at full (amazingly, over 1200px!) width.

screen shot 2017-10-30 at 1 53 00 pm

### DIFF
--- a/src/stylesheets/components/_sidenav.scss
+++ b/src/stylesheets/components/_sidenav.scss
@@ -6,3 +6,10 @@
 .usa-sidenav-sub_list {
   @include usa-sidenav-sublist;
 }
+
+.usa-layout-docs-sidenav {
+  padding-top: 2.4rem;
+  @include media($large-screen) {
+    padding-top: 0;
+  }
+}
--- a/src/stylesheets/components/_sidenav.scss
+++ b/src/stylesheets/components/_sidenav.scss
@@ -1,17 +1,19 @@
 
 .usa-sidenav-list {
   @include usa-sidenav-list;
-  @include margin(null (-$site-margins-mobile) null (-$site-margins-mobile));
   border-bottom: 1px solid;
   border-top: 1px solid;
 
-  @include media($medium-screen) {
-    @include margin(null 0);
+  @include media($large-screen) {
+    border-bottom: none;
+    border-top: none;
   }
 
-  @include media($large-screen) {
-    border: none;
-    margin: 0;
+  .usa-grid & {
+    @include margin(null (-$site-margins-mobile) null (-$site-margins-mobile));
+    @include media($medium-screen) {
+      @include margin(null 0);
+    }
   }
 }
 
--- a/src/stylesheets/components/_sidenav.scss
+++ b/src/stylesheets/components/_sidenav.scss
@@ -11,6 +11,7 @@
 
   .usa-grid & {
     @include margin(null (-$site-margins-mobile) null (-$site-margins-mobile));
+
     @include media($medium-screen) {
       @include margin(null 0);
     }
--- a/src/stylesheets/components/_sidenav.scss
+++ b/src/stylesheets/components/_sidenav.scss
@@ -1,6 +1,18 @@
 
 .usa-sidenav-list {
   @include usa-sidenav-list;
+  @include margin(null (-$site-margins-mobile) null (-$site-margins-mobile));
+  border-bottom: 1px solid;
+  border-top: 1px solid;
+
+  @include media($medium-screen) {
+    @include margin(null 0);
+  }
+
+  @include media($large-screen) {
+    border: none;
+    margin: 0;
+  }
 }
 
 .usa-sidenav-sub_list {
--- a/src/stylesheets/components/_sidenav.scss
+++ b/src/stylesheets/components/_sidenav.scss
@@ -9,6 +9,7 @@
 
 .usa-layout-docs-sidenav {
   padding-top: 2.4rem;
+
   @include media($large-screen) {
     padding-top: 0;
   }
--- a/src/stylesheets/core/_utilities.scss
+++ b/src/stylesheets/core/_utilities.scss
@@ -174,6 +174,17 @@
 
 @mixin usa-sidenav-list {
   @include unstyled-list();
+  @include margin(null (-$site-margins-mobile) null (-$site-margins-mobile));
+  border-bottom: 1px solid;
+  border-top: 1px solid;
+  @include media($medium-screen) {
+    @include margin(null 0);
+    max-width: none;
+  }
+  @include media($large-screen) {
+    border: none;
+    margin: 0;
+  }
 
   > li {
     background-color: transparent;
@@ -191,7 +202,7 @@
     display: block;
     font-family: $font-sans;
     line-height: 1.3;
-    padding: 0.85rem 1rem 0.85rem 1.8rem;
+    padding: 0.85rem 1rem 0.85rem $site-margins-mobile;
     text-decoration: none;
 
     &:hover {
@@ -207,10 +218,11 @@
     }
 
     &.usa-current { /* stylelint-disable-line selector-no-qualifying-type */
-      border-left: 0.4rem solid $color-primary;
+      $current-border: 0.4rem;
+      border-left: $current-border solid $color-primary;
       color: $color-primary;
       font-weight: $font-bold;
-      padding-left: 1.4rem;
+      padding-left: $site-margins-mobile - $current-border;
     }
   }
 }
--- a/src/stylesheets/core/_utilities.scss
+++ b/src/stylesheets/core/_utilities.scss
@@ -207,11 +207,10 @@
     }
 
     &.usa-current { /* stylelint-disable-line selector-no-qualifying-type */
-      $current-link-border: 0.4rem;
-      border-left: $current-link-border solid $color-primary;
+      border-left: $current-highlight-border-width solid $color-primary;
       color: $color-primary;
       font-weight: $font-bold;
-      padding-left: $site-margins-mobile - $current-link-border;
+      padding-left: $site-margins-mobile - $current-highlight-border-width;
     }
   }
 }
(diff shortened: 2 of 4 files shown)


thisisdano:
Oops. I see this has affected `.usa-nav-primary`. 
ONE MOMENT PLEASE.

maya:
@thisisdano also noticed this effects the preview of the sidenav component at small sizes bc it disappears a little from the negative margins: https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/dw-sidenav-separation/components/preview/sidenav--default.html

thisisdano:
Is the fractal display question one that could/should be addressed with a preview layout?

maya:
@thisisdano it seems like there are some inline styles causing this in the preview usa-grid div: `style="padding: 2em 0"` which removed the grid left/right padding. I wonder if we can remove this and makes me wonder if it should be there in the first place. This is also an issue in the internal page that resizes: https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/dw-sidenav-separation/components/detail/sidenav--default.html

maya:
@thisisdano the negative margins are a bit heavy-handed to add it directly to the sidenav component. Would adding it to `.usa-grid .usa-sidenav` make more sense? We should think through all the use cases of this component to make sure we're not introducing any regressions.

Also, do we think the left bar is still as readable when it's on the left edge?

thisisdano:
If I use the `@uswds-content` preview template, all is well. It has an override of 2em padding on each side. Perhaps these preview templates could be given more functional names (like `@padding-y` and `@padding-xy`) — either way, I got it to display better. Updates incoming soon.

Also, this line of code gave me pause:
```
.usa-nav-primary {
  @include usa-sidenav-list;
```

I don't need to get into it rn, but if this is an abstracted pattern, it should probably have a more abstracted name. It's definitely confusing that the sidenanv is imported into the primary nav, and maybe speaks to some quirks in component abstraction.

Ether way, I'm pulling that code out into the most sidenav-specific location.

And yeah, the flush margin _does_ minimize the active bar — which I would probably categorize as an upgrade. Full disclosure, indented content in the sidenav (or any y-bordered indented content) is def a pet peeve of mine. But the active content is _also_ boldface and blue, so I was hoping that the combination of all those things together would get the point across. Maybe not?

thisisdano:
Also, regarding the flush highlight bar position: at this width, by the time you even see this nav, you've already read the page, so hopefully that'll also add to the disambiguation of the current page.

maya:
>Also, this line of code gave me pause:
```
.usa-nav-primary {
  @include usa-sidenav-list;
```
The mobile navigation uses the sidenav pattern style. Reusability ftw :) Would it make sense to abstract this into something else?

screen shot 2017-10-30 at 3 23 29 pm

src/stylesheets/components/_sidenav.scss
```diff
@@ -1,8 +1,28 @@
 
 .usa-sidenav-list {
   @include usa-sidenav-list;
+  @include margin(null (-$site-margins-mobile) null (-$site-margins-mobile));
+  border-bottom: 1px solid;
+  border-top: 1px solid;
```
maya: In the future, is the top and bottom borders something we'd like to do in desktop as well or only mobile?

thisisdano:
Re: reusability — I'm :100: for it. My only concern is that it can be confusing to see something like 
```
.usa-nav-primary {
  @include usa-sidenav-list;
```

For me, this is because if I see `@mixin usa-sidenav-list` my brain connects it strongly to the sidenav, and, by a kind of exclusionary principle, doesn't expect to see it in the primary nav. What could be more "abstract" (dunno if that's the right word in this case) might be something like `nav-stacked`, which is neither sidenav nor primary nav, but something that might conceivably apply to either.

This feels less dissonant:
```
.usa-nav-primary {
  @include nav-stacked;
```

Now, I don't necessarily know if we should do something like this right now, but I hope it's a data point worth considering.

- - -

Now, about the top and bottom border. I think that a top border can look good in a sidenav. In a sketch for plainlanguage.gov, I worked up a sidebar with a border on the top of the nav, but also with a "nav hed" above it, without a border. 

screen shot 2017-10-30 at 4 00 52 pm

At mobile width, there's a border on the top and the bottom:

screen shot 2017-10-30 at 4 01 49 pm

I tend to think that's a good pattern: border on top and bottom at narrow widths, and on the top as a sidebar.

thisisdano:
@maya Do you think this PR is OK to merge?

maya:
@thisisdano not yet, I need to look at it again. Do you want this to go into the release or can/should it wait?

thisisdano:
It _could_ wait, but I think it's solid enough and limited enough to be ready for this one.

src/stylesheets/components/_sidenav.scss
```diff
@@ -1,8 +1,28 @@
 
 .usa-sidenav-list {
   @include usa-sidenav-list;
+  @include margin(null (-$site-margins-mobile) null (-$site-margins-mobile));
```
maya: The negative left and right margins are contextual related to the space of it's parent container, specifically it's dependent on being inside a `usa-grid` that provides the margins. For this reason, I think it may be safer to make this relationship explicit in the code, i.e. `.usa-grid .usa-sidenav-list` for the negative margins. What if a user isn't using our grid?

src/stylesheets/components/_sidenav.scss
```diff
@@ -1,8 +1,28 @@
...
+  border-bottom: 1px solid;
+  border-top: 1px solid;
+
+  @include media($medium-screen) {
+    @include margin(null 0);
+  }
+
+  @include media($large-screen) {
+    border: none;
```
maya: This also removes left and right borders, even though there's no left and right borders present. Can we be explicit with `border-top: none` and `border-bottom: none`?

src/stylesheets/components/_sidenav.scss
```diff
@@ -1,8 +1,28 @@
...
+    margin: 0;
```
maya: When I removed this code and it didn't have an effect on the margins because the top and bottom margins are already removed in the `$medium-screen` media query above. Can this be removed?

src/stylesheets/core/_utilities.scss
```diff
@@ -191,7 +191,7 @@
     display: block;
     font-family: $font-sans;
     line-height: 1.3;
-    padding: 0.85rem 1rem 0.85rem 1.8rem;
+    padding: 0.85rem 1rem 0.85rem $site-margins-mobile;
```
maya: Is the left padding of `$site-margins-mobile` something we want across all sizes or only mobile screen? The current sidenav has a slightly larger left margin and want to make sure.

src/stylesheets/core/_utilities.scss
```diff
@@ -207,10 +207,11 @@
     }
 
     &.usa-current { /* stylelint-disable-line selector-no-qualifying-type */
-      border-left: 0.4rem solid $color-primary;
+      $current-link-border: 0.4rem;
```
maya: I don't know if we need to add this at this variable at this point, but maybe could think about it in the future. I'd want to have have a bit of research to back up what people are needing from a variable for this border and then determine where it should live (in `variable.scss` etc...) and documentation for it. The name `$current-link-border` implies the entire border value (`0.4rem solid $color-primary;`). If it were just the width, I'd expect the name `$current-link-border-width`.

src/stylesheets/components/_sidenav.scss
thisisdano: will do

src/stylesheets/components/_sidenav.scss
thisisdano: will do

src/stylesheets/core/_utilities.scss
thisisdano: Yes, let's leave this as it. I don't think the extra 3px do enough to justify deviation from the "standard" indent

src/stylesheets/core/_utilities.scss
thisisdano: I wasn't thinking of this as a user-modifiable var, just a placekeeping var to keep our math consistent on line 214. But maybe this is a useful var to expose. We could include it in the side navigation docs (since the primary nav seems not to have its own docs).

What about `$current-highlight-border-width` as a var name?

src/stylesheets/components/_sidenav.scss
thisisdano: good point. will do.

thisisdano:
@maya is this ready to demo?

maya:
@thisisdano were the fixes made? I haven't re-reviewed this yet.

src/stylesheets/core/_variables.scss
```diff
@@ -143,6 +143,7 @@ $box-shadow:              0 0 2px $color-shadow !default;
 $focus-outline:           2px dotted $color-gray-light;
 $focus-spacing:           3px;
 $nav-width:               951px !default;
+$current-highlight-border-width: 0.4rem !default; // must be in rem for math
```
maya: @thisisdano for this new variable, a few things....
- Each variable section's values should align, here it extends past the other values.
- Is the variable applicable to more things besides the sidenav? 
- Would it work to call it `$current-border-width` or do we need `highlight`?
- I think we can remove the comment "// must be in rem for math" bc many other variables are in rem and we don't state this

PR #2228, opened 2017-11-14
Fix extended header navigation overlap

### LINKED ISSUE #2222: [Small screen navigation links overlap close button]: When the first link in the nav does not contain subnav links the close button for the off canvas overlaps the first link.
## Description

When viewing the navigation for smaller screens the first link in the navigation overlaps the close button in the menu when it is just a single link and not a dropdown. However if the first link is a dropdown - as seen on the [extended header components page](https://components.standards.usa.gov/components/detail/header--extended.html) this is not an issue. 

1. Go to the [extended header components page](https://components.standards.usa.gov/components/detail/header--extended.html) 
2. Move the single link `li` from the bottom of the list to the top
3. You will see the first link and the navigation close button are now overlapping. 

### Screenshots 
screen shot 2017-11-09 at 10 07 44 am

## Possible fix 
A possible fix would be to set a `clear:both;` on the `usa-nav-primary` style while at smaller screen sizes, making sure that whatever is inside of the UL is always below the close button and not overlapping. 
screen shot 2017-11-09 at 10 09 52 am

maya:
Fix extended header navigation overlap between the close button and next first item by clearing anything immediately following the navigation close button.

Fixes #2222.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/18f/web-design-standards/fix-nav-overlap/components/detail/header--extended.html)

**Vizdee question** 🌈: While I was reviewing this, I noticed the component sticker sheet (Sketch) has a tighter spacing between the close button and nav, while the original designs (which our current code follows) had wider spacing and want to confirm the desired spacing, which I can tack on here if we want it tighter like the sketch file. What do you think @thisisdano?

## Current Sketch file
screen shot 2017-11-14 at 9 57 13 am

## Previous designs (and current code)
screen shot 2017-11-14 at 9 57 32 am

### DIFF
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -23,11 +23,8 @@ $z-index-nav:     9000;
   }
 
   .usa-search {
-    margin-bottom: 1.5rem;
-
     @include media($nav-width) {
       float: right;
-      margin-bottom: 0;
       max-width: 21.5rem;
     }
   }
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -123,7 +123,6 @@
 
   @include media($nav-width) {
     display: inline;
-    margin-top: 0;
   }
 
   li {
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -118,10 +118,12 @@
 
 .usa-nav-primary {
   @include usa-sidenav-list;
+  margin-top: 1.5rem;
   order: 2;
 
   @include media($nav-width) {
     display: inline;
+    margin-top: 0;
   }
 
   li {
@@ -303,14 +305,17 @@
 // Secondary navigation ----------- //
 
 .usa-nav-secondary {
+  margin-top: 1.5rem;
+
   @include media($nav-width) {
+    margin-top: 0;
     position: absolute;
     right: $site-margins;
     top: -5.7rem; // XXX magic number
   }
 
   .usa-search {
-    @include margin(3rem null);
+    @include margin(1.5rem null 0 null);
 
     @include media($nav-width) {
       @include margin(-0.9rem null 0 0);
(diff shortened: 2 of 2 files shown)


thisisdano:
I think the sketch file is too tight and our current style is too loose.

- The close box should something like 16px (1.6rem) of `margin-bottom` to separate the target area from a target below-adjacent (like a search box).
- The secondary section needs about 16px whitespace on top if there's no search section above it. If there's a search, it should have no whitespace on top (as it does now).

screen shot 2017-11-14 at 2 20 31 pm

maya:
Great, looks like header already has 1.5rem / 15px of bottom margins between the search button and mobile nav so we're covered there. Lmk if we should change these to 1.6rem / 16px or use the existing spacing amount.

screen shot 2017-11-15 at 12 15 00 pm
screen shot 2017-11-15 at 12 15 08 pm

maya:
Spacing changes made in a9d7061.

thisisdano:
yeah, 15px is ok for the top. I do still think the secondary section needs about 16px whitespace on top if there's no search section above it.

Currently I see when I move/remove the search box:

screen shot 2017-11-15 at 1 03 07 pm

And there isn't enough separation between the secondary and primary navs

maya:
Ah ok, I had thought you wanted 16px between all sections. I changed it from 30px, but will change it right back.

maya:
@thisisdano reverted that change in 066b8aa so it should be more spacious like it was before. Lmk if that works.

maya:
Actually hold that thought, looks like I may need to add in addition margins to the secondary nav for when there's no search form.

maya:
@thisisdano to clarify, 30px or 15px/16px between the primary nav + search + secondary links:

### 15px
screen shot 2017-11-15 at 1 30 08 pm

### 30px
screen shot 2017-11-15 at 1 30 55 pm

thisisdano:
I guess the issue it that 16px is OK for the separation between search bar and primary, but it too little between the primary abutting the secondary.

maya:
@thisisdano changes made in d40db45, let me know if this works. Using with the 1.5rem for consistency, and can change this out if/when we do a larger spacing scale change.

thisisdano:
Once this passes circle, I'm good.

maya:
Tests fixed in 052125a, will merge.

PR #2267, opened 2017-12-12
Fixing dropdown arrow display in IE10/11

### LINKED ISSUE #2265: Select input arrow missing in IE10 & IE11
## Description

The arrow on the right side of the select input is missing in IE10 & 11 starting in USWDS V1.4.x
I believe it's actually tiny and not missing. The SVG isn't rendering properly.

1. Go to https://standards.usa.gov/components/form-controls/ in IE10
2. Scroll down to "dropdown"

screen shot 2017-12-11 at 4 03 50 pm

donjo:
Fixes #2265 

Found [this fix](https://www.broken-links.com/2010/06/14/using-svg-in-backgrounds-with-png-fallback/) and it appears to be working for me in IE11. 

I wasn't quite sure the best way to confirm it was using the SVG in browsers that support it but I unchecked that rule (forcing the PNG to load) and zoomed in to notice the render wasn't quite as sharp which would be the expected behavior for a non-svg in a case like this.

### DIFF
--- a/src/stylesheets/elements/_inputs.scss
+++ b/src/stylesheets/elements/_inputs.scss
@@ -112,7 +112,8 @@ select {
   appearance: none;
   background-color: $color-white;
   background-image: url('#{$image-path}/arrow-both.png');
-  // Ensure browsers that don't support SVG in background-image (IE 11 and below) fall back to PNG. See https://www.broken-links.com/2010/06/14/using-svg-in-backgrounds-with-png-fallback/
+  // Ensure browsers that don't support SVG in background-image (IE 11 and below) fall back to PNG. 
+  // See https://www.broken-links.com/2010/06/14/using-svg-in-backgrounds-with-png-fallback/
   background-image: none, url('#{$image-path}/arrow-both.svg'), url('#{$image-path}/arrow-both.png');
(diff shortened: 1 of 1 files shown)


src/stylesheets/elements/_inputs.scss
```diff
@@ -112,7 +112,7 @@ select {
   appearance: none;
   background-color: $color-white;
   background-image: url('#{$image-path}/arrow-both.png');
-  background-image: url('#{$image-path}/arrow-both.svg');
+  background-image: none, url('#{$image-path}/arrow-both.svg'), url('#{$image-path}/arrow-both.png');
```
maya: Can you add this comment here on line 115 (above this rule): `// Ensure browsers that don't support SVG in background-image (IE 11 and below) fall back to PNG. See https://www.broken-links.com/2010/06/14/using-svg-in-backgrounds-with-png-fallback/.` ?

maya: @donjo minor thing: I don't know if it's documented but I think we try to do a 80 character limit for the code rule. Could probably drop "See" to another line.

donjo: @maya good call, fixed!

PR #2271, opened 2017-12-14
Move Circle config to .circle folder

donjo:
This PR moves the circle config to the file/folder structure recommended in the [Circle 1.0 -> 2.0 migration guide](https://circleci.com/docs/2.0/migrating-from-1-2/#steps-to-configure-required-20-keys). It works either way but we should probably stay consistent with the docs site and the preferred method of doing things.


maya:
Ready for review?

donjo:
@maya Yep!

Issue #2272, opened 2017-12-14
Choose utility classes to add


maya:
Utility classes (or helper classes) provide a lightweight way to customize your markup with single-purpose classes. These classes are simple, composable, and reusable. For example: `.usa-red { color: $color-red; }`, `.usa-h1 { font-size: $h1-font-size; }`. We've seen the web community moving in this direction and think that providing helpful classes can enable teams to quickly remix these basic building blocks. 

## 📣 Tell us what you need! 
**Comment on this issue with utilities you'd like to see added.** 

---

Below are utility classes we've heard people ask for, ordered from most to least requested.

### Top requested
- **color** (text, background)
- **space** (margin, padding)
- **typography** (sizing)

### Also requested
- display (none, inline, block)
- float (left, right)
- typography styles (caps, bold, underline)
- position (relative, absolute)
- border (border-bottom, rounded, etc)
- grid (column sizing at diff viewports)
- size

*Note: We're also aware they can be problematic when taken to the extreme or when teams don't communicate how they'll be using them (and will need to think through how we [document](#2263) this). 

Related #2264, #2214.

evetsreklaw:
This is a good list. If grid is referring to something like bootstrap's breakpoints (xs-6 or md-8) that would be number one for me. Many of the others I have created and I am sure others have too so I hope others share what they have done. Colors - usa-fill- and usa-text- + color at https://github.com/evetsreklaw/USWDS-Extended/blob/master/_colors.scss.

maya:
We have a good sense of which ones to add first, closing this as we begin implementation.

Issue #2264, opened 2017-12-11
Research Sass organization


maya:
[🔒 Working document](https://docs.google.com/document/d/1tdWVt7AYznidpNwYBeBInWUt9TxXbeesicMnEHeP_Lo/edit)

maya:
## Action items
- [ ] Create a `utilities/` folder for utilities which are imported after everything
- [ ] Move utility mixins out of `core/utilities.scss` and into their own files inside a `mixins/` folder (alternatively could be called `tools/` folder).

Future: consider breaking variables into their own files in a `variables/` or `settings/` folder if a single file becomes hard to use.
 
(Breaking up this file makes it harder to plugin to something like CodePen, so consider how what the user experience of this would look like. Already could be an issue with breaking up mixins but seems worth it bc the it's hard to see what's in that file.)

Issue #2214, opened 2017-11-07
Naming variables and classes


maya:
**User story**: As a developer using the USWDS, I'd like to use variables to create consistent websites with ease.

Variables are the foundation of any design system. This issue will begin to address the design of our current variables, their pain points, and ways we can improve them.

### Types of sizes 
- t-shirt sizing: xs, sm, m, l, xl
- numbered sizing: sp-1, sp-2, sp-3, sp-4

### Ways of naming (brevity vs. descriptive)
- small, sm, s
- background, bg, fill

## Examples to look at
- NAVA
- tachyons
- basscss
- tailwind

## Acceptance criteria

- [x] Comparative analysis: research how other design systems document and work with variables @maya
- [x] Understand how we name variables
- [ ] Assess how our variables are organized
- [ ] Determine if we're missing variables
- [ ] Assess weaknesses and pain points of variables

maya:
[🔒 Working document](https://docs.google.com/document/d/1CdzR4EhF1OK2NO-EbsqmdpekNpGjaBZ2u_IcxBoIHdY/edit#heading=h.ulunx6x144lz) - may move to markdown when complete.

maya:
We met and agreed to go with an approach that's clear, readable, and nimble to work with. A brief summary:

## Colors
Text color: `.usa-blue`
Background color: `.usa-bg-blue`

*TBD the amount of colors we'll release.

## Space
An abbreviated style similar to [tachyons](http://tachyons.io/) for margin and padding: `.mt0`

## Typography
`.usa-h1`  or `.usa-sans-s6` 

We explored a few options, but have not come to a conclusion yet. Something like `.usa-h1` is recognizable and familiar.  `.usa-sans-s6` removes heading terminology from the size and provides more than 6 heading sizes. We may have to try out a few things as we start to build this out and see what works best for our users.

PR #2287, opened 2017-12-22
[WIP] Add utilities

maya:
### ⚠️🚧 work in progress 🚧 ⚠️

This begins work for adding in utilities as well as rearchitecting the Sass to support this.

- Add utilities folder
- Break up mixins into individual files
- Rearrange Sass imports in a more logical order (resembling ITCSS)

Related #2264, #2272, #2214.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-utilities)

---
## Notes
- There are two deprecated files: `core/_grid-settings.scss` has been deprecated for a while and `core/_utilities` (just deprecated). Do we need them for any reason?
- `core/mixins/font-smoothing.scss` has a TODO note that the file can be removed once PostCSS has been added and it has been. We can remove this mixin from the source.


src/stylesheets/uswds.scss
```diff
@@ -39,3 +58,8 @@
 @import 'components/section';
 @import 'components/sidenav';
 @import 'components/skipnav';
+
+// Utilities ------------ //
+// @import 'utilities/color';
+// @import 'utilities/space';
+// @import 'utilities/typography';
```
maya: Should this be removed (or added to)?

thisisdano:
My latest work should address the issues mentioned in https://github.com/uswds/uswds/pull/2378

It should 
- [fix the grid](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-utilities/components/detail/grid--default.html)
- [update to the correct blue color](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-utilities/components/detail/buttons--default.html)
- [update to the correct light gold color](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-utilities/components/detail/alerts--warning.html)
- [use normalize as expected in tables and elsewhere](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-utilities/components/detail/tables.html)

thisisdano:
I can merge in `develop` to resolve those conflicts.

thisisdano:
@hursey013 In resolving merge conflicts in `package-lock.json` when merging `develop`, I just chose everything in `develop` — assuming that reflected (y)our new work. But, honestly, not 100% sure. Can you check the diff and validate my choices?

thisisdano:
Shoot, it looks like my merge broke something!

hursey013:
@thisisdano I noticed the compiled versions of the stylesheets in `/dist` have a lot of comments in them, and not necessarily in a way that makes sense. `dist/uswds-utilities.css` has 1600+ lines of comments at the top of the file but I'm thinking we are probably safe to just remove them since they make more sense contextually in the Sass files, thoughts?

thisisdano:
Yeah, all that comment stuff is really ugly. Is there a flag we can use to strip those comments from the un-minified css?

src/stylesheets/core/_font-definitions.scss
```diff
@@ -0,0 +1,146 @@
...
+      300: false,
+      400: 'merriweather-italic-webfont',
+      500: false,
+      600: false,
+      700: false,
+      800: false,
+      900: false,
+    ),
+  ),
+  roboto-mono: (
```
hursey013: I know we explicitly recommend using Source Sans Pro and Merriweather now - how do Roboto Mono and United Sans work into that recommendation? Are these just to extend the number of available options or are there specific instances when these new font families should be chosen? (Sorry if this has already been covered elsewhere).

thisisdano: Hasn't been covered yet, but I consider this a topic for upcoming guidance. But also even in the absence of guidance, by starting to build a library of font definitions, we can begin to use the system's new ability to normalize font sizes — and be more simpler and more programmatic about building our `@font-face` rules

maya:
We started with 2 SCSS files in the root and here it's grown to 6. I'm finding this a lot to parse and hard to see which files depend on each other. I think a flat ITCSS structure in a single file could be a lot more linear, simpler, and easier to see. Sticking with a structure that's known and used in the industry could also be easier to maintain. Would this work using the ITCSS structure?

thisisdano:
I think this work is consistent with ITCSS, so maybe we can talk about what would make the organization simpler.

In general, though, there are four files instead of one to make it simpler to include only the parts of the system you want to include:

`uswds-components` is all the components
`uswds-utilities` is the utilities
`uswds-fonts` is the `@font-face` declarations — broken out to make it simpler to inline critical styles in the `head`
`uswds` packages all these together as one

`required.scss` is imports common to all of these — this file could live somewhere else.

I've used it in a project like 
```sass
// project.scss
@import "uswds-utilities"
```
when I didn't want to include the components. You could do a similar thing if you didn't want the utilities. (You can't put `import` directives inside a conditional and control output using, say, an `$output-utilities` bool.)

That said, there's probably another way to organize everything that makes it extra clear. Let's talk it through.

joshbruce:
@Maya and @donjo: Here's an interesting integration we just put on another project I maintain. Having actually put it into practice but I like the idea of it: https://github.com/apps/wip

PR #2306, opened 2018-01-06
Updated gitignore

BalajiJBcs:
Made some updates to the .gitignore file which exclude configuration file of IDE when making git commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ coverage/
 # ignore the generated dist directory
 dist/
 
-# ignore the Fractal build directgory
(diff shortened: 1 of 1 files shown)


maya:
Hi @BalajiJBcs, thanks for opening this. I checked with another developer, @toolness, who said this PR might make sense if there were a recommended/required IDE for contributors to use, but there isn't. As a result, we'd recommend you put those entries in a global gitignore instead: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore 

Closing for now, let us know if you find anything else.

PR #2308, opened 2018-01-08
Optimizing PNGs with ImageOptim

line47:
## Description

Used ImageOptim to reduce PNG images without losing quality.
Some images could not be reduced any further and therefore are not in the PR.

## File size saved screenshot
screen shot 2018-01-08 at 12 27 01 pm

🎩  Hat tip to @fulldecent for opening a similar PR in https://github.com/18F/uswds-jekyll/pull/92


fulldecent:
Looks good, thank you.

Issue #2289, opened 2017-12-23
Add utilities


maya:
Utility classes (or helper classes) provide a lightweight way to customize your markup with single-purpose classes. These classes are simple, composable, and reusable. For example: `.usa-red { color: $color-red; }`, `.usa-h1 { font-size: $h1-font-size; }`. We've seen the web community moving in this direction and think that providing helpful classes can enable teams to quickly remix these basic building blocks.

francisrupert:
Recommend **responsive** flexibility, as with any given grid system, e.g...
```html
class="usa-text-size--foo usa-text-size--bar@s usa-text-size--zazz@m"
```
Aside: I'm extremely averse to utility styles in markup, but I recognize the wide variety of audiences USWDS caters to will appreciate a convenient way to do so -- and I've been plenty guilty to using them; besides, most grid systems are nothing more than utility classes. There's a valid discussion to be had that utility styles are "too easy, possibly lazy" and will be used far too liberally. It can be habit-forming and prevents users from growing their style skillsets, let alone understanding what exactly it is they're doing. At any rate...

maya:
Hi @francisrupert, thanks so much for your feedback. Part of why the USWDS didn't launch with color helper classes was for the very reasons you mentioned: we didn't want to promote bad practices within the code. After a couple of years, it's become clearer that folks would find this helpful (especially for prototyping), and thanks to the influence of utility systems like [Tachyons](http://tachyons.io/). We hope to keep this utility set limited in scope (#2272), starting with color, space, and typography -- and extend it to a few layout items later on. 

That's a good point about responsive flexibility. We haven't released responsive typography yet but it's something we're thinking about and have some designs created. We've mostly heard the responsive requests for a grid system, and that would be tackled later on. 

The use of utility classes would have to be taken on by individuals using the system - some folks may prefer to use them, some may choose not to, others may use them for only prototyping, etc. We're hoping each team will make the decision that works best for them. We can likely provide some documentation of best practices though. If you have more thoughts about this, keep them coming.

s2t2:
Hi, I was looking for float utilities like `pull-right` and `pull-left`. See Twitter Bootstrap: https://getbootstrap.com/docs/4.0/utilities/float/.

maya:
@s2t2 thanks for the note. We'll be including float utilities as described here: #2272. You can also get a sneak peek here: #2316, but it may be more useful once we build out the documentation. We'll keep you posted!

maya:
[...]

PR #2316, opened 2018-01-17
Adding unit and setttings utilities

thisisdano:
This PR establishes a wide palette of values (unit values) available to the system as project settings variables. Many of these values can now be accessed not only as variables in project sass, but as utility classes in markup. 

`uswds-u-units.scss` (unit utilities) and `uswds-u-settings.scss` (settings utilities) are new sass files that compile into `dist/uswds-u-units.min.css` and `dist/uswds-u-settings.min.css`, respectively.

This work is based off the utility class builder here: https://github.com/18F/united

More to come...

Related: #2289.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-feature-add-utilities)


joshbruce:
https://tailwindcss.com

joshbruce:
@carlosvalle

maya:
Perhaps a more understandable metaphor than "units and settings" would be "tokens and theme." I was wondering if "units and settings" would be more abstract for folks to comprehend. 

Salesforce's Lightning Design System [defines tokens](https://www.lightningdesignsystem.com/design-tokens/) as:
>Design tokens are the visual design atoms of the design system — specifically, they are named entities that store visual design attributes. We use them in place of hard-coded values (such as hex values for color or pixel values for spacing) in order to maintain a scalable and consistent visual system for UI development.

Then themes use a subset of tokens (or not) for product-specific visual styles.

Issue #2300, opened 2018-01-02
Changing our name to more accurately describe what we are and what we do


thisisdano:
This epic addresses the need to be a good player in government by being extra clear about what and who we are. We know that the word _Standards_ is a sticking point as well as a useful lever, so our challenge is to maintain our authority and identity as we change to something more accurate and less controversial.

### Issues related to this epic

- Choose a new name for the system
- Create an implementation path for the new name
- Audit all places our old name is used
- Work to change references to the old name — first in our own site, code, and documentation

maya:
[:lock: Naming decisions doc](https://docs.google.com/document/d/1obEmQLNDhvMFuzcw6H7oxKKhae1HxZZpBpFVkA_vC5Y/edit#) is complete.

maya:
- [ ] Decide what to change it to: U.S. Web Design System
  - [ ] Shorthand: the Design System
  - [ ] Change name and shorthand throughout main repo and docs site
- [ ] Write a blog post to explain the change as relevant to the work we’re doing and what we’re providing
  - Less about the name, more about the benefit, pedagogy — benefits of system design thinking, similar to standardization without the baggage, rigidity
- [ ] Pick a URL: designsystem.gov / designsystem.digital.gov
  - [ ] Register URL
  - [ ] Set up redirects from old URL
- [ ] Rename repo to uswds
- [ ] Establish our own GitHub org for uswds? (i.e. uswds/uswds)
  - [ ] Reach out to #github-admins
- [ ] Change our team email to uswds@gsa.gov (alt designsystem@gsa.gov)
- [ ] Change Twitter display name, #uswds and #uswds-public Slack channel description (any other social media stuff?)

jeremyzilar:
I think this would be an opportunity to put out a few cards* for Twitter and Facebook that talk about
- The name change
- Why the U.S. Web Design System is important
- How to contribute
🇺🇸 

```
The U.S. Web Design System is ready to help you make better digital services.
Become a part of the #uswds.
```

```
Over 400 people have contributed to The U.S. Web Design System. 
Become a part of the #uswds.
```

```
Build faster, and with greater assurance using The U.S. Web Design System. 
Become a part of the #uswds.
```

```
A design system for the public.
Accessible. Flexible. Fast.
Become a part of the #uswds.
```

_* cards == jpgs with text on them_

bruce-usab:
> * cards == jpgs with text on them

[...]

PR #2320, opened 2018-01-19
Add new name "System" and new domain

maya:
**To be merged right after the domain name change**

This adds the new name, shorthand, project domain, GitHub URL, and email address.

Related: https://github.com/18F/web-design-standards-docs/pull/487.

See #2300

### DIFF
--- a/WHO_IS_USING_USWDS.md
+++ b/WHO_IS_USING_USWDS.md
@@ -1,6 +1,6 @@
-# Website and applications that use the Design System
+# Websites and applications that use the Design System
 
-Below are a list of website and applications currently using the U.S. Web Design System. If your project is currently using the Design System and you do not see it on this list, please feel free to submit a pull request or email the core team at uswebdesignstandards@gsa.gov.
+Below are a list of websites and applications currently using the U.S. Web Design System. If your project is currently using the Design System and you do not see it on this list, please feel free to submit a pull request or email the core team at uswebdesignstandards@gsa.gov.
 
 - [18F](https://18f.gsa.gov)
 - [18F Automated Testing Playbook](https://automated-testing-playbook.18f.gov/)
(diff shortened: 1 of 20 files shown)


WHO_IS_USING_USWDS.md
```diff
@@ -1,6 +1,6 @@
-# Website and applications that use the Standards
+# Website and applications that use the Design System
```
thisisdano: `Websites`?

WHO_IS_USING_USWDS.md
```diff
@@ -1,6 +1,6 @@
...
 
-Below are a list of website and applications currently using the U.S. Web Design Standards. If your project is currently using the Standards and you do not see it on this list, please feel free to submit a pull request or email the core team at uswebdesignstandards@gsa.gov.
+Below are a list of website and applications currently using the U.S. Web Design System. If your project is currently using the Design System and you do not see it on this list, please feel free to submit a pull request or email the core team at uswebdesignstandards@gsa.gov.
```
thisisdano: `websites`?

thisisdano:
Seeing all these together makes me wonder whether we should use capitals when referring to the design system in shorthand. As in 

`If your project is currently using the Design System and you do not see it on this list...`
vs
`If your project is currently using the design system and you do not see it on this list...`

I see how caps is consistent with our earlier usage, but I wonder if it's the right call now.

maya:
@thisisdano caps should be used if Design System is a proper noun, but lowercase if it's more of a general thing.

thisisdano:
Yeah, I get the proper noun thing, but that usage does have a whiff of incorrectness even though it's technically correct, like.... Internet or... Kleenex. But I'm comfortable moving forward with it now. I don't think this is a decision that can't be modified at any later date if it seems more wrong/weird.

maya:
Oh, I wasn't saying that it _is_ a proper noun. It could be a proper noun in some cases and not in other cases. We have to make the call.

WHO_IS_USING_USWDS.md
maya: Fixed in 2f2aee4.

WHO_IS_USING_USWDS.md
maya: Fixed in 2f2aee4.

maya:
@thisisdano is this also good for approval?

Issue #18, opened 2015-06-10
[Charts/Graphs] How should we think about designing these?


mollieru:
Looking from teams to give us a bit more insight on what components of charts and graphs people are looking for. Like styled out bar charts, line graphs, etc? 

Would be helpful to get guidance on whether this is about visual guidance and/or also building customizeable interfaces, which will require muchmore front end lift.

mjovel:
@mollieru Here are a couple of different examples of data visualizations that currently exist in the FDA world -
* http://www.fda.gov/drugs/developmentapprovalprocess/druginnovation/ucm381397.htm
* http://www.accessdata.fda.gov/scripts/oce/inspections/oce_insp_searching.cfm

I don't think the team would have to reinvent the wheel but possibly just create some standardizations of appearance on existing libraries.

cew821:
Seems to me that getting closer to actual implementation of a data
visualization would be a lot of work, and would add a dependency on some
type of JS visualization library like D3. I also think charting
implementation is often dictated by (or helps dictate) how the back end
data API is structured, increasing the level of difficulty for building
generic components that are usually useful.

However, if done well, I could see how this could be a huge time saver for
a team looking to implement data visualization.

Still, I would be surprised if this level of effort would be worth doing in
the early stages of the project. Definitely I wouldn't take it on unless
you can find a clear customer of the tool who will imement it in a real
app as we build.

Scotchester:
In our breakout group, I was the one who mentioned this item. I agree with @cew821 that offering implementation of it is probably outside the scope of the first release, maybe outside the scope forever, but I think that offering some design best practices advice would be a useful thing to do.

cew821:
Yeah that sounds totally reasonable.

Or maybe just a link to Tufte's Visual Display of Quantitative
Information book 📊📈📉📕

Scotchester:
Also, to call it out explicitly, advice for making charts & graphs **accessible** would be extremely valuable, too.

mjovel:
Agree with with everything @cew821 and @Scotchester have said here

amycesal:
I'm pulling together a design manual issue for CFPB on this: https://github.com/cfpb/design-manual/issues/230 and I put together a data vis style guide for Sunlight Foundation when I worked there: http://sunlightfoundation.com/blog/2014/03/12/datavizguide/

[...]

PR #2326, opened 2018-01-24
Addition of charts and graphs

scottkellum:
## Description

This adds the ability to easily generate simple charts with tables. Simply put your content into a well formatted table and add the class `.usa-chart`. The goal was to make this as simple and easy to use as possible while also remaining accessible to those without the use of JavaScript.

## What problem does this solve?

Agencies already use graphs to display information to audiences and this allows them to embed them directly into their sites more easily and with better accessibility than images.

### Bar chart example

screen shot 2018-01-29 at 2 42 52 pm

```html
<div class="usa-chart" data-chart-type="bar"> <!-- char-type: bar, line -->
  <table class="usa-chart-data" data-min="0" data-max="100" data-increment="20" data-grid="true"> <!-- all values optional -->

    <!--
    label datasets with header row
    unnessisary unless multiple values are assigned
    -->
    <tr>
      <td></td>
      <th class="usa-chart-dataset">set one</th>
      <th class="usa-chart-dataset">set two</th>
    </tr>

    <tr class="usa-chart-item">
      <td class="usa-chart-label">Thing one</td>
      <td class="usa-chart-value">8%</td>
      <td class="usa-chart-value">28%</td>
    </tr>
    <tr class="usa-chart-item">
      <td class="usa-chart-label">Thing two</td>
      <td class="usa-chart-value">12%</td>
      <td class="usa-chart-value">24%</td>
    </tr>
    <tr class="usa-chart-item">
      <td class="usa-chart-label">Thing three</td>
      <td class="usa-chart-value">26%</td>
      <td class="usa-chart-value">40%</td>
    </tr>
    <tr class="usa-chart-item">
      <td class="usa-chart-label">Thing four</td>
      <td class="usa-chart-value">60%</td>
      <td class="usa-chart-value">42%</td>
    </tr>
  </table>
</div>
```

### Line graph example

screen shot 2018-01-29 at 2 43 09 pm

```html
<div class="usa-chart" data-chart-type="bar"> <!-- char-type: bar, line -->
  <table class="usa-chart-data" data-min="0" data-max="100" data-increment="20" data-grid="true"> <!-- all values optional -->

    <!--
    label datasets with header row
    unnessisary unless multiple values are assigned
    -->
    <tr>
      <td></td>
      <th class="usa-chart-dataset">set one</th>
      <th class="usa-chart-dataset">set two</th>
    </tr>

    <tr class="usa-chart-item">
      <td class="usa-chart-label">Thing one</td>
      <td class="usa-chart-value">8%</td>
      <td class="usa-chart-value">28%</td>
    </tr>
    <tr class="usa-chart-item">
      <td class="usa-chart-label">Thing two</td>
      <td class="usa-chart-value">12%</td>
      <td class="usa-chart-value">24%</td>
    </tr>
    <tr class="usa-chart-item">
      <td class="usa-chart-label">Thing three</td>
      <td class="usa-chart-value">26%</td>
      <td class="usa-chart-value">40%</td>
    </tr>
    <tr class="usa-chart-item">
      <td class="usa-chart-label">Thing four</td>
      <td class="usa-chart-value">60%</td>
      <td class="usa-chart-value">42%</td>
    </tr>
  </table>
</div>
```


maya:
This is looking good! We've heard charts and graphs are something many agencies would need (see #18). We're still figuring out what our component submission workflow will look like and will let you know when we have more info. Appreciate the simplicity and focus on progressive enhancement here.

scottkellum:
@maya I’m feeling comfortable for this to be reviewed now when y’all are ready.

- I removed the pie chart because [they aren’t very good at displaying information](http://www.businessinsider.com/pie-charts-are-the-worst-2013-6)
- Consolidated everything into a single component to reduce duplication
- Added support for more than one dataset on a graph

maya:
Thanks so much @scottkellum! We'll probably be able to start reviewing this (in _some_ way) in the next sprint (or the one following that), as this sprint we're focused on building out utilities, and we've tentatively planned the next sprint to be a technical debt sprint. The process may take some time as you're our first major component submission, so we appreciate the patience.

Breaking up the review for the team, which could happen async:

- [ ] UX -- @donjo 
- [ ] Visual design -- @thisisdano 
- [ ] Front-end
   - [ ] HTML, CSS, a11y -- @maya or @donjo 
  - [ ] JS -- [TBD engineer]

---
For this to go on the main website, we'll need documentation.

Open question:
Is this included in the main package, behind a `experimental` flag that's not included in the package unless you opt-in, or through CodePen?

scottkellum:
Thanks @maya ! Let me know if you need anything from me in the process and I’ll sit tight as you move through your sprints.

maya:
Hey @scottkellum, could you fix the linting errors in CircleCI? You can also run `npm test` in your code.

scottkellum:
@maya Yup! Sorry about that should pass now

src/components/11-infographics/chart.njk
```diff
@@ -0,0 +1,35 @@
+<div class="usa-chart" data-chart-type="bar"> <!-- char-type: bar, line -->
+  <table class="usa-chart-data" data-min="0" data-max="100" data-increment="20" data-grid="true"> <!-- all values optional -->
+
```
maya: Can you add a `<caption>` as the title giving a brief description of the table? See: https://webaim.org/techniques/tables/data#caption

src/components/11-infographics/chart.njk
```diff
@@ -0,0 +1,35 @@
...
+    <!--
+    label datasets with header row
+    unnessisary unless multiple values are assigned
+    -->
+    <tr>
+      <td></td>
+      <th class="usa-chart-dataset">set one</th>
```
maya: In various places: can you add the appropriate scope attributes to identify your headers? See: https://webaim.org/techniques/tables/data#headers

scottkellum:
I’m working on this feedback, I need to figure out a way to surface the `caption` in the rendered graph.

In terms of accessibility please review the rendered SVG as well. I tested it with VoiceOver and it reads through the data appropriately but you might have insights on things I’m missing.

donjo:
Hey @scottkellum, got a chance to look at this a bit more thoroughly today and offer some general UX/design feedback. I'll mention ahead of time that a lot of this is based on things I learned in working on a mini discovery project with census (github.com/18F/uswds-data) where we were trying to establish the baseline of some data design guidelines. That said here are some suggestions and I'm curious to hear what you think. 

**Add titles and source placeholder text**
- One thing the Census team kept driving home was the importance of linking to data sources. While it won't be necessary 100% of the time, it could be useful to include a placeholder here to nudge users into including a link to the raw data if they have it. As for titles, looks like @maya already has that covered. To go with that I'd suggest moving the chart keys to below the chart

**Separate `njk` for line chart**
- Would be helpful for folks looking at the component library to see there are multiple options without having to read the comments. 

**Adjust font size on label**
- May just be a personal preference but the labels seem a bit large and overpowering. What do you think about going from 4.25px to 4px in 

```
.usa-chart-gridLabel {
+  text-anchor: middle;
+  font-size: calc(4.25px - .25vw);
}
``` 

I think that's all I got for now. Great work!

maya:
@scottkellum I thought when I previously tested the rendered SVG it read it like the table, in that I was able to progress through each value one by one, but now it seems to read it all smushed together. Was that intentional?

screen shot 2018-02-20 at 2 27 36 pm

---
For the title, it seems like it should render a `<title>` in the SVG and a `<caption>` in the table.

maya:
For the font size, could we just use the `$base-font-size`?

maya:
Oh, nevermind about the font-size. Looks like it's responsive to the size of the chart 👍.

scottkellum:
@maya @donjo Thanks for your feedback! I made revisions and feel like the accessability is much improved. For a chart title I just dropped an `h2`.

maya:
Thanks @scottkellum! 

I've talked with a few people about this and heard the best practice is to make the SVG hidden for screen readers and include the table. The table would be visually hidden but accessible to screen readers and you could have the option to show it visibly too along with the chart.

Karl Groves mentioned including tables allows you to identify table headers and explore the table using typical table reading commands.

Some background from my colleague @toolness: 
>as far as i can tell, [the best practice] is to provide a table as additional content, so that the bar chart is basically just an image (though its general shape could be provided as alt text) but the user can deep-dive into the data of the table
>
>heydon pickering gave a talk where he actually argues that showing the table _only_ to certain users is actually non-inclusive, b/c even some sighted users (like himself) prefer tables to bar charts
>
>i find that SVGs are very unpredictable with screen readers--some readers essentially treat them opaquely like image elements, while others actually let you navigate through individual elements of the SVG

scottkellum:
Thanks @maya! I was using [info from CSS-Tricks](https://css-tricks.com/accessible-svgs/) when making these accessible but this makes a lot more sense. I’ll try to get to this today.

scottkellum:
@maya I’m playing with this and it doesn’t feel like it would be a great experience if the user has low vision. The visual indicators on screen are no longer aligned with the text. Navigation does feel much better for users who can’t use the screen.

Updates pushed though, let me know how these feel to you.

maya:
I see what you're saying and am checking if that's an issue. A little more feedback I've gotten:

Someone had mentioned using a toggle to show the table instead of a chart. For someone using a magnifier, the chart may be harder to read than the table. And he half-expected the link to data sources to show the table. 

Whether there's a toggle to replace the chart with the table or just show both may depend on the audience and am not sure which approach is the best way.

scottkellum:
This is really good feedback @maya thank you! I think I’m going to take a step back and diverge a bit thinking about the various ways to display this data and giving both authors and users more flexibility in how that data is presented.

maya:
I've shared this with our Section 508 listserv and am getting some helpful responses (will keep them coming as I get them):

---
Maya,

I know it’s not quite the question, but I’m sighted and often prefer a table to a graph. Also, if I really want to get into the data, if there’s a table, I can copy it into Excel and sort, filter, etc.

So, that’s the general usability argument for keeping the table for everyone. As for accessibility, people with low vision or cognitive disabilities might be negatively affected by not having access to the table.

People with low vision might also want to copy and paste into Excel to have more control over what they see. Some people with low vision have a very small field of vision, so looking at the middle of a bar chart could leave them without the ability to see the X or Y axis. A table might be easier to follow, even if the header row and first column are out of the field of vision. I don’t know, but it’s probably worth investigating. Also, some people with low vision have to zoom to 200% or more—SVG is a great start, but that can still put the X and Y axes out of view.

I also imagine there must be cognitive disabilities for which a table is preferable. I know for myself that the likelihood that I’d prefer a table goes up with the noise or other distractions around me, so perhaps people with attention-related disorders?

They’re just thoughts, really, but my immediate gut reaction to your idea was “No! Don’t take my table away!” Charts are great for most of us for a quick look, but as soon as you start really thinking about the data, tables are just way easier to use.

Bethany Letalien
Section 508 Coordinator
U.S. Office of Personnel Management
Office of the Chief Information Officer

---

Maya,

I prefer your last two display options: (1) show the table below the graph or (2) add a toggle to show the table instead of the graph.

The reasons:

- You do not want actionable hidden content (e.g. the first “data sources” link) to be accessible via the keyboard, because keyboard navigation is not always paired with using a screen reader to hear the hidden content.  Users may access content using only the keyboard for a variety of reasons, and a visual indication of where the keyboard focus is currently located is important.
- People absorb information in different ways. If the data is already prepared in both graph and table format, why not allow all users to access both?
- Maintenance… Hidden content often gets overlooked when updates are made.

Side note:  I’m using IE11. When I open your example link, the bar chart is empty and I do not see a table. If I view in Compatibility Mode, then the chart disappears and I only see the table.  It could be my locked down environment, or there could be some browser troubleshooting needed on your end.  Good luck!

Best wishes,
-Dawn

Dawn Wyse Hurto
Enterprise Business Solutions  |  Office of the Chief Information Officer  |  U.S. Department of the Treasury
DHS Section 508 Trusted Tester - Certification Number 300297

scottkellum:
Hey @maya, Unfortunately I don’t have bandwidth to work on this right now or in the near future.

maya:
Ah OK, thanks so much for the work so far! Closing this PR for now until we have the bandwidth to pick it up and continue building on this great work.

PR #2330, opened 2018-01-30
Remove text color and decoration from usa-footer-return-to-top

line47:
Remove text color and decoration from usa-footer-return-to-top section.

Let the styles for the link element do their thing :)

### Reasoning
I'm wanting to put some other content into the `usa-footer-return-to-top ` div that is not a link and therefore don't want the text to be blue and underlined. I'm thinking of using this return to top section as a page footer with meta information like last updated date and a link to take you to GitHub for editing.

### Screenshot
The first item and last item are links and the last updated date/time is just text

screen shot 2018-01-30 at 12 36 12 pm

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -41,10 +41,8 @@
 }
 
 .usa-footer-return-to-top {
-  color: $color-primary;
   padding-bottom: $spacing-medium;
   padding-top: $spacing-medium;
-  text-decoration: underline;
 }
 
 .usa-footer-primary-section {


donjo:
I think this makes sense but I also think that maybe this should get a different class name that describes what is being styled since we're not really styling the word "return to top" as much as we're styling a div that provides some spacing above the rest of the footer. Maybe `usa-footer-anchor`, `usa-footer-page_nav`, `usa-footer-pre_section`? 

Something that communicates that it's for things that are intended to be at the bottom of the page but context-aware of the page you are currently on.

line47:
I agree 💯 . I like the idea of `usa-page-footer` or something similar as you mentioned above.

maya:
Agree, it could have a different class name. How about `usa-footer-meta` (or first, top, etc...)?

The other two sections below it are `usa-footer-primary-section` `usa-footer-secondary_section` which is inconsistent with the dividers preceding `section` and should be fixed too (e.g. `usa-footer-section-primary`, `usa-footer-section-secondary`). The modifiers above could be added after `usa-footer-section-*`.

Issue #2255, opened 2017-12-01
Proposal for an accessible color system


thisisdano:
Choosing appropriate, accessible colors can be a challenge and the Standards could help make this easier.

We currently support a very limited color palette (blue, red, gray, and a couple special states) by design. We intend to support only a baseline of color options and expect modifications to the standards will conform to Section 508 contrast guidelines, as tested with the many tools available for this purpose. 

I propose that the Standards provide a fuller color palette, in combination with color guidance and sensible defaults, designed to allow a much broader expressive gamut with clear accessible combinations. This broader palette would be more useful design tool out of the box and provide a consistent pedagogy for understanding the contrast implications of color choices.

The broad framework for this palette is directly inspired by now a now-deprecated (I believe) palette developed at IBM. 
- Distinct hues with descriptive names (like `blue`, `red`, or `violet`)
- Numbered level (brightness) variants within the hue
- A color level scale that ranges from `0` (`#fff` white) to `100` (`#000` black) in increments of 10
- Color combinations that differ by `40` or more would  achieve WCAG AA Large contrast
- Color combinations that differ by `50` or more would conform to the Section 508 contrast standard (WCAG AA); thus, values of `50` would be Section 508 conformant against pure white and pure black.
- Color combinations that differ by `60` or more would achieve WCAG AAA contrast
- Vivid variants with suffix `v` at some levels for a more complete expressive gamut (like `violet-40v`)

This results that colors that can be expressed like `red-40`, `warm-gray-90`, `teal-50v` as variables (`$red-40`), tokens, elements of a utility classes (`.bg-gray-90`).

- - -

#### Benefits

This rubric would allow designers and developers to design with confidence, understanding by color name alone whether the contrast would be conformant (whether the combination is useful, practical or aesthetically appropriate is another matter).

This system would provide a strong baseline for design, and a clear baseline for explicit guidance.

- - -

#### How might this look in practice?

The following system of 27 named hues and 308 total colors is a proposal for how the above rubric could be implemented. I consider this proposal something between a proof of concept and a draft. It's a palette I've consciously designed, based on the existing Standards palette and other palettes used by sites developed on top of the Standards. It can still use some work. *It is definitely a work in progress.* 

That said, I've used variants of this system and this palette for all my visual design work over the last year and it has proven a useful tool, worthy of exposing to a wider community.

Check my work at http://jxnblk.com/colorable/demos/text/ or help develop a tool to make this checking easier!

[...]

PR #2333, opened 2018-02-05
add some fractal examples for utility classes

donjo:
Adds some initial structure for documenting utility classes. The docs folder is a place where some of the overviews for utility things can live. I was hoping to create an additional group (like Components) for utilities but that may not be possible in this version of fractal.

With that said, I think we could do some initial cleanup and grouping on the fractal components that we document to make a shorter list of high level categories to make utilities easier to find. As it stands now they pretty much mirror how things are documented on `uswds-site` but I don't think that's necessary anymore given how we use the fractal site. 

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fractal-utility-structure)


src/components/utilities/color/colors.config.yml
```diff
@@ -0,0 +1,43 @@
...
+      - gold
+      - orange
+      - peach
+      - terra
+      - red
+      - magenta
+      - purple
+      - gray-cool
+      - gray
+      - gray-warm
```
maya: @thisisdano mentioned he needs to change the color names, so this would need to change.

donjo:
Hmmm so while this was just intended to be a PR for some placeholder content I'm realizing that it's not going to be able to pass any of the aXe tests unless we manually change the text color some of these color examples. I'd vote for turning off the testing on all the utilities examples since their job is to just be a visualization for the tokens in the system and not actual endorsed implementations. However, I'm not quite sure how to do that (this problem is blocking us on uswds/uswds-site#450 too)

maya:
I'd check with @thisisdano with how he wants colors displayed. If we moved the text off of the color, then the errors won't flag. Want to go through this a bit as we know folks will be looking towards this for accessibility guidance as well.

screen shot 2018-02-06 at 3 03 36 pm
via #2255.

donjo:
Oh yeah, that style makes sense. Just pushed a temporary change to move the text out of the backgrounds so it will past tests but we can talk about better ways to display that (maybe like the sketch example above)

PR #2337, opened 2018-02-08
Add circle utility for both settings and units

thisisdano:
This also modifies the build function to allow multiple properties in a utility, as well as a utility-level extend.

- **Prototype utility** accepts all grid unit values and pixel values
- **Production utility** accepts spacing settings values

- - -

This also removes the percentage values from `square` since they don't work.

- - -

```
/*
Circle
usage:
  .circle-[value]
output:
  height: [value];
  width: [value];
  border-radius: 50%;
example:
  .circle-6 {
    height: 3rem;
    width: 3rem; }
*/
```

- - -

- This also adds a `-circle` value to prod and proto `radius` utilities
- Changes the value of `pill` to `9999em`... _it's the only way to be sure._
- It adds top (`t`), right (`r`), bottom (`b`), and left (`l`) modifiers to the proto `radius` utility to allow setting border radius on only one side of a box

### DIFF
--- a/src/stylesheets/utilities/settings/rules/border-radius.scss
+++ b/src/stylesheets/utilities/settings/rules/border-radius.scss
@@ -3,17 +3,17 @@
 Border Radius
 ----------------------------------------
 usage:
-  .br[-]*[value]
+  .radius-[value]
 ----------------------------------------
 output:
   border-radius: [value];
 ----------------------------------------
 example:
-  .br0 {
-    border-radius: 0; }
+  .radius-md {
+    border-radius: $s-spacing-md; }
 
-  .br-pill {
-    border-radius: 10em; }
+  .radius-pill {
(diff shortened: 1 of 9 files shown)


src/stylesheets/utilities/settings/rules/border-radius.scss
```diff
@@ -23,7 +23,8 @@ $border-radius: (
     modifiers: null,
     values: (
       0: '0',
-      pill: ('10em', isReadable),
+      pill: ('9999em', isReadable),
+      circle: ('50%', isReadable),
```
maya: Why is it 50% and not 100%?

thisisdano: They are effectively the same, since a roundness of 50% is all you need to get to the midpoint of the square. (I have checked this!)

Issue #2215, opened 2017-11-07
[Typography] Improve typography spacing


maya:
**User story**: As a developer using the USWDS, I'd like typography to consistently display proper margins in different use cases (from simple layouts to complex grids).

### ✅ Ideal visual, margins collapse, single div. 
Top and bottom margins on headings and paragraph text
<img width="925" alt="screen shot 2018-08-16 at 1 39 31 pm" src="https://user-images.githubusercontent.com/5249443/44233865-e1f6a880-a159-11e8-9e49-8a58ab3c7160.png">

### ✅with divs wrapped around the heading and p text, with a single background color
<img width="966" alt="screen shot 2018-08-16 at 1 46 13 pm" src="https://user-images.githubusercontent.com/5249443/44234130-a90b0380-a15a-11e8-917c-f9193c772edb.png">

- pretty good with white background

### ❌but when you want to add color to the section, the margins mess it up and don't allow the divs to be filled with color equally 
<img width="970" alt="screen shot 2018-08-16 at 1 45 34 pm" src="https://user-images.githubusercontent.com/5249443/44234152-ba541000-a15a-11e8-813d-fef080726f3b.png">

### ✅removing top and bottom margins of divs fixes this (currently behavior in USWDS)
![36122745-59710eba-0fff-11e8-9c00-73650204475c](https://user-images.githubusercontent.com/5249443/44234227-f25b5300-a15a-11e8-87dd-b7f840e6dcfa.png)
- works well when you add padding between sections

### ❌But when you use a single background without additional sections to add space between, they are collapsed with each other 
<img width="968" alt="screen shot 2018-08-16 at 1 49 44 pm" src="https://user-images.githubusercontent.com/5249443/44234346-377f8500-a15b-11e8-8ad9-c68ae1626a55.png">

## Potential solutions
1. #2343 - only use top margin on `p` text, and remove top and bottom margins of first elements inside certain elements https://github.com/uswds/uswds/issues/2215#issuecomment-365080405
2. GOV.UK uses contextual adjustments to add top padding to headings that appear directly after paragraphs. https://github.com/uswds/uswds/issues/2215#issuecomment-405679277. The code seems only added to `@extends` and classes, not directly to the elements.

## Reference materials
- [Slack convo [Design systems Slack]](https://design-systems.slack.com/archives/C0ENULNQZ/p1509396679000313) 
[...]

Issue #2241, opened 2017-11-16
Request: provide css that only applies to classes, not tags


dpgalloway:
I want to use specific components of the uswds CSS, for example alerts. However, when I link to the uswds.css stylesheet, it overrides almost all of the styles on the site, no matter what order the link tag is placed. This happens because there are several styles in the CSS that apply to tags as a whole, e.g. html {...}, body {...}, so the style is applied no matter what.

I would like to request an alternate CSS file/repo that only applies styles to elements with the specified classes, i.e. only apply styles when an element has a class that starts with "usa-" and leave everything else as-is.

thisisdano:
I've run into this scoping issue myself (adding the government banner to digitalgov.gov) and it hurts. Our base element styling (and our current use of rems) effectively prevents integrating the USWDS into existing sites. I had to do a lot to limit the scope of the USWDS styles to only my component.

I would also hope that this kind of change would be appropriate for our primary css as well (not just a variant), or possibly configured through something like a sass variable. But it may be tricky! (It always is.)

maya:
TODO:
- Document what elements (files) have sizing styles applied to them. 
  - Take a look at normalize and font sizes.
- Think about applying mixins to these elements to either apply to base elements or add'l classes
- Question: should default be applied to base elements or not

maya:
Here's a proposal that gets to this:

1. By default, typography styles are not applied to directly to elements with `$typography-base: false !default;`
2. If users would like typography styles applied to base typography elements, they'll add `$typography-base: true`
3. To use the typography styles, nest the elements inside a `usa-content` element (could also be called `.prose-*` like [Mapbox's Assembly](https://www.mapbox.com/assembly/documentation/#Prose), but we already use `usa-content`)
4. Use the styles as classes with [utility classes](https://v2.designsystem.digital.gov/utilities/font-size-and-family/) e.g. `.font-body-md`, `.font-heading-lg`

```
$typography-base: false !default; // Sets the default typography base styles to false, will be wrapped in usa-content

@mixin typography-base($typography-base) {

  @if $typography-base == false {
    // Opening .usa-content tag
    .usa-content {
  } 
      // Text styles go here
      p {
        @include paragraph-text;
      }  

      h1 {
        @include h1;
      }

      // Additional text styles go here (e.g. headings, lists, etc...)

  @if $typography-base == false {
    } // Closing tag
  }
}

// Class-based paragraph text styling
.usa-p {
  @include paragraph-text;
}

[...]

PR #2343, opened 2018-02-12
[WIP] Fix typography spacing

maya:
Seeks to ensure typography consistently displays proper margins in different use cases (from simple layouts to complex grids). There were some issues with typography spacing in different contexts (grid, columns, section classes).

Changes include:
* Going one-directional in paragraph text `p` margins (by using top margins only) to help with typography stacked inside multiple elements
* Remove top margins of elements in the first grid item column items and leaves the following grid items with their natural spacing
* When using the section classes (`usa-section`) which provide top and bottom spacing, remove the top and bottom margins of first and last elements inside it respectively, so that the section's spacing provides the intended spacing. 

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-typography-spacing/components/detail/type-spacing.html)

This is very much a work-in-progress and tries to show examples of possible solutions.

Screenshot of the preview linked: black outline is the `usa-grid` class and red outline is the `usa-width-*` classes:
127 0 0 1_3000_components_preview_type-spacing

Fixes: #2215.


alex-page:
We recently solved this problem using lobotomised owls in our [design system](https://gold.service.gov.au/), you can take a look at an example here:
https://github.com/govau/uikit/blob/master/packages/accordion/src/sass/_module.scss#L27

Our inspiration came from this detailed article:
https://alistapart.com/article/axiomatic-css-and-lobotomized-owls

If you want to collaborate on solving this issue let me know! Happy to help.

maya:
@alex-page Thanks! The technique that we use is similar to a scoped version of that but it still causes some issues which is the reason for this PR. I've also heard from Heydon Pickering previously that even he doesn't recommend that technique on existing projects, it's much easier to add from the beginning. 

P.S. Great job on GOLD. Looks SOLID! 🥇

melwoodard:
:wave: a bunch of us on the Vets.gov team at DSVA are super excited to see this PR happening! 

We've been wanting to approach spacing for some time and add some more sophisticated rules to allow content to line up right in different scenarios, and were really happy to see this happening within USWDS instead of rolling our own overrides. 

Any sense of timing for when this might be available in the design system?

maya:
@melwoodard thanks for your interest in this work! We're excited that this is something you're looking for. This work would likely make its way into our next major release (2.0) which is around a couple of months away. If you are able to use this work in your project and let us know how it's working for you, that would be great feedback for us and go a long way to making this update stronger and validated.

el-mapache:
@maya How is this coming along? Its been open for a bit so I just wanted to give it a lil bump

maya:
No updates yet, will probably work on this after #2241. If you'd like to get a better understanding of the issue, check out #2215 and GOV.UK's solution: https://github.com/uswds/uswds/issues/2215#issuecomment-405679277.

maya:
Closing this PR in favor of #2696.

/cc @alex-page @melwoodard let us know what you think!

PR #2347, opened 2018-02-13
Trap focus to mobile navigation menu and other accessibility fixes

### LINKED ISSUE #2031: Mobile navigation menu allows tab focus outside of menu
I'm not sure if this is actually an unintended bug or an intended feature, but I figured I'd document it here just in case.

Basically, when in the mobile navigation menu, the user can <kbd>Tab</kbd> the keyboard focus outside of the menu, into the dimmed content of the actual page.

The menu essentially seems to function like a dialog box, and behaves much like one from a mouse/touch standpoint, since taps/clicks outside the menu collapse it, so we might want to make sure it behaves like one from a keyboard standpoint as well.

Doing this is a bit of a lift--see Rob Dodson's [Modals and Keyboard Traps](https://www.youtube.com/watch?v=BoAsayPVogE) video of the [A11y Udacity class](https://www.udacity.com/course/web-accessibility--ud891) for full details--but could help provide the infrastructure we need for providing accessible modals in the future, too (see #911 / #1477).

While we're at it, we might also want to add support for using <kbd>Esc</kbd> to get out of the nav menu too.

maya:
This does a few things:
* Trap focus to mobile navigation menu so that the user cannot <kbd>Tab</kbd> outside the menu
* Add ability to close navigation by <kbd>Esc</kbd> key
* Change nav display to none in order to properly skip navigation when tabbing through
    * Change back to display flex when is-visible is active
    * Use keyframes animation in order to re-add animation that was not working bc of display none

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-mobile-nav-a11y/components/preview/layout--landing.html)

## TODO 

- [x] When it's mobile size and I tab through, then make it full size again, focus is stuck in the menu. How can I untrap it when moving from mobile to desktop? -- worked on this with @el-mapache ✅ 
- [x] Review JavaScript quality and style, this was written in a specific manner and want to make sure it's consistent -- worked on this with @el-mapache ✅ 

Fixes #2031

### DIFF
--- a/src/js/components/navigation.js
+++ b/src/js/components/navigation.js
@@ -9,7 +9,7 @@ const PREFIX = require('../config').prefix;
 
 const NAV = `.${PREFIX}-nav`;
 const NAV_LINKS = `${NAV} a`;
-const OPENERS = `.${PREFIX}-menu-btn`;
+const OPENERS = `.${PREFIX}-menu-btn`; // Will hold previously focused element
 const CLOSE_BUTTON = `.${PREFIX}-nav-close`;
 const OVERLAY = `.${PREFIX}-overlay`;
 const CLOSERS = `${CLOSE_BUTTON}, .${PREFIX}-overlay`;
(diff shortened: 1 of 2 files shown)


maya:
>Move to a standalone component?
@el-mapache yeah, if we offer a modal component, we'd use a lot of the same code, right?

el-mapache:
@maya Yes exactly, we did some similar stuff for login.gov. Probably don't need to break it out till we need to, though.

PR #2353, opened 2018-02-14
Fix build command instructions

### LINKED ISSUE #2352: "npm run build:package" doesn't work
Our [CONTRIBUTING](https://github.com/uswds/uswds/blob/develop/CONTRIBUTING.md#building) file  says we can run `npm run build:package` or `gulp release` to build the release, but when I tried `npm run build:package`, it didn't work, but `gulp release` did work. Should we remove the npm command or is there another one that works? cc @donjo

maya:
Fixes #2352.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,9 +123,7 @@ any conflicting screenshots.
 To build the `uswds` package in preparation for releases, run:
 
 ```sh
-npm run build
-# or
-gulp release
+npm run prepublish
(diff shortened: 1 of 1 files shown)


maya:
Getting this error when I try that:

`npm ERR! missing script: prerelease`

donjo:
@maya oops, the command is actually `prepublish` – try that one

maya:
Fixed in 9609c4c.

PR #2354, opened 2018-02-14
Add experimental components

maya:
Add experimental components by ignoring Sass files with an underscore. Sass commented out from `uswds.scss`.

## TODO
- [ ] Do this for JS: ignore JS files that start with an underscore


config/gulp/javascript.js
```diff
@@ -70,6 +70,7 @@ gulp.task('eslint', function (done) {
 
   return gulp.src([
       'src/js/**/*.js',
+      '!src/js/**/_*.js',
```
donjo: following the pattern of the sass example did you mean for this to be `!src/js/_**/*.js`?

maya: For the JS file, I tried leaving it in the  `src/js/components/` folder and added an underscore to the file like this:  `src/js/components/_chart.js`. The Sass files all have underscores already which is why I moved it into the ``src/scss/_experimental` folder.

Issue #2049, opened 2017-07-19
[HTML5 tags]: Prefer HTML5 tags in code examples


jordangov:
## Description

There are many code examples across the UI Components section which use outdated HTML tags (mostly a lot of `div`s). This issue is a holding place for a task to sweep the code examples and replace `div` tags with the semantic HTML5 appropriate tag (`nav`, `main`, `section`, `aside`, `header`, `footer`, etc)

This will set a precedent for proper modern HTML standards and also supports older browsers in case required. Browsers default unknown tags with `div` styles.

## Additional information

Some specific examples:
* [side navigation](https://standards.usa.gov/components/sidenav/)
* [accordions](https://standards.usa.gov/components/accordions/)

AndreaSigz:
Thanks @jordandds. @donjo @atul

msecret:
I definitely think this could be done in some areas. We'd have to be careful of using context-based tags, like `aside`, `main` for certain components because the components could be used in different contexts. For example, somebody might use the accordion for nav, but it could also be used in main content.

maya:
I know somewhere it was mentioned the side navigation using the `nav` element, but what did the accordion need?

jordangov:
The accordion is all div's now... but probably should be a `<main>` along with `<section>`'s and then (IMO) `<header>`'s versus using a button.

maya:
A `main` makes sense (and perhaps `section` too). I think the button is fine to get the accessible features from it, ideally, we could use a heading but we don't know what heading level someone would need and could see it leading wrong heading levels from copy + paste.

See Heydon Pickering's Collapsable Sections:

```
<main>
  <h2>
    <button aria-expanded="false">
      Section 1
      <svg aria-hidden="true" focusable="false" viewBox="0 0 10 10">
        <rect class="vert" height="8" width="2" y="1" x="4"/>
        <rect height="2" width="8" y="4" x="1"/>
      </svg> 
    </button> 
  </h2>
  <div hidden>
[...]

PR #2360, opened 2018-02-15
Update accordion to use HTML5 elements

maya:
This has some breaking changes so possibly would make more sense for 2.0.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-accordion-html5/components/preview/accordion--default.html)

Related: #2049.


jordangov:
Hmm... I'm not sure what that specific warning is for, but I do see that there isn't a single node for each accordion panel, which I think is going to be necessary. Basically, the `{% for item in accordion.items %}` needs a single element inside of it that represents the entire accordion panel?

maya:
The error is referring to this ["kitchen sink" page](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-accordion-html5/components/preview/kitchen-sink.html) and is saying "you can't have a `<main>` inside a `<main>`" since the main content of the page is wrapped in a `<main>`. Is that error accurate?

maya:
>Basically, the {% for item in accordion.items %} needs a single element inside of it that represents the entire accordion panel?
Hmm... I didn't see each panel wrapped in an element in Heydon's example: https://codepen.io/heydon/pen/QqzRvQ

joshbruce:
Can a page have more than one `main`? Thought this was going to use the `summary` + `details`.

joshbruce:
Oh! My bad @Maya - was looking at this on my phone - thought each accordion was a `main` for some reason. I dig it. 

In case you didn't see this conversation before - not sure if there's anything that might change anything here: https://github.com/uswds/uswds/issues/1136

maya:
>Thought this was going to use the `summary` + `details`.
I would love to use that element. The problem is I don't really know the state of how accessible it is. Would we ultimately have to write custom JS to make it accessible? So rather than getting the accessibility of a native button like here, we'd have to write JS to do that? So you end up either having to write JS to open and close it and get a11y for free, or have it open + close natively and write JS to make it accessible. We'd also need to provide a polyfill as it looks like it's not supported in any IE.

Thanks for linking to that issue, as you can see in here, we're not using lists for it :)

maya:
Also, posting this to look at: https://scottaohara.github.io/accessible_accordions/

maya:
@jordangov I think `<header>` would have to be changed to a heading level  (e.g. h2) to be more accurate. I worry it could be misused with copy + paste, so we can add a comment in the code to alleviate this.

jordangov:
sure, I think we're on the right track... and anything is better than div's ;)

joshbruce:
I feel like we’ve been here before, but I missed where we went a different route. Seriously getting a sense of de ja vu.

The JS used to run on load to set `aria-hidden` to true, which handles the accordions expanded by default thing, because they would collapse when JS ran...think I actually did a PR for something related to it; could be wrong.

Then @maya and I played a PR iteration dance that was fun related to an `aria-hidden` thing.

Glad to the `div` and list go away and trying to get back to possibly a more simplified implementation.

maya:
It sounds familiar, but when I ran the accordions with JS turned off, they're all closed.

joshbruce:
Yeah? 

So weird. Check this file...and related things (including CSS): https://github.com/joshbruce/web-design-standards/blob/staging/src/js/components/accordion.js

For the CSS, check PR history. Their was like three of us involved. A user with”duck” in his handle is coming to mind.

And all the conversation re `aria-hidden` and how it shouldn’t be associated with CSS - specifically `display: none`.

joshbruce:
Feels like JS maybe started expanding a bit more than it maybe needed to (compared to progressive enhancement)?? 

But I wasn’t paying attention to the conversations that got that going...this general trend on the web is partially the inspiration behind: https://m.8fold.pub/on-constraints-internet-bandwidth-eab05c20e218?source=linkShare-f4442260749b-1518840173 - shameless plug...a little sorry about that.

src/components/06-accordion/accordion.njk
```diff
@@ -1,16 +1,17 @@
...
       aria-controls="{{ accordion.id_prefix }}{{ item.id }}">
       {{ item.title }}
     </button>
-    <div id="{{ accordion.id_prefix }}{{ item.id }}" class="usa-accordion-content">
-      {{ item.content }}
-    </div>
-  </li>
+  </h2>
+  <div id="{{ accordion.id_prefix }}{{ item.id }}" class="usa-accordion-content">
+    {{ item.content }}
```
donjo: In an effort to keep HTML tags out of the config we should probably move the `<p>` here to wrap {{ item.content }}

maya: I'm testing other text elements (e.g. headings and lists) to ensure resilience and flexibility in the accordion panels, so keeping the text in the config for now.

maya:
Merging into 2.0 branch... _bombs away_

PR #2363, opened 2018-02-20
USWDS  - navigation.js

el-mapache:
* Necessary since all the js components are included everywhere
* In the future we might want to audit how these components are
instantiated, components such as `navigation` should probably be opt-in

## Description

Fixes issue in which the navigation.js component throws an error on each page without the expected DOM nodes.

## Additional information

* Necessary since all the js components are included on each page.
* In the future we might want to audit how these components are
instantiated, components such as `navigation` should probably be opt-in.

Ran through this locally and pages appear to be error free!

### DIFF
--- a/src/js/components/navigation.js
+++ b/src/js/components/navigation.js
@@ -20,8 +20,7 @@ const VISIBLE_CLASS = 'is-visible';
 
 const isActive = () => document.body.classList.contains(ACTIVE_CLASS);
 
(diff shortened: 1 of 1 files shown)


maya:
Adding a preview link:

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/ab-check-nav-node)

PR #2365, opened 2018-02-21
Simplify Sass asset path variable overrides 

### LINKED ISSUE #1797: Simplify Sass path variable overrides
Currently it's necessary to set two different variables to change the path to the font and image files relative to the CSS output path. For instance, if you were to copy the whole `dist` directory to your project's `assets/uswds` directory and wanted your custom CSS to live in `assets/css/main.css`, like so:

```
assets/
├── css/
│   └── main.css
└── uswds
    ├── css/
    ├── img/
    ├── fonts/
    └── ...
```

Then the Sass that generates `assets/css/main.css` would need to do something like this:

```scss
$font-path: '../uswds/fonts';
$img-path: '../uswds/img';

@import 'path/to/uswds/all';
```

This feels redundant and a bit silly. I think that we can simplify this for users by introducing a new variable that establishes the USWDS base path under which we'd expect to find `fonts` and `img` directories:

```scss
$uswds-path: '../' !default;
$font-path: '#{$uswds-path}fonts' !default;
$img-path: '#{$uswds-path}img' !default;
```

This way, you can just override `$uswds-path`, and both `$font-path` and `$img-path` will by default be set to the respective expected paths.

maya:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/simplify-asset-path)

@donjo @thisisdano Does the variable `$uswds-path` make sense? Could also be `$asset-path`.

TODO:

- [x]  Add documentation in https://github.com/uswds/uswds-site, https://github.com/uswds/uswds-site/pull/504 (this PR is not dependent on it)

Fixes #1797.

### DIFF
--- a/src/stylesheets/core/_variables.scss
+++ b/src/stylesheets/core/_variables.scss
@@ -125,7 +125,7 @@ $uswds-path: '../' !default;
 
 // Relative font and image file paths
 $font-path: '#{$uswds-path}fonts' !default;
-$img-path: '#{$uswds-path}img' !default;
+$image-path: '#{$uswds-path}img' !default;
 
 // Set $asset-pipeline to true if you're using the Rails Asset Pipeline
 $asset-pipeline:      false !default;
--- a/src/stylesheets/core/_variables.scss
+++ b/src/stylesheets/core/_variables.scss
@@ -121,11 +121,11 @@ $medium: new-breakpoint(min-width $medium-screen $grid-columns-medium) !default;
 $large: new-breakpoint(min-width $large-screen $grid-columns-large) !default;
 
 // Set the base path for assets (used for font and image paths below)
-$uswds-path: '../' !default;
+$asset-path: '../' !default;
 
 // Relative font and image file paths
-$font-path: '#{$uswds-path}fonts' !default;
-$image-path: '#{$uswds-path}img' !default;
+$font-path: '#{$asset-path}fonts' !default;
+$image-path: '#{$asset-path}img' !default;
 
 // Set $asset-pipeline to true if you're using the Rails Asset Pipeline
 $asset-pipeline:      false !default;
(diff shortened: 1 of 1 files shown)


thisisdano:
hm, hm, hm!
I'd originally thought that `$uswds-path` was more clear and correct, though now I'm not so sure.

So this var is set so the compiled CSS points to the proper spots for CSS-referred files like background images and font files (are there any other kinds of files CSS might refer to?). It is the relative path _from_ the compiled CSS _to_ the referred assets.

I get it confused in my head because something like `$uswds-path` is _also_ the kind of var I might need to set in my gulpfile to make sure that the sass transpiler knows where to find the USWDS sass source files (in addition to the override sass files I've created to modify the defaults).

So, since this is really scoped to "where to find the USWDS assets", now I tend to think that `$asset-path` in the context of a USWDS sass project setting var (with the other path vars listed below, and maybe some helper text as well) is the most clear and correct option.

```
$asset-path: '../uswds/' !default; 
// path to font and image assets from the compiled USWDS CSS
// example:
// ├── css
// │   └── main.css
// ├── other-stuff/
// └── uswds
//     ├── img/
//     ├── fonts/
//     └── ...
$font-path: '#{$asset-path}fonts' !default;
$img-path: '#{$asset-path}img' !default;
```

🤷‍♀️

maya:
Sounds good, happy to update the variable name. 

I'd prefer to keep sample code in the documentation in this PR https://github.com/uswds/uswds-site/pull/504 not in the variable file.

Also, `$asset-path: '../uswds/' !default;` won't work it needs to stay as `$asset-path: '../' !default;`

thisisdano:
Fine with leaving documentation on the site for this, but I wonder if including/repeating _some_ documentation in the code might make it more useful

maya:
Updated variable name in 5d40a3e.

PR #2368, opened 2018-02-23
Change input optional text to HTML content from CSS content

### LINKED ISSUE #2367: 508 error content inserted with css
We got a 508 error from SortSite on these lines:

https://github.com/uswds/uswds/blob/master/src/stylesheets/elements/_inputs.scss#L91
https://github.com/uswds/uswds/blob/master/src/stylesheets/elements/_inputs.scss#L96
 
Level A
 
2 issues on 6 pages

``` 
Content inserted with CSS is not available to some screen readers, and to people with low vision who turn off style sheets.
Move the content from CSS to the HTML page so that all users can see it.CSS content causes problems in:

• Firefox when the No Style menu item is chosen (CSS content disappears)
• NVDA 2016.2 with IE11 (CSS content not read)
• JAWS 17 with IE11 (CSS content not read)
• JAWS 16 with IE11 (CSS content not read)
• JAWS 15 with IE11 or Firefox 29 (CSS content not read)
• WindowEyes 9.2 with IE11 (CSS content not read)
```

Here's the 508 guideline. I guess the question is whether the "optional" and "required" text is decorative or not. I would say it conveys information and isn't merely decorative.
https://www.w3.org/TR/WCAG20-TECHS/F87.html

maya:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-input-label-helper/components/detail/name-form)

Question: do we leave the CSS content CSS in or remove it? Is it a breaking change?

Fixes #2367.

### DIFF
--- a/src/components/07-form/templates/address-form.njk
+++ b/src/components/07-form/templates/address-form.njk
@@ -5,8 +5,7 @@
       <label for="mailing-address-1">Street address 1</label>
       <input id="mailing-address-1" name="mailing-address-1" type="text">
 
-      <label for="mailing-address-2" class="usa-input-optional">
-        Street address 2</label>
+      <label for="mailing-address-2">Street address 2 <span class="usa-input-label-helper">(optional)</span></label>
       <input id="mailing-address-2" name="mailing-address-2" type="text">
 
       <div>
--- a/src/components/07-form/templates/name-form.njk
+++ b/src/components/07-form/templates/name-form.njk
@@ -2,17 +2,17 @@
   <form class="usa-form">
     <fieldset>
       <legend>Name</legend>
-      <label for="title" class="usa-input-optional">Title</label>
+      <label for="title">Title <span class="usa-input-label-helper">(optional)</span></label>
       <input class="usa-input-tiny" id="title" name="title" type="text">
 
       <label for="first-name">First name</label>
-      <input id="first-name" name="first-name" type="text" required="" aria-required="true">
+      <input id="first-name" name="first-name" type="text" required aria-required="true">
 
-      <label for="middle-name" class="usa-input-optional">Middle name</label>
+      <label for="middle-name">Middle name <span class="usa-input-label-helper">(optional)</span></label>
       <input id="middle-name" name="middle-name" type="text">
 
       <label for="last-name">Last name</label>
-      <input id="last-name" name="last-name" type="text" required="" aria-required="true">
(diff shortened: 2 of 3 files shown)


src/components/07-form/templates/address-form.njk
```diff
@@ -5,8 +5,7 @@
       <label for="mailing-address-1">Street address 1</label>
       <input id="mailing-address-1" name="mailing-address-1" type="text">
 
-      <label for="mailing-address-2" class="usa-input-optional">
-        Street address 2</label>
+      <label for="mailing-address-2">Street address 2 <span class="usa-input-label-helper">(optional)</span></label>
```
maya: Do we like this class name?

donjo: that sounds good to me. I could see wanting to keep it more descriptive like `usa-input-label-required` but there are probably multiple uses here.

src/components/07-form/templates/address-form.njk
```diff
@@ -5,8 +5,7 @@
       <label for="mailing-address-1">Street address 1</label>
       <input id="mailing-address-1" name="mailing-address-1" type="text">
```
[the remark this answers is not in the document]
donjo: I think one example of the `usa-input-label-required` class would be good somewhere

donjo:
To answer the question: I'd say leave them in as deprecated since it would be a breaking change otherwise. Maybe elaborate on the note to say why they were deprecated so we (or someone else) remembers what exactly is being cleaned up when 2.0 comes along.

src/components/07-form/templates/address-form.njk
maya: Could also be `-note`, `-info` - open to something better.

maya:
Elaborated on the deprecated text in 9d63021.

src/components/07-form/templates/address-form.njk
thisisdano: i like `input-label-helper`

thisisdano: are there distinct styles for helper text for the `required` text and other (non-`required`) helper text?

maya: Yes, required is red and helper is gray.

hursey013: Only alternative I can think of would be `input-label-text` to be slightly more descriptive.

Issue #2348, opened 2018-02-13
Audit the build process


donjo:
With the addition of utility classes our circleCI build time has increased by quite a bit. It's been a while since we've taken a look at the build/test process and I believe that the package may be building itself more times than necessary

maya:
Looked at post and pre-utilities build times to find the bottlenecks:
![chart 1](https://user-images.githubusercontent.com/5249443/36455340-868c99fe-1654-11e8-87a2-d82a87704b57.png)
[🔒 Source](https://docs.google.com/spreadsheets/d/18PQ1mmra3W22yTa_Pb2Ek4pjc_nGMI5Z5-YCMr6fhcA/edit?usp=sharing)

donjo:
A few things we learned here:

- There was an extra unnecessary build step happening, that is being removed in #2371 
- There is a certain amount of node startup time that can't really be avoided, it takes a bit longer in Circle than it does on local machines but that is to be expected considering our laptops are supercomputers compared to the instances that Circle spins up.
- There may be gains to be had by embracing parallelism, a feature offered in Circle 2.0, but will require some refactoring of our test suite to potentially see gains. We may look into this once an engineer is added to the team.
- There are still mystery variables we have yet to uncover. For example: why does the aXe test suite need to finish in 20000ms to avoid timeout errors?

maya:
Re-opening as this work is continuing.

maya:
Closing this via: #2407. 

With the help of caching and removing unneeded steps, CircleCI build times are now around 1 minute (down from 2-4 minutes) and Federalist build times are 3 minutes (down from 4-6 minutes). Good work @hursey013! 

Updated chart
![chart 4](https://user-images.githubusercontent.com/5249443/37793834-011c2f56-2dcd-11e8-87d4-fb7b21f2b046.png)

PR #2371, opened 2018-02-27
Remove unnecessary build step from circle build process

donjo:
As part of #2348 I learned that this step is no longer necessary. I think it was left over from the time before the circle 2.0 migration where building/testing/deploying were all part of one process and not two separate parts of a workflow that happen one after another.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,6 @@ jobs:
   build:
     docker:
       - image: circleci/node:6-browsers
-    parallelism: 4
     steps:
       - checkout
       - run:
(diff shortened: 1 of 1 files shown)


maya:
Is it bc `npm run build` is only needed in the deploy process?

donjo:
@maya 

Yeah I believe we had it in there as a quick way to make sure we were starting with a fresh build right before deploying to NPM but since that happens separately in the deploy process, we're good to go on that front. 

The project itself gets built as a subset of the `npm install` process and the build that is produced there is the one that tests and code coverage reports are run on.

Issue #2344, opened 2018-02-12
Clicking footer topic heading incorrectly hides other topic lists on desktop


maya:
When I click on the first "Topic" heading in the footer, all the other elements disappear.

What I expect:
- Clicking on the heading does nothing.
- Hovering over heading does not have an underline.

This looks like it happened the footer JS was refactored and caused a regression? See #1659. 

Footer: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/develop/components/preview/footer--big.html

### Gif showing behavior
![](http://g.recordit.co/ePA2iaSfrq.gif)

Is the footer JavaScript appearing on the desktop size when it shouldn't?

maya:
This looks like it happened when the topic headings were moved inside the list possibly addressing #1659 and caused a regression.

maya:
Should also note that the mobile behavior is fine (aside from an unrelated issue: #1857), the issue is just in desktop.

PR #2386, opened 2018-03-12
Disable hiding footer menus on desktop

hursey013:
It looks like the javascript for the mobile footer accordion was set up in a way that it will run regardless of screen size. This adds a conditional statement to make sure we are in a mobile layout before running. Fixes https://github.com/uswds/uswds/issues/2344

I also added a quick test to confirm it's working (but may revisit later once I get more familiar with how testing is being done).


maya:
Adding a [preview link](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/bh-footer-fix/components/preview/footer--big.html) for easy viewing.

PR #2393, opened 2018-03-14
Utility updates

thisisdano:
- This addresses more of the issues from the first round of utilities docs. 
- It also introduces a new way to access values from the system palette with project settings that load before system variables
- It introduces plugins that extend the range of values utilities can access
- It adds a new way of storing metadata about specific typefaces for more automated generation of `@font-face` rules, as well as normalized font sizing
- It adds a `@font-face` renderer that accesses this metadata
- It introduces a number of new project value names using a lot of the relative sizing rubric in the `#design-systems` slack and https://medium.com/@sitaggart/how-we-use-sass-maps-for-design-tokens-and-developer-happiness-8bae50393114
- It adds many new functions and mixins to help devs access system palette tokens from component CSS — connecting those values to the values used in utilities
- It splits the USWDS CSS intro three base files `uswds-fonts`, `uswds-components`, and `uswds-utilities`  that are concatenated into a single `uswds.css`
- It prepares us not only to use these utilities, but use the values in component CSS
- It updates the docs 🔒 [here](https://docs.google.com/document/d/1VsHtYGWyzurwGrOj0S-dMZTgeUnl4xBf3HaajivL9o0/edit#)

- - -

The code is solid and workmanlike. Could it be an order of magnitude more efficient and elegant? — probably. Is it a good working baseline for future work? — I'm counting on it.


hursey013:
@thisisdano Added a commit to update the paths to a couple of the project sass files in `_required.css` to appease CircleCI

PR #2397, opened 2018-03-17
Change h1 copy to "Page heading" on typography page

maya:
Changes the `<h1>` copy to "Page heading" from "Section heading" as `<h1>` s are used once per page. 

This would keep it consistent with what we use on the [docs template](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/develop/components/detail/layout--docs.html).

Alternatively, could also be called "Page title."

### DIFF
--- a/src/components/01-type/typesetting.njk
+++ b/src/components/01-type/typesetting.njk
@@ -13,7 +13,7 @@
 
   <h6 class="usa-heading-alt">Spacing</h6>
 
-  <h1>Section heading</h1>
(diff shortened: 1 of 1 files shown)


maya:
From @thisisdano: 
>I think `title` is a better, more accurate descriptor, but `heading` is more consistent with our use of "heading" in class names, vars, etc — so I'd probably tilt toward `heading` for its pedagogical consistency :tomato: :shrug: :tomahto:

So going to keep this as is.

PR #2399, opened 2018-03-19
Adds color maps

thisisdano:
This defines the unit colors with sass maps formatted like 
```sass
// violet-warm
$violet-warm: (
  5:   #fef7ff,
  10:  #f5dcf7,
  20:  #e0bae3,
  30:  #d096d6,
  40:  #bd73c7,
  50:  #b24bbf,
  60:  #854180,
  70:  #573455,
  80:  #332431,
  90:  #211a21,
  vivid: (
    5:  #fdedff,
    10: #f9c9ff,
    20: #f3a8ff,
    30: #ee82ff,
    40: #e05ef7,
    50: #c132d4,
    60: false,
    70: false,
    80: false,
    90: false,
  ),
);
```
it also adds a function (`color()`) to get the value from the map:

- - -

EDIT:
I've modified this function to avoid ambiguous units.

Now, for a vivid variant, use `color($hue, $value, vivid)` like `color($indigo, 50, vivid)`

- - -

I have not rewritten the builder and the color palette presets to pull directly from these maps. Instead, they continue to pull from hardcoded vars like `$violet-warm-50v` which are, themselves defined with the `color()` function. 

Rewriting the palette and builder is possible, but nontrivial — I think we should wait for a compelling use-case for rewrite before going any farther down that road.

- - -

As an aside, the following map throws errors due to sass miscasting `##v` values as numbers (with `v` as the unit), thus seeing these maps as containing duplicate keys. Since it seemed good to keep the keys as numbers for potential use with math, I didn't stringify them or manually cast them as strings with `'`s.
```sass
$violet-warm: (
  5:   #fef7ff,
  10:  #f5dcf7,
  20:  #e0bae3,
  30:  #d096d6,
  40:  #bd73c7,
  50:  #b24bbf,
  60:  #854180,
  70:  #573455,
  80:  #332431,
  90:  #211a21,
  5v:  #fdedff,
  10v: #f9c9ff,
  20v: #f3a8ff,
  30v: #ee82ff,
  40v: #e05ef7,
  50v: #c132d4,
  60v: false,
  70v: false,
  80v: false,
  90v: false,
);
```

### DIFF
--- a/src/stylesheets/core/_functions.scss
+++ b/src/stylesheets/core/_functions.scss
@@ -229,6 +229,26 @@ to a set target
 
 /*
 ----------------------------------------
+color()
+----------------------------------------
+Derive a color from its $family and
+value
+----------------------------------------
+ex:
+background-color: color($red-warm, 50v);
+----------------------------------------
+*/
+
+@function color($hue, $value){
+  @if unit($value) == '' {
+    @return map-get($hue, $value);
+  } @else {
+    @return map-deep-get($hue, vivid, strip-unit($value));
+  }
+}
+
+/*
+----------------------------------------
 @render-utility
 ----------------------------------------
 Build a utility from values calculated
--- a/src/stylesheets/core/_functions.scss
+++ b/src/stylesheets/core/_functions.scss
@@ -239,13 +239,14 @@ background-color: color($red-warm, 50v);
 ----------------------------------------
 */
 
-@function color($hue, $value){
-  @if unit($value) == '' {
+@function color($hue, $value, $variant...){
+  @if length($variant) == 0 {
     @return map-get($hue, $value);
   } @else {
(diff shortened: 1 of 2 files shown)


thisisdano:
Now — to squash some color bugs...

thisisdano:
Yeah, passed Circle!

maya:
Looking good to me. Is the way its ordered for any particular reason? It seems more logical to me to order them starting with reds and ending with purples (_a la_ ROYGBIV), that would make it easier for me to find the colors I'm looking for.

joshbruce:
https://caniuse.com/#feat=css-variables
https://stackoverflow.com/questions/46429937/ie11-does-a-polyfill-script-exist-for-css-variables

I'm starting to consider switching over to CSS variables to reduce dependency on Sass going forward. IE11, of course, is the problem child.

Issue #2401, opened 2018-03-20
Automatically determine site origin for proper 'official indicator' text


arichiv-usds:
## Description

On https://code.mil we noticed that the banner at the top displays "The .gov means it’s official", but given that we have a .mil website this seems to imply we're not official :-P

We're currently using some hacky JS to change this, but should really just add a determiner to the library https://github.com/deptofdefense/code.mil/issues/182

maya:
Hi, thanks for opening this issue. Agree that folks with a `.mil` domain should change this text accordingly. How are you implementing the banner on your site?

PR #2402, opened 2018-03-20
Add variable to indicate if site is on .mil TLD to [banner]

arichiv-usds:
## Description

See #2401, deptofdefense/code.mil#182

I went with this approach (context boolean) as only two TLDs are supported now (.mil and .gov). If a third option is ever added it might be worth considering defaulting to a list, and using JS to determine the site's TLD

## Additional information

screen shot 2018-03-20 at 7 36 36 am

### DIFF
--- a/src/components/banner/banner.config.yml
+++ b/src/components/banner/banner.config.yml
@@ -1,1 +1,5 @@
 label: U.S. Government banner
+
+context:
+  banner:
+    is_dot_mil: False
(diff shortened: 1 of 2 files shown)


maya:
Hi @arichiv-usds, thanks for this PR. I'll have to check with the team, if the fractal site is intended to be used for building the client-side components or not. I wasn't aware that they were. Are you using the fractal site to construct your components on code.mil? Would it make sense to add documentation to our site instead to change that text in the markup instead?

hursey013:
Hi @arichiv-usds, just to echo what @maya mentioned above - the configuration files in this repo are used primarily to demonstrate components in our [component library](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/develop/).

With that said, it looks like you are using https://github.com/18F/uswds-jekyll though, and this pull request make a lot of sense in that context - would you be interested in opening up a PR in that repo?

arichiv-usds:
I'll move to https://github.com/18F/uswds-jekyll, thanks!

Issue #2321, opened 2018-01-21
Return to top footer link not behaving as expected


arvalarva:
## Description

Using the keyboard, the link behaves correctly and takes the user visually to the top of the page, but a second tab navigation shoots the user back down to the next footer link.

I believe the intent is that keyboard users can return to the top and then tab through that navigation, without tabbing through every link in the long footer. As it is now, this is not the case. See amazon.com for proper behavior in its "return to top" footer link.

## Steps to reproduce the issue

1. Tab to the "Return to top" link at the beginning of the footer.
2. Tab again.
3. Notice that your focus is back down to the footer.

maya:
Thanks for bringing our attention to this @arvalarva. Yes, it seems like moving focus to the top of the page would be the expected and most accessible user experience.

maya:
Fixed in #2403.

PR #2403, opened 2018-03-20
Make return to top link set focus on main content

hursey013:
This enables the `Return to top` link to set the focus on the main content area so that tabbing does not skip back to the footer links.

It is backwards compatible so if the return to top link is set to `#` (which it is in all of our documentation), it will override that and instead set it to `#main-content` (if it exists). If folks have manually set the link to target a different element, it should still also work as it did before, but will also reset the focus to the specified target.

This addresses https://github.com/uswds/uswds/issues/2321

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/bh-return-top)

### DIFF
--- a/spec/unit/skipnav/skipnav.spec.js
+++ b/spec/unit/skipnav/skipnav.spec.js
@@ -8,12 +8,14 @@ const TEMPLATE = fs.readFileSync(__dirname + '/template.html').toString();
 describe('skip nav link', function () {
   const body = document.body;
 
-  let link;
+  let skipLink;
+  let topLink;
   let main;
 
   beforeEach(function () {
     body.innerHTML = TEMPLATE;
-    link = body.querySelector('.usa-skipnav');
+    skipLink = body.querySelector('.usa-skipnav');
+    topLink = body.querySelector('.usa-footer-return-to-top a');
     main = body.querySelector('main');
     skipnav.on(body);
   });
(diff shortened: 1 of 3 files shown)


hursey013:
I think we may also want to update our documentation and default HTML in the footer component to change the link from `#` to `#main-content` so that it is more transparent on what the intended functionality is - what do you think @maya?

maya:
Agree with these changes... a few things:

1. I'm testing the changes on the [docs page](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/bh-return-top/components/preview/layout--docs.html#main-content). Everything is good the first time. After I click it the first time, then scroll back to the bottom and click it again, nothing happens.

2. Can we remove the visible focus from the `#main-content`? It's not necessary for content and find it doesn't look clean.
screen shot 2018-03-20 at 9 16 35 am

hursey013:
@maya I took another pass, let me know if it's 👍

maya:
@hursey013 did you push it up yet? Not seeing any new commits.

hursey013:
@maya Oops, there now 😬

maya:
I tested this with VoiceOver. 

What I expected was how it works with the skipnav link which announces that I'm on the main:

screen shot 2018-03-20 at 12 39 06 pm

The way I experienced the back to top link was this (rattles off the sidenav items):

screen shot 2018-03-20 at 12 44 19 pm

Can we bring the two in line with each other to get the back to top link behavior consistent with the skipnav behavior?

Issue #2361, opened 2018-02-20
Duplicate CSS


maya:
via Slack:

Two max widths:

```
.usa-grid,
.usa-grid-full {
 max-width: 1200px;
 margin-left: auto;
 margin-right: auto;
 max-width: 1040px;
}
```

this is occurs twice:

```
html {
 box-sizing: border-box;
}

*,
*::before,
*::after {
 box-sizing: inherit;
}
```

stphnwlkr:
Asked the same question on Slack a year ago (March 12)... Shawn Allen replied "it's mostly an artifact of how we've structured our Sass—specifically, using `@include` vs. `@extend` :confused:
there are tools that you (or we) could use to post-process the CSS and optimize selectors and properties a _lot_ more 

the flip side is that debugging the CSS becomes a _lot_ harder when you optimize it like that."

maya:
@stphnwlkr I assume you're referring to the first issue about grids, not the second issue? 

I think it could be fixed by making better use of the neat `outer-container` mixin it includes:

```
@mixin outer-container($local-max-width: $max-width) {
  @include clearfix;
  max-width: $local-max-width;
  margin: {
    left: auto;
    right: auto;
  }
}
```

However, I don't want to get too deep into it in case we end up upgrading Bourbon + Neat to the latest version or use a different grid system.

maya:
Removing the box-sizing CSS from base.scss fixes the first issue, but I don't know where the box-sizing code is coming from that gets added as the first thing in the CSS file and would like to know.

stphnwlkr:
Yes. Sorry I wasn’t clear on that.

maya:
Closed via #2404.

Issue #2137, opened 2017-08-29
Rethinking the grid system research


msecret:
After having talked to some users about the WDS grid, I've found the following problems consistently mentioned:

### Documented offset classes

While there are offset classes for the grid system, they aren't documented on the main WDS site, leading to confusion as to whether they even exist.

**Why it's a problem**: Users expect offset classes to be there due to them being prevalent in other grid systems and frameworks.

### More then max columns requires additional `usa-grid` containers

**Why it's a problem**: 

- In dynamic code systems, content could be of variable amounts, so it's much easier to code for a grid system that will automatically adapt 5 vs 3 one-third columns.
- Even when it's not a dynamic code system, the users shouldn't have to think about when they need to add a grid container

### Can't nest grid columns

**Why it's a problem**:
- Certain layouts are much easier to do with nesting. A 1/4 column next to a 3/4 column, where the first element in the 3/4 column expands the full width, and it then splits up into 3 1/3 columns to fill the rest of the 3/4 column. Creating this layout with nested grids is as simple as:

```
.row-4          .row-4
content         large element
                .row-3  .row-3  .row-3
```

With the WDS grid, where columns can't be nested, it would mean:

```
.width-one-fourth     .width-three-fourths
content               large element
.usa-grid
.offset-one-forth       .width-one-fourth       .width-one-fourth       .width-one-fourth
```

### Can't target specific breakpoints

**Why it's a problem**: In general, elements is supposed to break and flow based on it's content, not predefined breakpoints. The Standards offers no way to do different things with the grid at different breakpoints, leading to inflexibility with the content.

### The naming convention is confusing

User's have stated:
- Unsure of when an "s" is needed on the end of width classes, `usa-two-halves`, `usa-one-half`.
- Hard to remember the name of things, such as `usa-grid-full`.

**Why it's a problem**: Difficulty in remember names of classes means it's a harder experience for developers to work with the grid, requiring looking up documentation to complete code.

msecret:
Criteria:
A. Documented offset classes
B. More then max columns requires additional usa-grid containers
C. Can't nest grid columns
D. The naming convention is confusing

[BASSCSS](http://basscss.com/#basscss-grid)
A: **fail**, there are not offset classes in the grid.
B: **pass**, can add as many columns and things will flow correctly to next lines
C: **pass**, can nest
D: **unsure**, the system is more widely used but the naming convention is somewhat atomic, which isn't necessarily in the WDS style.

msecret:
[...]

PR #2404, opened 2018-03-20
Remove duplicate css

hursey013:
This removes the duplicate css identified in https://github.com/uswds/uswds/issues/2361

The `box-sizing` at the top of the compiled  stylesheet was being brought in by Bourbon Neat, so in order to remove it the `$border-box-sizing` has to be set to false. I created a vendor variables partial in order to set that before Bourbon in added. The alternative would be to continue to rely on Bourbon Neat, and remove the duplicate code from `core/_base.scss` but it seems more flexible for us to be able to control those rules vs it being injected by an a third party library.

Bourbon Neat also sets a default `$max-width` that is used with the `outer-container` mixin, which is why we were having to specify a second `max-width` with `$site-max-width` to override it. The `outer-container` actually takes the max-width as an argument though, so I updated all the occurrences of it to pass in `$site-max-width` which makes adding the second max-width unnecessary.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/bh-duplicate-css)

### DIFF
--- a/src/stylesheets/components/_banner.scss
+++ b/src/stylesheets/components/_banner.scss
@@ -36,9 +36,8 @@
 }
 
 .usa-banner-inner {
-  @include outer-container();
+  @include outer-container($site-max-width);
   @include padding(null $site-margins-mobile);
-  max-width: $site-max-width;
 
   @include media($nav-width) {
     @include padding(null $site-margins);
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -154,11 +154,10 @@ $z-index-nav:     9000;
 
   .usa-navbar {
     @include media($nav-width) {
-      @include outer-container();
+      @include outer-container($site-max-width);
       @include padding(null $site-margins);
       display: block;
       height: auto;
-      max-width: $site-max-width;
     }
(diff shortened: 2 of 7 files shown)


maya:
Sounds good, thinking future friendly, if/when we upgrade to Neat 2.0 do the names of the variables have to change?

maya:
This creates a new top-level folder in stylesheets `vendor/` and I wonder if it's confusing since we already have a `lib`. Would we be able to add a `variables-vendor.scss` in `core` instead?

hursey013:
Unfortunately it looks like both `$max-width`, `$border-box-sizing`, and the `outer-container` mixin are removed in later versions so I don't think we can count on any of those. Is the plan to update Neat to 2.0 in the near future? Because if that is the case I think this issue is larger than duplicate code.

I agree a new folder is probably overkill to set one variable, I'll refactor it as you suggested.

maya:
Upgrading to Neat 2.0 is TBD, but I think we definitely want to improve grids in general (some thinking started in #2137), likely as part of the USWDS 2.0 release. It's good to keep on our radar its changes, in any case, thanks!

PR #2416, opened 2018-03-27
Fix typos/errors in README

maya:
Fixes a typo and provides a more accurate file structure example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,11 +69,12 @@ There are a few different ways to use the Design System within your project. Whi
   │   ├── uswds.min.css
   │   └── uswds.css
   ├── fonts/
-  ├── html/
-  └── js/
-      ├── uswds.min.js.map
-      ├── uswds.min.js
-      └── uswds.js
+  ├── img/
+  ├── js/
+  │   ├── uswds.min.js.map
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +54,6 @@ There are a few different ways to use the Design System within your project. Whi
 
 *Use the Design System `npm` package if:*
 - You are familiar with using `npm` and package management.
-- You would like to leverage USWDS [Sass](#sass) files.
```
hursey013: This is still accurate isn't it?

maya: In my description above: "Remove the line about using npm if you want to use Sass bc we provide Sass files in the zip." so if you want Sass you can use the zip file too.

hursey013: Yeah, I see what you're saying, I guess part of the confusion is that later on in the readme, where we describe using Sass it's specifically in reference to it being brought in by npm and the code residing in `node_modules`.

maya: Ah, I see... Yeah that is confusing. We could add the Sass info in the main Download section. What's confusing is that there's more general Sass info in the npm Sass section.

maya: Added a line about it here: 5b4e81d.

Issue #2173, opened 2017-10-11
Enhancements - grid viewport classes and visibility utility classes


ncksllvn:
I'm a developer on the team for vets.gov and if possible we'd love some insight into what the road ahead looks like for USWDS. Our project was initially started leveraging the Foundation CSS framework and although we've begun migrating to USWDS, there are certain features of Foundation that we can't do without and now we're deciding how to proceed. 

The main features of Foundation that we're invested in and that are not already implemented in USWDS are the visibility utility classes and control over the grid for certain viewports (i.e. `usa-medium-width-one-half` or something like that.) Are there any plans for these to be included in the project, and if so are you open to working together to implement them? If not, we'll just backfill them in our own project as we remove Foundation, but we thought it could be a useful collaboration if these are planned features for USWDS and it would really help us out.

Note - I sent an email last week about this but didn't receive a response.

maya:
Hey Nick, thanks for opening this issue. Just realized you didn't get my email so re-posting here as well.

That's exciting to hear about how you're using the Standards. We've heard of folks wanting more viewport control in the past like you mentioned so we'll likely take a look at it in the future. I don't know when we'll be working on rethinking grids, but can check with the core team and see how it fits in our roadmap and if/when would be good time to collaborate. In the meantime, feel free to go ahead with your modification and let us know how it progresses.

ncksllvn:
Thanks for your response Maya! Okay, we'll figure something out for our project and check back periodically. Great job with the project so far by the way!

emes001:
Fellow dev over at @USOPM here. I second more granular control of the grid system, based on a specific viewport. In lieu of this missing feature, I'm relying more on custom classes at the moment (more-so because this is for a prototype). Do you have a timeline for 2.0 and/or is this currently being worked on?

stphnwlkr:
A third. I have started looking at alternatives as well because the grid is too limiting. The CMS team has done a nice job as well as the more popular frameworks.

maya:
@emes001 the timeline for 2.0 is likely in the next few months, in which updates to the grid would be included. The main focus right now is on our utility classes feature and will likely be taking a closer look at grids soon after that feature is complete.

annekainicUSDS:
Our team has done some thinking about the specific requirements for any enhancements to the grid that we would benefit from, and how those might be implemented. We thought we would share them here to get your thoughts and feedback!

#### Desired Enhancements:
- Mobile-First Grid System
- Responsive Utilities

[...]

PR #2422, opened 2018-04-11
[WIP] Add new layout grid

thisisdano:
Demo: https://federalist-proxy.app.cloud.gov/preview/gsa/10x/dw-design-refinements/layout-grid/

#### TODO
- [ ] more responsive testing
- [ ] abstract and `@mixin`ize some of the common code patterns
- [ ] componentize some of the demo page code from utilities
- [ ] get feedback
- [ ] test implementation on existing components

Fixes: #2173, #2447


maya:
In the preview, it uses the `tablet` to denote size but tablet is so subjective and has a huge range. I think I prefer to use small, medium, large. This also keeps it in-line with how it is currently.

annekainicUSDS:
Thank you @thisisdano and @maya for all of the work on this! Our team is really excited about using this new functionality!

We reviewed the new implementation as a team and compiled a few questions for you:

#### Implementation questions

1. How do we accommodate different layouts at different breakpoints? Do we just add additional classes? E.g., `tablet:g-col-6 desktop:g-col-12`
2. Are visibility classes used by just adding `:display-none` to other grid classes?
3. There seems to be a few different places where `!important` is used. Is there any concern for downstream effects this might cause?
4. Are you planning to maintain existing `usa-` grid classes in the new version? Keeping these classes, while perhaps deprecated, would enable us to upgrade to the new system incrementally instead of having to convert everything in one go.
5. Are there plans for a screenreader-only class and what might that look like? What are the general plans for screenreader support? Currently the grid system includes support for screenreaders. Specific grid-related concern: changing the source order in the layout.

#### Thought-process questions

1. We're curious about the choice to name the breakpoints after devices (e.g., `tablet`) instead of sizes. We'd love to know the reasoning for this.
2. Also curious on why you decided to go with flexbox for the grid system, and have you done research into performance-related issues with this.
3. Why the use of the colon (`tablet:g-col-6`) in the grid classes?

thisisdano:
Hi @annekainicUSDS — thanks so much for looking this through and asking thoughtful questions.

> How do we accommodate different layouts at different breakpoints? Do we just add additional classes? E.g., `tablet:g-col-6 desktop:g-col-12`

Yes, the grid classes and utility classes both use a mobile-first methodology where responsive prefixing applies to the breakpoint and higher. And the classes are additive, just as you suggest — adding ` desktop:g-col-12` would give the element a 12-column width at `desktop` size and higher.

> Are visibility classes used by just adding `:display-none` to other grid classes?

Yes, visibility classes and other utility classes are available and would do just what you say. Adding `u-display-none` to a tag would apply `display: none`, just as `tablet:u-display-block` would add `display:block` to a media query of `min-width: [tablet]`.

> There seems to be a few different places where `!important` is used. Is there any concern for downstream effects this might cause?

The utility classes can be output with `!important` in project settings or you can output them without `!important`. In general, this would assure, by design, that they will always override any component CSS in a mixed utility/component environment. This is a feature of some utility class libraries (Buzzfeed Solid, and the override utilities in the new GDS design system are the first that come to mind) and we think the downstream effects will be understandable and intentional — but perhaps further testing is necessary to bear this out.

> Are you planning to maintain existing `usa-` grid classes in the new version? Keeping these classes, while perhaps deprecated, would enable us to upgrade to the new system incrementally instead of having to convert everything in one go.

Probably? I tend to think it's a good idea. We haven't hashed this out yet, but I'm inclined to make as few hard breaking changes as we can reasonably accommodate. Incremental updating and implementation is one of our roadmap goals, so this would fit squarely within that goal.

> Are there plans for a screenreader-only class and what might that look like? What are the general plans for screenreader support? Currently the grid system includes support for screenreaders. Specific grid-related concern: changing the source order in the layout.

Yes, to screenreader-only class(es) but we haven't built them yet. Something like `u-sr-only` would be useful. Honestly, I don't know what screenreader support for grids means — happy to learn — but the utility class system includes `u-order-[n]` utilities for setting, say `tablet:u-order-last`.

> We're curious about the choice to name the breakpoints after devices (e.g., tablet) instead of sizes. We'd love to know the reasoning for this.

There are pros and cons to sizing methodologies and I tend to fall on the side of being more explicit rather than more abstract. I know, as I develop, that I am often looking at reference material when I'm trying to remember exactly what abstract breakpoints like `sm` or `md` actually refer to. It is very true that `tablet` is shorthand for myriad sizes — no need to post any of those Luke Wroblewski photos! — but it's at least a more discrete shorthand that immediately signifies a range.

While version 2.0 makes it possible to add custom sizing to breakpoint values, I do not encourage it. We're working to build a system that's easy to build from and contribute back into — a more immutable range of breakpoints helps component submissions by making responsive behavior more consistent and reliable. 

In general, we've made design decisions that favor explicitness over concision where possible, especially in whitespace sizing and, by extension, breakpoints. 

> Also curious on why you decided to go with flexbox for the grid system, and have you done research into performance-related issues with this.

We've not done research into performance issues with flexbox — if there are performance issues, please let us know. We're following the lead of something like Bootstrap 4 (and others) that value the benefits of flexbox (flexible layout, ordering, and alignment are some of the big, showstopper benefits) over compatibility issues. Practically, we have to see flexbox layout as progressive enhancement — understanding that the oldest browsers will not see it and building layouts and components that are functional regardless.

> Why the use of the colon (tablet:g-col-6) in the grid classes?

We use the colon for state-based prefixes to help distinguish states from the utility itself. This applies not only to responsive prefixes, but to `hover:`, `active:`, `visited:`, and `focus:` states. 

This serves as disambiguation as well. `tablet`, for instance, is a width as well as a breakpoint. A utility like `u-max-width-tablet` is a good example and something like `desktop:u-max-width-desktop` is a potentially useful utility for setting application width. Colon-prefixing helps reinforce the usage of the particular prefix and help build, over time, a faster understanding of a utility, its value, and its state.

This is useful not only for clarity in markup, but also connects to a the pathway by which you might implement utility markup into component CSS.

If you were including the grid styling into CSS, you could do so by using `@include g-col-6` in component markup. Pseudostates and media queries are handled differently. Treating them differently from the beginning helps develop a consistent pattern and a connection with component integration.

```sass
.my-component {
  @include g-col(12);
  &:hover {
    @include u-background-color(black);
  }
  @include at-media(tablet) {
    @include g-col(4);
  }
}
```

All this said — there are many issues at play here and with the work we're doing in support of version 2.0. Your feedback is very very helpful and it pushes us to make the best, most useful changes in support of real-life needs.

Let's keep talking.

maya:
In the existing grid, all columns are built as 100% width at the smallest size and then trigger the declared column width at a larger size. E.g. `usa-width-one-third` gives you a 33% column which becomes a 100% column at the smallest size.

In the new grid, you have to write the width at the mobile size each time, e.g. `g-col-12 tablet:g-col-4`

Would the new grid support the same kind of sensible default functionality of a 100% mobile column that's in the current grid?

thisisdano:
I would argue, with somewhat of a hedge, no — that it is clearer and more predictable to require explicitness and less opinion, even if it means more typing and redundancy. I think the defaults of our current grid are an opinionated and unpredictable net negative.

That said, I think there's something of a compromise. Set `g-col` equivalent to `g-col-12`. 

Currently `g-col` gives you a flexible column of `flex: 1`, but this could easily be rewritten to be `g-col-fill` and that's consistent with flex utilities (like `u-flex-fill`).

Now, if you had a column, you could set its classes as `g-col tablet:g-col-4` and it would have the functionality you outline _and_ address my own predictability concerns. I kinda love it.

What do you say?

maya:
That could work, but let's keep it as is for now and see how it does in testing. There's something nice about the simplicity of `g-col` filling. Maybe I'll get used to it. Or maybe `g-col-full` could be a mobile full class, even though it's two extra characters. Or maybe there's a setting that you could set for a mobile 100% columns.

thisisdano:
yoops — by the time I saw this I'd already made the change since I thought it was such a good idea! :/

I still kinda think it is — though I do agree about `g-col` — the consistency with `flex-fill` though was deeply appealing to me...

thisisdano:
Relatedly/unrelatedly — I'd never actually really used grid utilities until building this and prototyping with them (!), and now I can unreservedly say I love them.

Live and learn! :)

PR #2425, opened 2018-04-12
Remove submit button block display and extra styling

### LINKED ISSUE #2409: Issue with floating and alignment of buttons introduced in v1.5.0.
## Description

Version 1.5.0 introduced a "breaking" change by adding `display: block` to all buttons.  This resulted in misalignment of buttons when floated or aligned inside of containers.  I know it's too little too late now, but for a project that has been using USWDS for a while and would like to continue to use the most up to date versions, it is difficult to have to include overrides and track down issues when an update comes out.

maya:
This removes the `display: block` from submit buttons. While doing this I also realized we likely don't need any of the added styles so removed them.

**A few things I found in this work:**
1. The sketch file has two different font size for "show password" (17px) and "show my typing" (15px) toggles. Since they function in the same way should they be the same sizes? The code has them at 15px.
2. This code removes the extra padding on the submit button, leading to a less wide Sign in button. Is it too short this way? We can add a min-width of 134px on the submit buttons bc the extra padding is not desired on buttons that are already wide from longer text.

Fixes #2409.

### DIFF
--- a/src/stylesheets/components/_forms.scss
+++ b/src/stylesheets/components/_forms.scss
@@ -2,7 +2,6 @@ $usa-form-width: 32rem;
 
 [type=submit],
 [type=submit] {
-  display: block;
   margin-bottom: 1.5em;
   margin-top: 2.5rem;
 
--- a/src/stylesheets/components/_forms.scss
+++ b/src/stylesheets/components/_forms.scss
@@ -1,20 +1,8 @@
 $usa-form-width: 32rem;
 
-[type=submit],
-[type=submit] {
-  margin-bottom: 1.5em;
-  margin-top: 2.5rem;
-
-  @include media($medium-screen) {
-    padding-left: 2.7em;
-    padding-right: 2.7em;
-    width: auto;
-  }
-}
-
 [name=password],
 [name=confirmPassword] {
-  margin-bottom: 1.1rem;
+  margin-bottom: 1rem;
 }
 
 fieldset {


thisisdano:
1. Yes, I'd say they helper toggles should be the same type size.
2. I think the button looks fine with less padding.

In general, I do think it looks a little funny being as close as it is to the field and there's a bit of dissonance with the helper text toggle, but I consider these things minor, and better to fix the breaking change than mess about with spacing and introduce a new one.

It sucks to go back and forth though, since anyone building with the `block` button might find their layouts affected with this fix. Again, I don't think this is reason not to make this fix as implemented.

maya:
>Yes, I'd say they helper toggles should be the same type size.

which did you prefer 15px or 17px?

We should probably note to fix these changes in the Sketch file.

maya:
The Sketch file has some inconsistencies around the button spacing. The Sign in button has the button closer to the input than the password reset, which has it after the helper toggle (also "helper toggle" might be a better class name than `usa-form-note`). I'm guessing you'd prefer the Reset password distance @thisisdano?

screen shot 2018-04-12 at 2 15 55 pm
screen shot 2018-04-12 at 2 16 04 pm

maya:
@thisisdano updated the styling to reduce the dissonance:
screen shot 2018-04-12 at 2 27 45 pm

maya:
Updated so the margin bettween the submit button and input is 30px in order to maintain vertical rhythm between inputs.
screen shot 2018-04-12 at 2 42 07 pm

src/stylesheets/components/_forms.scss
```diff
@@ -1,23 +1,11 @@
...
 [type=submit] {
-  display: block;
-  margin-bottom: 1.5em;
-  margin-top: 2.5rem;
-
   @include media($medium-screen) {
-    padding-left: 2.7em;
-    padding-right: 2.7em;
-    width: auto;
+    margin-top: calc(3rem - .2em); // Label margin minus input bottom margin
```
maya: note for future: these should probably be variables.

Issue #2375, opened 2018-03-02
Add documentation for running USWDS Sass w/o PostCSS


maya:
The USWDS runs PostCSS as part of its build process to add vendor prefixes. However, some people may be building the Sass directly from the SCSS we provide, without PostCSS. This means that their CSS does not have vendor prefixes added. 

What should the workflow look like for these developers?

## Ideas
- Offer a prefixed version of Sass
- Add documentation for adding it themselves

Discovered in #2373.

thisisdano:
We use PostCCSS? I _do_ see we use autoprefixer, and I feel that's a fine best practice. I'm pretty sure we don't want to prefix manually, but we could give some guidance about what post-compilation steps our sass requires if someone has their own pipeline.

maya:
Yes, autoprefixer is a PostCSS plugin.

maya:
Offering SCSS already prefixed could be a nice option for folks who don't use autoprefixer.

PR #2427, opened 2018-04-13
Add note about browser prefixes + Sass

hursey013:
Autoprefixing is not included when using the Sass files directly, so adding a note to make folks aware of the fact that they will need to add that functionality into their own projects.

Fixes https://github.com/uswds/uswds/issues/2375


maya:
For the future, I wonder if we should have this all in a single place so we don't have to update the content twice like we're doing now: in the docs site and README.

Issue #2429, opened 2018-04-16
Jekyll build error using release-2.0 branch from npm


hursey013:
When attempting to import `uswds` into a sass stylesheet in Jekyll:

```
Conversion error: Jekyll::Converters::Scss encountered an error while converting 'assets/css/main.scss':
                    $string: false is not a string for `quote' on line 695
             Error: $string: false is not a string for `quote' on line 695
             Error: Run jekyll build --trace for more information.
```

thisisdano:
Have a _feeling_ I know the kind of thing this is referring to. It's not an error I've gotten before, though. Is there an open branch I could look at? (Probably) somewhere a `quote()` function is getting a `false` value — I'll need to find out where it's coming from...

Does `jekyll build --trace` actually give you any more info?

hursey013:
@thisisdano think I tracked it down here https://github.com/uswds/uswds/pull/2430

PR #2430, opened 2018-04-16
Update spacing function

hursey013:
Looks like you can't run `quote()` on a string that is a boolean like `false`.  For now, moving the check to see if the value is false to the beginning of the conditional statement so it does not accidentally get passed through.

Fixes https://github.com/uswds/uswds/issues/2429


thisisdano:
Cool — there are only 8-9 places where `quote()` is used in the functions — I wonder if there are any other instances where a false value might creep in?

PR #2483, opened 2018-05-09
Update Code of Conduct in CONTRIBUTING

maya:
Updates CONTRIBUTING doc Code of Conduct link and language to more closely match the language in the new CoC. Also, changes a reference from 18F to TTS.

Maybe review this with the diversity guild and the writing lab (https://github.com/18F/writing-lab/issues/345).


maya:
Got the 👍 from @thebestsophist in https://github.com/18F/writing-lab/issues/345: 
>Looked good to me, the folks from the diversity guild who saw it also said 👍🏼 So I think it’s good to go!

PR #2522, opened 2018-05-26
Add new banner

thisisdano:
This attempts to rewrite the banner component with the new system vars, utilities, and practices — and also tries to address some of its existing problems:

**original:** Jumping flag and variable font size

banner-original

**new:** An action-oriented message

banner-new

**original:** Absolutely positioned text will break at small widths or magnification

screen shot 2018-05-29 at 7 05 59 am

**new:** Text will flow more naturally

screen shot 2018-05-29 at 7 06 36 am

**original text:** `An official website of the United States government`
**new text:** `An official United States government website`

Cutting those two words allows us to use a slightly larger font size at 320px width without text breaking over the line.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-new-banner)


maya:
This has a lot more changes that I realized. Going to put placeholders for what would need review:

- Content
- UX design
- Accessibility
- Front-end

Also, I think I'd like to push back on changes that don't have a demonstrable need from our users -- i.e. if they didn't tell us something was difficult about the current UX and content of the current banner, we shouldn't change it.

thisisdano:
Well I am a user and I am saying that the UX is not so hot.

thisisdano:
I _do_ think that burying the lede in a six-word sentence is pithy, but a _little_ more pithy than realistic.

andrewmaier:
Do we need to say "official" US government website? Does the US government have _unofficial_ websites?

stphnwlkr:
Is it even required? How many implementers of the USWDS are actually using it and what value does it serve the citizenry if they even notice it is there? I suspect most people know what https and .gov indicate.

thisisdano:
So, the banner is not required, but probably helpful. I think many folks probably get the `.gov` part, but probably not the `https` part. All of it is something of a kludge for something more reliable — like something built into the browser that communicates the same message [like where it currently has a little lock and some text like 'secure'].

That said, I think you could get away with `A United States government website` and it would be an improvement. 

This ain't `The official snack chip of Major League Baseball` or somesuch. `Official` has a whiff of security theater razzle-dazzle and hand-waveyness. The copy equivalent of one of those graphic design rubber-stamp things that says "VALID" at a 20° angle.

When you look at the site, you ask yourself "Is this a government website?" not "Is this an official website of the United States government?" — which sounds more like it belongs on a commemorative coin.

That said, I am a radical, and I don't really think the flag is the right thing to put up there either.

thisisdano:
:fire:

thisisdano:
Or `This is a United States Government website` since that leading `A United` has an awk feel.

el-mapache:
@thisisdano Should we try and knock this out, or close for now?

thisisdano:
Think we're going to close for now, since the solution I'm proposing changes the markup and we're trying to avoid big, unnecessary markup changes in the 2.0 release.

maya:
Also, if I remember correctly, the content guild preferred the old text to the new.

thisisdano:
That is true, but I don't think that's the end of the story on that text.

PR #2524, opened 2018-05-30
Use px for utility media queries until we can determine what's wrong with em

thisisdano:
This change should render the utility media queries with px values instead of em. Something about our 10px base isn't playing well with the em-based media query (I don't know why — the em keeps referencing 16px and I don't know where that value is coming from).

### DIFF
--- a/src/stylesheets/core/_functions.scss
+++ b/src/stylesheets/core/_functions.scss
@@ -625,7 +625,7 @@ is false. Warn if it is.
       @return unquote($value);
     }
     @else {
-      @return $value;      
+      @return $value;
     }
   }
   @else {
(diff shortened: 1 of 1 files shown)


thisisdano:
Hm! Per this article the em-based mq is based on the user's browser settings and has nothing to do with the markup!?!
http://blog.w3conversions.com/2015/03/all-your-base-are-belong-to-us-%E2%80%94-or-what-base-size-do-em-based-media-queries-use/

So, let's not do that.

thisisdano:
`px4eva`

thisisdano:
Conversely, we _could_ just set our own base font size to 16px/100% (which we _should_ do regardless) and use an em base for the mqs. But it still feels odd to base your mqs off a value that may have no relationship to the layout

PR #2527, opened 2018-05-31
Escape parentheses in showcase link

hursey013:
This link is appearing as a table on docs site: https://designsystem.digital.gov/getting-started/showcase/all/


maya:
@hursey013 we use parentheses on a lot of other links on the page and it works fine. Do you know why it's only happening here?

maya:
Oh I think it's out of date on the site, fixed it here: https://github.com/uswds/uswds/commit/4c9d54539bbacef8df1f5211744ffd37ea567974#diff-29983973f8a9f1936ee33ab86ebb6b7a

hursey013:
Ah, that makes sense. Closing this PR.

PR #2530, opened 2018-05-31
Opacity and text-indent tweaks

thisisdano:
- Remove very small text-indent (1px and 2px)
- Use numeric scale for opacity


hursey013:
My only other thought would be to represent the opacity as a percentage, as in `opacity-30` vs `opacity-03`. For some reason `opacity-100` is more clear to me the element is 100% opacity vs `opacity-1`. Purely personal preference though.

thisisdano:
@hursey013 That makes sense to me — I was initially inclined to hew to the way you'd write the css (from 0-1) but clearly we're not heading down the path of 100% parity with css. I agree that `opacity-100` feels way more sensical and immediately understandable. I'll make the change.

PR #2532, opened 2018-06-01
Add quotes to token

hursey013:
To fix build error:
```
Message:
    src/stylesheets/core/_functions.scss
Error: Duplicate key 1 in map ((1px: 0.8rem, 2px: 2px, "05": 0.4rem, 1: 0.8rem, "105": 1.2rem, 2: 1.6rem, "205": 2rem, 3: 2.4rem, "neg-1px": -1px, "neg-2px": -2px)).
        on line 87 of src/stylesheets/core/_functions.scss
>>     $collection: map-merge($collection, $map);
   -----------------^
```

### DIFF
--- a/src/stylesheets/core/_functions.scss
+++ b/src/stylesheets/core/_functions.scss
@@ -96,7 +96,7 @@ source: https://gist.github.com/bigglesrocks/d75091700f8f2be5abfe
 smart-quote()
 ----------------------------------------
 Quotes strings
-Inspects numbers
+Inspects `px`, `xs`, and `xl` numbers
 Leaves bools as is
 ----------------------------------------
 */
@@ -106,7 +106,10 @@ Leaves bools as is
     @return quote($value);
   }
   @if type-of($value) == 'number' {
-    @if unit($value) == 'xl' or unit($value) == 'xs' {
+    @if unit($value) == 'px' {
+      @return inspect($value);
+    }
+    @elseif unit($value) == 'xl' or unit($value) == 'xs' {
       @return inspect($value);
     }
     @else {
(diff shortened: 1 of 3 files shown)


thisisdano:
@hursey I made a couple additional changes to stringify `1px` and `2px`values across the board in maps — and to adjust a quoting/stringify function to stringify and quote `1px` and `2px` values.

Has some hacky vibes, but is pretty straightforward, I think.

src/stylesheets/core/_functions.scss
```diff
@@ -106,7 +106,10 @@ Leaves bools as is
     @return quote($value);
   }
   @if type-of($value) == 'number' {
-    @if unit($value) == 'xl' or unit($value) == 'xs' {
+    @if unit($value) == 'px' {
```
hursey013: Looks like the `if` and `elseif` can be combined into one statement here since they are returning the same value. 
```
@if unit($value) == 'px' or unit($value) == 'xl' or unit($value) == 'xs' {
```
Might be able to you `index` also (not tested):
```
@if index(('px', 'xl', 'xs'), unit($value)) {
```

thisisdano:
@hursey013 oops, wrong username above

src/stylesheets/core/_functions.scss
hursey013: Are there any drawbacks to quoting all `px` values?

thisisdano: I _do_ wonder if there's a more succinct way to write this! Curious about that index technique

thisisdano: Re drawbacks: _probably_ not, particularly in the long term. It could be a short-term issue and throw an error if some older USWDS custom value were being passed to the spacing function. But this seems like a pretty small slice, and one that we'd probably want to iron out by 2.0 anyway

thisisdano: It _might_ be an issue for some other custom implementations, too — but that also feels like a stretch

thisisdano: I'm gonna try that index technique (it worked!) and re-commit.

PR #2535, opened 2018-06-04
Add a `u-font()` mixin that combines both font-size and font-family

thisisdano:
We can now set just font size with `size()`, just family with `family()`, or both with `font()`

### DIFF
--- a/src/stylesheets/core/mixins/utilities/_font.scss
+++ b/src/stylesheets/core/mixins/utilities/_font.scss
@@ -6,7 +6,7 @@ Get a font-family stack
 ----------------------------------------
 */
 
-@mixin u-family($family) {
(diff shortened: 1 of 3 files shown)


thisisdano:
This now makes me want to change the `size` utility to `font` — since this is what that utility does — set both font-family and size.

But in component CSS, we probably don't want to _always_ be setting both — that'd be a lot of redundant font-family — which is why there are these three mixins available.

thisisdano:
Better, more consistent:
- Utility for both family and size is `font` (i.e. `font-sans-sm`)
- Utility for family is `font-family`(i.e. `font-sans` — rare)
- Mixin for both family and size is `font()`
- Mixin for family is `font-family()`
- Mixin for font-size is `font-size()`

PR #2544, opened 2018-06-11
Change banner parent container div to section element

maya:
Currently, there's a `header` element within the banner parent container `div`, and this needs to be inside a sectioning element to be conformant. This ensures the markup is conformant to HTML standards. 

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-gov-banner/components/detail/banner.html)

Feedback:
>But `header` tags, per the 5.1 spec, “can only contain a header or footer if they are themselves contained within sectioning content.” (https://www.w3.org/TR/html51/sections.html#example-6f9fce34) As these `header` tags are not contained within `article`, `aside`, `nav`, or `section`, this use seems to be non-conforming (and is causing my tests to yell at me).

### DIFF
--- a/src/components/banner/banner.njk
+++ b/src/components/banner/banner.njk
@@ -1,4 +1,4 @@
-<div class="usa-banner">
+<section class="usa-banner">
   <div class="usa-accordion">
     <header class="usa-banner-header">
       <div class="usa-grid usa-banner-inner">
@@ -33,4 +33,4 @@
       </div>
     </div>
   </div>
-</div>
+</section>


waldoj:
I appreciate this feedback should have come 21 hours ago, but I don't think that this fixes the problem. Per the HTML 5.1 spec:

> For cases where an developer wants to nest a `header` or `footer` within another `header`: The `header` element can only contain a `header` or `footer` if they are themselves contained within sectioning content.

So I think _both_ `header` elements must be within e.g. a `section` container. Not only is that my read of the spec itself, but the W3's new validator rejects a `header` > `section` > `header` hierarchy. (As tested by html5validator which is powered by the [Nu Html Checker](https://github.com/validator/validator) which, in turn, powers [the new W3 validator](https://validator.w3.org/nu/).

This PR created a `header` > `section` > `header` hierarchy, when it appears that what is required is  `section` > `header` > `header`.

maya:
I just checked the source code and it looks like the government banner is not part of the main document header: https://components.designsystem.digital.gov/components/detail/layout--docs.html

The page template Waldo is using is out of date. Once we can get fractal pulling in these templates, it should have the accurate HTML on our docs site. I still think it's semantically alright for the gov banner to be a section element though so should be fine to leave it as is.

thisisdano:
Oof. I feel like there's some real ambiguity in that guidance spec. Can't say it makes much sense from a semantic perspective to favor `section` > `header` > `header` over `header` > `section` > `header` — I don't see from the spec how conforming markup could have an unsectioned `header` within another parent `header`. 😕 

And I can't seem to make the [validator](https://validator.w3.org/nu/) throw the expected errors with either the first solution or the existing code. 🤷‍♀️

maya:
The current code is 
`section > header`
`header` (new block element)

Not section > header > header, it's working on the existing code.

PR #2549, opened 2018-06-11
[UI component]: Adding best-practice ruleset and fix for heading-level error

1Copenut:
## Description

Adding best-practice ruleset and  exceptions to axe test script. Conversation in the USWDS public Slack channel asked about best-practice rules showing up in axe web browser plugin, but not on automated test runs. I added a `runOnly` object to the axe-tester.js configuration object that adds best-practice rules as defined by Deque.

I changed the first `<h4>Grand Canyon</h4>` to `<h5>Grand Canyon</h5>` to fix 2 best-practices errors calling out heading order. This felt like a better approach than trying to add an exception to the axe-tester.js, as I felt letting users know of a heading order issue was a good thing. 

Deque `axe-core` list of rules, including best practices:
https://github.com/dequelabs/axe-core/blob/develop/doc/rule-descriptions.md

### DIFF
--- a/spec/axe-tester.js
+++ b/spec/axe-tester.js
@@ -14,12 +14,22 @@ const AXE_CONTEXT = JSON.stringify({
 });
 
 const AXE_OPTIONS = JSON.stringify({
+  runOnly: {
+    type: 'tag',
+    values: [ 'section508', 'wcag2a', 'wcag2aa', 'best-practice' ],
+  },
   rules: {
     // Not all our examples need "skip to main content" links, so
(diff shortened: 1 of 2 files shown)


maya:
Thanks so much, @1Copenut! We'll need to review the  `best-practice` ruleset to ensure it's what we want to run. By law, we need to ensure Section 508 and WCAG 2.0 AA are met, but of course, we can go above our requirements too. Looking at the list, `cat.semantics` might be something we want to add in the future as well. Stay tuned!

PR #2551, opened 2018-06-11
Code sample improvements and prepare components to export to uswds-site

maya:
This removes extraneous lines and indentations from code samples.

This also prepares various components so the fractal component can be used on the uswds-site repo:
- Adds a validation component
- Updates sidenav content
- Creates an inner content fractal component for landing and documentation page templates (necessary so we can properly link to JS and CSS on the fractal and docs site
- Preps grid for inclusion

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/remove-extra-code-lines)


maya:
The axe failure is appearing on the components layout--landing-inner and layout--docs-inner bc it thinks it's a full page (I think). We need a way to exclude axe from running on them, it already runs on the full page which includes the elements it's looking for: layout--docs and layout--landing. 

cc @toolness for any ideas.

toolness:
Hmm, unfortunately I don't have much extra time to dive into this, but here are some options:

* Add the components to the [`SKIP_COMPONENTS`](https://github.com/uswds/uswds/blob/a6cc5fe17de4dd22d2f0dfc78f0c2fbdf706570a/spec/headless-chrome.spec.js#L29-L32) array in `spec/headless-chrome.spec.js`.

* Modify [`getHandles()`](https://github.com/uswds/uswds/blob/a6cc5fe17de4dd22d2f0dfc78f0c2fbdf706570a/spec/chrome-fractal-tester.js#L57-L59) in `spec/chrome-fractal-tester.js` to filter out any Fractal components that have `hidden: true` in their configs. I don't actually know if this is possible, but it seems like it _ought_ to be?

maya:
@toolness thanks so much for pointing us in the right direction! 

I couldn't see any examples of components that were excluded so was wondering if you could show what that would look like in `SKIP_COMPONENTS`?

The second idea you posted sounds like it could be the most sustainable so may give that a go too. What would the JS look like to exclude components with `hidden: true`?

toolness:
Oh try looking at the history of that file; `SKIP_COMPONENTS` used to be populated but it's not anymore.

Unfortunately I don't know what the JS would look like to exclude `hidden: true` because I don't know anything about how fractal's data structures work :(

maya:
Thanks! I found this version of the file with a lot of options enabled and components excluded: https://github.com/uswds/uswds/blob/73c586b26ba3643b04936361dc14a761d4281644/spec/axe.spec.js

maya:
Andddd the test is fixed in 1c4c1ec 🎉

thisisdano:
The `border-(left|top|bottom|right)` examples are all currently showing `border-top`.

I have been arranging colors from lightest to darkest, including the base or core value (like `border-secondary` or `border-base`) in the center instead of at the start of the section. I find that gives a good sense of continuity without adding additional weight to the base value — which I don't necessarily think is more important than its variants.

The grayscale palette only includes
- white
- gray-5
- gray-10
- gray-30
- gray-50
- gray-70
- gray-90
- black

PR #2555, opened 2018-06-15
Fix link path to npm script example

maya:
We don't have a `develop` branch anymore on the uswds-site, this changes it to the `master` branch.

Though I don't know what line 4 in https://github.com/uswds/uswds-site/blob/master/package.json#L4 is intended to link to. That file has likely changed.


maya:
I've linked it to the scripts in line 28, however this line will change as soon as anything is added above it like contributors.

PR #2567, opened 2018-06-27
Improvements to grid and gap

thisisdano:
- Add `$theme-grid-container-max-width` variable to set default container max width
- Fix `grid-gap()` to allow `.5`
-  Scope gap padding to `grid-col*` classed items
- Allow `grid-offset(none)` and `.grid-offset-none`


maya:
The title of the PR is `[WIP]`. Did you mean to remove that instead and leave the title?

PR #2569, opened 2018-07-03
Allow variables and functions to be passed to style functions

thisisdano:
These changes allow core styling functions to allow and properly parse variables, such as the following intentionally somewhat convoluted code:

```scss
$face: sans;
$size: 6;
$line-height: 2;

$color-family: 'red-warm';
$color-grade: 50;

$this-color: $color-family, $color-grade;

$my-scale: $face, $size + 1;
$my-lh: $face, $line-height + 1;
$my-color: append($this-color, vivid);


.foo {
  @include u-text(color($my-color), bold);
  font-size: scale($my-scale);
  line-height: lh($my-lh);
  background-color: color(append($this-color, vivid));
}
```

This will allow more abstracted code to be written for components, possibly helping templatization down the road.

- - -

Also renames a couple core styling functions for more consistency with our naming rubric:

- rename `font()` --> `scale()`
- rename `spacing()` --> `units()`

### DIFF
--- a/src/stylesheets/core/_functions.scss
+++ b/src/stylesheets/core/_functions.scss
@@ -828,7 +828,7 @@ and scale
     @error "Font size requires a valid FAMILY and SIZE in the form `font('FAMILY', 'SIZE')`";
   }
   $our-family: smart-quote(nth($props, 1));
-  $our-scale: smart-quote(smart-quote(nth($props, 2)));
+  $our-scale: smart-quote(nth($props, 2));
   @if not map-has-key($project-cap-heights, $our-family) {
     @error "#{$our-family} is not a valid font family. Valid values: #{map-keys($project-cap-heights)}";
   }
--- a/src/stylesheets/core/_functions.scss
+++ b/src/stylesheets/core/_functions.scss
@@ -467,13 +467,18 @@ a family and a line-height scale unit
 ----------------------------------------
 */
 
-@function lh($family, $scale){
-  $scale: smart-quote($scale);
+@function lh($props...){
+  $props: unpack($props);
+  @if length($props) != 2 {
+    @error "lh() needs both a valid face and line height unit in the format `lh(FACE, HEIGHT)`.";
+  }
+  $family: smart-quote(nth($props, 1));
+  $scale: smart-quote(nth($props, 2));
   @if not map-has-key($project-cap-heights, $family) {
     @error "#{$family} is not a valid font family. Valid values: #{map-keys($project-cap-heights)}";
   }
   @elseif not map-get($uswds-line-height, $scale) {
-    @error "`#{$scale}` is not a valid line-height scale. Valid values: #{map-keys($all-type-scale)}";
+    @error "`#{$scale}` is not a valid line-height unit. Valid values: #{map-keys($uswds-line-height)}";
   }
   @elseif map-get($project-cap-heights, $family) == false {
     @return false;
@@ -731,6 +736,7 @@ colors: project and system
 */
 
 @function color($props...){
+  $props: unpack($props);
   @if length($props) == 0 {
     @return false;
   } @elseif nth($props, 1) == false {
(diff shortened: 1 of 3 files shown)


src/stylesheets/core/_functions.scss
```diff
@@ -467,13 +467,18 @@ a family and a line-height scale unit
...
+    @error "lh() needs both a valid face and line height unit in the format `lh(FACE, HEIGHT)`.";
+  }
+  $family: smart-quote(nth($props, 1));
+  $scale: smart-quote(nth($props, 2));
   @if not map-has-key($project-cap-heights, $family) {
     @error "#{$family} is not a valid font family. Valid values: #{map-keys($project-cap-heights)}";
   }
   @elseif not map-get($uswds-line-height, $scale) {
-    @error "`#{$scale}` is not a valid line-height scale. Valid values: #{map-keys($all-type-scale)}";
+    @error "`#{$scale}` is not a valid line-height unit. Valid values: #{map-keys($uswds-line-height)}";
```
thisisdano: Fixed this bug: output the wrong keys in the error message.

src/stylesheets/core/_functions.scss
```diff
@@ -731,6 +736,7 @@ colors: project and system
 */
 
 @function color($props...){
+  $props: unpack($props);
```
thisisdano: FYI: `unpack()` converts nested arglists into single one-dimensional lists for proper parsing:

((A, B), C) --> (A, B, C)

src/stylesheets/core/_functions.scss
```diff
@@ -809,74 +815,36 @@ Gets a valid USWDS flex value
...
-    @if not map-has-key($project-font-stacks, $our-family){
-      @error "#{$our-family} is not a valid font family. Valid values: #{map-keys($project-font-stacks)}";
-    }
-    @elseif not map-get($project-font-stacks, $our-family) {
-      @error "Font family `#{$our-family}` is disabled in your project's theme settings. Set its value to `true` to use this family.";
-    }
-    @else {
-      @return map-get($project-font-stacks, $our-family);
-    }
+@function scale($props...){
```
thisisdano: Removed the `family` and `weight` options from this function. Vestigial functionality — converted to single-use (typescale only).

src/stylesheets/core/_functions.scss
```diff
@@ -978,14 +946,14 @@ system orders
 
 /*
 ----------------------------------------
-spacing()
+units()
 ----------------------------------------
 Converts a spacing unit multiple into
```
el-mapache: I know you didn't write this comment but feels unclear. Should it be multiplier?

src/stylesheets/core/_functions.scss
```diff
@@ -809,74 +815,36 @@ Gets a valid USWDS flex value
...
-      $this-scale: map-get($all-type-scale, $our-scale);
-      @if $this-scale and $this-cap {
-        @return type-scale($this-cap, $this-scale);
-      }
-      @else {
-        @error "The scale `#{$our-scale}` is disabled in your project's theme settings. Set its value to `true` to use this family.";
-      }
-    }
+  $our-family: smart-quote(nth($props, 1));
+  $our-scale: smart-quote(smart-quote(nth($props, 2)));
```
el-mapache: Where is the `smart-quote` fn coming from? I cannot find it anywhere in bourbon, the sass reference, or this repo. Asking because the double-nesting seems weird to me, but I don't really have any context :D

maya: is it this https://github.com/uswds/uswds/blob/release-2.0/src/stylesheets/core/_functions.scss#L96 ?

el-mapache: ah yep, I was searching the wrong branch. thanks!

src/stylesheets/core/_functions.scss
thisisdano: It definitely works like a multiplier, but since we've been using `spacing units` as the term for our standard whitespace palette, it started to make more sense to use `units()` — I think I'm open to other names for this function, but since this is one of the primary user-facing functions (like `color()`) I'd like a name that works to connects the function to the concept.

thisisdano: Oh, and I did write that comment, so I can accept full responsibility!

src/stylesheets/core/_functions.scss
thisisdano: The double-nesting does seem redundant

thisisdano: But the intention of `smart-quote()` is to format specific types in a specific way (many of these functions can accept a range of types) — and to avoid the errors that come from doing something like quoting a number or a bool.

src/stylesheets/core/_functions.scss
el-mapache: Yeah, sorry, I meant the word `multiple` should be replaced with `multiplier`. The name of the function is totally clear

PR #2572, opened 2018-07-06
[ci] test filtering out pull requests from ci build

el-mapache:
## Filter PR branches from CI

The env variables specific to this project are not available when building from a forked branch, causing those builds to fail

## Description

Check for branches prefixed with `pull/`, which should indicate a PR originating from a fork of this repo. Hopefully this resolves build failures on those branches


el-mapache:
@maya Yep, my mistake, I applied the branch filter to the whole CircleCI set of tests. It now seems to be properly skipping only the code climate checks

PR #2574, opened 2018-07-09
Update eslint to use airbnb style guide

### LINKED ISSUE #2310: Use Airbnb JS linter config
We're using a custom rule set for our [JS linter config](https://github.com/18F/web-design-standards/blob/develop/.eslintrc.yml). We should use [Airbnb's JS linter config](https://www.npmjs.com/package/eslint-config-airbnb-base) instead as [recommended by the 18F Front End Guild](https://frontend.18f.gov/javascript/style/) for consistency. Discovered while researching #2307. Thanks to @joshbruce for bringing this to our attention.

el-mapache:
## Description
Use airbnb eslint config as a base, as per the front end guild's recommendations, rather than a series of incrementally added ad-hoc lint rules

## Additional information

I updated `gulp-lint` to the latest version so it plays nicely with the airbnb configs

Closes #2310.


maya:
Adding a link for a preview branch:

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/ab-2310)

el-mapache:
Thanks @maya! I don't think a JS review is necessary for this, I mostly just moved functions around / renamed them when eslint complained.

PR #2583, opened 2018-07-11
Fix footer class naming inconsistency

### LINKED ISSUE #1505: [UI Component] - Footers class name refactor
## Description

In reviewing the upcoming page template work I noticed that some of the class naming in the footers component is inconsistent. For example, the footers that use both primary and secondary sections use the classes

`usa-footer-primary-section`
`usa-footer-secondary_section`

I propose that they either be changed to 

`usa-footer-section-primary` and `usa-footer-section-secondary`
or 
`usa-footer-primary_section` and `usa-footer-secondary_section`

depending on how you interpret our suggested BEM style*

*someone who more regularly contributes to the frontend here could probably have a better explanation on which one would be correct

maya:
Changes the class `usa-footer-secondary_section` to `usa-footer-secondary-section` to follow what we use in `usa-footer-primary-section`. 

I was originally going to suggest we go with `usa-footer-section-secondary` and `usa-footer-section-primary` but then found we have a lot of additional CSS classes following the `usa-footer-primary-*` pattern and would rather not change that many class names.

I'm not sure how the unit tests work but realized we have the footer template appearing manually in `spec/unit/footer/template.html` and wasn't sure if it should be pulling it in from a source file so we don't have to keep an additional file up-to-date.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-footer-naming/components/detail/footer--default.html)

Fixes #1505.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8994,13 +8994,14 @@
       "integrity": "sha512-MC94mHZRvJ3LfykJlTUipBqenZz1pacOZEMhhQ8dMGcDHs0SBE5GbsavUXV7YtP3icBW17W0Zy1I0lfASmo9Pg==",
       "requires": {
         "inherits": "^2.0.1",
+        "isarray": "^2.0.4",
         "stream-splicer": "^2.0.0"
       },
       "dependencies": {
         "isarray": {
-          "version": "0.0.1",
-          "resolved": "https://registry.npmjs.org/isarray/-/isarray-0.0.1.tgz",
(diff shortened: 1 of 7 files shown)


el-mapache:
@maya I can't actually view the preview link, can you double check the URL?

maya:
Fixed! Was missing `/components/` in the URL.

maya:
Also, noticed the footer had the status of `wip` in fractal which didn't sound right, so changed it to `ready` in 00d373bf1643b1cf8204bac6cc5a76b433f4506d.

PR #2584, opened 2018-07-11
Fix 1 mixed content NetSparker issue

maya:
https://components.designsystem.digital.gov/components/detail/links.html: Change external link from http to https

Not 100% positive what the issue is on https://components.designsystem.digital.gov/components/detail/footer--default.html is, but we changed the http in the SVGs to https.

/cc @wslack


maya:
@wslack I'm not sure if the SVG change was correct. According to https://www.w3.org/2000/svg, `http://www.w3.org/2000/svg` is an XML namespace.

wslack:
I might be totally wrong; want me to circle back and ask them?

maya:
Yes, please bc `http://www.w3.org/2000/svg` is the valid XML namespace for SVG.

wslack:
KK. Waiting for GSA IT to tell me what the mixed content issue was.

maya:
Merging this, @wslack feel free to give us a heads up if they find out what the second issue was.

PR #2599, opened 2018-07-18
[WIP] Add global styles control, apply CSS to classes and remove styles from HTML elements

### LINKED ISSUE #2241: Request: provide css that only applies to classes, not tags

maya:
`$uswds-global-styles` will add base typography styles inside `.usa-prose` class, as global elements applied to base HTML, or none at all.

### Element styles added to scoping
- [x] Paragraph text (`p`)
- [x] Links (`a`)
- [x] Headings (`h1`-`h6`)
- [x] Lists (`ul`, `ol`, `li`)
- [x] Table (`table`)

### New or updated elements and classes
- `.usa-button` (button) (class already existed)
- `.usa-list` (ul, ol)
- `.usa-table` (table)
- `.usa-label` (label) for HTML `<label>`s changed from Label component which is now tag 
- `.usa-input` (input)
- `.usa-select` (select) - dropdown
- `.usa-textarea` (textarea)
- `.usa-legend` (legend)
- `.usa-tag` changed from `.usa-label` which we're now using for HTML `<label>`
- `.usa-checkbox-input` [type="checkbox"]
- `.usa-checkbox-label` [type="checkbox"]
- `.usa-radio-input` [type="radio"]
- `.usa-radio-label` [type="radio"]
- `.usa-range` [type="range"]

### Questions

- Link styles are [not appearing on forms](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-global-styles-control/components/detail/sign-in-form.html). Is this weird? Should link styles be turned on by default?
  - [x]  `a` and `p` on by default with `$uswds-global-styles-basic: true !default;`
-  Should we add: `.usa-link` class as well?
    - [x]  Added
- What's the relationship between this and `.usa-content` which sets a max-width?
- Best name for the scoping class: `.usa-prose`, `usa-content`?
- Do we want to add in add'l controls for fine-tuning the scoping, i.e. adding in paragraph and link styles

### Nesting
- For tables, I ended up nesting the internal CSS to make the `@mixin` and `@extend` easier to manage
- For lists, I kept them unnested by adding an additional `@extend`
- Which approach do we prefer?

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-global-styles-control)

Fixes #2241.


maya:
@el-mapache @thisisdano this is coming along! I have some questions above that would be great to get your thoughts on.

maya:
Should we have a switch for link and paragraph styles (like GOV.UK) and turn them on by default? Components will look strange without these basic styles.

thisisdano:
I could definitely see some kind of default link style

thisisdano:
Have to see what the effect on `p` would be...

thisisdano:
But yeah, I think trying the gov.uk approach makes sense

maya:
I'm not sure I follow those first thoughts, I think we should have a solid scoped set of type styles that you can turn on/off with a class: `usa-content` or some other name. And that we should be simpler to start, and can add/tweak as we go. The refinement of what's scoped would be in the Sass mixin or in settings (that can be built in later on), but we would have a sensible default set of styles to start.

maya:
OK let's go with the link and paragraph text to start. I think we'll be updating `p` styles in #2215 to have less adverse effects, but that would be in a future PR.

thisisdano:
My first comment is an attempt to be less ambiguous and try to address the following bullets:

- What's the relationship between this and `usa-content` which sets a max-width?
- Best name for the scoping class: `usa-prose`, `usa-content`?

So, what is the relationship between the existing component (`usa-content`) and the hook for the suite of scoped styles?

**Case 1: They are the same**
- What should the name be?
- What is the technique for enabling or disabling the scoped styles associated with the component?

**Case 2: They are different**
- What should the names be?
- What distinguishes the components?

My proposal argues for enabling Case 1 but that Case 2 is valid and could be controlled with settings:
- There are two classes, one for nonscoped content styles (`usa-content`) and one for scoped content styles that is explicit in its naming that it's for scoping and thus may cause specificity issues (`usa-content-scope`)
- There is a flag/setting for applying the scoped content styles to `usa-content` directly for those who _always_ want `usa-content` to apply the scoped styles. This could be active by default — subject to feedback.
- There are be valid use cases of `usa-content` that want the scoped styles only in _some_ instances — in which case you would set the deactivate the above flag and use both `usa-content` and `usa-content-scope` on an element that needs both
- This is distinct from the _global_ scope styling that we could set on `a` and possibly also `p`, which could also be disabled in settings

**So what this means:** Essentially, do what you've been doing, but add a new style (called something like `usa-content-scope` that has the scoped element styling. This _potentially_ has the benefit of using class-based techniques to enable functionality that could previously only be activated with compiling from source, which we know is occasionally not possible.

src/stylesheets/elements/_inputs.scss
```diff
@@ -222,79 +220,79 @@ legend {
...
   background: transparent;
   color: transparent;
   cursor: pointer;
   height: 1.2rem;
   width: 100%;
 }
 
-[type=range]::-webkit-slider-thumb {
+.usa-range::-webkit-slider-thumb {
   -webkit-appearance: none;
```
thisisdano: Again, unrelated to this PR, but should the webkit-thumb explicitly have the same bg color as the other two?

src/stylesheets/elements/_inputs.scss
```diff
@@ -222,79 +220,79 @@ legend {
...
-[type=range] {
+.usa-range {
   appearance: none;
   border: none;
   padding-left: 0;
   width: 100%;
 }
 
-[type=range]::-webkit-slider-runnable-track {
+.usa-range::-webkit-slider-runnable-track {
```
thisisdano: Unrelated to this PR, but I see that these three tracks have different rules and it makes me want to see a cross-browser slider example — for instance, it strikes me as strange that webkit's bg is light gray, moz has primary, and ms is transparent....

src/stylesheets/elements/_inputs.scss
```diff
@@ -323,13 +321,13 @@ legend {
   width: 2.2rem;
 }
 
-[type=range]::-ms-fill-lower {
+.usa-range::-ms-fill-lower {
```
thisisdano: Ah... this solves my transparent question :)

src/stylesheets/elements/_tags.scss
```diff
@@ -12,8 +12,7 @@
   }
 }
 
-.usa-label-big {
-  @extend .usa-label;
```
thisisdano: I see this use of an `@extend` here — are we consciously avoiding this technique now in favor of requiring both classes?

src/stylesheets/elements/_typography.scss
```diff
@@ -31,6 +31,99 @@ a {
...
+    }
+  } @else if $usa-content-styles == 'global' {
+    @extend %usa-content-styles;
+  } @else {
+    // Default nothing if it’s neither
+  }
+}
+/* stylelint-enable */
+
+@include usa-global-styles-basic;
```
thisisdano: I really like this placeholder --> mixin --> conditional --> include technique. Clean!

src/stylesheets/uswds.scss
```diff
@@ -36,9 +36,9 @@ Output all USWDS
 @import 'elements/embed';
 @import 'elements/figure';
 @import 'elements/inputs';
-@import 'elements/labels';
 @import 'elements/list';
 @import 'elements/table';
+@import 'elements/tags';
```
thisisdano: 🎉🎉🎉 
Now, it might make sense to make a note of the breaking markup changes so we can include them in the migration guide

src/components/04-tables/tables.njk
```diff
@@ -31,7 +31,7 @@
   </tbody>
 </table>
 
-<table class="usa-table-borderless">
+<table class="usa-table usa-table-borderless">
```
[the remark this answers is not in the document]
maya: I was wondering this too so I spoke to some people in the Design Systems slack about this and the consensus seems to be to double class it. tl;dr it seems to be more maintainable. See comments after this: https://design-systems.slack.com/archives/C0ENULNQZ/p1533067442000463

src/stylesheets/core/_settings.scss
```diff
@@ -11,6 +11,33 @@ PROJECT THEME SETTINGS
...
+Basic styles
+----------------------------------------
+Sets the default typography styles for
+paragraph text and links
+
+Accepts true or false
+----------------------------------------
+*/
+
+$usa-global-styles-basic: true !default;
```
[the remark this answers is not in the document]
maya: We can use theme.

src/components/07-form/templates/address-form.njk
```diff
@@ -78,7 +78,7 @@
       </div>
     </div>
 
-    <label for="zip">ZIP</label>
-    <input class="usa-input-medium" id="zip" name="zip" type="text" pattern="[\d]{5}(-[\d]{4})?">
+    <label class="usa-label" for="zip">ZIP</label>
+    <input class="usa-input usa-input-medium" id="zip" name="zip" type="text" pattern="[\d]{5}(-[\d]{4})?">
```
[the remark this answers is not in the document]
maya: See answer above.

src/components/07-form/templates/name-form.njk
```diff
@@ -1,16 +1,16 @@
 <form class="usa-form">
   <fieldset>
-    <legend>Name</legend>
-    <label for="title">Title <span class="usa-input-label-helper">(optional)</span></label>
-    <input class="usa-input-tiny" id="title" name="title" type="text">
+    <legend class="usa-legend">Name</legend>
+    <label class="usa-label" for="title">Title <span class="usa-input-label-helper">(optional)</span></label>
+    <input class="usa-input usa-input-tiny" id="title" name="title" type="text">
```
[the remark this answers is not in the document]
maya: See answer above.

src/components/07-form/controls/text-input.njk
```diff
@@ -1,19 +1,19 @@
...
+  <input class="usa-input" id="input-type-text" name="input-type-text" type="text">
 
-  <label for="input-focus">Text input focused</label>
-  <input class="usa-focus" id="input-focus" name="input-focus" type="text">
+  <label class="usa-label" for="input-focus">Text input focused</label>
+  <input class="usa-input usa-focus" id="input-focus" name="input-focus" type="text">
 
   <div class="usa-input-error">
-    <label class="usa-input-error-label" for="input-error">Text input error</label>
+    <label class="usa-label" class="usa-input-error-label" for="input-error">Text input error</label>
```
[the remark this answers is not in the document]
maya: Fixed two `class` attribtues issue in cb7f266.

thisisdano:
@maya are you able to resolve the merge conflict? Otherwise this is probably good to go!

maya:
Yes, merge conflicts resolved in ce1d8c2.

src/stylesheets/uswds.scss
maya: Yeah, we should add any notable changes to 2.0 in https://github.com/uswds/uswds/releases/tag/v2.0.0-alpha

PR #2604, opened 2018-07-19
Add new, accessible focus style

### LINKED ISSUE #2142: Problems with new gray dashed focus style
Hi all, I have some concerns about changing focus from vibrant blue to light gray: https://github.com/18F/web-design-standards/pull/2105

I agree that the current blue halo is problematic, but I design *a lot* on gray backgrounds and this will be an issue to have visible focus for many of my use cases. I think a more resilient approach is something like GOV.UK's solid gold outline and solid gold background on text links. This is also used on USA.gov (orange) /cc @davidstenger. While it's always hard to find the right focus color bc of different background colors, not many designs are bright orange/yellow, so it shouldn't conflict often. The dashes are also visually busy (personal opinion). Using a color with vibrance will help its visibility.

Was this approach tested with low-vision end-users? Is there research pointing to this approach?

Just some thoughts and would love to hear about the approach and what others think.

maya:
This updates the focus color and style to a solid blue outline (**Blue 40v** `#009afa`) from a light gray dotted outline.
- Meets WCAG 2.1's requirements of AA large text (3:1 contrast ratio) for focus on white and black background.
- Adds a focus offset only to buttons (not links) to make sure it's visible on blue buttons.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-new-focus/components/preview/layout--landing.html)

#### Blue 40v `#009afa`
3 on white (contrast ratio)
7 on black (contrast ratio)
screen shot 2018-07-19 at 3 13 55 pm
One designer mentioned the original focus being too thick so I changed it to 3px thickness and 3px offset.

### Blue focus on landing page template
Blue focus

Fixes #2142

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "uswds",
-  "version": "2.0.0-alpha",
+  "version": "2.0.0-alpha.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


maya:
@thisisdano here is what it looks like in 2, 3, 4:

2px
screen shot 2018-07-19 at 3 40 53 pm

3px
screen shot 2018-07-19 at 3 41 44 pm

4px
screen shot 2018-07-19 at 3 41 01 pm

maya:
@thisisdano I used 3px for the outline and offset bc they seemed the most balanced. Lmk if you find a system value that works better.

PR #2605, opened 2018-07-20
Hide date autocomplete icons on Safari

### LINKED ISSUE #1972: Small text inputs become very confusing when AutoFill on Safari Desktop activates
## Description

Safari's AutoFill prompt takes up the entire input field if the field is short, like in the date input template. I was very confused about how to overcome this and enter a date that was not part of the AutoFill list. I don't imagine this would test very well with a broader audience.

This feels like a problem with Safari's implementation – putting the prompt within the text field – but the pain will be felt by the user anyway.

This can be avoided by [disabling autocompletion](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion) if appropriate in the context of use, but is overkill of the template.

1. Enable AutoFill on Safari `preferences->AutoFill->Using info from my contacts` and ensure it's using a contact card with a DoB.
2. Visit [form-controls/date-input on standards.usa.gov](https://standards.usa.gov/components/form-controls/#date-input)
2. Attain focus on first input field
3. Figure out what to do next if you don't want to enter you own birthday

Screen shot of AutoFill Activated and consuming entire input field

maya:
This hides the autocomplete icons from appearing in Safari when you click on a field.

To test, you'd need to go to a computer where you have a Safari "My card" in your contacts with your birthdate (my personal machine in this case), and click on the field.

This also cleans up the SCSS of the date field component by nesting it better and removing some code that didn't do anything - was based on an older CSS Tricks article which I assume that Chrome may have changed how it displays auto-populate icons, so made it unnecessary.

I initially tried a simple `appearance: none` but it was not strong enough bc Safari's user agent adds a `display: none !important`, so I based the code on this: https://jrnv.nl/hiding-safaris-contact-auto-fill-when-autocomplete-is-turned-off-c415b2257df8

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/hide-date-autocomplete/components/preview/date-input.html)

Fixes #1972.

### DIFF
--- a/src/components/07-form/controls/date-input.njk
+++ b/src/components/07-form/controls/date-input.njk
@@ -1,18 +1,20 @@
-<fieldset>
-  <legend>Date of birth</legend>
-  <span class="usa-form-hint" id="dobHint">For example: 04 28 1986</span>
-  <div class="usa-date-of-birth">
-    <div class="usa-form-group usa-form-group-month">
-      <label for="date_of_birth_1">Month</label>
-      <input class="usa-input-inline" aria-describedby="dobHint" id="date_of_birth_1" name="date_of_birth_1" type="number" min="1" max="12" value="">
+<form>
+  <fieldset>
+    <legend>Date of birth</legend>
+    <span class="usa-form-hint" id="dobHint">For example: 04 28 1986</span>
+    <div class="usa-date-of-birth">
+      <div class="usa-form-group usa-form-group-month">
+        <label for="date_of_birth_1">Month</label>
+        <input class="usa-input-inline" aria-describedby="dobHint" id="date_of_birth_1" name="date_of_birth_1" type="number" min="1" max="12" value="">
+      </div>
+      <div class="usa-form-group usa-form-group-day">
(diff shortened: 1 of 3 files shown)


src/components/07-form/controls/date-input.njk
```diff
@@ -1,18 +1,20 @@
-<fieldset>
-  <legend>Date of birth</legend>
-  <span class="usa-form-hint" id="dobHint">For example: 04 28 1986</span>
-  <div class="usa-date-of-birth">
-    <div class="usa-form-group usa-form-group-month">
-      <label for="date_of_birth_1">Month</label>
-      <input class="usa-input-inline" aria-describedby="dobHint" id="date_of_birth_1" name="date_of_birth_1" type="number" min="1" max="12" value="">
+<form>
```
maya: Adding a form element was the only way to test this, i.e. to get the autofill to show up on the fractal site. Do we want to keep this in or remove?

PR #2606, opened 2018-07-20
Make extended header search open by default

### LINKED ISSUE #1640: Search Component design feedback from USA.gov
Issue more for consideration. Also, wondering if other users have provided feedback on this.

In USA.gov, we customized the WDS search box to reflect:

- a wider input field to give desktop users enough space for their query and make it easier for them to review or edit it. 
- visible expanded search box: now that we have a less cluttered header, we kept a visible input field (as opposed to just having the icon). Users tend to find it faster and easier. 

Finally, we are starting to conduct low fidelity UT on new header/footer. So far, we hear that users are having a hard time finding the search in mobile and some don't expect to find it under "Menu". We will continue to test this with a more functional prototype and report back. Just sharing some early feedback.

You might have seen some of these resources, but figured it would not hurt to include them here. Or they could perhaps be listed under the "For further reading" section: 
- https://www.nngroup.com/articles/search-visible-and-simple/
- https://www.nngroup.com/articles/magnifying-glass-icon/
- https://www.smashingmagazine.com/2009/09/10-useful-usability-findings-and-guidelines/#6-the-ideal-search-box-is-27-characters-wide)

maya:
- Puts secondary (utility) links above search
- Allows you to remove secondary links or search w/o effecting layout
- Makes search 27 letters wide as recommended on the [search bar page](https://designsystem.digital.gov/components/search-bar/).

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-extended-header-search/components/preview/header--extended.html)

Fixes #1640.

### DIFF
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -25,7 +25,7 @@ $z-index-nav:     9000;
   .usa-search {
     @include media($nav-width) {
       float: right;
-      max-width: 72ch;
+      max-width: 30em;
     }
   }
 }
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -25,7 +25,7 @@ $z-index-nav:     9000;
   .usa-search {
     @include media($nav-width) {
       float: right;
-      max-width: 27em; // max-width of 27 letters
+      max-width: $search-min-width;
     }
   }
 }
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -25,7 +25,7 @@ $z-index-nav:     9000;
   .usa-search {
     @include media($nav-width) {
       float: right;
-      max-width: 30em;
+      max-width: 27em; // max-width of 27 letters
     }
   }
 }
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -25,7 +25,7 @@ $z-index-nav:     9000;
   .usa-search {
     @include media($nav-width) {
       float: right;
-      max-width: $search-min-width;
+      max-width: calc(#{$search-min-width} + #{$usa-btn-small-width});
     }
   }
 }
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -25,7 +25,7 @@ $z-index-nav:     9000;
   .usa-search {
     @include media($nav-width) {
       float: right;
-      max-width: 21.5rem;
+      max-width: 72ch;
     }
   }
 }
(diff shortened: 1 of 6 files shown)


maya:
@thisisdano could you take a look at the design of this, I know the search bar looks a little wide but was following our own guidance of making it at least 27 letters wide.

src/stylesheets/components/_header.scss
```diff
@@ -25,7 +25,7 @@ $z-index-nav:     9000;
   .usa-search {
     @include media($nav-width) {
       float: right;
-      max-width: 21.5rem;
+      max-width: 27em; // max-width of 27 letters
```
el-mapache: we use this in two places, and have a comment explaining it, so it seems like a good candidate for a variable!

maya: Added in a2c60bb.

thisisdano:
It's definitely dramatic — for better or for worse, it totally captures your attention, perhaps overshadowing the links(?). I think it works, but it gives the header a very different feel — more practical I'd say. Let's go for it.

thisisdano:
I put a little "real" fake content in to get a better sense of how you might scan and read it. It's different, but successful.

screen shot 2018-07-25 at 7 44 40 am

src/stylesheets/core/_variables.scss
```diff
@@ -774,6 +774,7 @@ $focus-outline:                 units(0.5) solid $color-focus !default;
 $focus-spacing:                 0 !default;
 $nav-width:                     951px !default;
 $sidenav-current-border-width:  0.4rem !default; // must be in rem for math
+$search-min-width:              27em !default; // minimum width of 27 letters for comfortable search bar
```
thisisdano: I've already approved (and I like the way it looks right now) but perhaps `27ch` better reflects the 27-character guidance?

src/stylesheets/core/_variables.scss
```diff
@@ -774,6 +774,7 @@ $focus-outline:                 units(0.5) solid $color-focus !default;
...
 
 // 44 x 44 pixels hit target following Apple iOS Human Interface Guidelines
 $hit-area: 4.4rem !default;
```
thisisdano: Good ol `$hit-area` — always brightens my day. Good to see you again, friend.

maya:
It is dramatic! I think it's right on that ems might not be the best units to use as it accounts for the widest letter (m), not the average of all the letters, which ch's (0) might be closer. However, when I used 27ch it was dramatically small and didn't seem to actually allow for 27 letters:
screen shot 2018-07-25 at 9 07 35 am

Changing it to 34-35ch got it closer or 190-200px. Thoughts?
screen shot 2018-07-25 at 9 22 30 am

maya:
Made the changes to search width and merging since this was previously approved.

thisisdano:
1) I think this is OK as is [and it's already merged :) ]
2) Buuuut.... you _could_ use something like new grid to do something like...

```scss
[role=search] {
  @include grid-row;
  @include u-flex(justify-end);
}

.usa-search-small [type=search] {
  max-width: $search-min-width;
  @include grid-col;
}

.usa-search-small [type=submit] {
  @include grid-col(auto);
}
```

Then there's no calc-ing required.

maya:
We can always open new PRs to improve :) feel free to post any screenshots too so we can see what it would look like.

PR #2608, opened 2018-07-23
[app] unnest h4 from li element in footer

### LINKED ISSUE #1659: Footer - Topic part of list items
**Issue**:
The H4 Topic is part of the list items

**Solution**:
Remove the H4 for the list items

Example:
 ```html
<h4>Topic</h4>
<ul class="usa-unstyled-list usa-width-one-fourth usa-footer-primary-content">
  <li><a href="javascript:void(0);">Secondary link</a></li>
  <li><a href="javascript:void(0);">Secondary link</a></li>
  <li><a href="javascript:void(0);">Secondary link</a></li>
  <li><a href="javascript:void(0);">Secondary link</a></li>
</ul>
```

el-mapache:
## Description

Header elements cant be children of `ul`s, this PR promotes the heading in a footer nav to be a sibling of the `ul`.

## Additional information

I had to make some minor CSS additions to get the nav to display consistently

[PREVIEW 😎 ](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/ab-1659/components/preview/footer--big.html)

Fixes #1659

### DIFF
--- a/src/components/10-footer/footer--big.njk
+++ b/src/components/10-footer/footer--big.njk
@@ -5,42 +5,42 @@
   <div class="usa-footer-primary-section">
     <div class="usa-grid">
       <nav class="usa-footer-nav usa-width-two-thirds">
-        <ul class="usa-unstyled-list usa-width-one-fourth usa-footer-primary-content">
-          <li class="usa-footer-primary-link">
-            <h4>Topic</h4>
-          </li>
-          <li class="usa-footer-secondary-link"><a href="javascript:void(0);">Secondary link</a></li>
-          <li class="usa-footer-secondary-link"><a href="javascript:void(0);">Secondary link</a></li>
-          <li class="usa-footer-secondary-link"><a href="javascript:void(0);">Secondary link that's a bit longer than most of the others</a></li>
-          <li class="usa-footer-secondary-link"><a href="javascript:void(0);">Secondary link</a></li>
-        </ul>
-        <ul class="usa-unstyled-list usa-width-one-fourth usa-footer-primary-content">
-          <li class="usa-footer-primary-link">
-            <h4>Topic</h4>
-          </li>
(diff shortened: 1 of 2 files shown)


maya:
@el-mapache There looks like some issues in mobile.

### This is what I see (nothing happens, no arrow appearing):
screen shot 2018-07-23 at 11 11 10 am

### This is what I expect:

maya:
@el-mapache on desktop I think the spacing may be off between the heading item and list. Can you run the [visual regression testing](https://github.com/uswds/uswds/blob/develop/CONTRIBUTING.md#visual-regression-testing) to make sure nothing's changed?

el-mapache:
@maya hmm, the `secondary-link` class has a ton of specificity attached to it, which is probably why trying to move the tag out of the list is breaking things. Do you think a better approach would be to style the primary link element to look more like an `h4`? Or do we specifically want that level of heading in this component pattern?

maya:
The link heading has the class `usa-footer-primary-link`, so as long as we apply styles and functionality to that, it should be good, right? `usa-footer-secondary-link` are the links within the list.

el-mapache:
yes, but I believe the issue is that the mobile styles etc all depend on that link also being inside the list

el-mapache:
I guess what I'm thinking is, its going to be a lot more work to pull all that stuff out, and make sure it doesnt break anything, than it is to just style the primary link to look like a header

maya:
Yes, I was thinking we would want to pull that out the list bc it shouldn't have been in a list in the first place. We'd just want it applied to the class, but I haven't taken a look at the JS code in a while.

el-mapache:
@maya, ok, we'll have to keep the old JS as well, otherwise it will be a breaking change. Or should this be going into 2.0?

maya:
@el-mapache I'm fine for this to go into 2.0

PR #2618, opened 2018-07-23
Accept named values in units() fn

thisisdano:
- Now the `units()` function can accept the USWDS named spacing values, like `units(card)` or units(`mobile-lg`)`
- Rewrite `u-padding()` mixin to use `get-uswds-value()` not `units()` fn

### DIFF
--- a/src/stylesheets/core/mixins/utilities/_padding.scss
+++ b/src/stylesheets/core/mixins/utilities/_padding.scss
@@ -8,18 +8,18 @@
     $important: ' !important';
   }
   @if $side == all {
-    padding: units($value...)#{$important};
+    padding: get-uswds-value(padding, $value...)#{$important};
   }
   @elseif $side == x {
-    padding-left: units($value...)#{$important};
-    padding-right: units($value...)#{$important};
+    padding-left: get-uswds-value(padding, $value...)#{$important};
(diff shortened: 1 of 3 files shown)


maya:
Will `units(1)` still work? Bc I remember using that on the doc site so will want to update it as well.

thisisdano:
Yeah, any numeric value should still work on `units()` — like `units(1)` or `units(.5)`

src/stylesheets/core/mixins/utilities/_padding.scss
```diff
@@ -8,18 +8,18 @@
     $important: ' !important';
   }
   @if $side == all {
-    padding: units($value...)#{$important};
+    padding: get-uswds-value(padding, $value...)#{$important};
```
el-mapache: This seems fine, but I'm not sure I understand why this change was made. Can you give me a little context?

thisisdano: Mostly to make it consistent with the other utility mixins that are built with the same function. It handles errors a little differently — that is, `get-uswds-value` accepts non-USWDS values (like `17px`) with a warning and `units` throws an error. `get-uswds-value()` also "knows" the standard values for padding and will reject something like `neg-2` or `-2` since those negative values are invalid for padding.

PR #2623, opened 2018-07-24
Fix palette registry, etc.

thisisdano:
Converting `plugins` --> `palette` bolloxed a few things up in the palette registry (that is, the big list of available palettes). I took the opportunity to clear up that registry and remove a bunch of complexity that wasn't helping us out. The new registry is a simple map, and the palette names are also a bit clearer now.


src/stylesheets/core/_functions.scss
```diff
@@ -1053,154 +1053,15 @@ elegant
...
-    }
-    @else {
-      $registry: map-get($palette-registry, standard);
-    }
-    // now the registry is set
-    @if not map-has-key($registry, #{$palette}){
-      @error "#{$palette} isn't in the registry.";
-    }
-    @else {
+    @if map-has-key($palette-registry, $palette){
```
thisisdano: ✂️!

el-mapache: wow, good change

PR #2628, opened 2018-07-25
Update Showcase page

maya:
An ongoing list of new sites using the USWDS.


maya:
Probably a good time to update this page. Moved to "review."

PR #2644, opened 2018-08-01
Add range slider to form controls

### LINKED ISSUE #2374: Add range slider example and documentation to "Form controls"
We currently have a range slider code (#259) in the USWDS, but not an example of it. We should add it. 

We even have usability/implementation documentation for it:  

> • Whenever appropriate, label each end of the slider with what it represents (whether a number range, “large” or “small,” “high” or “low”).
> • Set the `min` and `max` attribute of the `input` element to match whatever instructions or labels accompany the slider.
> • Set the `step` attribute so the slider is not too granular. Sliders should let users input approximate values. By setting the step to “10” or “20,” you keep users from inadvertently selecting values that are more precise than they intended.

Relevant issues: #244, #701

maya:
- Adds the range slider that already exists in the CSS into fractal to display the component.
- Cleans up range slider Sass

This is a CSS-only range input that has a single background color.

To get the left side blue with CSS-only, we'd need to change the design so that the height of the thumb control is the same height as the track, bc we'd do this with a negative left box shadow (via [Stack Overflow](https://stackoverflow.com/questions/18389224/how-to-style-html5-range-input-to-have-different-color-before-and-after-slider)). Which we could do with a square-shaped thumb control. 

Here's a JavaScript version of the slider that meets all design expectations (larger circle, shorter track, blue to the left): https://codepen.io/mbenari/pen/JBLdrr via [CSS Tricks](https://css-tricks.com/sliding-nightmare-understanding-range-input/). We could link to this or offer it up for people that need this fancier version.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-range-slider/components/detail/range-input.html)

Fixes #2374.

### DIFF
--- a/src/stylesheets/elements/_inputs.scss
+++ b/src/stylesheets/elements/_inputs.scss
@@ -273,17 +273,17 @@ legend {
   background: $color-gray-lighter;
   border: 1px solid $color-gray-medium;
   cursor: pointer;
-  height: 1.2rem;
+  height: 1.6rem;
   width: 100%;
 }
 
 @mixin range-thumb {
   background: $color-gray-lightest;
-  border: 1px solid $color-gray-medium;
+  box-shadow: 0 0 0 1px $color-gray-medium;
   border-radius: 1.5rem;
   cursor: pointer;
-  height: 2.2rem;
-  width: 2.2rem;
+  height: 2.5rem;
+  width: 2.5rem;
 }
 
 @mixin range-ms-fill {
@@ -330,7 +330,7 @@ legend {
   &::-webkit-slider-thumb {
     @include range-thumb;
     appearance: none;
-    margin-top: -0.65rem;
+    margin-top: -0.6rem;
   }
 
   &::-moz-range-thumb {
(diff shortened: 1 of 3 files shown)


src/stylesheets/elements/_inputs.scss
```diff
@@ -263,76 +263,90 @@ legend {
...
   background: $color-gray-light;
   border: 1px solid $color-gray-medium;
   border-radius: 2rem;
 }
 
-[type=range]::-ms-fill-upper {
-  background: $color-gray-light;
-  border: 1px solid $color-gray-medium;
-  border-radius: 2rem;
+[type=range] {
```
el-mapache: Did we want to use a class here, instead of targeting the actual HTML element? Or is targeting a specific subset of elements ok?

I don't remember exactly where we landed but I thought that we were trying to move away from tag-based css

maya:
Thanks! Also, feel free to clean up the JS and markup in the CodePen, I don't totally know what it's doing, only that it works :)

PR #2646, opened 2018-08-03
Remove unused JavaScript

el-mapache:
## Description

Removes components and tests which are no longer in use ahead of the 2.0 release.


el-mapache:
@maya My mistake, it is ready for review. I don't think the accordion updates are affected at all by the PR you linked.

PR #2663, opened 2018-08-09
[WIP] Banner / Fonts / Bugs

thisisdano:
**UPDATE: Changing to WIP until I can keep the scope reasonable.**

Here are a few changes related to the work I'm doing to add USWDS 2.0 to Digital.gov. The new style scoping work was a success, and the following work was all that was necessary to get USWDS 2.0 working with Digital.gov.

Here's the work [on digital.gov](https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/dw-add-uswds-2.0/).

- - -

**Update banner to USWDS 2.0 tokens**
- Uses new tokens and mixins for the banner
- Minimal styling changes from the the 1.x banner
- No markup changes

- - -

**Start converting 1.0 variables to 2.0 variables**
- This starts to move the user-modifiable variables to `settings.scss` and use the `theme-` prefix
- Updates these variables to 2.0 tokens
- Clones the new variable to the old variable (`$old-var: $theme-new-var`)
- NEXT: There's more work to do here, both with organization and a better, more backward compatible cloning strategy, but I think this is OK for now

- - -

**Add Tahoma font definition and restrict font face generation to provided faces**
- Adds a font definition for Tahoma for normalization
- Restricts `@font-face` generation to only our provided fonts (that have `generate: true` set)
- NEXT: This PR does not convert the normalization values to aspect value. This is a bit more work and can come in a subsequent PR

- - -

**Fix some bugs**
- Fix output of color values
- Allow non-2.0-token values in the old float grid (at least for now)
- Use a `px` value for a legacy media query calculation


maya:
The old grid on desktop had a width of 1040px with 30px margins on each side making the content width 980px. On mobile, the margins are 15px. This is how all the other components are built on the site. 

This new grid has a width of 1024px with 32px margins making the content 960px. On mobile the margins are also 32px.

This will make it look disjointed from the other components unless it's all done together. Was this the intention?

thisisdano:
OK, left the old var names as is for now to help scope this PR.

thisisdano:
Closing in favor of #2668

PR #2665, opened 2018-08-10
Fix errors generated from coverage command

### LINKED ISSUE #2662: Code Climate test coverage not appearing
https://codeclimate.com/github/uswds/uswds
screen shot 2018-08-08 at 11 52 59 am

On the README
screen shot 2018-08-08 at 11 53 07 am

el-mapache:
## Description

I noticed we were getting some errors with `npm run cover`, which may be affecting the test coverage display in code climate.

I also adjusted the `if` statement in our circle config, it was causing the test coverage to never be sent, instead of only skipping the report when `pull` is present in the branch name

Fixes #2662.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
       - run:
           name: Checking build
           command: |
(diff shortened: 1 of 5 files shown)


maya:
Yeah I opened this issue #2662 around that.

PR #2668, opened 2018-08-10
Banner / Fonts / Bugs [Take 2]

thisisdano:
**New PR, replacing the old one with the same name...**

Here are a few changes related to the work I'm doing to add USWDS 2.0 to Digital.gov. The new style scoping work was a success, and the following work was all that was necessary to get USWDS 2.0 working with Digital.gov.

Here's the work [on digital.gov](https://federalist-proxy.app.cloud.gov/preview/gsa/digitalgov.gov/dw-add-uswds-2.0/).

- - -

**Update banner to USWDS 2.0 tokens**
- Uses new tokens and mixins for the banner
- Minimal styling changes from the the 1.x banner
- No markup changes

- - -

**Start converting 1.0 variables to 2.0 variables**
- This starts to move the user-modifiable variables to `settings.scss` and use the `theme-` prefix
- Updates these variables to 2.0 tokens
- Clones the new variable to the old variable (`$old-var: $theme-new-var`)
- NEXT: There's more work to do here, both with organization and a better, more backward compatible cloning strategy, but I think this is OK for now

- - -

**Add Tahoma font definition and restrict font face generation to provided faces**
- Adds a font definition for Tahoma for normalization
- Restricts `@font-face` generation to only our provided fonts (that have `generate: true` set)
- NEXT: This PR does not convert the normalization values to aspect value. This is a bit more work and can come in a subsequent PR

- - -

**Fix some bugs**
- Fix output of color values
- Allow non-2.0-token values in the old float grid (at least for now)
- Use a `px` value for a legacy media query calculation


thisisdano:
From @maya in the old PR:

> The old grid on desktop had a width of 1040px with 30px margins on each side making the content width 980px. On mobile, the margins are 15px. This is how all the other components are built on the site.
>
> This new grid has a width of 1024px with 32px margins making the content 960px. On mobile the margins are also 32px.
>
> This will make it look disjointed from the other components unless it's all done together. Was this the intention?

src/stylesheets/components/_banner.scss
```diff
@@ -1,160 +1,171 @@
...
       }
     }
   }
 }
 
 .usa-banner-inner {
-  @include outer-container($site-max-width);
-  @include padding(null $site-margins-mobile);
+  @include grid-container($theme-site-max-width);
+  @include u-padding-x($theme-site-margins-mobile);
```
maya: Why does this use `$theme-site-margins-mobile` and https://github.com/uswds/uswds/pull/2668/files#diff-f387466a429ac4e7ace4438ebb06393bR19 and https://github.com/uswds/uswds/pull/2668/files#diff-d44185fbb54964385b1a02ae836e77efR8 uses `$site-margins-mobile`? 

When would you use `$theme-site-margins-mobile` over `$site-margins-mobile` and vice versa?

thisisdano: In general, I was trying to scope the work in this PR as much as possible to this specific component. `$theme-site-margins-mobile` should be the new variable going forward, used anywhere `$site-margins-mobile` is used. We could begin to globally replace the spacing unit variables now, I suppose, but I started to go in that direction and found myself in deeper, broader PR water in my initial PR. 

Spacing units would be a good candidate for global replacement (in most instances) — something like `font-size` variables is definitely a bigger job and one I'd like to save for (possibly) a PR of its own or (more likely) something that's only integrated piece by piece.

src/stylesheets/core/_variables.scss
```diff
@@ -606,19 +618,19 @@ settings
...
+$title-font-size:     $theme-title-font-size;
+$h1-font-size:        $theme-h1-font-size;
+$h2-font-size:        $theme-h2-font-size;
+$h3-font-size:        $theme-h3-font-size;
+$h4-font-size:        $theme-h4-font-size;
+$h5-font-size:        $theme-h5-font-size;
+$h6-font-size:        $theme-h6-font-size;
+$base-line-height:    $theme-base-line-height;
+$heading-line-height: $theme-heading-line-height;
+$lead-line-height:    $theme-lead-line-height;
```
thisisdano: Actually, I should revert these to the old values for now — to allow us to more specifically scope our work. Globally changing these vars will cause unexpected problems right now.

PLEASE HOLD ON MERGING UNTIL I FIX THIS

src/stylesheets/core/_variables.scss
```diff
@@ -777,18 +789,20 @@ calculated variables
...
 $text-narrow-width:             40ch !default;
 $text-wide-width:               72ch !default;
 $lead-max-width:                77rem !default;
-$site-max-width:                1040px !default;
-$site-margins:                  3rem !default;
-$site-margins-mobile:           1.5rem !default;
+$site-max-width:                $theme-site-max-width;
+$site-margins:                  $theme-site-margins;
+$site-margins-mobile:           $theme-site-margins-mobile;
 $article-max-width:             600px !default;
```
thisisdano: I guess these should be OK to replace with new vars... I thought I'd reverted this, tho.
This means we can replace all instances of `$site-margins` and `$site-margins-mobile` with there `$theme-` counterparts.

src/stylesheets/components/_banner.scss
thisisdano: OK, I changed the vars in the old grid to use the new margins theme vars

src/stylesheets/core/_variables.scss
```diff
@@ -616,9 +628,9 @@ $h3-font-size:        2rem !default;
 $h4-font-size:        1.7rem !default;
 $h5-font-size:        1.5rem !default;
 $h6-font-size:        1.3rem !default;
-$base-line-height:    1.5 !default;
-$heading-line-height: 1.3 !default;
-$lead-line-height:    1.7 !default;
+$base-line-height:    1.5rem !default;
+$heading-line-height: 1.3rem !default;
+$lead-line-height:    1.7rem !default;
```
maya: These values should all be unitless to scale with font size. Can you change this back?

thisisdano: Oh ha — I knew this looked wrong.

thisisdano:
OK, fixed that line height mistake

src/stylesheets/components/_banner.scss
```diff
@@ -1,160 +1,171 @@
...
   text-decoration: underline;
   top: 0;
   width: 100%;
 
-  @include media($small-screen) {
-    @include padding(0 null null 0);
+  @include at-media('mobile-lg') {
+    @include u-padding-top(0);
+    @include u-padding-left(2px);
+    @include u-margin-left(.5);
```
maya: Add a leading 0 here, since that's what we follow in our Sass coding styleguide and show apply to mixins as well.

maya:
🚀 @thisisdano do the bug fixes in this PR close any open issues in the backlog?

PR #2669, opened 2018-08-13
[app] close top navigation dropdown on body click

### LINKED ISSUE #2617: [Header]: Close expanded dropdown navigation when user clicks away
via @mariamarrero in #1905
>Once the menu is opened, it does not close unless a user clicks on the tab. Have you considered changing it so that when a user clicks somewhere on the page (for example in the search box), the menu closes without the click, but also keeping the option of closing it when the user clicks on the tab?
>
>PS: We have gotten direct comments from users on these 2 issues.

el-mapache:
## Description

Previously, when a desktop top navigation menu was open, it would remain open as other elements on the page were interacted with, which can be confusing. Now it should close when other elements are clicked.

Fixes #2617

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,11 +10,11 @@
       "integrity": "sha1-qjwUYz6IdvxgKYBkvdTp4/F33CI=",
       "dev": true,
       "requires": {
-        "commander": "2.15.1",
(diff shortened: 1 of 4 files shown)


maya:
LGTM

PR #2674, opened 2018-08-14
Add utilities folders to stylelint and fix stylelint issues

maya:
Adds folders utilities/, packages/, project/ to be linted.

However, this only goes one level deep. Since there are nested folders it's not going deeper than one level. TODO: need to go deeper to make sure all nested files are linted.

Fixes a few lint errors.

I didn't change the commenting bc it wasn't flagging (for now). We're using two different linting rulesets on uswds-site and this repo (the 18F ruleset vs. a custom ruleset). Though I think we may end up rewriting comments like that anyway bc if the comment is not necessary to be viewed in compiled CSS, it should use the standard invisible Sass comments.


thisisdano:
FWIW, all comments — with the exception of `/*! important comments */` — are currently stripped from the compiled css

maya:
Oh, I didn't realize that they were stripped out. In that case, I don't see a real reason to change.

Issue #1469, opened 2016-09-09
Media queries should probably be in em units


kaelig:
To respect the users' font-size setting, it is typically a good thing to set media queries in `em` (instead of `px` currently).

This article by @zellwk  sums it up very well: http://zellwk.com/blog/media-query-units/

> As you can see from our tests above, the only unit that performed consistently across all four browsers is em. There aren’t any differences between em and rem with the exception of bugs found on Safari.
> 
> `px` media queries performed well in two of the three experiments (with the exception of Safari, again). Unfortunately, `px` media queries remained at `400px` in the third experiment, which makes it a no-go if you intend to support users who change their browser’s font-size value.
> 
> Hence, my conclusion after these experiments is: **Use `em` media queries.**
> 
> If you’re using a library that doesn’t do em media queries, point the developer to this article so they know the implications of their code. Otherwise, feel free to switch to a em based library like [Mappy-breakpoints](https://github.com/zellwk/mappy-breakpoints), [Breakpoint-sass](http://breakpoint-sass.com/) or [sass-mq](https://github.com/sass-mq/sass-mq).

Being the author of sass-mq, I am obviously biased… But you may want to know that it was inspired by the way gov.uk handles media queries. It was built with content websites (News sites in particular) in mind – with two constraints: encourage a mobile-first design, and support IE8 with the minimum effort required. Happy to help answer any of your questions about it!

PS: thank you for open sourcing your code, it's awesome to have visibility on your workflow, practices, and code! Very inspiring.

maya:
Hi @kaelig, thanks so much for opening this issue and bringing to our attention common practices in content/news websites. We'll be able to consider this more when we update our grids component.

kaelig:
Closing this as it is stale and I'm cleaning up my issues.

PR #2677, opened 2018-08-14
Add switch to respect user font size settings

thisisdano:
The changes we've made and are making for USWDS 2.0 allows us to do something previously impossible: respect users' browser font size settings.

This PR adds a new boolean variable to theme settings: `$theme-respect-user-font-size`.  When set to `true` the system does a few things:
- Sets the font size of the `html` element to `100%`
- Uses `em` values for media queries (for media queries, `em` is a multiplier of the user's browser default font size)

When set to `false` the system acts as it does now:
- Sets the font size of the `html` to the value of `$theme-root-font-size`
- Uses `px` values for media queries

- - -

Currently, I've left the value of `$theme-respect-user-font-size` as `false` because our components are not all rebuilt with the new tokens. They'll all tend to get _really_ big if this is set to `true` right now. However, once the system is rebuilt in the new tokens (as the banner is now), we should set our system default to `true` — and our components will scale related to the user's settings, _which is the expected behavior_. Our mobile-first, responsive components will be well suited to this change.

**Examples**
Medium font size set in Chrome:

screen shot 2018-08-14 at 3 05 43 pm

- - -

Large font size set in Chrome:

screen shot 2018-08-14 at 3 01 53 pm
(The discrepancy in widths between the header and the content above is due to the old grid still using a px value as its max-width)

- - -

In the 1.0 system, these components would appear the same size regardless of the user settings. The `false` setting above would preserve this behavior, which might be the appropriate setting for some USWDS projects.

### DIFF
--- a/src/stylesheets/project/_uswds-project-settings.scss
+++ b/src/stylesheets/project/_uswds-project-settings.scss
@@ -71,12 +71,34 @@ $theme-namespace: (
 ----------------------------------------
 Root font size
 ----------------------------------------
-Sets the font size of the html root
-for rem calculations
+Setting $theme-respect-user-font-size to
+true sets the root font size to 100% and
+uses ems for media queries
+----------------------------------------
+$theme-root-font-size only applies when
+$theme-respect-user-font-size is set to
+false.
 
-Accepts [n]px or 100%
+This will set the root font size
+as a specific px value and use px values
+for media queries.
+
+Accepts true or false
 ----------------------------------------
 */
+
+$theme-respect-user-font-size: false;
+
+
+// $theme-root-font-size only applies when
+// $theme-respect-user-font-size is set to
+// false.
+
+// This will set the root font size
+// as a specific px value and use px values
+// for media queries.
+
+// Accepts values in px
 
 $theme-root-font-size: 10px;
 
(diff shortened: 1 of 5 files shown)


src/stylesheets/project/_uswds-project-settings.scss
```diff
@@ -149,33 +170,33 @@ defined in the system:
 $theme-font-definitions: (
   /*
   font-new: (
-    name: 'Font Display Name',
-    dir: 'folder-name', // relative to settings.$theme-font-path
+    name: "Font Display Name",
```
maya: Should be single quotes.

thisisdano: Oh huh, it must not have picked up the last changes

thisisdano:
It is somewhat based on that epic, largely triggered by thinking about docs and the way we communicate the relative size of things. It's also related to https://github.com/uswds/uswds/issues/1469 and the problems I previously had with https://github.com/uswds/uswds/pull/2524

But after thinking it through, it seemed like we were in a good position to do this now, and come up to speed on this accessibility best practice.

maya:
Sounds good, just wanted to make sure it was linked to the right issue/conversations, thanks!

PR #2678, opened 2018-08-15
Fix bugs in italic font builder

thisisdano:
- Remove unnecessary `@if` in font builder mixin
- Use proper filename for source sans pro italic
- Use proper map keys for italic weights loop

### DIFF
--- a/src/stylesheets/core/_functions.scss
+++ b/src/stylesheets/core/_functions.scss
@@ -1495,11 +1495,9 @@ through all possible variants
     }
 
     @if $italic {
-      @debug 'it me! italic! #{$font-name}';
       @each $weight, $filename in $italic {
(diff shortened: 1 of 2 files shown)


src/stylesheets/core/_functions.scss
```diff
@@ -1473,41 +1473,38 @@ through all possible variants
...
-                #{$weight},
-                italic,
-                $file-formats: eot woff2 woff ttf
-              );
-            }
+    @if $italic {
+      @each $weight, $filename in $italic {
+        @each $key, $outputweight in $these-weights {
+          @if $outputweight == $weight and $filename {
+            @include font-face(
```
thisisdano: Looks like this is a Bourbon mixin — just to keep in mind as we move to possibly deprecate

PR #2679, opened 2018-08-15
Update font size theme vars to use current token names

thisisdano:
Note: this change will temporarily break `uswds-site`. Once this is merged, we can update the var names in `uswds-site` and all will be well again.


thisisdano:
I added the new vars into `uswds-site/v2` to prevent breakage when this is merged. 
We'll delete the old vars from site's settings once this is merged.

PR #2685, opened 2018-08-17
SCSS refactoring

el-mapache:
## Description

WIP PR trying to remove/refactor some of out duplicated logic


thisisdano:
So, I really complexed up this PR and managed to break the Circle build as well 🎉. Unfortunately, I seem to be completely unable to parse Circle errors these days and I can't tell what broke.

But what the heck did I do?

The process of commenting the `color()` function raised a number of issues:

- Color vars had a different naming scheme from the rest of the style vars. They didn't use the `$color-` prefix, like `$units-` and `$scale-` vars.
- The color functions were configured to accept any hex and rgb(a) color because the color mixins accepted color vars like `$red-50` that, themselves, output as hex/rgb(a). They should probably only accept system/theme color values.
- Existing `$uswds-` prefixing of token vars was inconsistent with the `system` designation we're giving these values in the documentation.

So, how did I address these:

- I commented and cleaned up the `color()` fn
- I commented and cleaned up `get-system-color()` fn
- I used `system` instead of `uswds` for the system var prefix
- I used `$color-` as a color var prefix
- I used utility shortcodes (text strings) instead of variables (hex, rgb) in color functions. Now the color fns check for a string match with shortcodes instead of just passing a hex/rgb color through
- I disallowed the use of  "basic colors" (single-named system colors, like `red` or `orange`) in color functions because they are inappropriate for component Sass.

I still have some more work to do to make all the color functions and mixins work properly. That work is coming soon. I'm sorry that this muddied up what was previously a nice, light PR!

el-mapache:
@dano no worries! Maybe we can look into making a new PR with just the color changes? We can also look to see if there is a logical place to break up the some of the new work as well; for example, all the `system` prefixing could be a separate PR.

Thanks for helping out on this, it's looking really great!

maya:
@uswds/core it's showing [aXe errors](https://circleci.com/gh/uswds/uswds/3120?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) that are not accurate. The foreground color (text) is not `#adadad`. Where is that coming from and how do we resolve?

maya:
Oops wrong PR, ignore.

thisisdano:
@el-mapache I temporarily changed all the theme color vars to have a `$color-theme` prefix to prevent collision with the old vars. This fixes our test errors. I'll do a more permanent fix (using 2.0 color functions in our components) in a subsequent PR.

I also tried to do a better job commenting the `get-system-color()` fn to better explain the reasoning and potential TODO fixes.

I _think_ we can consider the PR complete at this point. Thanks for your help shepherding it through.

thisisdano:
This will break `uswds-site` until https://github.com/uswds/uswds-site/pull/629 is merged

PR #2688, opened 2018-08-20
Bump alpha release to .2

el-mapache:
## Description

Update the `package.json` with the new version number


maya:
@el-mapache Do we need to publish tags as well?

PR #2691, opened 2018-08-20
[test] Adds better file globbing to stylelint configs

el-mapache:


maya:
@el-mapache I checked the CircleCI test and it looks like lib isn't being ignored anymore: https://circleci.com/gh/uswds/uswds/3115?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Also, should we ignore `codeclimate/total-coverage` issue? The issue it's linked to is about the footer which is unrelated to this PR.

maya:
Actually, it looks like some lib files are ignored, but this was what I saw being checked:  
`/home/circleci/project/src/stylesheets/lib/*` (many of these)

Also, can we not output each ignored file like we currently do? or group it so it doesn't take as much space.

maya:
I'm not sure that I understand that it wasn't being ignored before. I noticed that before we had 45 SCSS files checked (not inluding the lib files I mentioned) and in this PR it shows 134 sources checked (including ~89 lib files). Do we know why they're showing up now and not before?

el-mapache:
@maya Nothing was being ignored explicitly, stylelint was only ignoring the files in `lib` because it wasn't one of the folders being passed to the `linter` function.

The path being provided to the `ignore` option wasn't being recognized at all.

Now that it is, it seems like stylelint's default behavior is to dump a message into the console whenever a file is ignored, which is pretty annoying

maya:
@el-mapache yeah I think that error is bc we're using a super old stylelint (2 years old), if we upgrade it, this should go away.

```
Changed: files matching ignore patterns no longer receive an "info"-severity message, which was always printed by the string formatter. Instead, the file's stylelint result object receives an ignored: true property, which various formatters can use as needed. The standard string formatter prints nothing for ignored files; but when the verbose formatter is used, ignored files are included in the filelist.
```
via https://stylelint.io/CHANGELOG/

el-mapache:
@maya hm ok, I'm a little hesitant to upgrade as it necessitates a change to the `stylelint-rules` package and a publish to npm, and I don't know how many other projects depend on it. I suppose we can just go for it!

maya:
we'll see what happens! 😀 

It also sounds like upgrading will hide the ignored files by default unless the `verbose` formatter is used, which will be nice.

maya:
Would this also require a publish to npm if we use `release-2.0` as the base branch?

el-mapache:
the default formatter is actually the `verbose` level, the other ones have really ugly output. I'll play around with ti.

el-mapache:
The `stylelint-rules` live in their own repo [here](https://github.com/18F/stylelint-rules), so we'll need to publish an update regardless.

maya:
OK, if the `verbose` formatter is helpful for the error reporting then maybe we should leave it.

re: `stylelint-rules` - does it make sense to depend on that repo? I noticied it hasn't changed since 2016. Would we be better off using our own stylelint config file, like we do on the docs site? https://github.com/uswds/uswds-site/blob/master/.scss-lint.yml I guess it would be good to know what changes would need to be made.

/cc @thisisdano for visibility

el-mapache:
I'm ok with that, the only caveat being that we lose the custom rules our package provides. However, maybe we don't need to care about those anymore?

maya:
Our own config file could still be based on the rules of the 18F one if it's sensible and that's what we'd want, right?

el-mapache:
@maya yep definitely! The rules in place already reflect the 18F linting config choices. Again, the only difference is we lose the custom plugins for rules that stylelint didn't handle.

maya:
Once the base branch is set to `release-2.0`, this is good to go. We will clean up in another PR (use newer stylelint, use our own config file on our repo).

el-mapache:
@maya looks like I based this off of develop, and bringing it up to date with release-2.0 produces many linting errors in our new `utilities` folder. I'm going to close this and open a new one that fixes the errors, and also just uses `gulp-stylelint`, rather than our custom version

PR #2695, opened 2018-08-21
Nav menu items should re-open after closing

### LINKED ISSUE #2692: Header navigation won't re-open after opening and closing
## Steps to reproduce
1. Visit https://v2.designsystem.digital.gov/components/headers/basic/
2. Open, close, and open the same navigation dropdown item right after each other
3. Observe you can't re-open it

Tagging @el-mapache as it seems JS-related

el-mapache:
## Description

* We have to explicitly add a check to see if the currently clicked menu
item is the same as the active menu item, otherwise we get a weird
interaction between the accordion and nav behaviors

* Move the body event handler after the nav control handler. We do this
because the body event is guaranteed to fire first, since any event that
bubbles will fire that listener

fixes #2692

### DIFF
--- a/config/gulp/javascript.js
+++ b/config/gulp/javascript.js
@@ -31,9 +31,13 @@ gulp.task(task, function (done) {
     .pipe(gulp.dest('dist/js'));
 
   stream
-    .pipe(sourcemaps.init({ loadMaps: true }))
-    .pipe(uglify())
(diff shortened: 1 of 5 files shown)


maya:
@el-mapache doesn't look like CircleCI is passing, can that be fixed?

PR #2696, opened 2018-08-22
Fix typography spacing - take 2

maya:
Updates our typography spacing rules to ensure content consistently displays proper margins in different scenarios (from simple layouts to complex grids). We expect folks to adjust margins as needed to sections in their project.

## Changes include
- Uni-directional margins on paragraph text (top margin)
- Applies top margins to headings that follow other elements using the `* + %usa-heading` selector. This is scoped to the the `theme-content-styles` mixin which is set in a user's settings file to be scoped or global.
- Removes previous code that removed top margins from the first element and bottom margins from the last element in `usa-section` and `usa-grid` and grid items

TODO
- [x] Check page templates and components to make sure it looks OK and we didn't cause any unintended issues
- [ ] When the review is complete, should we remove the added test page: [Type spacing](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-typography-spacing-2/components/detail/type-spacing.html)?

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-typography-spacing-2/components/detail/type-spacing.html)

## Sample page
127 0 0 1_3000_components_preview_type-spacing 1

Supercedes #2343.

Fixes: #2215.

### DIFF
--- a/src/components/01-type/type-spacing.njk
+++ b/src/components/01-type/type-spacing.njk
@@ -1,0 +1,58 @@
+<div class="usa-prose">
+  <div class="usa-grid">
+    <div class="usa-width-one-half">
+      <h1>Heading 1</h1>
+      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam sodales eros a diam fermentum, id pulvinar lacus tincidunt. Ut varius ipsum at ex maximus molestie. Quisque tincidunt posuere ligula, ut dictum quam.</p>
+    </div>
+    <div class="usa-width-one-half">
+      <h2>Heading 2</h2>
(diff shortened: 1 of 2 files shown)


alex-page:
Awesome work @maya let me know if you have any issues with this in the future, happy to contribute. 

Must feel great removing all those `margin-top` resets 👍

maya:
@kangax thanks so much, and exactly! It's like a weight is lifted 😄

thisisdano:
Maybe not in scope here, but should `usa-paragraph` have default measure applied?

maya:
@thisisdano possibly, but agree we should do it in another PR.

thisisdano:
Aight, let's merge and move onward.

PR #2703, opened 2018-08-27
A few bug fixes

### LINKED ISSUE #2700: minh utility should use `auto` not `none`
And `[class*=grid-col]` should not have a `min-height` set. This is causing layout errors in flex-column layouts.

### LINKED ISSUE #2701: `grid-gap` not working properly
The following grid gap utilities seem not to be working:
- `grid-gap-lg`
- `grid-gap-sm`
- `grid-gap-2`
- `grid-gap-5`

thisisdano:
- Add `0`, `2`, and `5` to list of system column gaps
- Use `sm` and `lg` as column gap variable suffixes (ex: `$theme-column-gap-sm`)
- Fix bug that prevented `grid-gap-sm` and `grid-gap-lg` from outputting
- Add explicit `grid-gap-md` as a clone of `grid-gap`
- Use `0` for min-height and min-width reset value
- Rename internal grid-units fn to prevent unintended use (`grid-units()` --> `spacing-multiple()`)
- A little file tidying

- - - 
Fixes #2701 
Fixes #2700


maya:
Question about this:
- Rename internal grid-units fn to prevent unintended use (`grid-units()` --> `spacing-multiple()`)

Will `units(1)` and `$units-1` still work? as shown here: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds-site/dw-add-utility-mixins/style/spacing-units/

maya:
The answer to the question above:
>yes `units()` is still the public function and the units var remains the same. This change is just for an internal function. this change is to prevent users from mistakenly using the internal function

PR #2713, opened 2018-08-31
Update colors to normalized values

### LINKED ISSUE #2712: Update system colors to luminance-normalized values
If we normalize each color at each grade around the `gray` luminance value for that grade, we should have a color system that delivers on its promise: to achieve AA Large contrast for magic numbers of 40+ and AA contrast for magic numbers of 50+.

Our new luminance conversion tools will assist in this. Most colors will see only minor adjustments, but some may be a bit more noticeable. See [this chart](https://v2.designsystem.digital.gov/style/color/system-palette-adjust/) for the first pass of automatic adjustments. About 80% of the automatic adjustments are OK as is, the rest can be hand tuned as needed.

thisisdano:
This applies the colors from https://github.com/uswds/uswds-site/pull/639
This normalizes color grades around the luminance of the `gray` family to assure that magic numbers work as advertised.

Fixes #2712.


maya:
@thisisdano the preview branch is not showing styles on any pages. See: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-update-colors/components/detail/fonts.html

src/stylesheets/core/_system-tokens.scss
```diff
@@ -198,9 +198,9 @@ Measure
 
 $system-measure-smaller:	    40ch;
 $system-measure-small:	      60ch;
-$system-measure-base:	      66ch;
+$system-measure-base:	        66ch;
 $system-measure-large:	      72ch;
-$system-measure-larger:	    77ch;
+$system-measure-larger:	      77ch;
```
maya: This section's alignment is off.

thisisdano:
Fixed — I think

Issue #2660, opened 2018-08-07
 Integrate 2.0 palette values into components


maya:
### Core
- [x] variables @maya @thisisdano #2760
  - review the variables at the end of the file (move or remove)

### Elements
- [x] buttons @thisisdano #2737 
- [x] inputs @thisisdano #2729
- [x] list (was aready done)
- [x] table @maya #2720
- [x] tags @maya #2717
- [x] typography @thisisdano #2731 

### Components
- [x] accordions @el-mapache #2722
- [x] alerts @el-mapache #2728
- [x] banner @thisisdano #2668
  - can we rewrite this with our new approach (property + funciton, less mixins)? -maya
  - yes, will do -dw
- [x] footer @thisisdano #2741
- [x] forms @thisisdano #2743
- [x] graphic-list @maya #2735
- [x] header @maya #2746 
- [x] hero @el-mapache #2726 
- [x] layout (documentation page template) @maya #2733
- [x] media-block @maya #2734
- [x] navigation @thisisdano #2750 
- [x] search @maya #2745
- [x] section @el-mapache #2727
- [x] sidenav @maya #2740
- [x] skipnav @maya #2724

### Generic 
- [x] grid - remove
- [x] mixins (should have been updated after each component but double check at the end) - @maya #2748 (removes unneeded mixin)

---

Our existing components should use the 2.0 palette.

- Spacing units
- Colors
- Type scale steps

## Size key: conversion table

1.0 | 2.0
--- | ---
1px | 1px —
2px (2-3px) | 2px —
5px (3-6px) | 4px `units(0.5)`
10px (6-12px) | 8px `units(1)`
12px (10-14px) | 12px `units(1.5)`
15px (14-18px) | 16px `units(2)`
20px (18-22px) | 20px  `units(2.5)`
25px (22-28px) | 24px  `units(3)`
30px (28-36px) | 32px `units(4)`
40px (36-44px) | 40px `units(5)`
50px (44-52px) | 48px `units(6)`
55px (52-60px) | 56px `units(7)`
60px (60-68px) | 64px `units(8)`
70px (68-76px) | 72px `units(9)`

1.0 | 2.0 shortcode | 2.0 var | 2.0 system palette
--- | --- | --- | ---
`$color-black` | `'black'` | `$color-black` | `'black'`
`$color-base` `$color-black-light` | `'base-darkest'` `'ink'` | `$color-theme-base-darkest` `$color-theme-ink` | `'gray-90'`
[...]

PR #2720, opened 2018-09-05
Use 2.0 palette values for table

maya:
Part of #2660.

[😎 Preview](https://federalist-proxy-staging.app.cloud.gov/preview/uswds/uswds/update-table-v2/components/detail/tables.html)

### DIFF
--- a/src/stylesheets/elements/_table.scss
+++ b/src/stylesheets/elements/_table.scss
@@ -1,6 +1,6 @@
 %usa-table {
   border-spacing: 0;
-  margin: 2em 0;
+  margin: units(2.5) 0;
 
   thead {
     th {
@@ -9,7 +9,7 @@
 
     th,
     td {
-      background-color: $color-gray-lightest;
+      background-color: color('base-lighter');
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/stylesheets/elements/_table.scss
```diff
@@ -19,16 +19,16 @@
 
   th,
   td {
-    background-color: $color-white;
-    border: 1px solid $color-gray;
+    @include u-border(1px, 'base-dark');
+    background-color: color('white');
     font-weight: $font-normal;
-    padding: 1rem 1.5rem;
+    padding: units(1) units(2);
```
maya: y padding was 10px before, thought that looked better.

src/stylesheets/elements/_table.scss
```diff
@@ -19,16 +19,16 @@
...
   }
 
   caption {
-    @include h5;
-    font-family: $font-serif;
-    margin-bottom: 1.2rem;
+    @include u-font('serif', 'xs');
+    font-weight: $font-bold;
```
maya: what function do we use for font-weight?

src/stylesheets/elements/_table.scss
```diff
@@ -19,16 +19,16 @@
...
-    border: 1px solid $color-gray;
+    @include u-border(1px, 'base-dark');
+    background-color: color('white');
     font-weight: $font-normal;
-    padding: 1rem 1.5rem;
+    padding: units(1) units(2);
   }
 
   caption {
-    @include h5;
```
maya: Should I keep using this h5 mixin?

src/stylesheets/elements/_table.scss
maya: Updated in 1420978.

PR #2721, opened 2018-09-05
Update tags v2

maya:
Removes some unneeded mixins from tags.

### DIFF
--- a/src/stylesheets/elements/_tags.scss
+++ b/src/stylesheets/elements/_tags.scss
@@ -1,11 +1,10 @@
 .usa-tag {
-  @include u-bg('base-dark');
   @include u-font('sans', '2xs');
-  @include u-margin-right(0.5);
-  @include u-padding-x(1);
-  @include u-padding-y(1px);
   @include u-radius('sm');
   @include u-text('white', 'uppercase');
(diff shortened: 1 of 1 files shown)


src/stylesheets/elements/_tags.scss
```diff
@@ -1,11 +1,10 @@
 .usa-tag {
-  @include u-bg('base-dark');
   @include u-font('sans', '2xs');
-  @include u-margin-right(0.5);
-  @include u-padding-x(1);
-  @include u-padding-y(1px);
   @include u-radius('sm');
```
maya: Should this be `border-radius: units(1px);` or the mixin?

maya: Should this be `border-radius: units(2px);` or the mixin?

maya: Is there a border radius function a la `border-radius: radius(sm);`?

thisisdano: no, but there should be. made an issue here: https://github.com/uswds/uswds/issues/2723

maya: Added the function d0a8c88, seems to work.

Issue #2633, opened 2018-07-28
2.0 Alpha - Theme fonts not based on new font families


stphnwlkr:
I attempted to set $theme-font-heading to $theme-font-sans in the settings file, but in typography.scss, the font families are not based on the theme values. When I made the change in typography.scss to use the theme value, the font family is set as source-sans-pro and not "Source Sans Pro Web","Helvetica Neue",Helvetica,Roboto,Arial,sans-serif".

thisisdano:
You're right — we haven't gotten to this yet. Hoping to address variable usage this sprint.

thisisdano:
This work is scheduled for the current sprint

maya:
Done in #2731 .

PR #2722, opened 2018-09-05
Accordion component  2.0 updates

el-mapache:
## Description

Part of #2660

Attempts to update existing variables to new tokens/functions/mixins/etc

[Preview here](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/ab-sass-accordion-update/components/detail/accordion--bordered.html) — I'm not 100% if the hover gray color is correct, it seems a bit lighter

### DIFF
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -1,12 +1,12 @@
 // Variables
 
-$accordion-border: 3px solid color('gray-5');
+$accordion-border: 3px solid color('base-lighter');
 
 // Accordion Styles
 
 @mixin accordion-list-styles {
   @include unstyled-list;
-  color: $color-base; // what is this in the new system
+  color: color('base-darkest');
   margin: 0;
   padding: 0;
   width: 100%;
@@ -15,24 +15,25 @@ $accordion-border: 3px solid color('gray-5');
 // scss-lint:disable PropertyCount
 @mixin accordion-button-styles {
   @include button-unstyled;
-  background-color: color('gray-5');
+  @include u-font-family('sans');
+
+  background-color: color('base-lighter');
   background-image: url('#{$theme-image-path}/minus.png');
   background-image: url('#{$theme-image-path}/minus.svg');
   background-position: right units(2.5) center;
   background-repeat: no-repeat;
   background-size: units(2);
-  color: $color-base; // theme-base or something?
+  color: color('base-darkest');
   cursor: pointer;
   display: inline-block;
-  font-family: $theme-font-sans;
   font-weight: $theme-font-weight-bold;
   margin: 0;
   padding: units(2) (units(2.5) * 2 + units(2)) units(2) units(2.5);
   width: 100%;
 
   &:hover {
-    background-color: color('gray-5');
-    color: $color-base; // again who knows
+    background-color: color('base-lighter');
+    color: color('base-darkest');
   }
 }
 // scss-lint:enable PropertyCount
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -1,6 +1,6 @@
 // Variables
 
-$accordion-border: 3px solid color('base-lighter');
+$accordion-border: 3px solid $color-theme-base-lighter;
 $text-color: color('base-darkest');
 
 // Accordion Styles
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -16,12 +16,6 @@ $accordion-border: 3px solid $color-theme-base-lighter;
 @mixin accordion-button-styles {
   @include button-unstyled;
   @include u-font-family('sans');
-  @include u-padding(
-    units($theme-spacing-sm),
-    units($theme-spacing-md) * 2 + units($theme-spacing-sm),
-    units($theme-spacing-sm),
-    units($theme-spacing-md)
-  );
 
   background-color: color('base-lighter');
   background-image: url('#{$theme-image-path}/minus.png');
@@ -34,6 +28,7 @@ $accordion-border: 3px solid $color-theme-base-lighter;
   display: inline-block;
   font-weight: $theme-font-weight-bold;
   margin: 0;
+  padding: units($theme-spacing-sm) units($theme-spacing-md) * 2 + units($theme-spacing-sm) units($theme-spacing-sm) units($theme-spacing-md);
   width: 100%;
 
   &:hover {
@@ -97,15 +92,9 @@ $accordion-border: 3px solid $color-theme-base-lighter;
 }
 
 .usa-accordion-content {
-  @include u-padding(
-    units($theme-spacing-sm),
-    units($theme-spacing-md),
-    calc(#{units($theme-spacing-sm)} - #{units($theme-spacing-2xs)}),
-    units($theme-spacing-md)
-  );
-
   background-color: color('white');
   overflow: auto;
+  padding: units($theme-spacing-sm) units($theme-spacing-md) calc(#{units($theme-spacing-sm)} - #{units($theme-spacing-2xs)}) units($theme-spacing-md);
 
   > *:first-child {
     margin-top: 0;
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -17,18 +17,18 @@ $accordion-border: 3px solid $color-theme-base-lighter;
   @include button-unstyled;
   @include u-font-family('sans');
   @include u-padding(
-    units($theme-spacing-md-small),
-    units($theme-spacing-medium) * 2 + units($theme-spacing-md-small),
-    units($theme-spacing-md-small),
-    units($theme-spacing-medium)
+    units($theme-spacing-sm),
+    units($theme-spacing-md) * 2 + units($theme-spacing-sm),
+    units($theme-spacing-sm),
+    units($theme-spacing-md)
   );
 
   background-color: color('base-lighter');
   background-image: url('#{$theme-image-path}/minus.png');
   background-image: url('#{$theme-image-path}/minus.svg');
-  background-position: right units($theme-spacing-medium) center;
+  background-position: right units($theme-spacing-md) center;
   background-repeat: no-repeat;
-  background-size: units($theme-spacing-md-small);
+  background-size: units($theme-spacing-sm);
   color: $color-theme-base-darkest;
   cursor: pointer;
   display: inline-block;
@@ -47,7 +47,7 @@ $accordion-border: 3px solid $color-theme-base-lighter;
   background-image: url('#{$theme-image-path}/plus.png');
   background-image: url('#{$theme-image-path}/plus.svg');
   background-repeat: no-repeat;
-  background-size: units($theme-spacing-md-small);
+  background-size: units($theme-spacing-sm);
 }
 
 @mixin accordion-nested-list {
@@ -69,7 +69,7 @@ $accordion-border: 3px solid $color-theme-base-lighter;
 
   + .usa-accordion,
   + .usa-accordion-bordered {
-    margin-top: units($theme-spacing-small);
+    margin-top: units($theme-spacing-xs);
   }
 }
 
@@ -78,7 +78,7 @@ $accordion-border: 3px solid $color-theme-base-lighter;
     border-bottom: $accordion-border;
     border-left: $accordion-border;
     border-right: $accordion-border;
-    padding-bottom: units($theme-spacing-md-small);
+    padding-bottom: units($theme-spacing-sm);
   }
 
   .usa-accordion-heading {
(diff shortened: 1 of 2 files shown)


el-mapache:
Please tell me all the things I'm doing wrong 👼

src/stylesheets/components/_accordions.scss
```diff
@@ -1,16 +1,12 @@
 // Variables
 
-$accordion-border: 3px solid $color-gray-lightest;
+$accordion-border: 3px solid color('gray-5');
```
thisisdano: - Let's use the theme palette var here: `color('base-lighter')`
- Let's look at using `2px` instead of `3px` to align with system palette

src/stylesheets/components/_accordions.scss
```diff
@@ -1,16 +1,12 @@
...
 // Accordion Styles
 
-$icon-size: $spacing-md-small;
-$h-padding: $spacing-medium;
-$v-padding: $spacing-md-small;
-
 @mixin accordion-list-styles {
   @include unstyled-list;
-  color: $color-base;
+  color: $color-base; // what is this in the new system
```
thisisdano: `base-darkest` See https://github.com/uswds/uswds/blob/ca2b10fef9278a334715b6179e069a4a89e37a8e/src/stylesheets/core/_variables.scss#L1047

src/stylesheets/components/_accordions.scss
```diff
@@ -1,16 +1,12 @@
 // Variables
 
-$accordion-border: 3px solid $color-gray-lightest;
+$accordion-border: 3px solid color('gray-5');
 
 // Accordion Styles
 
-$icon-size: $spacing-md-small;
```
thisisdano: You _could_ set this to `units(2.5)` and use the var as it was used previously

thisisdano: And we _could_  build a new set of vars like those that start at https://github.com/uswds/uswds/blob/ca2b10fef9278a334715b6179e069a4a89e37a8e/src/stylesheets/core/_variables.scss#L1139

...and use _them_ in these var assignments. Like:
```sass
$theme-spacing-x-small: units(0.5); // was 0.5rem
$theme-spacing-small: units(1); // was 1rem
$theme-spacing-md-small: units(2); // was 1.5rem;
$theme-spacing-medium: units(2.5); // 2rem;
$theme-spacing-large: units(4); // 3rem;
```

```sass
$icon-size: $theme-spacing-md-small;
$h-padding: $theme-spacing-medium;	
$v-padding: $theme-spacing-md-small;
```

src/stylesheets/components/_accordions.scss
```diff
@@ -19,24 +15,24 @@ $v-padding: $spacing-md-small;
...
+  color: $color-base; // theme-base or something?
   cursor: pointer;
   display: inline-block;
-  font-family: $font-sans;
-  font-weight: $font-bold;
+  font-family: $theme-font-sans;
+  font-weight: $theme-font-weight-bold;
   margin: 0;
-  padding: $v-padding ($h-padding * 2 + $icon-size) $v-padding $h-padding;
+  padding: units(2) (units(2.5) * 2 + units(2)) units(2) units(2.5);
```
thisisdano: now that's a complicated little fellow!

src/stylesheets/components/_accordions.scss
maya: For colors, it's always best to use the highest level of abstraction aka theme values. In this case, gray 5 is base lighter. See: https://v2.designsystem.digital.gov/style/color/theme-palette/

maya: 3px would need to be changed to a 2.0 palette value.

src/stylesheets/components/_accordions.scss
```diff
@@ -19,24 +15,24 @@ $v-padding: $spacing-md-small;
...
   width: 100%;
 
   &:hover {
-    background-color: $color-gray-lighter;
-    color: $color-base;
+    background-color: color('gray-5');
```
thisisdano: Use `color('base-lighter')`

src/stylesheets/components/_accordions.scss
```diff
@@ -19,24 +15,24 @@ $v-padding: $spacing-md-small;
...
+    color: $color-base; // again who knows
```
thisisdano: Use `color('base-darkest')`

src/stylesheets/components/_accordions.scss
maya: In the old system, this corresponded to the darkest base color or what was off-black. In 2.0, I think this is `$theme-color-ink` which is base darkest.

src/stylesheets/components/_accordions.scss
thisisdano: Whoops, nm. It's goes against the paradigm a bit to use functions in settings. The more reliable thing would be more like:

```sass
$theme-spacing-x-small: 0.5; // was 0.5rem
$theme-spacing-small: 1; // was 1rem
$theme-spacing-md-small: 2; // was 1.5rem;
$theme-spacing-medium: 2.5; // 2rem;
$theme-spacing-large: 4; // 3rem;
```

```sass
$icon-size: units($theme-spacing-md-small);
$h-padding: units($theme-spacing-medium);	
$v-padding: units($theme-spacing-md-small);
```

src/stylesheets/components/_accordions.scss
```diff
@@ -67,7 +63,7 @@ $v-padding: $spacing-md-small;
 
   + .usa-accordion,
   + .usa-accordion-bordered {
-    margin-top: $spacing-small;
+    margin-top: units(1);
```
thisisdano: It's worth investigating using the technique I'd mentioned on line 7 to use variables as before, except assign those vars system tokens

src/stylesheets/components/_accordions.scss
```diff
@@ -76,7 +72,7 @@ $v-padding: $spacing-md-small;
     border-bottom: $accordion-border;
     border-left: $accordion-border;
     border-right: $accordion-border;
-    padding-bottom: $spacing-md-small;
+    padding-bottom: units(2);
```
thisisdano: See comment on line 70

src/stylesheets/components/_accordions.scss
```diff
@@ -90,17 +86,14 @@ $v-padding: $spacing-md-small;
   margin: 0;
 
   &:not(:first-child) {
-    margin-top: $spacing-x-small;
+    margin-top: units(0.5);
```
thisisdano: See comment on line 70

src/stylesheets/components/_accordions.scss
```diff
@@ -45,7 +41,7 @@ $v-padding: $spacing-md-small;
   background-image: url('#{$theme-image-path}/plus.png');
   background-image: url('#{$theme-image-path}/plus.svg');
   background-repeat: no-repeat;
-  background-size: $icon-size;
+  background-size: units(2);
```
thisisdano: See comment on line 70

src/stylesheets/components/_accordions.scss
```diff
@@ -19,24 +15,24 @@ $v-padding: $spacing-md-small;
 // scss-lint:disable PropertyCount
 @mixin accordion-button-styles {
   @include button-unstyled;
-  background-color: $color-gray-lightest;
+  background-color: color('gray-5');
   background-image: url('#{$theme-image-path}/minus.png');
   background-image: url('#{$theme-image-path}/minus.svg');
-  background-position: right $h-padding center;
+  background-position: right units(2.5) center;
```
thisisdano: See comment on line 70

src/stylesheets/components/_accordions.scss
```diff
@@ -19,24 +15,24 @@ $v-padding: $spacing-md-small;
...
   background-repeat: no-repeat;
-  background-size: $icon-size;
-  color: $color-base;
+  background-size: units(2);
```
thisisdano: See comment on line 70

maya:
Can you add a preview link for the new work?

src/stylesheets/components/_accordions.scss
```diff
@@ -19,24 +15,24 @@ $v-padding: $spacing-md-small;
...
+  color: $color-base; // theme-base or something?
   cursor: pointer;
   display: inline-block;
-  font-family: $font-sans;
-  font-weight: $font-bold;
+  font-family: $theme-font-sans;
```
thisisdano: This won't work properly in this format. `$theme-font-sans` is a key/token/string (like `'source-sans-pro'`) rather than a value/stack (like `"Source Sans Pro", Helvetica, sans` or whatever).

I realize this is confusing (see https://github.com/uswds/uswds/issues/2633).

For now, use `@include u-font-family('sans')` — but I can make a function as well... something like `family()` as well as a var (`$font-family-sans`)

src/stylesheets/components/_accordions.scss
maya: The Sketch file is using a gap of 10px for this, do we want to bump this up to 8px/1 unit?

thisisdano: Sounds good

maya:
Not seeing any hover styles working on this.

src/stylesheets/components/_accordions.scss
el-mapache: I think I'm not a huge fan of that? Like, we have all of the tokens, and this is an internal component, so why add the additional unnecessary abstraction of a new variable for this component? 

Unless we add them to variables that get used throughout the components?

src/stylesheets/components/_accordions.scss
el-mapache: Yep, the original was pretty complex too! What suggestions do you have to improve it? Should I explicitly list out each padding value (left, right, etc)

src/stylesheets/components/_accordions.scss
el-mapache: actually, forget this! I totally misread what you wrote, adding those spacing vars to the variables file seems like a good idea!

el-mapache:
Will probably need more work but removing the WIP to make it clear this is in a mostly done state

src/stylesheets/components/_accordions.scss
```diff
@@ -1,16 +1,13 @@
 // Variables
 
-$accordion-border: 3px solid $color-gray-lightest;
+$accordion-border: 3px solid color('base-lighter');
+$text-color: color('base-darkest');
```
maya: See this https://github.com/uswds/uswds/pull/2722#discussion_r215398212, we already have a variable for text color, so it's not needed.

el-mapache: oh ok, but we do still need a call to the `color` fn for `base-lighter`? Or is there a variable for that too?

src/stylesheets/components/_accordions.scss
el-mapache: I was just going off of what @thisisdano wrote above. If there is a different standard I'll switch to that.

src/stylesheets/components/_accordions.scss
maya: Good question, I think `$color-theme-base-lighter` may work as well. I'm just looking through the variables file. I don't know if there's a preference for fn vs the variable.

thisisdano: Yes, `$color-theme-base-lighter` will work

el-mapache: @dano Essentially then, the pattern is:

1. Always default to a theme variable if it exists
2. If not, use the string key of the token (not sure if that is the right word here)

el-mapache: Because, I see, for example, in @maya's PR [here](https://github.com/uswds/uswds/pull/2720/files) that she is using 'base-lighter' (the string) vs the variable/

thisisdano: I think this is not yet settled law, but it could be a good start — 
1. if the variable is sufficient, use the variable
2. else if the function is sufficient, use the function
3. else use the mixin

I'm hesitant to carve this in stone right now, but it's a useful decision tree

thisisdano: 0. If a standard rule is sufficient, use a standard rule

el-mapache: Super helpful, thanks! I feel like arriving at a decision soon is fine, I'd rather know exactly what the accepted style is, so we can maintain consistency among contributors

src/stylesheets/core/_variables.scss
```diff
@@ -1142,3 +1142,9 @@ $spacing-small: 1rem;
 $spacing-md-small: 1.5rem;
 $spacing-medium: 2rem;
 $spacing-large: 3rem;
+
+$theme-spacing-x-small: 0.5; // was 0.5rem
```
el-mapache: @thisisdano If we want to use tshirt sizing here how would that look?

perhaps:
```
theme-spacing-xs
theme-spacing-sm
theme-spacing-md
theme-spacing-lg
```

thisisdano: Let's go with this — I have some qualms about imposing a project-level tshirt spacing scale, but let's lay them aside for right now

src/stylesheets/components/_accordions.scss
```diff
@@ -19,24 +15,30 @@ $v-padding: $spacing-md-small;
 // scss-lint:disable PropertyCount
 @mixin accordion-button-styles {
   @include button-unstyled;
-  background-color: $color-gray-lightest;
+  @include u-font-family('sans');
+  @include u-padding(
```
thisisdano: Sorry, the `u-padding()` mixin cant take multiple arguments in this form (yet?) — use a conventional padding rule (or rules) instead:

el-mapache:
@thisisdano updated to remove to `u-padding` include

thisisdano:
OK! Cut and print! Thanks for working through all the back-and-forth on this

PR #2724, opened 2018-09-05
Update skipnav to use 2.0 palette

maya:
- Removed the black text for the link so it uses standard link color.
- Removed outline zero so focus appears

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-skipnav-v2/components/preview/layout--landing.html) - hit <kbd>Tab</kbd>

Part of #2660

### DIFF
--- a/src/stylesheets/components/_skipnav.scss
+++ b/src/stylesheets/components/_skipnav.scss
@@ -1,17 +1,15 @@
 .usa-skipnav {
   background: transparent;
-  color: $color-base;
   left: 0;
-  padding: 1rem 1.5rem;
+  padding: units(1) units(2);
   position: absolute;
-  top: -4.2rem;
+  top: -3.8rem; // skipnav link height
   transition: all 0.2s ease-in-out;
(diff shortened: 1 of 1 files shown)


src/stylesheets/components/_skipnav.scss
```diff
@@ -1,17 +1,15 @@
...
   background: transparent;
-  color: $color-base;
   left: 0;
-  padding: 1rem 1.5rem;
+  padding: units(1) units(2);
   position: absolute;
-  top: -4.2rem;
+  top: -3.8rem; // skipnav link height
   transition: all 0.2s ease-in-out;
   z-index: 100;
```
maya: What's the variables/function for z-index?

thisisdano: There isn't a function for that one yet either

maya: Updated in e6add3b.

PR #2726, opened 2018-09-06
Update hero component to use 2.0 system

el-mapache:
## Description

[Preview here](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/ab-2660-hero/components/detail/hero.html)

### DIFF
--- a/src/stylesheets/components/_hero.scss
+++ b/src/stylesheets/components/_hero.scss
@@ -19,15 +19,15 @@
 }
 
 .usa-hero-callout {
-  background-color: color('primary-darker');
-  padding: units(4);
+  background-color: $color-theme-base-darker;
+  padding: units($theme-spacing-lg);
 
   @include media($small-screen) {
     max-width: 30rem;
   }
 
   > *:first-child {
-    @include margin(0 null unit(2) null);
+    @include margin(0 null unit($theme-spacing-sm) null);
   }
 }
 
--- a/src/stylesheets/components/_hero.scss
+++ b/src/stylesheets/components/_hero.scss
@@ -8,30 +8,30 @@
   background-size: cover;
 
   .usa-button {
-    color: $color-white;
+    color: color('white');
   }
 }
 
 .usa-section-dark {
   .usa-hero-link {
-    color: $color-primary-alt;
+    color: color('accent-cool');
   }
 }
 
 .usa-hero-callout {
-  background-color: $color-primary-darkest;
-  padding: 3rem;
+  background-color: color('primary-darker');
+  padding: units(4);
 
   @include media($small-screen) {
     max-width: 30rem;
   }
 
   > *:first-child {
-    @include margin(0 null $spacing-medium null);
+    @include margin(0 null unit(2) null);
   }
 }
 
 .usa-hero-callout-alt {
-  color: $color-white;
+  color: color('white');
   display: block;
 }


src/stylesheets/components/_hero.scss
```diff
@@ -2,36 +2,36 @@
...
 }
 
 .usa-hero-callout {
-  background-color: $color-primary-darkest;
-  padding: 3rem;
+  background-color: color('base-darker');
+  padding: units(4);
 
-  @include media($small-screen) {
+  @include media(units('mobile-lg')) {
```
el-mapache: Technically `small-screen` covers 481px, and our new `mobile-lg` starts at 480px, so not 100% sure if this is the right sub

thisisdano: It's not 100% exact, but I wonder if it will matter? Also, I recommend using the new mixin for media queries: `@include at-media('mobile-lg')`

maya: Looks like this still needs to be done.

el-mapache: nah, I changed it to use the `at-media` rule

maya: units is still there, it should look like how @thisisdano wrote it above.

src/stylesheets/components/_hero.scss
```diff
@@ -2,36 +2,36 @@
 // ==========================
 
 .usa-hero {
-  @include u-padding-y($site-margins);
+  @include u-padding-y(units(4));
```
thisisdano: Let's use `$theme-site-margins` here

thisisdano:
I think I'm going to need us to pause on this component (and perhaps most other components) until I get through typography and I can make a better suggestion about how to consistently style text.

thisisdano:
If you didn't want to change the markup, you could try adding the following after https://github.com/uswds/uswds/pull/2731 is merged. I _think_ it would work...

```sass
.usa-hero-callout h2 {
  @include typeset-h2;
}
```

maya:
I actually think we should change the markup though so it's not tied to the h2 element / someone may want to use a different heading level.

src/stylesheets/components/_hero.scss
```diff
@@ -2,36 +2,36 @@
...
+  padding: units(4);
 
-  @include media($small-screen) {
+  @include at-media('mobile-lg') {
     max-width: 30rem;
   }
 
   > *:first-child {
-    @include margin(0 null $spacing-medium null);
+    @include margin(0 null units(2) null);
```
maya: We are going to deprecate using Bourbon I believe and this is a Bourbon mixin, so this should probably be written in a 2.0 margin way.

thisisdano:
Yeah, I agree with that

maya:
@thisisdano we need to get that 1em of padding back to the top margin of the button, any idea on the best way to do that?

maya:
Also, probably want to remove the bottom margin of the button / or write it so it's the last element in the box, so that there's even padding all around.

el-mapache:
Sure, I can try that too! The spacing also seems slightly off between the text now, though :(

src/components/hero/hero.njk
```diff
@@ -1,7 +1,7 @@
 <section class="usa-hero">
   <div class="usa-grid">
```
maya: `usa-grid` will be deprecated, we'll want to use the new grid classes and margins to achieve this effect. What do you think is the best way of doing this @thisisdano?

thisisdano:
Dunno — something like the following gets us pretty close?

```sass
.usa-hero {
  .usa-button {
    margin: 0;
  } 

  * + .usa-button {
    margin-top: units(2);
  }
}
```

maya:
Do we want to wait until @thisisdano finishes the button component? Maybe how the margins are applied will change in that?

el-mapache:
@maya yeah that feels better than adding more css if it isn't necessary

maya:
Also, adding `usa-prose` to the box gets the top margins applied to the button correctly, however it adds an underline to the button which we don't want.

src/stylesheets/components/_hero.scss
```diff
@@ -27,7 +27,9 @@
   }
 
   > *:first-child {
-    @include margin(0 null units(2) null);
+    margin-bottom: units(2);
+    margin-top: 0;
+    @include u-margin-x(null);
```
maya: `null` just means "don't do anything" so this line can be removed. It's specific to that Bourbon mixin, I don't even know if it works in ours.

el-mapache: yeah I already removed it, since it was failing my local builds. I just haven't pushed it yet

src/components/hero/hero.njk
```diff
@@ -1,20 +1,22 @@
...
-  <div class="usa-grid">
-    <div class="usa-hero-callout usa-section-dark">
-      <h2>
-        {% if hero.callout %}
-          <span class="usa-hero-callout-alt">{{ hero.callout }}</span>
-        {% endif %}
-        {{ hero.title }}
-      </h2>
+  <div class="grid-container">
+    <div class="grid-row grid-gap-lg grid-gap padding-x-4">
```
maya: Not using columns in this component, remove grid gaps.

maya: padding to be added in a separate pr so can be removed

maya:
Added a few updates, this should be ready for re-review, cc @thisisdano.

maya:
When I added `usa-prose` to the callout, the link styles were applying to the button with an underline. I tried to add a `:not(.usa-button)` to the link styles but this somehow made the basic link styles to strong and override other things like the light link style on dark text, so removed that. Now this adds typesetting to the p text in the callout instead of using `usa-prose` to get the proper spacing.

thisisdano:
Looking at this one now.

maya:
Done in 3275e34.

src/stylesheets/components/_hero.scss
```diff
@@ -2,36 +2,45 @@
...
+
+  p {
+    @include typeset-p;
   }
 }
 
 .usa-section-dark {
   .usa-hero-link {
-    color: $color-primary-alt;
+    color: color('accent-cool');
```
thisisdano: @maya what's this used for? I don't see it in the example

src/stylesheets/components/_hero.scss
```diff
@@ -2,36 +2,45 @@
...
 
   > *:first-child {
-    @include margin(0 null $spacing-medium null);
+    margin-top: 0;
   }
 }
 
 .usa-hero-callout-alt {
-  color: $color-white;
+  color: color('white');
```
thisisdano: What's `usa-hero-callout-alt` used for? I don't see it in the example

maya: The heading has two colors: white and blue:

screen shot 2018-09-12 at 11 38 31 am

src/stylesheets/components/_hero.scss
maya: In case someone wants to add a link. It's not in the example, but may have been at one point.

screen shot 2018-09-12 at 11 38 07 am

src/stylesheets/components/_hero.scss
maya: It's in the code https://github.com/uswds/uswds/pull/2726/files#diff-d55517edab231c0570e713ee03a4d9dbR7

src/stylesheets/components/_hero.scss
maya: I think the default link color for this section is fine though, so removed in 5c1a955.
screen shot 2018-09-12 at 11 42 54 am

src/stylesheets/components/_hero.scss
thisisdano: oh yeah duh 🙃

thisisdano:
OK, made a couple changes of my own. (Sorry @maya I didn't realize you were actively working on this when I started.) The most significant:

- removed `grid-row` and `grid-gap` from markup since they didn't seem to be doing anything?
- added explicit `usa-hero-content` section to enclose the `p`(s) — this should allow more flexible spacing control
- removed `section-dark` from the component, as it wasn't really supporting it well
- tweaked the `theme.scale.xl` size and `system.line-height.2` line height to improve appearance

PR #2727, opened 2018-09-06
Update section component to use 2.0 system

el-mapache:
## Description

[Preview here](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/ab-2660-section/components/detail/layout--docs.html)

### DIFF
--- a/src/stylesheets/components/_section.scss
+++ b/src/stylesheets/components/_section.scss
@@ -1,7 +1,7 @@
 .usa-section {
   @include u-padding-y(units(4));
 
-  @include at-media(units('tablet')) {
+  @include at-media('tablet') {
     @include u-padding-y(units(8));
   }
 }
--- a/src/stylesheets/components/_section.scss
+++ b/src/stylesheets/components/_section.scss
@@ -1,7 +1,7 @@
 .usa-section {
   @include u-padding-y(units(4));
 
-  @include media(units('tablet')) {
+  @include at-media(units('tablet')) {
     @include u-padding-y(units(8));
   }
 }
@@ -11,7 +11,7 @@
 }
 
 .usa-section-dark {
-  background-color: color('base-darkest');
+  background-color: color('primary-darker');
   color: color('white');
 
   h1,
--- a/src/stylesheets/components/_section.scss
+++ b/src/stylesheets/components/_section.scss
@@ -1,17 +1,17 @@
 .usa-section {
-  @include u-padding-y($site-margins);
+  @include u-padding-y(units(4));
 
-  @include media($medium-screen) {
-    @include u-padding-y(units(6));
+  @include media(units('tablet')) {
+    @include u-padding-y(units(8));
   }
 }
 
 .usa-section-light {
-  background-color: $color-theme-base-lightest;
+  background-color: color('base-lightest');
 }
 
 .usa-section-dark {
-  background-color: $color-theme-base-darkest;
+  background-color: color('base-darkest');
   color: color('white');
 
   h1,
@@ -20,7 +20,7 @@
   h4,
   h5,
   h6 {
-    color: $color-primary-alt;
+    color: color('accent-cool');
   }
 
   p {
@@ -28,7 +28,7 @@
   }
 
   a {
-    color: $color-theme-base-lighter;
+    color: color('base-lighter');
 
     &:hover {
       color: color('white');
--- a/src/stylesheets/components/_section.scss
+++ b/src/stylesheets/components/_section.scss
@@ -2,17 +2,17 @@
   @include u-padding-y($site-margins);
 
   @include media($medium-screen) {
-    @include u-padding-y(6rem);
+    @include u-padding-y(units(6));
   }
 }
 
 .usa-section-light {
-  background-color: $color-gray-lightest;
+  background-color: $color-theme-base-lightest;
 }
 
 .usa-section-dark {
-  background-color: $color-primary-darkest;
-  color: $color-white;
+  background-color: $color-theme-base-darkest;
+  color: color('white');
 
   h1,
   h2,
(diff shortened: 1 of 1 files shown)


src/stylesheets/components/_section.scss
```diff
@@ -2,17 +2,17 @@
   @include u-padding-y($site-margins);
 
   @include media($medium-screen) {
-    @include u-padding-y(6rem);
+    @include u-padding-y(units(6));
```
maya: Use the appropriate size listed in the conversion table #2660

src/stylesheets/components/_section.scss
```diff
@@ -2,17 +2,17 @@
...
   }
 }
 
 .usa-section-light {
-  background-color: $color-gray-lightest;
+  background-color: $color-theme-base-lightest;
```
maya: Use the function + token style.

src/stylesheets/components/_section.scss
```diff
@@ -2,17 +2,17 @@
...
 }
 
 .usa-section-dark {
-  background-color: $color-primary-darkest;
-  color: $color-white;
+  background-color: $color-theme-base-darkest;
```
maya: Use the function + token style.

src/stylesheets/components/_section.scss
```diff
@@ -24,14 +24,14 @@
   }
 
   p {
-    color: $color-white;
+    color: color('white');
   }
 
   a {
-    color: $color-gray-lighter;
+    color: $color-theme-base-lighter;
```
maya: Use the function + token style.

src/stylesheets/components/_section.scss
```diff
@@ -2,17 +2,17 @@
   @include u-padding-y($site-margins);
```
maya: Update this to the function + token style using units instead of the variable.

src/stylesheets/components/_section.scss
```diff
@@ -2,17 +2,17 @@
...
 
   @include media($medium-screen) {
```
maya: I think we have new media queries to use in v2.

src/stylesheets/components/_section.scss
```diff
@@ -24,14 +24,14 @@
   }
```
maya: Update to function + token style.

src/stylesheets/components/_section.scss
el-mapache: is that 8, in this case?

el-mapache:
@maya updated, with one more question!

src/stylesheets/components/_section.scss
maya: Yep!

src/stylesheets/components/_section.scss
```diff
@@ -1,37 +1,37 @@
 .usa-section {
-  @include u-padding-y($site-margins);
+  @include u-padding-y(units(4));
```
thisisdano: We have a v2 var for this one.
You can use `u-padding-y($theme-site-margins)` here.

src/stylesheets/components/_section.scss
```diff
@@ -1,37 +1,37 @@
...
 .usa-section-light {
-  background-color: $color-gray-lightest;
+  background-color: color('base-lightest');
 }
 
 .usa-section-dark {
-  background-color: $color-primary-darkest;
-  color: $color-white;
+  background-color: color('primary-darker');
+  color: color('white');
```
thisisdano: Can we can just use `black` or `white` for those two colors instead of a function?

src/stylesheets/components/_section.scss
```diff
@@ -1,37 +1,37 @@
...
   h4,
   h5,
   h6 {
-    color: $color-primary-alt;
+    color: color('accent-cool');
   }
 
   p {
-    color: $color-white;
+    color: color('white');
```
thisisdano: see line 15

src/stylesheets/components/_section.scss
```diff
@@ -1,37 +1,37 @@
...
   }
 
   a {
-    color: $color-gray-lighter;
+    color: color('base-lighter');
 
     &:hover {
-      color: $color-white;
+      color: color('white');
```
thisisdano: see line 15

src/stylesheets/components/_section.scss
maya: This is on me, I recommended the units here bc I don't think there's a correlation between the site margins and the section component. I think it's fine as is.

src/stylesheets/components/_section.scss
maya: We could, but I'm not sold that we should, as this ties it back to the variable we defined as white and black.

thisisdano: I don't feel strongly about this, but I _generally_ feel that the system shouldn't set a definition of white of black that isn't `white` or `black` — that is to say, I'm comfortable that we have white and black defined in the system but i don't think they should ever change from pure white and pure black, so maybe they're redundant but also who knows it's been a day 🤷‍♀️

thisisdano: so let's just leave them as a function, at least for consistency's sake

src/stylesheets/components/_section.scss
```diff
@@ -1,37 +1,37 @@
 .usa-section {
-  @include u-padding-y($site-margins);
+  @include u-padding-y(units(4));
 
-  @include media($medium-screen) {
-    @include u-padding-y(6rem);
+  @include at-media('tablet') {
```
maya: Should be mobile large (mobile-lg)

el-mapache: It looks like `medium-screen` is defined at `600px`, so the closest analog to that seems to be `tablet`, at `640px`. mobile-lg is 480; that seems like a pretty significant difference.

thisisdano: I'm inclined to choose the closer value unless there's a better reason to go down to `mobile-lg`

PR #2728, opened 2018-09-06
Update alerts components to use 2.0 system

el-mapache:
## Description

[Preview here](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/ab-2660-alert/components/detail/alerts--default.html)

The major difference I'm noticing is that the font sizes appear smaller...I guess this is because we are using a different system font?

The other values look pretty accurate to my eye.


src/stylesheets/components/_alerts.scss
```diff
@@ -3,40 +3,40 @@ $usa-custom-alerts: () !default;
...
+  background-size: units(7);
+  padding-bottom: units($theme-spacing-sm);
   padding-left: $left-padding;
   padding-right: $h-padding;
   padding-top: $h-padding;
   position: relative;
 
   &::before {
-    background-color: darken($color-gray-lightest, 40%);
+    background-color: color('gray-40');
```
el-mapache: I ended up going with the specific grade of gray. I don't actually know when this is relevant, since all the alerts are overriding the background color with their own. I guess it might come into play if someone uses a `usa-alert` with no modifier?

thisisdano:
Checking this out now

src/stylesheets/components/_alerts.scss
```diff
@@ -88,14 +88,14 @@ $icon-size: 3rem;
 }
```
thisisdano: Use `@extend %usa-heading;` in `.usa-alert-heading` to explicitly get the heading styles

src/stylesheets/components/_alerts.scss
```diff
@@ -3,40 +3,40 @@ $usa-custom-alerts: () !default;
...
 
 $usa-alerts-bar: (
-  success: $color-green,
-  warning: $color-gold,
-  error: $color-secondary,
-  info: $color-primary-alt
+  success: color('success'),
+  warning: color('warning'),
+  error: color('error'),
+  info: color('accent-cool')
```
thisisdano: I think I will add `info` to the state palette. OK to just do it in this branch?

el-mapache: sure thing!

src/stylesheets/components/_alerts.scss
el-mapache: Is that in lieu of calling margin-top/bottom? Like we want to add all the additional styles (like `line-height`) to this header as well?

thisisdano: yup — it needs font size/face/line-height etc now that these properties aren't being applied via the `h3` element itself

thisisdano: Also, i made this change :)

src/stylesheets/components/_alerts.scss
```diff
@@ -124,5 +125,5 @@ $icon-size: 3rem;
 
 // Deprecated
```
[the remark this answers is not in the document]
el-mapache: No, I was hoping you had some insight 😄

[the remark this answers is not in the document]
el-mapache: the comment was added by marco, for what it's worth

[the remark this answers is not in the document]
thisisdano: I'm gonna remove that rule

PR #2729, opened 2018-09-07
Update inputs to v2 / Associated fixes

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-input-v2/components/detail/address-form.html)

The primary task of this PR was to update the inputs component(s). Which I did! But in doing so, I also ran across a number of bugs and other things that I've tried to address here.

Most of the changes were visually minimal, but a couple are more significant.

**Check boxes are bolder**
This makes them more visually similar to the radio buttons.
checkbox

**Simplified range slider**
Tried to bring the range slider design closer to that of the other input elements
slider

- - -

- Adjusted the colors of the `base` family to bring them more in line with the existing values (and updated the table in https://github.com/uswds/uswds/issues/2660)
- Updated colors in other v2 components to be in line with the new values
- Added missing `scale()` function 😬 
- Added `2em` of x-padding to the preview frame in fractal to improve preview display
- Added missing `u-shadow()` utility mixin
- Added light and dark grades to the `disabled` color family and lightened the default value
- Reduced based font size ('sm') to `5` to get it closer to the v1 size


thisisdano:
@maya Reverted range slider to something closer to the original

thisisdano:
Ah crap, I broke something

thisisdano:
Oh good, I fixed something

maya:
The red error color is looking a little light, we may want to downshift to a darker red to more closely match what we had before. We want are reds a little darker and greens a little lighter so that people with color blindness issues can tell them apart.

src/stylesheets/components/_accordions.scss
```diff
@@ -1,6 +1,7 @@
 // Variables
 
-$accordion-border: units(0.5) solid color('base-lighter');
+$accordion-border: units(0.5) solid color('base-lightest');
+$accordion-font-family: 'sans' !default;
```
maya: is this a pattern that we plan to do on all components?

thisisdano: I'm not really sure yet — but because font size and family are fundamentally coupled in v2 I've been playing with different techniques to make sure each component is accessing the proper value. I think after I muddle through typography I'll have a better attempt at guidance here

thisisdano:
With `red-50` and `green-40v` these are the colorblind variations I see with https://github.com/LeonardoCardoso/Colorblinding

Does this look ok?

colorblind-errors

maya:
No, we should use `red-60v` for the error state so there's more contrast between the `green-cool-40v`. I tested this with someone who has colorblindness who said, "It's easier to see the darker one bc there's a bigger difference." 

screen shot 2018-09-08 at 4 03 03 pm

thisisdano:
OK, updated the error state colors

PR #2730, opened 2018-09-07
Fix v2 accordion style palette and issues

maya:
- Changes the border thickness to a 2.0 palette value (2px)
- Makes the gap between accordion items larger (8px) as discussed in https://github.com/uswds/uswds/pull/2722#discussion_r215414778
- Use the function + shortcode style instead of variables


src/stylesheets/components/_accordions.scss
```diff
@@ -1,12 +1,12 @@
 // Variables
 
-$accordion-border: 3px solid $color-theme-base-lighter;
+$accordion-border: units(2px) solid color('base-lighter');
```
thisisdano: In this instance I think it might look better rounding up to 0.5 units (4px)

maya: Done in 64b23da. The latest Sketch file has it at 1px, so we'll need to update there as well.

PR #2731, opened 2018-09-09
Update typography to v2

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-typograhy-v2/)

- - -

This PR adds a bit of functionality and a little overhead to our component Sass.

Specifically, it removes default styling from the `<body>` element to keep our components as modular as possible — and gives the option to add it back with a feature flag.

```sass
// _settings-typography.scss

// Base settings are the equivalent of setting the <body> element
$theme-font-family-base:           'body' !default;
$theme-font-size-base:             'sm' !default; // was 1.7rem
$theme-line-height-base:           5 !default; // was 1.5

// If true, explicitly style the <body> element with the base styles
$theme-style-body-element:         false !default;
```

Our components should display properly regardless of `<body>` styling. To achieve this, we need to explicitly set typesetting at the component level with the `typeset()` mixin.

```sass
.usa-component {
  @include typeset;
  ...
}

// compiles as

.usa-component {
  font-family: font-family($theme-font-family-base);
  font-size: scale($theme-font-family-base, $theme-font-size-base);
  line-height: line-height($theme-font-family-base, $theme-line-height-base);
}
```

`typeset()` also accepts 1-3 tokens as arguments: `[font-family]`, `[scale]`,  and `[line-height]`. By default, those arguments are set to `$theme-font-family-base`, `$theme-font-size-base`, and `$theme-line-height-base`.  If an argument is explicitly `null`, that argument will get the default. 

```
typeset('sans')        == typeset('sans', $theme-font-size-base, $theme-line-height-base)
typeset('sans', 'xs')  == typeset('sans', 'xs', $theme-line-height-base)
typeset(null, null, 2) == typeset($theme-font-family-base, $theme-font-size-base, 2)
```

We can also set explicit font families for our components in `settings-typography`:

```sass
$theme-font-family-accordion:      'body' !default;
$theme-font-family-alerts:         'ui' !default;
$theme-font-family-banner:         'ui' !default;
$theme-font-family-footer:         'body' !default;
$theme-font-family-input:          'ui' !default;
$theme-font-family-lead:           'heading' !default;
$theme-font-family-prose:          'body' !default;
```

Then use that variable as the family arg in `typeset()`, like `typeset($theme-font-family-banner)`.

_Your component will probably appear very small until you do this!_

- - -

To apply default `p`, `a`, `h1-h6` styles to a class, use the `typeset-[element]` mixin (one exception: use `typeset-link` not `typeset-a`):

```sass
.usa-component-title {
  @include typeset-h3;
}

.usa-component-description {
  @include typeset-p;

  a {
    @include typeset-link;
  }
}

```

- - -

This PR also adds `ui` as one of the role-based font families.
- `ui`
- `heading`
- `body`
- `code`
- `alt`

### DIFF
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -15,7 +15,7 @@ $accordion-border: units(0.5) solid color('base-lightest');
 // scss-lint:disable PropertyCount
 @mixin accordion-button-styles {
   @include button-unstyled;
-  @include typeset-default($theme-accordion-font-family);
+  @include typeset-default($theme-font-family-accordion);
 
   background-color: color('base-lightest');
   background-image: url('#{$theme-image-path}/minus.png');
@@ -61,7 +61,7 @@ $accordion-border: units(0.5) solid color('base-lightest');
 .usa-accordion-bordered {
   @include accordion-list-styles;
   @include accordion-nested-list;
-  @include typeset-default($theme-accordion-font-family);
+  @include typeset-default($theme-font-family-accordion);
 
   + .usa-accordion,
   + .usa-accordion-bordered {
@@ -83,7 +83,7 @@ $accordion-border: units(0.5) solid color('base-lightest');
 }
 
 .usa-accordion-heading {
-  @include typeset($theme-accordion-font-family, $theme-base-font-size, 1);
+  @include typeset($theme-font-family-accordion, $theme-font-size-base, 1);
   margin: 0;
 
   &:not(:first-child) {
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -1,7 +1,6 @@
 // Variables
 
 $accordion-border: units(0.5) solid color('base-lightest');
-$accordion-font-family: 'sans' !default;
 
 // Accordion Styles
 
@@ -16,6 +15,7 @@ $accordion-font-family: 'sans' !default;
 // scss-lint:disable PropertyCount
 @mixin accordion-button-styles {
   @include button-unstyled;
+  @include typeset-default($theme-accordion-font-family);
 
   background-color: color('base-lightest');
   background-image: url('#{$theme-image-path}/minus.png');
@@ -26,7 +26,6 @@ $accordion-font-family: 'sans' !default;
   color: color('ink');
   cursor: pointer;
   display: inline-block;
-  font-family: font-family($accordion-font-family);
   font-weight: font-weight('bold');
   margin: 0;
   padding: units(2) units(2.5) * 2 + units(2) units(2) units(2.5);
@@ -62,6 +61,7 @@ $accordion-font-family: 'sans' !default;
 .usa-accordion-bordered {
   @include accordion-list-styles;
   @include accordion-nested-list;
+  @include typeset-default($theme-accordion-font-family);
 
   + .usa-accordion,
   + .usa-accordion-bordered {
@@ -83,7 +83,7 @@ $accordion-font-family: 'sans' !default;
 }
 
 .usa-accordion-heading {
-  @include typeset($accordion-font-family, $theme-base-font-size, 1);
+  @include typeset($theme-accordion-font-family, $theme-base-font-size, 1);
   margin: 0;
 
   &:not(:first-child) {
--- a/src/stylesheets/components/_accordions.scss
+++ b/src/stylesheets/components/_accordions.scss
@@ -15,7 +15,7 @@ $accordion-border: units(0.5) solid color('base-lightest');
 // scss-lint:disable PropertyCount
 @mixin accordion-button-styles {
   @include button-unstyled;
-  @include typeset-default($theme-font-family-accordion);
+  @include typeset($theme-font-family-accordion);
 
   background-color: color('base-lightest');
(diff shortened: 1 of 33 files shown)


maya:
@thisisdano is this ready for review? If so, do we want to remove the WIP in the issue title?

maya:
Looks like there's other work in this PR that got included (e.g. [forms](https://github.com/uswds/uswds/pull/2731/files#diff-55017277393aab7a35c19af6c95f2602)). Is there any way to remove that from the PR so it's only related to new typography work?

src/stylesheets/components/_accordions.scss
```diff
@@ -62,6 +61,7 @@ $accordion-font-family: 'sans' !default;
 .usa-accordion-bordered {
   @include accordion-list-styles;
   @include accordion-nested-list;
+  @include typeset($theme-font-family-accordion);
```
maya: Is this bc someone would want the font of the bordered accordion to be different from the default borderless accordion?

thisisdano: No, this looks like a mistake

thisisdano: nm, this rule is for `.usa-accordion, .usa-accordion-bordered` so it applies to both. I think I can remove the `typeset` from line 18, though

src/stylesheets/elements/_inputs.scss
```diff
@@ -69,19 +81,21 @@ $input-height: floor($input-height-exact * 10) / 10;
...
   color: color('error-dark');
   display: block;
-  font-size: $input-font-size;
+  font-size: $input-font-size-calc;
   font-weight: $theme-font-weight-bold;
 }
 
 // Deprecated: Some screen readers can't read CSS content.
 // Will be removed in 2.0.
+// TODO: remove?
```
maya: sure

src/stylesheets/elements/_inputs.scss
```diff
@@ -69,19 +81,21 @@ $input-height: floor($input-height-exact * 10) / 10;
...
 .usa-input-required:after {
   color: color('error-dark');
   content: ' (*required)';
 }
 
 // Deprecated: Some screen readers can't read CSS content.
 // Will be removed in 2.0.
+// TODO: remove?
```
maya: sure

thisisdano:
I don't know if I can strip individual threads of work out of this PR — but I'm happy to flag/revert anything that looks suspicious or unusual

src/stylesheets/elements/_typography.scss
```diff
@@ -184,76 +149,81 @@ dfn {
...
-  }
+  max-width: measure($theme-measure-text);
 }
 
 .usa-display {
-  @include h3;
+  @include typeset-h3;
   margin-bottom: 0;
 
   @include media($small-screen) {
```
maya: Can you change all the media queries to use the new 2.0 vars?

src/stylesheets/elements/_typography.scss
```diff
@@ -184,76 +149,81 @@ dfn {
...
 
     &:hover {
-      color: $color-white;
+      color: color('white');
     }
   }
 }
 
+// TODO: How should we keep or adapt these?
+// Strong candidates for deprecation
```
maya: If you can find where this is being used and replace it with something in 2.0, then ok to remove.

src/stylesheets/elements/_typography.scss
```diff
@@ -184,76 +149,81 @@ dfn {
...
+  a,
+  li,
+  span {
+    font-family: $font-sans;
+  }
+
+  // TODO: This is an unexpected effect of .usa-sans
+  a {
+    border-bottom: none;
+    font-weight: $font-bold;
```
maya: Links should not be bold.

thisisdano: This whole rule is strange — why should a link in a sans face have no `border-bottom`? Wh would it have `border-bottom` in the first place? Should we just nuke the whole `a` rule here?

maya: Idk where this came from, guessing vestigial code? If we can remove it, remove it.

src/stylesheets/components/_accordions.scss
thisisdano: Kept this one and removed the one on line 18

thisisdano:
@maya any other fixes on this one?

src/stylesheets/components/_accordions.scss
maya: Oh I see, I think that we should split this up though, to be in line with how we're styling our components, i.e. bordered would need `usa-accordion usa-accordion-bordered` so only the bordered style are added.

maya: But should we do it now or in a different PR?

thisisdano: Maybe a separate PR — there's a similar issue with `usa-form` and `usa-form-large`

maya: Added an issue #2736 to track this work. Add in more as you find them.

maya: Ideally, if we could tackle this in the 2.0 PRs we're working on and look for this, that would be good.

PR #2733, opened 2018-09-10
Update documentation page template v2

maya:
It will look slightly out of sync until header navigation is updated to v2.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-docs-layout/components/preview/layout--docs.html)


maya:
This is not done, but would like a review of the strategy, cc @thisisdano

maya:
On hold til sidenav is complete.

src/stylesheets/components/_layout.scss
```diff
@@ -1,36 +1,15 @@
...
-}
-
 .usa-layout-docs-sidenav {
   order: 2;
-}
-
-.usa-layout-docs-main_content {
-  margin-bottom: $site-margins;
-  order: 1;
+  padding-top: units(6);
```
maya: @thisisdano how much space should we have between the bottom of the page and the nav below?

thisisdano: 4 units seems like a good place to start

maya: Updated in 6e52d90, thanks! Also, this PR is ready to review 👓

PR #2735, opened 2018-09-10
Update graphic list to use 2.0 system palette

maya:
Question: `$site-margins` is set to `units($theme-site-margins)` in variables.scss, but should this component just use `units($theme-site-margins)` instead?

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-graphic-list-v2/components/detail/graphic-list.html)

### DIFF
--- a/src/components/graphic-list/graphic-list.njk
+++ b/src/components/graphic-list/graphic-list.njk
@@ -1,4 +1,4 @@
-<section class="usa-section usa-section-dark usa-graphic_list">
+<section class="usa-section usa-section-dark usa-graphic_list usa-prose">
   <div class="usa-grid usa-graphic_list-row">
     <div class="usa-width-one-half usa-media_block">
       <img class="usa-media_block-img"  src="{{ uswds.path }}/img/circle-124.png" alt="Alt text">
--- a/src/components/graphic-list/graphic-list.njk
+++ b/src/components/graphic-list/graphic-list.njk
@@ -12,7 +12,7 @@
         <img class="usa-media_block-img"  src="{{ uswds.path }}/img/circle-124.png" alt="Alt text">
         <div class="usa-media_block-body">
           <h3 class="usa-graphic_list-heading">Stick to 6 or fewer words.</h3>
-          <p>Keep body text to about 30. They can be shorter, but try to be somewhat balanced across all four. It creates a clean appearance with good spacing.</p>
+          <p>Keep body text to about 30 words. They can be shorter, but try to be somewhat balanced across all four. It creates a clean appearance with good spacing.</p>
         </div>
       </div>
     </div>
--- a/src/components/graphic-list/graphic-list.njk
+++ b/src/components/graphic-list/graphic-list.njk
@@ -4,14 +4,14 @@
       <div class="usa-media_block tablet:grid-col">
         <img class="usa-media_block-img"  src="{{ uswds.path }}/img/circle-124.png" alt="Alt text">
         <div class="usa-media_block-body">
-          <h3 class="font-heading-lg">Graphic headings can vary.</h3>
+          <h3 class="usa-graphic_list-heading">Graphic headings can vary.</h3>
           <p>Graphic headings can be used a few different ways, depending on what your landing page is for. Highlight your values, specific program areas, or results.</p>
         </div>
       </div>
       <div class="usa-media_block tablet:grid-col">
         <img class="usa-media_block-img"  src="{{ uswds.path }}/img/circle-124.png" alt="Alt text">
         <div class="usa-media_block-body">
-          <h3 class="font-heading-lg">Stick to 6 or fewer words.</h3>
+          <h3 class="usa-graphic_list-heading">Stick to 6 or fewer words.</h3>
           <p>Keep body text to about 30. They can be shorter, but try to be somewhat balanced across all four. It creates a clean appearance with good spacing.</p>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


thisisdano:
I would say yes: use the v2 syntax and var — I think the older way was a kludge

maya:
I reverted the `usa-prose` addition to the section and added `font-heading-lg` to the h3s in 31cf5b2 to keep it more specific.

maya:
@thisisdano updated to the 2.0 grid in d17fd86, but it's not working. Also, after merging the latest work, the font size is 10px. Can you take a look?

maya:
Fixed grid in c33bff2.

src/components/graphic-list/graphic-list.njk
```diff
@@ -1,33 +1,35 @@
...
-    <div class="usa-width-one-half usa-media_block">
-      <img class="usa-media_block-img"  src="{{ uswds.path }}/img/circle-124.png" alt="Alt text">
-      <div class="usa-media_block-body">
-        <h3>Stick to 6 or fewer words.</h3>
-        <p>Keep body text to about 30. They can be shorter, but try to be somewhat balanced across all four. It creates a clean appearance with good spacing.</p>
+      <div class="usa-media_block tablet:grid-col">
+        <img class="usa-media_block-img"  src="{{ uswds.path }}/img/circle-124.png" alt="Alt text">
+        <div class="usa-media_block-body">
+          <h3 class="usa-graphic_list-heading">Stick to 6 or fewer words.</h3>
+          <p>Keep body text to about 30. They can be shorter, but try to be somewhat balanced across all four. It creates a clean appearance with good spacing.</p>
```
thisisdano: 30 _words_?

src/stylesheets/components/_graphic-list.scss
```diff
@@ -1,18 +1,20 @@
...
-        margin-bottom: $site-margins * 2;
+      @include at-media(tablet) {
+        margin-bottom: units($theme-site-margins) * 2;
       }
     }
 
     &:last-child {
       .usa-media_block {
-        @include media($medium-screen) {
+        @include at-media(tablet) {
```
thisisdano: Quote the strings in functions and mixins: `@include at-media('tablet')`
Though the fns and mixins _will_ work without them...

src/stylesheets/components/_graphic-list.scss
```diff
@@ -1,18 +1,20 @@
 // A collection of media block objects
 
 .usa-graphic_list {
+  @include typeset;
+
   .usa-graphic_list-row {
     .usa-media_block {
-      margin-bottom: $site-margins;
+      margin-bottom: units($theme-site-margins);
```
thisisdano: I wonder about using `$theme-site-margins` as a bottom margin in these. Is it important that if the theme site margins change that this changes with it?

thisisdano:
I don't seem to be seeing a grid gap here. (see the two sections on the bottom row.)

screen shot 2018-09-12 at 12 40 51 pm

thisisdano:
When there is a gap between sections, should that gap be larger than the gap between the image and its text?

src/components/graphic-list/graphic-list.njk
maya: Done in 12127cd.

src/stylesheets/components/_graphic-list.scss
maya: That's interesting, I was following the documentation where they are not quoted: https://v2.designsystem.digital.gov/utilities/layout-grid/#example-usage also I have not been quoting any media queries in any components by following this example.

src/stylesheets/components/_graphic-list.scss
maya: This is similar to the section component where I asked the same thing. You recommended using that variable in a similar situation: https://github.com/uswds/uswds/pull/2727#discussion_r216104347

maya:
Added grid gap in e5ad6ba.

maya:
Not sure which gap you're referring to, feel free to add an image to show what you mean. But the spacing looks pretty balanced to me.

maya:
Should the default layout grid `grid-gap` be lg above tablet instead of desktop? This will give the component a larger padding at smaller sizes and more consistent with `grid-container` breakpoints.

src/stylesheets/components/_graphic-list.scss
thisisdano: Oops! Docs is wrong :(

src/stylesheets/components/_graphic-list.scss
thisisdano: I was wrong there :( You were right and I wasn't looking closely enough.
_Twice in a row!_

thisisdano:
> Should the default layout grid grid-gap be lg above tablet instead of desktop? This will give the component a larger padding at smaller sizes and more consistent with grid-container breakpoints.

Could be. Seems reasonable.

> Not sure which gap you're referring to, feel free to add an image to show what you mean. But the spacing looks pretty balanced to me.

I'd make the gap between the circle and the text 3 units instead of 4, so it's slightly narrower than the gap between the columns of media blocks. Like a gestalt grouping principle to keep like items closer together. It's minor, though!

maya:
Smaller margins added in e8f31bf.

src/stylesheets/components/_graphic-list.scss
maya: Updated to unit values in 1688ff6, but we will also need to update that file.

src/stylesheets/components/_graphic-list.scss
maya: Updated to unit values in 1688ff6, but we will also need to update that file.

src/stylesheets/components/_graphic-list.scss
maya: Updated to use quoted text in 185b7d8, but we will also need to go through all the components and update... not to mention updating the docs 😉

maya:
@thisisdano any other changes needed to this one?

thisisdano:
lgtm

PR #2737, opened 2018-09-10
Update buttons to v2

thisisdano:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-buttons-v2/components/detail/buttons--default.html)

### DIFF
--- a/src/stylesheets/elements/_buttons.scss
+++ b/src/stylesheets/elements/_buttons.scss
@@ -7,7 +7,7 @@ $button-stroke: inset 0 0 0 2px;
 // This mixin is only needed until we fully remove the
 // deprecated usa-button-disabled style.
 @mixin disabledesque {
-  background-color: $color-gray-lighter;
+  background-color: color('disabled-light');
   pointer-events: none;
 
   &:hover,
@@ -16,7 +16,7 @@ $button-stroke: inset 0 0 0 2px;
   &.usa-button-active,
   &:focus,
   &.usa-focus {
-    background-color: $color-gray-lighter;
+    background-color: color('disabled-light');
     border: 0;
     box-shadow: none;
   }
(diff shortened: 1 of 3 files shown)


src/stylesheets/elements/_buttons.scss
```diff
@@ -28,19 +28,18 @@ $button-stroke: inset 0 0 0 2px;
 .usa-button:visited,
 .usa-button-primary:visited {
   @include font-smoothing;
-  @include margin(0.5em 0.5em 0.5em null);
+  @include typeset($theme-font-family-button, null, 1);
+  @include u-margin-y(0.5)
```
maya: Since we're using a top margin model for paragraph text, can this be revisited so it doesn not have a bottom margin?

thisisdano: yeah, I'll try that

src/stylesheets/elements/_buttons.scss
```diff
@@ -172,38 +172,38 @@ $button-stroke: inset 0 0 0 2px;
...
+  color: color('base');
+  box-shadow: $button-stroke color('base');
 }
 
 @mixin button-unstyled {
   background-color: transparent;
   border: 0;
   border-radius: 0;
-  font-weight: $font-normal;
+  font-weight: $theme-font-weight-normal;
```
maya: Use function `font-weight('normal');`?

thisisdano: yeah, that's probably a better pattern

src/stylesheets/elements/_buttons.scss
```diff
@@ -7,7 +7,7 @@ $button-stroke: inset 0 0 0 2px;
 // This mixin is only needed until we fully remove the
 // deprecated usa-button-disabled style.
 @mixin disabledesque {
-  background-color: $color-gray-lighter;
+  background-color: color('disabled-light');
```
maya: I don't even know if disabled buttons are a good UX pattern or not, but I can barely read these. While I can read the existing ones better now. I know that these don't have to pass WCAG.

PR #2739, opened 2018-09-11
Remove redundant rules, re-color inverse buttons

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-buttons-v2/components/detail/buttons--secondary-inverse.html)

This cleans up button styles a bit more to remove redundant rules and improve the performance of inverse buttons:

- remove a couple duplicative/redundant rules
- makes the hover state the higher contrast state (like with regular buttons)
- makes the disabled state less bright


thisisdano:
Yeah, I didn't know — their deprecation was news to me!

maya:
Give it a look and see if they should even be deprecated or not.

PR #2740, opened 2018-09-11
Update sidenav v2

maya:
Part of #2660. 

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-sidenav-v2/components/preview/sidenav--compare.html)

### DIFF
--- a/src/stylesheets/components/_sidenav.scss
+++ b/src/stylesheets/components/_sidenav.scss
@@ -8,7 +8,7 @@
     border-top: none;
   }
 
-  .usa-grid & {
+  .grid-container & {
     @include u-margin-x(-$site-margins-mobile);
 
     @include at-media(tablet) {
--- a/src/stylesheets/components/_sidenav.scss
+++ b/src/stylesheets/components/_sidenav.scss
@@ -1,9 +1,9 @@
 .usa-sidenav-list {
   @include usa-sidenav-list;
-  border-bottom: 1px solid;
-  border-top: 1px solid;
+  border-bottom: units(1px) solid;
+  border-top: units(1px) solid;
 
-  @include media($large-screen) {
+  @include at-media(desktop) {
     border-bottom: none;
     border-top: none;
   }
@@ -11,7 +11,7 @@
   .usa-grid & {
     @include u-margin-x(-$site-margins-mobile);
 
-    @include media($medium-screen) {
+    @include at-media(tablet) {
       @include u-margin-x(0);
     }
   }
@@ -22,9 +22,9 @@
 }
 
 .usa-layout-docs-sidenav {
-  padding-top: 2.4rem;
+  padding-top: untis(3);
 
-  @include media($large-screen) {
+  @include at-media(desktop) {
     padding-top: 0;
   }
 }
--- a/src/stylesheets/components/_sidenav.scss
+++ b/src/stylesheets/components/_sidenav.scss
@@ -22,7 +22,7 @@
 }
 
 .usa-layout-docs-sidenav {
-  padding-top: untis(3);
+  padding-top: units(3);
 
   @include at-media(desktop) {
     padding-top: 0;
--- a/src/stylesheets/core/_variables.scss
+++ b/src/stylesheets/core/_variables.scss
@@ -1208,7 +1208,7 @@ $box-shadow:                    0 0 2px $color-shadow !default;
 $focus-outline:                 units(0.5) solid $color-focus !default;
 $focus-spacing:                 0 !default;
 $nav-width:                     951px !default;
-$sidenav-current-border-width:  0.4rem !default; // must be in rem for math
+$sidenav-current-border-width:  units(0.5) !default; // must be in rem for math
 $usa-btn-small-width:           4.5rem;
 $search-min-width:              27ch; // minimum length of 27 letters for typical search query
 
--- a/src/stylesheets/core/_variables.scss
+++ b/src/stylesheets/core/_variables.scss
@@ -1208,7 +1208,6 @@ $box-shadow:                    0 0 2px $color-shadow !default;
 $focus-outline:                 units(0.5) solid $color-focus !default;
 $focus-spacing:                 0 !default;
 $nav-width:                     951px !default;
-$sidenav-current-border-width:  units(0.5) !default; // must be in rem for math
 $usa-btn-small-width:           4.5rem;
 $search-min-width:              27ch; // minimum length of 27 letters for typical search query
 
--- a/src/stylesheets/core/mixins/_sidenav.scss
+++ b/src/stylesheets/core/mixins/_sidenav.scss
@@ -36,10 +36,10 @@ $sidenav-grandchild-padding-left: units($theme-site-margins-mobile) + units(1.5)
   }
 
   .usa-current {
-    border-left: $sidenav-current-border-width solid color('primary');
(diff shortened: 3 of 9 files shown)


src/stylesheets/core/mixins/_sidenav.scss
```diff
@@ -3,8 +3,8 @@
 
   > li {
     background-color: transparent;
-    border-top: 1px solid $color-gray;
-    font-size: $h4-font-size;
+    border-top: units(1px) solid color('base-dark');
```
thisisdano: I would say we probably don't _need_ to call `units()` on 1px and 2px. But maybe it's just better to be consistent and use it every time?

src/stylesheets/core/mixins/_sidenav.scss
```diff
@@ -3,8 +3,8 @@
...
+    font-size: scale('body', 5);
```
thisisdano: I _think_ we should try to use the theme tokens as much as possible: `'sm'` in this case

src/stylesheets/core/mixins/_sidenav.scss
```diff
@@ -13,65 +13,67 @@
...
+  $sidenav-child-padding-left: units($theme-site-margins-mobile) + units(1.5);
+  $sidenav-grandchild-padding-left: units($theme-site-margins-mobile) + units(1.5) * 2;
   @include unstyled-list();
   margin: 0;
   width: 100%;
 
   li {
     border: none;
-    font-size: $h5-font-size;
+    font-size: scale('body', 3);
```
thisisdano: Use the theme token `'2xs'`

src/stylesheets/components/_sidenav.scss
```diff
@@ -22,9 +22,9 @@
 }
 
 .usa-layout-docs-sidenav {
-  padding-top: 2.4rem;
+  padding-top: untis(3);
```
thisisdano: typo: `units`

src/stylesheets/core/mixins/_sidenav.scss
maya: Yeah that was my assumption, it's more verbose but at least it's consistent.

src/stylesheets/components/_sidenav.scss
```diff
@@ -1,17 +1,17 @@
 .usa-sidenav-list {
   @include usa-sidenav-list;
```
thisisdano: I feel like we should something like the following to `settings-typography`

```
$theme-font-family-sidenav:         'ui' !default;
```

...then typeset this component with `@include typeset($theme-font-family-sidenav, null, 3)`. This could take away the need to add explicit line height deeper in the component? It also isolates the sidenav from the body font choice, which I think could be useful.

Then use `$theme-font-family-sidenav` in subsequent font functions instead of `'body'`

src/stylesheets/core/mixins/_sidenav.scss
maya: Done in 8243f73.

src/stylesheets/core/mixins/_sidenav.scss
maya: Done in 8243f73.

src/stylesheets/components/_sidenav.scss
maya: Done in 8243f73.

src/stylesheets/core/_variables.scss
```diff
@@ -1208,7 +1208,7 @@ $box-shadow:                    0 0 2px $color-shadow !default;
 $focus-outline:                 units(0.5) solid $color-focus !default;
 $focus-spacing:                 0 !default;
 $nav-width:                     951px !default;
-$sidenav-current-border-width:  0.4rem !default; // must be in rem for math
+$sidenav-current-border-width:  units(0.5) !default; // must be in rem for math
```
thisisdano: If there are user-settable vars, we should migrate them to the appropriate settings file. I see `variables` as the home for mostly non-user-settable vars.

src/stylesheets/core/mixins/_sidenav.scss
```diff
@@ -13,65 +13,67 @@
...
     }
 
     &:focus {
       outline-offset: 0;
       position: relative;
-      z-index: 1;
+      z-index: z-index(100);
     }
 
     &.usa-current { /* stylelint-disable-line selector-no-qualifying-type */
```
thisisdano: Does this need the `a` specificity to work properly?

src/stylesheets/components/_sidenav.scss
maya: Done in f042eab.

src/stylesheets/core/mixins/_sidenav.scss
```diff
@@ -13,65 +13,67 @@
...
   }
 
   a {
-    padding-left: 2.8rem;
-    line-height: $heading-line-height;
+    padding-left: $sidenav-child-padding-left;
+    line-height: line-height($theme-font-family-base, 3);
 
     &:hover,
     &.usa-current { /* stylelint-disable-line selector-no-qualifying-type */
```
thisisdano: Does this need the `a` specificity to work properly?

src/stylesheets/core/mixins/_sidenav.scss
thisisdano: we'll probably want to do a find/replace at some point to make everything consistent

src/stylesheets/core/mixins/_sidenav.scss
maya: I'm honestly not sure what this is for. What comes to mind is maybe the mobile nav? Removed in 90c4912 and hopefully everything still works as expected.

maya: Woops, thought this was about the z-index above, well the z-index is still removed now 😂.

thisisdano: Probably for the best! :)

maya: Removed in e5047d1.

src/stylesheets/core/mixins/_sidenav.scss
maya: Removed in e5047d1. But I think it was just easier to write it this way.

src/stylesheets/core/_variables.scss
maya: Should this variable be moved to another file? If so, which file?

thisisdano: maybe `settings-spacing`?

maya: This is a component variable, not a general spacing variable. I wonder if it makes sense to live in the sidenav component file? Or to have a settings file with component vars that we pull out?

maya: Done in 60590cb. 

@thisisdano btw the change from `0.4rem` to `units(0.5)` is causing an issue in this code:
```
padding-left: units($theme-site-margins-mobile) - $sidenav-current-border-width;
```

It's outputting to:
```
padding-left: 1.6rem-units(.5);
```
Do you know how to get this to work?

maya: Fixed in 54fbd8e.

PR #2741, opened 2018-09-11
Update footer to v2

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-footer-v2/components/detail/footer--default.html)

There are some relatively small (I think) differences in appearance between v2 and v1. 

- `footer-big@mobile-lg` expands the primary nav into two columns and keeps the form in a separate row.
- `footer-big@tablet` should be back the the same as v1
- The js that appends `hidden` to collapsible primary nav elements deactivates at a different breakpoint than `mobile-lg` — @el-mapache can we use the v2 breakpoint here? (right now fixed with a sass kludge)
- Obviously (perhaps) rewriting the component with v2 grid means this is a big breaking change for the component as well

- - -

On the plus side, I was able to cut about 35% of the initial code!

### DIFF
--- a/src/components/10-footer/footer--big.njk
+++ b/src/components/10-footer/footer--big.njk
@@ -1,89 +1,123 @@
 <footer class="usa-footer usa-footer-big" role="contentinfo">
-  <div class="usa-grid usa-footer-return-to-top">
+  <div class="grid-container usa-footer-return-to-top">
     <a href="#">Return to top</a>
   </div>
   <div class="usa-footer-primary-section">
-    <div class="usa-grid">
-      <nav class="usa-footer-nav usa-width-two-thirds">
-        <section class="usa-width-one-fourth usa-footer-primary-content usa-footer-collapsible">
-          <h4 class="usa-footer-primary-link">Topic</h4>
-          <ul class="usa-unstyled-list">
-            <li class="usa-footer-secondary-link"><a href="javascript:void(0);">Secondary link</a></li>
-            <li class="usa-footer-secondary-link"><a href="javascript:void(0);">Secondary link</a></li>
-            <li class="usa-footer-secondary-link"><a href="javascript:void(0);">Secondary link that's a bit longer than most of the others</a></li>
-            <li class="usa-footer-secondary-link"><a href="javascript:void(0);">Secondary link</a></li>
-          </ul>
-        </section>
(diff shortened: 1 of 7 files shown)


thisisdano:
Cut 200 lines from `footer.scss` 😬

maya:
Should links look like links (underlined)?

stphnwlkr:
@maya Not trying to interrupt your workflow, but as it relates to the UWDS footer I would say no underlines. My rational, this is a NAV area and not mixed content.  VA's new preview site uses underline which I think are needed on the dark background, but it really adds to the noise in the footer (my opinion).

thisisdano:
What do you all think about underline-on-hover? Previously there was no underline at all, but I added it back to hover for clarity.

stphnwlkr:
I think underlining on hover or some other effect (i.e. background change) is ideal. I just created a new footer with no hover effect (color change too subtle) and I do not like it.

maya:
That's probably what the original designers had in mind re: navigation. Agree, underline on hover is probably a good call. 

As an enhancement, I wonder if there's a way to scope that to the navigation links. I remember making a custom footer for our main site and having to re-add an underline for something that I thought should be a default was odd.

screen shot 2018-09-12 at 11 12 03 am

maya:
I don't think we want the hover underline on the topic on large screen sizes since it only activates on small screens:
screen shot 2018-09-12 at 11 14 42 am

thisisdano:
Current no-underline scope:

```sass
.usa-footer-primary-link a,
.usa-footer-secondary-link a
```

Does that seem sufficient?

thisisdano:
Fixed the hover underline. You will still see it (erroneously) at some points just wider than `mobile-lg` because there's still a `.hidden` class applied. This will go away when we fix the js to stop applying  `hidden` at the appropriate (mobile-lg) breakpoint width.

maya:
@thisisdano how about we fix the JS right now, it's using 600px do you want it to be 480px? See: https://github.com/uswds/uswds/blob/dw-footer-v2/src/js/components/footer.js#L13

thisisdano:
That's probably the right fix rn, but it brings up a larger Q about using these values in js — since the explicit px value won't work properly with the em-based mqs triggered by `$respect-user-font-settings` — but this is not something to figure out in this PR

thisisdano:
Fixed js and removed the css kludge

PR #2743, opened 2018-09-13
Update forms to v2

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-forms-v2/components/detail/address-form.html)


src/stylesheets/uswds.scss
```diff
@@ -48,6 +48,7 @@ Output all USWDS
 @import 'components/accordions';
 @import 'components/alerts';
 @import 'components/banner';
+@import 'components/checklist';
```
maya: Isn't a checklist an input like radio buttons?

src/stylesheets/elements/_inputs.scss
```diff
@@ -1,7 +1,7 @@
 // scss-lint:disable QualifyingElement, PropertyCount
 
-$input-font-size-calc: scale($theme-font-family-input, $theme-font-size-base);
-$input-line-height-calc: lh($theme-font-family-input, $theme-line-height-input); // was 1.3
+$input-font-size-calc: scale($theme-font-family-forms, $theme-font-size-base);
+$input-line-height-calc: lh($theme-font-family-forms, $theme-line-height-input); // was 1.3
```
maya: I have been using verbose so it's clearer to people reading the code, let's save the shortened ones for prototyping.

src/stylesheets/uswds.scss
thisisdano: Hm, not in this case — this is a component (in v1 as well) called `.usa-checklist` which is essentially a validation checklist of "things you need to do". Definitely open to new names for it, though.

See:
https://components.designsystem.digital.gov/components/preview/validation.html
and
https://github.com/uswds/uswds/blob/e9249a0fb47733a065a35937eda06e5a47d00f22/src/stylesheets/components/_forms.scss#L113

thisisdano: It felt like it _could_ be used in more cases than just in a forms alert, though, which is why I think it should be abstracted into its own component

maya: Oh, I thought this was checkbox and didn't see the connection with validation.

src/stylesheets/components/_alerts.scss
```diff
@@ -41,6 +41,7 @@ $icon-size: units(4);
     margin-top: units(2);
   }
 
+  // TODO: why is this not simply a border?
```
maya: it's so you can have a plain alert with no border, try it with `usa-alert`.

thisisdano: Even `usa-alert` has a gray border. If there's a borderless alert, I can't track it down.  And even if there was one, the benefit of this technique (better text alignment between bordered and non-bordered alerts?) seems pretty minimal for the extra css required for an absolutely-positioned bar. 🤷‍♀️

thisisdano: Not saying this is a first-priority change by any means — just something I noticed

thisisdano:
It doesn't bother me, but I could probably give it 2px — something where the resulting space is a little less than that between the label and the input

src/stylesheets/components/_forms.scss
```diff
@@ -1,144 +1,49 @@
...
-  input {
-    margin-bottom: 3rem;
-  }
+  .usa-button {
+    margin-top: units(1);
 
-  select {
-    margin-bottom: 3rem;
+    @include at-media('mobile-lg') {
+      margin-top: units(2);
```
maya: `units(3)` to be the same as between input and the label?

maya:
if you use float right instead of text align on the show pw it will work.

src/stylesheets/components/_forms.scss
thisisdano: Done

maya:
@thisisdano looks like there's a conflict. Can you resolve?

PR #2745, opened 2018-09-13
Update search v2

maya:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-search-v2/components/detail/search--default.html)

Refactored, used flex instead of calc, ensure markup order is correct.

### DIFF
--- a/src/components/search/search.njk
+++ b/src/components/search/search.njk
@@ -2,7 +2,7 @@
 
 <form class="usa-search usa-search-big">
   <div role="search">
-    <label class="usa-sr-only" for="search-field-big">Search big</label>
+    <label class="usa-sr-only" for="search-field-big">Search</label>
     <input class="usa-input" id="search-field-big" type="search" name="search">
     <button class="usa-button" type="submit">
       <span class="usa-search-submit-text">Search</span>
@@ -14,7 +14,7 @@
 
 <form class="usa-search">
   <div role="search">
-    <label class="usa-sr-only" for="search-field">Search medium</label>
+    <label class="usa-sr-only" for="search-field">Search</label>
     <input class="usa-input" id="search-field" type="search" name="search">
     <button class="usa-button" type="submit">
       <span class="usa-search-submit-text">Search</span>
@@ -26,7 +26,7 @@
 
 <form class="usa-search usa-search-small">
   <div role="search">
-    <label class="usa-sr-only" for="search-field-small">Search small</label>
+    <label class="usa-sr-only" for="search-field-small">Search</label>
     <input class="usa-input" id="search-field-small" type="search" name="search">
     <button class="usa-button" type="submit">
       <span class="usa-sr-only">Search</span>
--- a/src/components/search/search.njk
+++ b/src/components/search/search.njk
@@ -1,35 +1,35 @@
 <h6>Search big</h6>
 
-<div role="search">
-  <form class="usa-search usa-search-big">
+<form class="usa-search usa-search-big">
+  <div role="search">
     <label class="usa-sr-only" for="search-field-big">Search big</label>
     <input class="usa-input" id="search-field-big" type="search" name="search">
     <button class="usa-button" type="submit">
       <span class="usa-search-submit-text">Search</span>
     </button>
-  </form>
-</div>
+  </div>
+</form>
 
 <h6>Search medium</h6>
 
-<div role="search">
-  <form class="usa-search">
+<form class="usa-search">
+  <div role="search">
     <label class="usa-sr-only" for="search-field">Search medium</label>
     <input class="usa-input" id="search-field" type="search" name="search">
     <button class="usa-button" type="submit">
       <span class="usa-search-submit-text">Search</span>
     </button>
-  </form>
-</div>
+  </div>
+</form>
 
 <h6>Search small</h6>
 
-<div role="search">
-  <form class="usa-search usa-search-small">
+<form class="usa-search usa-search-small">
+  <div role="search">
     <label class="usa-sr-only" for="search-field-small">Search small</label>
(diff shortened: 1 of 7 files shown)


src/stylesheets/components/_search.scss
```diff
@@ -11,69 +11,68 @@ $usa-btn-big-width:     11.6rem;
...
-  max-width: none;
+  @include typeset;
   position: relative;
 
+  [role=search] {
+    display: flex;
+  }
+
   [type=submit],
   .usa-search-submit {
```
thisisdano: Do we actually use `.usa-search-submit`? I don't see it.

src/stylesheets/components/_search.scss
```diff
@@ -11,69 +11,68 @@ $usa-btn-big-width:     11.6rem;
...
   @include clearfix;
-  max-width: none;
+  @include typeset;
   position: relative;
 
+  [role=search] {
+    display: flex;
+  }
+
   [type=submit],
```
thisisdano: Should we use `.usa-button` here instead?

src/components/search/search.njk
```diff
@@ -1,35 +1,35 @@
 <h6>Search big</h6>
 
-<div role="search">
-  <form class="usa-search usa-search-big">
+<form class="usa-search usa-search-big">
```
thisisdano: would it make sense to include the `role` with the `form` element, remove the inner `div`, and consolidate the styles?

thisisdano: Are we gaining something important by keeping them separate?

src/stylesheets/components/_search.scss
```diff
@@ -11,69 +11,68 @@ $usa-btn-big-width:     11.6rem;
...
+  $width: $usa-btn-big-width;
 
-      [type=search],
-      .usa-search-input {
-        font-size: $base-font-size;
-        height: $height;
-        width: calc(100% - #{$width});
-      }
+  @include at-media(mobile-lg) {
+    [type=search],
```
thisisdano: See comments on lines 17 and 18

src/stylesheets/components/_search.scss
```diff
@@ -11,69 +11,68 @@ $usa-btn-big-width:     11.6rem;
...
+      height: $height;
+    }
 
-      [type=submit],
-      .usa-search-submit {
-        font-size: $h3-font-size;
-        height: $height;
-        width: $width;
-      }
+    [type=submit],
```
thisisdano: See comments on lines 17 and 18

src/stylesheets/components/_search.scss
```diff
@@ -11,69 +11,68 @@ $usa-btn-big-width:     11.6rem;
...
+  $width: $usa-btn-small-width;
+  $lt-nav-width: rem-to-px($nav-width) - 1px;
 
-      [type=submit],
-      .usa-search-submit {
-        @include search-icon;
-        width: $width;
-      }
+  @include at-media(mobile-lg) {
+    [type=submit],
```
thisisdano: See comments on lines 17 and 18

src/components/search/search.njk
```diff
@@ -1,35 +1,35 @@
 <h6>Search big</h6>
 
-<div role="search">
-  <form class="usa-search usa-search-big">
-    <label class="usa-sr-only" for="search-field-big">Search big</label>
+<form class="usa-search usa-search-big">
+  <div role="search">
```
thisisdano: would it make sense to include the `role` with the `form` element, remove the inner `div`, and consolidate the styles? Do we gain some important isolation/control by keeping them separate?

src/stylesheets/components/_search.scss
```diff
@@ -11,69 +7,59 @@ $usa-btn-big-width:     11.6rem;
 
 .usa-search {
   @include clearfix;
-  max-width: none;
+  @include typeset;
```
thisisdano: Since you have the new settings var, use `typeset($theme-font-family-search)` here instead

src/components/search/search.njk
maya: No, it's for accessibility reasons, form has an inherent role. http://www.a11ymatters.com/pattern/accessible-search/

src/stylesheets/components/_search.scss
maya: Done in 355a34e.

src/components/search/search.njk
thisisdano: Gotcha. Can/should any of the rules tied to `[role=search]` be applied to `usa-search` instead?

src/stylesheets/components/_search.scss
maya: OK removed in 03c8a43, maybe we thought it would be good to have both a class and this at one point.

src/stylesheets/components/_search.scss
maya: I think it enforces good markup so don't think so.

src/components/search/search.njk
maya: Yeah for a11y reasons http://www.a11ymatters.com/pattern/accessible-search/

src/components/search/search.njk
maya: No bc `usa-search` is the form element and `[role=search]` is the inner div and they have different style needs.

src/stylesheets/components/_search.scss
maya: Answered above.

src/stylesheets/components/_search.scss
maya: Answered above.

src/stylesheets/components/_search.scss
maya: Answered above.

PR #2746, opened 2018-09-13
Update headers v2

maya:
There will be a little wonkiness with the media queries until nav is done.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-header-v2/components/preview/header--default.html)

### DIFF
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -158,10 +158,10 @@ $z-index-nav:     z-index(500);
 
   .usa-navbar {
     @include at-media($theme-nav-width) {
-      @include outer-container($site-max-width); // TODO: update
-      @include u-padding-x($theme-site-margins);
+      @include grid-container;
       display: block;
       height: auto;
+      overflow: auto;
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/stylesheets/components/_header.scss
```diff
@@ -142,83 +138,83 @@ $z-index-nav:     9000;
...
+      line-height: line-height($theme-font-family-header, 1);
     }
   }
 
   .usa-navbar {
-    @include media($nav-width) {
-      @include outer-container($site-max-width);
-      @include u-padding-x($site-margins);
+    @include at-media($theme-nav-width) {
+      @include outer-container($site-max-width); // TODO: update
```
maya: Tried to use `@include grid-container` instead of this, but it caused issues that I couldn't identify, cc @thisisdano.

thisisdano: I'll try to replicate and see what's up

thisisdano: That's unusual. It's not just `grid-container` that triggers this behavior, but using neither and just adding the padding and max width manually also results in wonky output.

thisisdano: Adding `overflow: auto;` helps — is there some float clearing thing happening?

maya: there's a clearfix on this https://github.com/uswds/uswds/pull/2746/files#diff-70c90cf98b07e2f38faec5eb1e9fa4e3R188

maya: Done in 30937b3. Had to add `!important` to change the grid's padding tho.

thisisdano:
Gonna merge this — though there may be some further adjustments as I work through the nav component

Issue #1729, opened 2017-02-22
.usa-logo-text only supports short titles


phae:
It seems that the pattern for headers doesn't work for text over a certain length.  As soon as the words are too long, the title pops off the top of the layout (or up against the banner if that's turned on, as seen in your docs). 

See screen shot.

<img width="554" alt="screenshot 2017-02-22 14 34 35" src="https://cloud.githubusercontent.com/assets/44568/23235911/1c715c14-f90c-11e6-89f8-4592c1b9200f.png">
<img width="839" alt="screenshot 2017-02-22 14 34 08" src="https://cloud.githubusercontent.com/assets/44568/23235913/1d8ced8e-f90c-11e6-93e5-9104886f6f4b.png">

SSs are Chrome 56.0.2924.87 OSX 10.10.5

juliaelman:
Thank you for your interest in the Draft U.S. Web Design Standards! We have received notification of your issue and will add it to our backlog for future feature consideration.

In the meantime, you can solve this issue by overriding the `.usa-navbar` styles. It looks like you are using the Basic Header, is that accurate?

Also, I am not sure if you are using Sass or CSS, but here are some solves I tried locally that seem to resolve the issue (see screenshot below FYR):

**Sass**
```
// path/to/my/scss/files/main/scss/my-custom_vars.scss

.usa-header-basic
.usa-navbar {
  @include media($nav-width) {
    width: 25%;
  }
}
```

**CSS**
```
@media screen and (min-width: 951px)
.usa-header-basic .usa-navbar {
  width: 25%;
}
```

![screen shot 2017-02-23 at 10 06 21 am](https://cloud.githubusercontent.com/assets/955558/23264569/c6d5aac2-f9af-11e6-808a-873ca7ca6d16.png)

phae:
Yes, the basic header.

In terms of what we're using in addition, so far literally neither CSS nor SaSS. We've only used the pattern library for a demo. 

Also, I don't think this is a feature request. It's a bug. Not sure what the longest Federal Agency name is, but might be worth putting that in any tests you have.

juliaelman:
@phae thanks for the clarification! Agreed that this is a bug not a feature and appreciate you pointing this out to us to help resolve.

maya:
Fixed in #2750.

Issue #2636, opened 2018-07-30
Header navbar spacing inconsistencies in Sketch file and code


maya:
There are spacing inconsistencies in the header component Sketch file and live code. The Sketch file navbar and current page border is shorter. It seems these visual design updates were made in Sketch but never made it into the code. You can see what needs to be done based on the unchecked items in #2086. These should match. 

Which is the source of truth that we want to follow: the Sketch file of live code? This will determine what needs to be changed. @thisisdano

maya:
Navigation updated in #2750 which will be the source of truth so can close this. At some point we'll want to update the Sketch files.

PR #2750, opened 2018-09-14
Update navigation (and related elements) to v2

thisisdano:
Whew! 😅 
There's a lot going on here — please stress-test on any browser you can find.

- - -

[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-nav-v2/components/detail/header--default.html) 🔍 
(Dunno why the extended header with mega menu looks strange on the compare page — it looks and performs ok elsewhere?)

- - -

- Rebuilt the header and navigation with new tokens and values
- Used flex where possible
- Removed font smoothing from all instances (See http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/)
- Where possible, moved abstracted vars to settings
- New mixin: `at-media-max()` for max-width mqs
- New mixin: `add-bar` for adding bars to things like nav items
- New mixin: `add-background-image` to simplify png/svg background images
- Simplified folder structure of social images
- Sadly, replaced `$hit-area` with `$touch-target-size` (RIP `$hit-area`) 💩 
- Rewrote nav list mixin
- Cut 150 lines from `navigation.scss`
- Tweaked appearance of `usa-current` in top nav
- Added a `usa-current` state to the preview
- Fixed the utility builder to use the at-media() mixin. Utilities now have the proper em-based output when `$theme-respect-user-font-size` is true

- - -

- Used a _slightly_ different indicator bar for sidenav
- Added light borders between all sidenav elements for clarity

screen shot 2018-09-18 at 2 03 36 pm

- - -

Fixes: #1729.


maya:
Let's consider issue #2636 as well in this.

maya:
Will give my impressions on the visual changes first. 

Adding lines to everything makes it harder for me to see the hierarchy. It makes it look like it's all one level. Adds more visual noise.
screen shot 2018-09-18 at 2 47 19 pm

The side nav current state looks incomplete bc it doesn't reach the top and bottom, there's an awkward space between.

The plus sign is too small - out of balance from the minus sign which is larger. Should be larger.
screen shot 2018-09-18 at 2 44 48 pm

I'm more in favor of keeping it closer to the existing visual design, unless we heard there were issues from users.

src/stylesheets/core/_variables.scss
```diff
@@ -1211,8 +1211,9 @@ $nav-width:                     951px !default;
 $usa-btn-small-width:           units(6);
 $search-min-width:              27ch; // minimum length of 27 letters for typical search query
 
-// 44 x 44 pixels hit target following Apple iOS Human Interface Guidelines
-$hit-area: 4.4rem !default;
+// Was 44 x 44 pixels hit target following Apple iOS HIG
+// Now 40px to conform with standard units — is this a problem?
```
maya: Yeah it's a problem bc it's out of conformance with WCAG, 44px is the minimum we should go https://www.w3.org/WAI/WCAG21/Understanding/target-size.html: why not go a step above?

maya:
The link and buttons are different heights, should be the same:
screen shot 2018-09-18 at 3 28 41 pm
screen shot 2018-09-18 at 3 28 51 pm

Potentially for another time, but these probably should meet the touch target minimums as well since it's only used on mobile devices.

maya:
The desktop arrow size may have been accidentally increased:
screen shot 2018-09-18 at 3 35 22 pm

maya:
FYI the reason the mobile nav buttons and links are different heights are bc the line heights are different.

thisisdano:
ahh, thanks

maya:
Do we think the distance of the hover state is close enough to the line now that it should just meet the line or is the 4px of distance desirable?
screen shot 2018-09-18 at 3 53 24 pm
screen shot 2018-09-18 at 3 54 07 pm

thisisdano:
It should probably just go down to the line.

maya:
Since we're using flexbox, do we think we can use it to also address the issue around supporting longer logo titles #1729? Happy to do it another time too.

maya:
Long titles seem to expand the header vertically so is probably in a better place than it was when the issue was opened:
screen shot 2018-09-18 at 4 14 55 pm

maya:
Sounds good, update the PR to close that issue on merge 👍

maya:
Do you think the current and hover states should cover the line like it does now? Like a fullbleed? You mainly notice it when there's a dark background used below.

This PR:
screen shot 2018-09-20 at 9 46 11 am

Fullbleed:
screen shot 2018-09-20 at 9 46 01 am

maya:
@thisisdano lmk when this one is ready too.

thisisdano:
To follow on to the indicator bar/line discussion — I'm beginning to think that the bottom border isn't even necessary or helpful on the basic header. 

One unit of bottom padding seems to be cleaner and just as effective.

screen shot 2018-09-20 at 12 33 22 pm

screen shot 2018-09-20 at 12 33 57 pm

thisisdano:
The extended nav could be tidied and made more legible by removing the line (at least with the hero) and increasing the size of the links:

screen shot 2018-09-20 at 12 54 14 pm

thisisdano:
But all this is probably a discussion for another day!

thisisdano:
The bold nav links aren't really doing us any favors...

maya:
I think the header's bottom border gives it a nice definition and that without it it looks lost and floating with no anchor.

screen shot 2018-09-20 at 12 44 10 pm

maya:
>The bold nav links aren't really doing us any favors...

I thought it looked fine with font smoothing, SSP without it looks a bit heavy. But understand that's an issue for other fonts. Can we add font smoothing just for SSP headers?

thisisdano:
> I thought it looked fine with font smoothing, SSP without it looks a bit heavy. But understand that's an issue for other fonts. Can we add font smoothing just for SSP headers?

Maybe that's an issue to file — I'm not inclined to make face-by-face smoothing exceptions, but worth considering.

thisisdano:
I'm not including any of these border/padding things in this PR. Another time. 

But I also think that biggest contributor to lost/floaty feeling is the generic placeholder copy. When the copy doesn't demand to be read, it occupies less "attention space" and the page loses dynamism.

thisisdano:
Lines are all back. I consider this done, pending further review notes

src/stylesheets/components/_header.scss
```diff
@@ -1,98 +1,118 @@
...
+    min-width: 0; // Fix a Firefox display quirk
+  }
+
+  + .usa-hero {
+    border-top: units(1px) solid color('white');
+  }
+
+  + .usa-section,
+  + main {
+    border-top: units(1px) solid color('base-lighter');
```
maya: @thisisdano can you help me understand why we're adding the border to the top of the next `.usa-section` or `main` rather than the bottom of the header itself. On the surface, it seems more fragile so just wanted to understand.

PR #2751, opened 2018-09-17
Remove old grid from some components

maya:
[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/remove-old-grid)


src/components/banner/banner.njk
```diff
@@ -10,8 +10,8 @@
       </button>
       </div>
     </header>
-    <div class="usa-banner-content usa-grid usa-accordion-content" id="gov-banner">
-      <div class="usa-banner-guidance-gov usa-width-one-half">
+    <div class="usa-banner-content usa-accordion-content grid-container grid-row" id="gov-banner">
```
maya: Would have liked to be able to add `grid-gap` but it messes up the margins - I'd prefer not to have to add another div.

maya: cc @thisisdano

thisisdano: Addressed in https://github.com/uswds/uswds/pull/2757/

src/components/layouts/landing-inner.njk
```diff
@@ -15,8 +15,8 @@
   {% render '@graphic-list' %}
 
   <section class="usa-section">
-    <div class="usa-grid">
-      <h2>Section heading</h2>
+    <div class="grid-container">
```
thisisdano: I feel like `usa-prose` _should_ work here, but it isn't. I'm not getting the `heading + *` spacing and the button text color is being overridden by the prose class...

screen shot 2018-09-20 at 10 21 19 am

I'll need to dig into this a bit more

src/components/layouts/landing-inner.njk
```diff
@@ -15,8 +15,8 @@
...
+      <h2 class="font-heading-xl">Section heading</h2>
```
thisisdano: This shouldn't be necessary if `usa-prose` were working properly

maya:
@thisisdano FYI I reverted the banner changes in this PR in 741abbf since they're addressed in another PR.

src/components/layouts/landing-inner.njk
maya: Do we expect this section to be used beyond the current design (heading + lead + button), e.g. customized with additional text?

thisisdano: Well, I guess I consider this section just an example of "something you could do". It's not really a component, per se, so it's not _really_ important whether it uses `usa-prose` or some other technique. It's certainly OK to leave as is.  

But it does reveal some of what I might consider bugs in `usa-prose`. We'll address them in a subsequent issue, so I'd say it's ok to leave this as-is and move this PR along.

src/components/layouts/landing-inner.njk
thisisdano: Just a note to self to address this in a subsequent issue.
https://github.com/uswds/uswds/issues/2758

PR #2756, opened 2018-09-19
First pass at adding a token formatter

el-mapache:
## Description

In an attempt to better share our design tokens between projects, we are investigating adding a json -> sass transform pipeline. For now, this is hardcoded to work with a single map.

[Preview, for debugging purposes](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/ab-2753/components/detail/buttons--primary.html)


el-mapache:
@dano or @maya does anything here look obviously incorrect? The build is failing on CI, but I get no aXe violations when I run this locally.

el-mapache:
hmm in the preview, it looks like all the buttons are white on white, which is definitely wrong 😄

el-mapache:
@uswds/core This PR is fixed I believe, just an issue with string vs number keys!

PR #2762, opened 2018-09-21
Add knockout font smoothing to reversed text

### LINKED ISSUE #2761: Re-add font-smoothing to buttons (dark background + light text)
Should we re-add font smoothing to buttons?

As stated http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/, they recommend using font smoothing on dark backgrounds + light text. Our buttons fall into this category (any others)? 

## w/o font smoothing
screen shot 2018-09-20 at 4 53 53 pm
Is legibility reduced bc the pixels spill out making the letters less defined?

## w/ font smoothing
screen shot 2018-09-20 at 4 53 41 pm

thisisdano:
🔎 [Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-smoothing/components/detail/buttons--default.html)
Adds smoothing back to components that use knockout text.

- `hero`
- `section-dark`
- `button`
- `tag`
- `background-dark`

- - -

Resolves #2761

### DIFF
--- a/src/stylesheets/components/_header.scss
+++ b/src/stylesheets/components/_header.scss
@@ -59,12 +59,16 @@ $z-index-overlay: 400;
   }
 
   + .usa-hero {
-    border-top: units(1px) solid color('white');
+    @include at-media($theme-nav-width) {
+      border-top: units(1px) solid color('white');
+    }
   }
 
   + .usa-section,
   + main {
-    border-top: units(1px) solid color('base-lighter');
+    @include at-media($theme-nav-width) {
+      border-top: units(1px) solid color('base-lighter');
+    }
   }
 }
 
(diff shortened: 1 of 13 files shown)


maya:
I'm wondering about the application on the normal font text and whether it's too thin, specifically the tags, which is something you probably want to stand out:
screen shot 2018-09-21 at 9 47 45 am
screen shot 2018-09-21 at 9 47 59 am

What do you think?

thisisdano:
Yeah, that does look too light — I think this also applies to the non-bold links in dropdown menus as well

thisisdano:
Perhaps it's best to scope this only to buttons and active nav items — it can make things look a little unbalanced where there's a mix of subpixel-smoothed text and antialiased text in a block.

PR #2767, opened 2018-09-22
Fix buttons naming scheme

### LINKED ISSUE #2171: Button changes, no documentation

maya:
Align button naming scheme with our themed colors, change back to "outline" button, returning to how they were named before.

Fixes #2171.

### DIFF
--- a/src/components/02-buttons/buttons.config.yml
+++ b/src/components/02-buttons/buttons.config.yml
@@ -2,41 +2,35 @@ label: Buttons
 status: ready
 
 variants:
-  - name: primary
-    context:
-      category: primary
-      label: Primary buttons
-      classes: 'usa-button'
-
   - name: secondary
     context:
-      category: secondary
+      category: alternate
       label: Secondary buttons
       classes: 'usa-button-secondary'
 
-  - name: secondary-inverse
-    context:
-      category: secondary-inverse
-      label: Secondary inverse buttons
-      classes: 'usa-button-secondary-inverse'
-
-  - name: aqua
+  - name: accent-cool
     context:
       category: alternate
-      label: Aqua buttons
-      classes: 'usa-button-primary-alt'
+      label: Accent cool buttons
+      classes: 'usa-button-accent-cool'
 
-  - name: gray
+  - name: base
     context:
       category: alternate
       label: Gray buttons
-      classes: 'usa-button-gray'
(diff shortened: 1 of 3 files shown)


maya:
Don't really understand this comment:
```
 // This mixin is only needed until we fully remove the
 // deprecated usa-button-disabled style.
@mixin disabledesque {
```

Are we removing it?

maya:
Wasn't done during my time on the project so don't know. Should we remove the comment and let it stand?

thisisdano:
@maya Do you see any reason to deprecate the class-based disabled button?

maya:
Not that I can think of. Maybe a dev would want to access it using a class?

thisisdano:
OK, let's remove the comment and let the current code stand

maya:
Also, the name of the mixin, `disabledesque`, is kind of interesting, maybe `button-disabled` or `disabled-button`?

maya:
Comment removed in 0a019e5.

thisisdano:
Is what they were going for something like the idea that our buttons should _only_ use the `disabled` attribute [link](https://www.w3schools.com/tags/att_button_disabled.asp) and selector [link](https://www.w3schools.com/cssref/sel_disabled.asp)? Is that something we should do?

As for the name of the mixin — I think `button-disabled` is clearer than the current name.

maya:
Done in 57f07b6. We have classes for the other states (hover, etc...) so don't see why disabled would be any different.

thisisdano:
So, just to summarize: 
- For now, we are not deprecating the `*-disabled` classed buttons. 
- We'll retain the `pointer-events: none` rule in `button-disabled()` to support the class-based disabled button
- We'll continue to support both `-disabled` and `:disabled` markup/styling for buttons

Does that sound right?

maya:
Yeah

thisisdano:
One last thing: is the `stylelint-disable` still necessary with the current code?

maya:
I think they meant it to be `stylelint-enable`

PR #2770, opened 2018-09-24
Update sidenav

thisisdano:
🔎 [Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-update-sidenav/components/detail/sidenav--compare.html)

- - -

Uses `usa-sidenav` for the primary class of the sidenav component

### DIFF
--- a/src/components/08-navigation/_nav-primary.njk
+++ b/src/components/08-navigation/_nav-primary.njk
@@ -3,7 +3,7 @@
 </button>
 <ul class="usa-nav-primary usa-accordion">
   {%- for link in nav.links -%}
-  <li class="usa-nav-item">
+  <li class="usa-nav-primary-item">
     {%- if link.links -%}
     <button class="usa-accordion-button usa-nav-link{% if link.is_current %}  usa-current{% endif %}" aria-expanded="false" aria-controls="{{ nav.id_prefix }}{{ link.id }}">
       <span>{{ link.text }}</span>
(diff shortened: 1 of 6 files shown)


maya:
What's the thinking behind leaving the list items as `usa-nav-item` rather than `usa-sidenav-item`. The latter would be following BEM.

thisisdano:
In general, this is because the sidenav and the top nav share the nav list mixin (`nav-list()`) — but I could use `&` statements in the mixin to pick up a parent class...

```sass
@mixin nav-list {
  &-item {
    rules...
  }
}

.usa-sidenav {
  @include nav-list;
  ...
}
```

maya:
Is there a way we can update the mixin to allow for both `.usa-nav-item` or `.usa-sidenav-item`. So we can use `.usa-nav-item` in the header and `.usa-sidenav-item` in the sidenav? Mixins shouldn't prescribe which classes we have to use, should they?

One idea is adding a second class to the mixin:

```
.usa-nav-item,
.usa-sidenav-item
```

Then you can use both in different situations? But maybe there's another cleaner way.

maya:
Btw the `nav-sublist` mixin also has `.usa-sidenav-sublist` as its grandchildren class, though you would use that class for the header as well (e.g. usa-nav-sublist), right?

maya:
Though if the case is `usa-nav-item` is semantically the correct name for this item in the sidenav, then that's a different story.

maya:
changing to `usa-nav-primary-item` adds a bit of clarity too, which is good.

thisisdano:
My first-ever use of `& &` in a rule!

PR #2774, opened 2018-09-24
Update and improve grid mixins

### LINKED ISSUE #2742: Assure that grid mixins work as documented
I think we've got some problems with how the layout grid mixins work in component code. Check to make sure that `grid-container()`, `grid-row`, `grid-gap()`, and `grid-col()` all work as expected in component Sass.

thisisdano:
🔎 [Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-update-grid-mixins/components/detail/multi-column-checkboxes.html)

Multi-column checkboxes in the example linked above are styled using markup and component code. This is also an example of how we might style components with variable columns — controlling the number of columns via a settings token.

- - -

Includes testing styles in `uswds.scss`. **I need to remove these before merging.**

- Includes support for responsive grid gap
- Includes support for column gap spacing in component code
- Abstracts adding responsive site margins into its own mixin: `add-responsive-site-margins()`
- Adds new function to calculate grid-col columns from desired column number: `columns()`
- Adds `grid-container` to some components that previously didn't use them

- - -

Resolves #2742

### DIFF
--- a/src/components/test/multi-column-checkboxes.njk
+++ b/src/components/test/multi-column-checkboxes.njk
@@ -15,6 +15,10 @@
           <input class="usa-checkbox-input" id="truth" type="checkbox" name="historical-figures-1" value="truth" checked>
           <label class="usa-checkbox-label" for="truth">Sojourner Truth</label>
         </div>
+        <div class="example-col">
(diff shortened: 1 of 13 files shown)


maya:
There's still a [WIP] label in the title, but saw we were assigned and it's in the review column. Is it ready?

PR #2780, opened 2018-09-25
Remove bourbon from packages

thisisdano:
Grrr. Forgot these.


el-mapache:
@thisisdano do we also need to remove the references to bourbon in `config/gulp/sass`?

maya:
Needs to be removed from `src/stylesheets/core/_required.scss`.

PR #2787, opened 2018-09-26
Remove component prefix var/refs and regularize theme var names

thisisdano:
- ~Removes vars and refs related to changing the component prefix~ Already done in #2785 
- Uses a `$theme-[component]-[property]` naming format for all vars that reference a specific component or type. A var like `$theme-color-primary` stays as is, as, in this case, `color` itself is the type. Examples:
  - `$theme-link-color`
  - `$theme-root-font-size`
  - `$theme-navigation-font-family`
  - `$theme-color-primary-darker`

### DIFF
--- a/src/stylesheets/components/_alerts.scss
+++ b/src/stylesheets/components/_alerts.scss
@@ -55,11 +55,11 @@ $icon-size: units(4);
   }
 
   a {
-    color: color($theme-color-link);
+    color: color($theme-link-color);
 
     &:focus,
     &:hover {
-      color: color($theme-color-link-hover);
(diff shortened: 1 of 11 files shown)


maya:
@thisisdano can you pull the latest changes into this? we merged at least _some_ of this in already in #2785.

maya:
Also, will needs to bring these variable changes into `uswds-site` if they're used there. Will you be able to make those updates on that repo as well?

thisisdano:
Looks like it's all up-to-date with `release-2.0` now. And yup, I'm working on getting this into `site`, too. I'll put a PR together after this is all merged.

PR #2803, opened 2018-09-28
Update mem to 4.0.0

maya:


thisisdano:
Is there any way to check and see if this change had the desired effect?

maya:
I checked package-lock.json and looks like 4.0.0 is showing up, but so it 1.1.0 - maybe best to way til they release the updated yargs like you mentioned.

PR #2811, opened 2018-10-03
[Checkbox] Add print styles for checked checkbox

### LINKED ISSUE #2810: Print styles for checkboxes
## Description
When printing a page containing a checkbox in the checked state, there is no clear indication of the fact that the fields are selected/checked -  the thin checkbox outline/box shadow is all that is visible (difficult to notice).

1. Print a page with a selected checkbox on it
(Chromes print preview pane allows you to see this behavior without having to actually print the page.)

screen shot 2018-10-03 at 21 34 56

## Additional information 
The reason this happened is because most browsers’ default behavior is to not print background colors and the visual state of the checkbox is dependent on a background color.

David-Way:
Fixes #2810 

Proposed changes
- Add print styles for checkbox
- Add alternative correct8 img files


thisisdano:
I'm going to start taking a look at this soon. I just want to make sure if it's something we add to `1.x` that it's also something we incorporate into `2.0`. It also reminds me that print stylesheets, in general, are a topic area we should address more fully in coming months.

src/img/correct8-alt.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 216 146"><path fill="#0071bc" d="M168.86 37.966l-11.08-11.08c-1.52-1.52-3.367-2.28-5.54-2.28-2.172 0-4.02.76-5.54 2.28L93.254 80.414 69.3 56.38c-1.52-1.522-3.367-2.282-5.54-2.282-2.172 0-4.02.76-5.54 2.28L47.14 67.46c-1.52 1.522-2.28 3.37-2.28 5.542 0 2.172.76 4.02 2.28 5.54l29.493 29.493 11.08 11.08c1.52 1.52 3.368 2.28 5.54 2.28 2.173 0 4.02-.76 5.54-2.28l11.082-11.08L168.86 49.05c1.52-1.52 2.283-3.37 2.283-5.54 0-2.174-.76-4.02-2.28-5.54z"/></svg>
```
thisisdano: Maybe this should be a pure black check? It may make sense to optimize for grayscale printing.

src/stylesheets/elements/_inputs.scss
```diff
@@ -261,6 +261,14 @@ legend {
   cursor: not-allowed;
 }
 
+@media print {
+  [type=checkbox]:checked + label::before,
+  [type=checkbox]:checked:disabled + label::before {
+    content: url('#{$image-path}/correct8-alt.png');
+    content: url('#{$image-path}/correct8-alt.svg');
```
thisisdano: - Add an explicit background-color of `white` here, just in case the user decides to print background images. 
- It might also be worth explicitly setting the background image to `none` as well. That could make this solution a bit more portable if we wanted to apply something similar elsewhere.
- Set `text-indent: 0` to center the new image in the checkbox

thisisdano:
Also, dunno if you're using it, but this is a useful way to force a `print` render mode in the Chrome inspector: https://stackoverflow.com/questions/9540990/using-chromes-element-inspector-in-print-preview-mode

It was new to me!

David-Way:
> Also, dunno if you're using it, but this is a useful way to force a `print` render mode in the Chrome inspector: https://stackoverflow.com/questions/9540990/using-chromes-element-inspector-in-print-preview-mode
> 
> It was new to me!

Yeah cheers, it's very useful and really buried away in there.

thisisdano:
Thanks @David-Way — I'll publish a new `1.x` release with this, and incorporate it into `2.0` as well.

FYI, I'm merging this into a different branch, just to finish some internal testing before this makes its way into production.

PR #2827, opened 2018-10-05
Remove :visited classing from buttons

thisisdano:
🔍 [Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-fix-buttons/components/detail/buttons--secondary.html)

This is a little difficult to preview and reproduce. You need to manually change the buttons in the preview to be `a`s instead of `button`s and then preview with `:visited`. This also fixes an additional bug I noticed with the `:active` state with `a` buttons.


thisisdano:
Updated to remove the `:not()` rule since it was adding too much specificity. I went back and just focussed on the `usa-button` styles.

thisisdano:
@saracope Tagging you in on this one for review!

PR #2848, opened 2018-10-14
Print styles for radio buttons

### LINKED ISSUE #2847: Print styles for radio buttons
## Description
When printing a page containing a radio button in the checked state, there is no clear indication of the fact that the fields are selected/checked -  the thin radio button outline/box shadow is all that is visible (difficult to notice).

1. Print a page with a selected radio on it
(Chromes print preview pane allows you to see this behavior without having to actually print the page.)

screen shot 2018-10-14 at 11 16 24

## Additional information 
The reason this happened is because most browsers’ default behavior is to not print background colors and the visual state of the checkbox is dependent on a background color.

David-Way:
Fixes #2847

### Proposed changes:
- Add print style for radio button
- 2 possible approaches attempted
  1. Remove the custom radio button styles when printing
  2. Use box shadow to emulate selected style

Not 100% sure what the best approached would be here.


saracope:
Thanks for the contribution, @David-Way! Reviewing this today.

Issue #2849, opened 2018-10-15
Border utility mixin is not working properly


thisisdano:
The `u-border()` utility mixin should accept border-style, units, and color tokens but is only working with border-style now. And it's giving a relatively unhelpful error message.

Thanks for pointing this out @jeremyzilar

arch-lord:
I’ll attempt a fix tomorrow morning after a good night’s sleep. Although I don’t know the project structure very well, I have some ideas as to this issue.

thisisdano:
Thanks @arch-lord — I think I probably know what's going on, but I'd love to have you take a swing at this — if only to approach the issue with fresh eyes.

thisisdano:
OK, I was having trouble reproducing this and can now clarify that the issue seems to be that the mixin is not accepting the string equivalent units values like `'05'`. The numeric values like `0.5` are OK.

PR #2860, opened 2018-10-28
Repair border() mixin to accept valid units string tokens

thisisdano:
Fixes the `border()` mixin to accept any valid border-width unit token and improves the error message if the mixin includes an invalid token.


saracope:
Fixes #2849

PR #2867, opened 2018-11-02
Fix compile errors with Dart Sass

thisisdano:
This fixes two errors and one warning in Dart Sass compile:

- - -

Fixes an error where certain numbers couldn't be looked up in the number-to-token conversion table. The product of a mathematical operations (like `(3 + 1)`) would fail in the lookup but regular numbers like (`4`) would be OK. 🤷‍♀️ I fixed this by using strings like `'4'` for keys in the lookup instead of numbers, and converting the numbers to strings in the `number-to-value()` function.

I also renamed the fn `number-to-token` for clarity.

- - -

Fixes an error where colors (like `red` or `transparent`) passed into the color function messed up the `string()` function. Since raw colors like this shouldn't be passed into functions and mixins at all, I made this an explicit error and pointed the user to the documentation.

I changed the code in `_banner.scss` to use `color('transparent')` instead of `color(transparent)`.

- - -

Rewrote some code to address the `interpolation near operators` warnings.


thisisdano:
This all checks out on `uswds-site`. No new errors.

PR #2906, opened 2019-02-07
[v2] [List styles] Add max width, remove bottom margin on last-child.

adunkman:
List items should have a maximum width, similar to their `<p>` counterparts. Additionally, when a list is the last child of a container, it should not contain a margin-bottom (see the left border in the screenshot below).

| Before | After |
|-------|-------|
| Screenshot of list items which do not have maximum width constraints and have extra margin-bottom | Screenshot of list items which are constrained by a maximum width, and do not contain extra margin-bottom |

Thoughts? Was going to open an issue, but figured it’d be easier to demonstrate with a PR directly.

### DIFF
--- a/src/stylesheets/core/mixins/_unstyled-list.scss
+++ b/src/stylesheets/core/mixins/_unstyled-list.scss
@@ -6,5 +6,6 @@
 
   > li {
     margin-bottom: 0;
+    max-width: unset;
   }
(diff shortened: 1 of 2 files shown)


thisisdano:
Excellent — this was on my list of things to do! I'll review asap.

Issue #2907, opened 2019-02-12
Change to print styling


saracope:
## Description

**Issue**: It is not clear to users that the USWDS does not "handle" print styling. 
**Result**: Users do not add a print stylesheet. Consequently, when the print version of their site looks wonky they equate this to a bug in the design system.
**Consider**: 
1.) Omitting the media type for the condensed style sheet or setting it to 'all'. 
2.) Documenting the descision here and a (possible) recommendation for users to create their own print stylesheets as needed.
**Reasoning**: This would at least allow the site to use any already defined styles it has for a better experience.

TODO
**Investigate**: Any side effects to making this change in the design system, especially with accessibility.

thisisdano:
Fixed in #2910

mgifford:
So is print styling no longer recommended as a best practice by the USWDS?  I looked at all the A-F websites that are using it and surprisingly few had print CSS definitions. Some of the best are:

- https://clinicaltrials.gov/ct2/html/css/print.css?v=43
- https://betobaccofree.hhs.gov/sites/default/files/css/css_z98YhE0GkqAFpYUw9F-u1Zfsq3DJ73sZ6ol42VtOvYE.css
- https://www.ars.usda.gov/css/print.css
- https://18f.gsa.gov/assets/css/print.css
- https://feedback.usa.gov/assets/css/print.css

Some are doing it inline like:

- https://www.fcsic.gov/

There really aren't many USWDS sites that are using print CSS. I don't think a change in documentation is going to do much.

arkjoseph:
@mgifford @saracope - flex box/uswds is not printer friendly and really is not up to par with what most clients expect in regards to printer output.
Ive resolved this on gsa.gov by forcing columns to render using specific widths in % values, in addition ive also replicated what bootstrap does and change the display of grid-row to display:table.  ive also created classes to address 'printer-only' out put or screen only output within markup, not css.

Issue #2883, opened 2018-11-26
"site is secure" and ".gov means it's official" icons filling page when printing


suprenant:
When I visit https://join.tts.gsa.gov/ as well as some 18F guides (including the [Agile Guide](https://agile.18f.gov/)) and print, I notice that the lock and courthouse icons fill an entire page. 

I do not notice this on some other sites using USWDS. 

From a colleague in TTS Slack:
_It would appear nothing is constraining the SVG sizes. I wonder where the shared print style for the guides comes from_

![screen shot 2018-11-16 at 10 51 09 am](https://user-images.githubusercontent.com/1452369/49038710-4ecf0500-f183-11e8-9f4f-143058e9d5f3.png)
![screen shot 2018-11-16 at 10 51 06 am](https://user-images.githubusercontent.com/1452369/49038711-4f679b80-f183-11e8-9017-1f6577ade282.png)

greensteph:
Hi @suprenant thanks for submitting this issue. We are exploring how to best provide a solution as well as improving print styling overall.

suprenant:
@greensteph sounds good, thanks for the update. I'll look forward to seeing what might come from this. If you ever want any feedback or anything, please let me know.

thisisdano:
This is fixed in USWDS 2.0 with https://github.com/uswds/uswds/pull/2910.
Once the dust clears from 2.0, we can look into getting this into a 1.x release.

PR #2910, opened 2019-02-15
Use `all` for media queries instead of screen only

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-print-mqs/components/preview/layout--landing.html)

This PR changes the scope of media queries to all media types instead of screen only. This allows media query styling to apply to print jobs, fixing a number of current issues with print performance. In testing, printing to standard 8.5 x 11" paper results in mobile-style output, which is perfect.

We haven't uncovered any potential problems with this new scope, but it's worth a bit more research time to make sure there aren't any unintended consequences, especially accessibility issues.

**Q:** are there some places where reverse text will be an issue?

Fixes https://github.com/uswds/uswds/issues/2907
Fixes https://github.com/uswds/uswds/issues/2883

- - -

screen shot 2019-02-15 at 9 32 45 am

- - -

With `Print background graphics` checked:

screen shot 2019-02-15 at 9 35 03 am

### DIFF
--- a/src/stylesheets/core/mixins/_at-media.scss
+++ b/src/stylesheets/core/mixins/_at-media.scss
@@ -1,4 +1,5 @@
 // Mobile-first media query helper
+
 @mixin at-media($bp) {
   $quoted-bp: smart-quote($bp);
   $our-breakpoints: map-deep-get($system-properties, breakpoints, standard);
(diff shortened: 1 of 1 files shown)


thisisdano:
Cool. It looks like the reverse text might print a little light, but it's probably an improvement from what it was doing before.

thisisdano:
Good to remember that we shouldn't put any critical images in backgrounds since they won't print!

thisisdano:
@saracope please hold on merging until we sort out some Federalist issues

PR #2918, opened 2019-02-22
Improve font settings

thisisdano:


thisisdano:
Set this up with the wrong base and triggered the wrong CI. Let's try again

Issue #2887, opened 2018-12-05
Layout grid should use a different !important switch from utilities


thisisdano:
You should be able to switch on `!important` for utilities separately from activating `!important` for the layout grid. Additionally, components should be able to import grid mixins without those rules rendering as `!important`, regardless of the switch setting. The switch should apply _only_ to the grid classes, not any derivatives.

maya:
Fixed in #2924.

PR #2924, opened 2019-02-25
Add a new var to control if layout grid gets !important

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-grid-important/)

Adds a `$theme-layout-grid-uses-important` bool to settings-general. Previously, layout grid would get `!important` if `$utilities-use-important` was set to true.

Fixes https://github.com/uswds/uswds/issues/2887

### DIFF
--- a/src/stylesheets/core/_variables.scss
+++ b/src/stylesheets/core/_variables.scss
@@ -602,6 +602,6 @@ Grid
 
 $grid-global: '';
 
-@if $utilities-use-important {
+@if $theme-layout-grid-uses-important {
   $grid-global: '!important';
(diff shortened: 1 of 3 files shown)


maya:
Any reason to have `use` for utilities and `uses` for grid or should they be the same?

thisisdano:
I guess I was going for grammatical correctness over consistency — do you think it makes better sense with `use`?

maya:
Yes, I'd go with `use `.

thisisdano:
OK, renamed

Issue #2928, opened 2019-03-07
Focus outline shows on non-focusable elements on IE 11


ncksllvn:
## Description
Using IE11, the focus outline works as expected when using the tab key. However, if I click a non-focusable element on the page, the dotted outline shows even though it's not focusable.

## Steps to reproduce the issue

1. Open IE11
2. Navigate to site
3. Click a non-focusable element, such as a div container or image
4. See outline appear

## Additional information [optional]

- This issue was discovered during an accessibility audit on VA.gov.
- Screenshot after clicking "Getting started" container
![image](https://user-images.githubusercontent.com/1915775/53964004-6a1fa980-40bc-11e9-89e3-2e7f62dcb953.png)

maya:
Hi, thanks for opening this issue. We'll review this with our team and see how it fits on our roadmap. Stay tuned!

ncksllvn:
Thank you for checking it out @maya!

PR #2929, opened 2019-03-07
[VA.gov] - [Focusable Elements] - Apply :focus behavior to focusable elements only

ncksllvn:
## Description
This pull request explicitly lists elements that are focusable and applies the focus mixin only to those elements, rather than applying the rule globally. The immediate goal of this is to patch the bug as described in Issue #2928.

Resolves https://github.com/uswds/uswds/issues/2928

## Additional information

- Even though this issue was discovered only in IE, I feel that this is also the more correct solution than applying the `:focus` rule globally.
- This solution is based on the solution described in this SO thread, https://stackoverflow.com/a/30753870.
- jQuery's `:focusable` selector also has interesting info, https://api.jqueryui.com/focusable-selector/.
- See issue for more context

    - May have messed up the `[Website]` part of this.

### DIFF
--- a/src/stylesheets/core/_base.scss
+++ b/src/stylesheets/core/_base.scss
@@ -30,17 +30,26 @@ body {
   display: none !important; /* stylelint-disable-line declaration-no-important */
 }
 
-a[href],
-area[href],
-input:not([disabled]),
-select:not([disabled]),
-textarea:not([disabled]),
-button:not([disabled]),
+input,
+select,
+textarea,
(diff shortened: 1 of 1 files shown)


thisisdano:
Merging into an internal branch for integration

thisisdano:
Work continuing here: https://github.com/uswds/uswds/pull/2983

ncksllvn:
Thank you @thisisdano!

PR #2931, opened 2019-03-08
Fix landing page tagline section

### LINKED ISSUE #2920: Add missing section to landing page template
This section "A tagline highlights your approach..." is missing below the hero from the fractal site, but appears on the docs site:

screen shot 2019-02-22 at 1 44 53 pm

https://components.designsystem.digital.gov/components/detail/layout--landing.html

maya:
Ensures that the tagline section appears.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-landing-page/components/preview/layout--landing.html)

---

There's an ongoing issue that the browser's user agent stylesheets are providing margins for various texts - paragraph text, headings, etc... so our top and bottom margins are not consistent. And there's no way to get proper em-based relative margins in the utilities. 

When adding `.usa-prose` to a text section at the bottom, the margins on the headings are correct but it negatively effects `.usa-lead`: it overrides the line-height and max-width. Also, it changes the button text to blue so it's unreadable. So, the **selector strength is too strong** `.usa-prose > p` overrides `.usa-font-lead` which is added directly to the `p` element.

Screen Shot 2019-03-08 at 3 13 00 PM

Currently, I added `margin-top-0` to those headings, but I don't think that's a great approach and it doesn't give me the 0.5em bottom margins that I'd expect. The bottom margins are the larger user agent stylesheet.

Fixes #2920.

### DIFF
--- a/src/components/layouts/landing-inner.njk
+++ b/src/components/layouts/landing-inner.njk
@@ -3,11 +3,13 @@
 
   {% if tagline %}
   <section class="grid-container usa-section">
-    <div class="usa-width-one-third">
-      <h2>{{ tagline.title }}</h2>
-    </div>
-    <div class="usa-width-two-thirds">
-      {{ tagline.content }}
+    <div class="grid-row grid-gap">
+      <div class="tablet:grid-col-4">
+        <h2 class="font-heading-xl margin-top-0">{{ tagline.title }}</h2>
+      </div>
+      <div class="tablet:grid-col-8 usa-prose">
+        {{ tagline.content }}
+      </div>
     </div>
   </section>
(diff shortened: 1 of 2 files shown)


maya:
Removed top and bottom margins from some headings in 6930c03. The 1em margins on paragraph text that Chrome provides works out as it's what we use/recommend, but there probably should be another way to do this in USWDS so we don't rely on user-agent sheets as there's probably some variance between them.

thisisdano:
I have some thoughts about what we could do as a longer-term solution for this kind of problem that I'll roll up into another PR.

PR #2932, opened 2019-03-09
Update external link example to use hyphen

maya:
Now it shows up.

Following BEM, should it be written as `usa-link--external`?

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/fix-external-link/components/detail/links.html)


thisisdano:
Yeah, this should be `usa-link--external` and also `usa-link--alt` — I just made this change.

maya:
OK, I noticed the manual visited link color is not appearing so I guess a change needs to be made on the docs site. Also, wondering if it's worth showing some of the hidden "alt" examples at some point.

PR #2934, opened 2019-03-10
Use most current versions of Source Sans Pro and Merriweather

thisisdano:
This updates the font packages to use the most current versions of Source Sans Pro and Merriweather from google fonts — with additional formats (eot, woff, woff2) generated by https://transfonter.org/

It also updates the license to include the provenance of these new versions.

I included all weights and styles of these new versions since they can be manually disabled in settings and only faces used on the page will actually load.

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-update-fonts)


maya:
I believe in one or both of these fonts, the font creators created a subset of the font for us, but I guess it's OK to start using the latest. Can you share what version of each you're using?

thisisdano:
Is there a way to get the version through google fonts? I'm not 100% sure anything has changed since our version, but I just figured I'd grab whatever was currently available...

maya:
Here are the latest releases of each:
https://github.com/adobe-fonts/source-sans-pro/releases
https://github.com/EbenSorkin/Merriweather/releases

Though I don't know what Google fonts is using.

maya:
@miguelsousa previously did some post-processing on Source Sans Pro in #802 #848 to keep them performant and a small footprint size.

thisisdano:
Comparing Merriweather 2.001 to 1.003
The newer one is a bit different and a bit nicer.

merriweather 2 x v 1 x

maya:
Yeah, that's quite a difference. Also, noticed a lot of changed on Merriweather since the last release https://github.com/EbenSorkin/Merriweather/commits/master. Should we check with @EbenSorkin to see if he plans on cutting a release anytime soon?

Is this ready to go in now? @miguelsousa and @EbenSorkin can always provide feedback.

thisisdano:
I'd love to get it in now, with follow up from the creators/maintainers.

miguelsousa:
I'd like to review the new font files before this PR is merged. Please allow me a couple days for that.

thisisdano:
@miguelsousa Thanks! We'll pause this PR until we get your feedback. Please let us know what you find. The Source Sans files in this PR are those from https://github.com/adobe-fonts/source-sans-pro/releases/tag/2.045R-ro%2F1.095R-it — subsetted into Latin and and converted into woff, woff2, and eot by https://transfonter.org/

I'd be curious to know if this is a proper workflow or does anything unexpected to the files.

miguelsousa:
@thisisdano I can tell you right now that the new Source Sans files have less character support (and less language support) and are lacking the small cap glyphs, so I expect several issues will regress if you merge this PR as is.
I would also like to understand why a few new styles (e.g. Black) are included.

maya:
@miguelsousa also created a [dedicated repository](https://github.com/miguelsousa/source-sans-pro-subset) for this project where he mentioned being able to publish updates whenever there was a new release of SSP. Is this something we should refer to going forward?

thisisdano:
As for why the new weights and styles are included: In general, it's to support a wider range of project needs.

USWDS 2.0 supports a broader range of weights, so it made sense to make those weights available in the distribution package. System users can now activate or deactivate weights depending on their project needs — and the system only builds the necessary `@font-face` rules based on these settings.

miguelsousa:
@maya that's right. I was about to mention that repo.
For the record, even though there are newer versions of Source Sans Pro available, I haven't felt the need to update the subset fonts used by USWDS because the changes have no impact on the subset. 
But I'm willing to generate new subsets from the latest version in order to get the version number in sync and to add the new weights.

thisisdano:
@miguelsousa That would be very helpful, thank you.

thisisdano:
Also, now that I know that our version of Source Sans has small caps support, we should add that support into our text utilities.

miguelsousa:
@thisisdano please fetch the Source Sans subsets from here https://github.com/miguelsousa/source-sans-pro-subset/tree/master/subset

maya:
Added new SSP subset font files in 59d71b2.

@thisisdano what is the new cap height value for Merriweather?

thisisdano:
@maya Let's use `371` for the Merriweather `cap-height` and be sure to update the license to reflect the new origin of Source Sans Pro

maya:
Updated `cap-height` in 9943ed5 and license in 49479c4.

maya:
Since the new font metrics of Merriweather seems a bit thicker, noting the difference of with and without antialiasing on headings on dark backgrounds:
## normal
normal

## antialiased
antialiased

thisisdano:
LGTM for now. Let's monitor how things change with the new Merriweather and respond as needed. @maya can you do the approving?

Issue #2949, opened 2019-03-12
Audit BEM notation


maya:
Make sure we are using accurate BEM notation throughout [USWDS 2.0](https://github.com/uswds/uswds/tree/release-2.0/src/stylesheets).

A helpful guide: https://seesparkbox.com/foundry/bem_by_example

## USWDS classes that begin with `.usa-*`

https://gist.github.com/maya/17f1b37d5320713877159cc0b4e3c8ff

## To investigate
```
.usa-button-unstyled
.usa-unstyled-list
.usa-input-error-message
.usa-input-label-helper
.usa-input-label-required
.usa-media-link
.usa-social-link
.usa-form-hint
.usa-input-list
.usa-content-list
.usa-font-lead
.usa-dark-background
.usa-footer-* ...
.usa-nav-* ...
.usa-mobile-nav-active
```

thisisdano:
That sparkbox guide is really nice. 
I _think_ we've implemented our own BEM consistent with that guidance, so I hope there are only a few instances where stuff is out of whack.

thisisdano:
One known inconsistency is with `usa-unstyled-list`. Currently this is a standalone block instead of a modifier of `usa-list` (à la `usa-list.usa-list--unstyled`). I believe this is for simplicity's sake. But there's a fair argument to be made that the benefits of simplicity are outweighed by the cognitive burden of the inconsistency.

`.usa-button-unstyled` is the same (but different). If we leave them as standalone blocks, they should both have the same naming system (probably something like `usa-unstyled-button` -- that would be consistent with `usa-unstyled-list` and `usa-dark-background` which are named that way to prevent being confused as modifiers).

thisisdano:
Of course, maybe `usa-button-unstyled` should just be regular BEM — `usa-button--unstyled`.

I'm inclined to leave `usa-dark-background` as is since `usa-background` is currently meaningless.

thisisdano:
@maya what do you think about `.usa-button-unstyled` and `. usa-unstyled-list`? 
I _think_ I'd like to just go ahead and BEM them but I'm not 100%.

maya:
`.usa-button--unstyled` and `.usa-list--unstyled` make the most sense to me for consistency sake.

thisisdano:
OK, working on unstyled lists and button here: https://github.com/uswds/uswds/issues/2956

maya:
Looking through some classes that didn't seem to use BEM, this is what I found:

[...]

PR #2951, opened 2019-03-12
Update media link to BEM notation

maya:
Part of #2949.


thisisdano:
I remember that I did not change this one because it was so unusual — it didn't have the `usa` prefix and seemed like some kind of outlier. Is there a reason why this particular class is unprefixed?

maya:
Added the prefix in 380bf9a. No good reason I could think of not to have it.

PR #2952, opened 2019-03-12
Update radio buttons and checkboxes to use BEM notation

maya:
In this case, we do not have a `.block` set explicitly even though it's the `<li>` element. We do have a `.block__element`. I've also heard that some folks add the `.block` even if there's no styling needed, but I left it out for now unless others feel strongly it should be added.

Part of #2949.


thisisdano:
Yeah, I think this confusion was why I left it out of the Beta 6 pass.
I think I would like to add `usa-radio` to the `<li>`s to make the relationship more clear.

maya:
All good, we can add that in.

maya:
Added in d09a4f7.

PR #2964, opened 2019-03-18
Use `ex` unit for measure values

### LINKED ISSUE #2960: Use `ex` unit for measure values
I'm finding that the `ch` unit is a poor unit for doing our measure calculation, since the `0` character on which it's based has significant variance from font to font. Since we're already normalizing our font size in such a way that fonts display with a similar x-height at each grade in our font-size system, we should use the `ex` unit for a more consistent and reliable output. The `ex` unit has universal support.

**This is also distorting our measure output tokens in the direction of being generally too wide** — far outside the 50-75 target character count in our guidance. That said, the guidance is maybe a little bit skewed narrow, as a popular readable site like medium.com has a line length of around 75-78 characters, and we could a bit longer than that for something like a small caption.

I'm also proposing that we add an additional token to accommodate some of gamut we'll lose by concentrating our coverage on the guidance range.

| token | desired output | proposed value | current value | current output in Public Sans 
| --- | --- | --- | --- | ---
| `1` | ~50 characters | `44ex` | `40ch` | 51 characters
| `2` | ~66 characters | `60ex` | `60ch` | 80 characters
| `3` | ~72 characters | `64ex` | `66ch` | 88 characters
| `4` | ~78 characters | `68ex` | `72ch` | 95 characters
| `5` | ~86 characters | `74ex` | `77ch` | 100 characters
| `6` | ~96 characters | `88ex` | — | —

| Beta 6 | Beta 7
| --- | ---
| '1' | '1'
| '2' | '4'
| '3' | '5'
| '4' | '6'
| '5' | '6'

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-update-measure/components/detail/typesetting.html)

- - -

This updates our measure tokens to use `ex` units instead of `ch` per the reasoning in #2960.

This rebalances our measure tokens to be concentrated within our guidance range of 45-90 characters (see https://practicaltypography.com/line-length.html) and should result in output that's more consistent from typeface to typeface because the `ex` is the height of the `x` character and our normalization scheme tends to make this value more consistent from face to face.

Now our docs and `usa-prose` measure are what they say they are: ~75 characters (`measure 4`).

Screen Shot 2019-03-18 at 10 02 15 AM

However, this is something of a breaking change in that the new measure tokens tend to be narrower than their old values and will need to be updated per the chart that follows.

| token | desired output | new value | old value | old output in Public Sans 
| --- | --- | --- | --- | ---
| `1` | ~50 characters | `44ex` | `40ch` | 51 characters
| `2` | ~66 characters | `60ex` | `60ch` | 80 characters
| `3` | ~72 characters | `64ex` | `66ch` | 88 characters
| `4` | ~78 characters | `68ex` | `72ch` | 95 characters
| `5` | ~86 characters | `74ex` | `77ch` | 100 characters
| `6` | ~96 characters | `88ex` | — | —

| Beta 6 | Beta 7
| --- | ---
| '1' | '1'
| '2' | '4'
| '3' | '5'
| '4' | '6'
| '5' | '6'

- - -

Fixes #2960

### DIFF
--- a/src/components/01-type/typesetting.njk
+++ b/src/components/01-type/typesetting.njk
@@ -2,7 +2,7 @@
   <div class="usa-line-length-example usa-prose">
     <h6 class="usa-heading-alt">Line length</h6>
     <section class="typography-example usa-prose margin-top-1">
-      <p><strong>75 characters (66ch) max-width:</strong> Yosemite National Park is set within California’s Sierra Nevada mountains. It’s famed for its giant, ancient sequoias, and for Tunnel View, the iconic vista of towering Bridalveil Fall and the granite cliffs of El Capitan and Half Dome.</p>
+      <p><strong>75 characters (68ex) max-width:</strong> Yosemite National Park is set within California’s Sierra Nevada mountains. It’s famed for its giant, ancient sequoias, and for Tunnel View, the iconic vista of towering Bridalveil Fall and the granite cliffs of El Capitan and Half Dome.</p>
     </section>
   </div>
 
--- a/src/stylesheets/core/_properties.scss
+++ b/src/stylesheets/core/_properties.scss
@@ -435,6 +435,7 @@ $system-properties: (
       3:	    $system-measure-base,
       4:	    $system-measure-large,
       5:	    $system-measure-larger,
+      6:	    $system-measure-largest,
       'none':   none,
     ),
     extended: (),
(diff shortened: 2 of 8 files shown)


thisisdano:
Hm, whuts this new Snyk thing? 🤔

saracope:
Snyk checks every PR for new vulnerabilities.
I just updated for these specific vulnerabilities. If you pull the latest version of 2.0 it should resolve this.

thisisdano:
Yeah, I should have a docs update by later today as part of a big docs/typesetting update

PR #2965, opened 2019-03-18
Tighten the letterspacing of `letterspacing 1`

### LINKED ISSUE #2962: Tighten letterspacing of `letterspacing-1`
`Letterspacing 1` is the appropriate token for setting the letterspacing for very small text. But 0.05em is a bit too loose for that purpose. I suggest using 0.025em.

**no letterspacing**
Screen Shot 2019-03-17 at 2 08 45 PM

**0.025em letterspacing**
Screen Shot 2019-03-17 at 2 11 19 PM

**0.05em letterspacing**
Screen Shot 2019-03-17 at 2 08 32 PM

thisisdano:
No preview

Uses 0.025 for the letter-spacing value of `letterspacing 1` to address the issue in #2962 

Fixes #2962

### DIFF
--- a/src/stylesheets/core/_properties.scss
+++ b/src/stylesheets/core/_properties.scss
@@ -289,7 +289,7 @@ $system-properties: (
       'ls-neg-3':	  -.03em,
       'ls-neg-2':	  -.02em,
       'ls-neg-1':	  -.01em,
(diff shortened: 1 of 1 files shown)


thisisdano:
You're right — we should check for non-leading zeroes.

PR #2966, opened 2019-03-18
Add new plus/minus variants

thisisdano:
Adding a new plus/minus variant for testing on `uswds-site`, consistent with the close box introduced in the new mobile banner. In general, this style is a 16px x 16px base rectangle with 2px stroke. 

close box
Screen Shot 2019-03-18 at 10 38 04 AM

plus/minus variant
Screen Shot 2019-03-18 at 10 39 37 AM


maya:
If this is just for testing on the `uswds-site` repo, why not add it there?

Do we intend to redraw all of our icons? Would they be in addition to or replace Font Awesome?

thisisdano:
Yeah, I'll just add to `uswds-site` for now — and not planning to redraw right now. Just looking at a couple case-by-case things at the moment, but we could look at the fontawesome set again at some point and see if we want to make any changes.

thisisdano:
Closing in favor of doing this work directly in `uswds-site`

Issue #2954, opened 2019-03-12
Note Supported Versions of Node


joekarasek:
# Supported Versions of Node

Add documentation for which versions of Node are supported (looks like its up to 8, nothing later). 

## Description

First time attempting to contribute and I had a trouble with `npm install`. I was using node v10. Bumping down to node v8 resolved the issue.

## Steps to reproduce the issue

1. Setup your local environment to be using node v10 or later.
2. Run `npm install` 

## Additional information [optional]

Optionally add a `.nvm` file so that developer who use `nvm` to manage multiple versions of node will get a prompt to change versions.

joekarasek:
Oops, accidentally hit close.

thisisdano:
Starting with USWDS 2.0 we will be using a `.nvmrc` file supporting the current LTS version of Node (10.15.1): https://github.com/uswds/uswds/blob/release-2.0/.nvmrc

maya:
I agree that we should probably have documentation on the what Node version to use as well as `nvm`. Glad you got it working eventually.

maya:
@joekarasek If you're interested in contributing to version 2.0, which is where all our latest work is, use the [release-2.0](https://github.com/uswds/uswds/tree/release-2.0) branch.

saracope:
I put together a Docker image for testing, but it could also help solve this problem. If that seems interesting, I can run it by the devops group to get more eyes on and publish it to the repo.

maya:
I know @toolness also did something similar here: https://github.com/toolness/docker-uswds

toolness:
Hi @maya!  Let me know if you have any questions about that docker-uswds thing, happy to answer anything you need.

saracope:
OH, that looks real nice!

PR #2968, opened 2019-03-18
Update node and npm versions in README [2.0]

maya:
Update node and npm versions to what we currently use in the README. 
Semi-related to #2954.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ And that’s it — you should now be able to copy our code samples into your `i
 
 1. Install `Node/npm`. Below is a link to find the install method that coincides with your operating system:
 
-    - Node v10.15.1+, [Installation guides](https://nodejs.org/en/download/)
+    - Node v10.15.1, [Installation guides](https://nodejs.org/en/download/)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -129,15 +129,15 @@ And that’s it — you should now be able to copy our code samples into your `i
 
 1. Install `Node/npm`. Below is a link to find the install method that coincides with your operating system:
 
-    - Node v4.2.3+, [Installation guides](https://nodejs.org/en/download/)
+    - Node v10.15.1+, [Installation guides](https://nodejs.org/en/download/)
```
maya: Do we know if anything above 10 works? e.g. does node 11 work?

thisisdano: Untested afaik

maya: Updated to be explicitly node 10 in f65e9fb.

saracope:
The circleci config is still using a node 8 image so we should bump that too. @maya  Do you want to make that change in the PR or do is separately?

maya:
@saracope We can do it in a separate PR.

maya:
Also, what's the proper way to do that? Right now, it's `circleci/node:8.10-browsers` does it need to change to `circleci/node:10.15.1-browsers`?

saracope:
Yeah

PR #2969, opened 2019-03-18
Update node and npm versions in README [1.x]

### LINKED ISSUE #2954: Note Supported Versions of Node

maya:
Update node and npm versions to what's currently supported. Before, we wrote `Node v4.2.3+` is supported implying everything above `Node v4.2.3` would be supported. However, @joekarasek noted in #2954 that node 10 was not supported and only node 8 was supported so this updates the README to reflect that.

Fixes #2954.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ And that’s it — you should now be able to copy our code samples into your `i
 
 1. Install `Node/npm`. Below is a link to find the install method that coincides with your operating system:
 
-    - Node v4.2.3+, [Installation guides](https://nodejs.org/en/download/)
+    - Node v8.14.1, [Installation guides](https://nodejs.org/en/download/)
 
     **Note for Windows users:** If you are using Windows and are unfamiliar with `Node` or `npm`, we recommend following [Team Treehouse's tutorial](http://blog.teamtreehouse.com/install-node-js-npm-windows) for more information.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -129,15 +129,15 @@ And that’s it — you should now be able to copy our code samples into your `i
 
 1. Install `Node/npm`. Below is a link to find the install method that coincides with your operating system:
 
-    - Node v4.2.3+, [Installation guides](https://nodejs.org/en/download/)
+    - Node v8.14.1, [Installation guides](https://nodejs.org/en/download/)
```
thisisdano: Is it worth saying that versions _between_ 4.2.3 and 8.14.1 are necessary? Or is it better as is?

maya: I think it's simpler and introduces fewer variables if we stick with one, the version we have in our `.nvmrc` file.

PR #2974, opened 2019-03-20
Update unstyled button and list to BEM

### LINKED ISSUE #2956: Update unstyled lists and buttons to BEM
Instead of our standalone unstyled class for buttons and lists, we should be consistent with our new BEM style and use modifiers:

| old | new
| --- | ---
| `usa-button-unstyled` | `.usa-button.usa-button--unstyled`
| `usa-unstyled-list` | `.usa-list.usa-list--unstyled`

thisisdano:
[Unstyled button preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-update-unstyled-classes/components/detail/buttons--default.html)
[Unstyled list preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-update-unstyled-classes/components/detail/lists.html)
- - -

- Changed `.usa-button-unstyled` to  `.usa-button.usa-button--unstyled` as the unstyled button variant
- Added unstyled button to example code
- Changed `.usa-unstyled-list` to  `.usa-list.usa-list--unstyled` as the unstyled list variant
- Added unstyled list to example code
- Updated `Footer big` component

- - -

Screen Shot 2019-03-20 at 9 16 28 AM

Screen Shot 2019-03-20 at 9 16 43 AM

- - -

Fixes #2956


src/stylesheets/elements/_list.scss
```diff
@@ -27,7 +27,9 @@
 
 // Unstyled lists
 @include override-prose {
-  .usa-unstyled-list {
+  .usa-list--unstyled {
     @include unstyled-list;
   }
 }
+
+// TODO: check lists to see if line height is working ok
```
maya: Has this been checked? Is the comment meant to stay?

src/stylesheets/elements/_buttons.scss
```diff
@@ -195,6 +212,14 @@ $button-stroke: inset 0 0 0 units($theme-button-stroke-width);
...
+.usa-button,
+.usa-button--accent-cool,
+.usa-button--base,
+.usa-button--big,
+.usa-button--inverse,
+.usa-button--outline,
+.usa-button--secondary {
+  &.usa-button--unstyled {
+    @include button-unstyled;
+  }
```
maya: Why is this necessary to add the variant button classes before `.usa-button--unstyled`. It looks like it works without this and keeps the CSS much smaller and shallower.

thisisdano: It was defensive specificity, but I suppose we could remove it

src/stylesheets/elements/_list.scss
thisisdano: Whoops, yeah, we checked and it's ok

src/stylesheets/elements/_buttons.scss
thisisdano: Actually, it does fix some specificity stuff around hover/active/focus

maya: I'm not able to reproduce when I check the button states? Is there something else I should do to see the issue?

maya:
I assume this is OK, but noticed the text on the unstyled big button is also large.

Screen Shot 2019-03-20 at 10 43 35 AM

thisisdano:
That was intentional, but maybe it shouldn't be?

src/stylesheets/elements/_buttons.scss
maya: I see one issue with the outline button on hover but can probably be resolved individually by removing the outline.

thisisdano:
As for hover issues — I see it in the `outline` and `accent-cool` variants. The `outline` version gets an outline, and the `accent-cool` version changes smoothing modes.

src/stylesheets/elements/_buttons.scss
maya: Do you want to resolve [those issues](https://github.com/uswds/uswds/pull/2974#issuecomment-474956088) individually, chain only those classes, or chain all classes (even if most don't need it)?

thisisdano: I'm inclined to leave it as is and apply the specificity across the board, if only to make it clear that the `unstyled` variant _should_ always override any other variant styling. But I don't feel super strongly about this.

maya: Made a few tweaks to the mixin in #2976 to make it more robust and fix these two issues. If that looks OK, we can merge it in.

thisisdano: LGTM

PR #2979, opened 2019-03-20
Update CircleCI node image to 10.15.1

maya:
As mentioned in https://github.com/uswds/uswds/pull/2968#issuecomment-475013277.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 references:
   container: &container
     docker:
-      - image: circleci/node:8.10-browsers
+      - image: circleci/node:10.15.1-browsers
         environment:
           CHROME_PATH: /usr/bin/google-chrome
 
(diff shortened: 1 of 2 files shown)


maya:
@saracope tried to update CircleCI node image like you suggested but looks like it made the Circle tests fail.

.circleci/config.yml
```diff
@@ -3,15 +3,14 @@ version: 2
...
+      - image: circleci/node:10.15.1-browsers
         environment:
           CHROME_PATH: /usr/bin/google-chrome
 
   restore_deps: &restore_deps
     restore_cache:
       keys:
-        - v1-uswds-dependencies-{{ checksum "package-lock.json" }}
-        - v1-uswds-dependencies-
+        - v2-uswds-dependencies-{{ checksum "package-lock.json" }}
```
maya: Working now after forcing the cache to update by changing the key. Removed `v1-uswds-dependencies-` bc it was using that cache even if `v2-uswds-dependencies-{{ checksum "package-lock.json" }}` was different, but would like someone to check this.

PR #2983, opened 2019-03-21
Apply :focus behavior to focusable elements only

thisisdano:
## Description
This pull request explicitly lists elements that are focusable and applies the focus mixin only to those elements, rather than applying the rule globally. The immediate goal of this is to patch the bug as described in Issue #2928.

Resolves https://github.com/uswds/uswds/issues/2928

## Additional information

- Even though this issue was discovered only in IE, I feel that this is also the more correct solution than applying the `:focus` rule globally.
- This solution is based on the solution described in this SO thread, https://stackoverflow.com/a/30753870.
- jQuery's `:focusable` selector also has interesting info, https://api.jqueryui.com/focusable-selector/.
- See issue for more context

    - May have messed up the `[Website]` part of this.

### DIFF
--- a/src/stylesheets/core/_base.scss
+++ b/src/stylesheets/core/_base.scss
@@ -30,6 +30,7 @@ body {
   display: none !important; /* stylelint-disable-line declaration-no-important */
 }
 
+// Apply :focus behavior to focusable elements only (For IE 11)
 input,
 select,
 textarea,


thisisdano:
Thanks @ncksllvn — This seems like a reasonable change, implemented well. I can confirm that this is an issue in IE11, and that it's even more of an issue with USWDS 2.0, since the focus highlight is stronger.

thisisdano:
[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/dw-limit-focus-develop/components/preview/layout--landing.html)

thisisdano:
### Before
Screen Shot 2019-03-21 at 12 52 55 PM

### After
Screen Shot 2019-03-21 at 12 53 08 PM

- - -

I haven't found, nor can I think of, any other instances where a focus indicator would be necessary that aren't included in the scope of this change.

Tested on Windows 10 / IE11 via saucelabs

@maya can you take a look at this?

thisisdano:
@ncksllvn I see you didn't include the `&:not([tabindex='-1'])` from the original SO technique. Is that correct? Can you give some context?

thisisdano:
nm — I see the answer in the SO thread...

thisisdano:
Anything we're missing? Perhaps `audio[controls]` and `video[controls]`?
Are there any instances where traditionally non-tabbable elements should receive focus?

Looking at this list: https://allyjs.io/data-tables/focusable.html
Should we be concerned about the difference between `tabbable` and `focusable`?

maya:
It's not building on Federalist now with this error:
```
ERROR:  Error installing jekyll:
	jekyll-watch requires Ruby version >= 2.4.0.
```

PR #3005, opened 2019-03-26
Add setup snyk command to deploy

maya:
Snyk is not installed before we run `snyk monitor --org=uswds` in the deploy step.

Also, removes `v1-uswds-dependencies-` cache key.


maya:
Same

Issue #2992, opened 2019-03-25
Use 2.0 color palette on icons


maya:
Icons (SVG and PNGs) still use 1.x colors.

maya:
Fixed in #3007.

PR #3007, opened 2019-03-27
Update icon colors SVG to 2.0 color palette and remove fallback PNGs

maya:
Fixes part of #2992 - removes PNGs since we no longer support the bowsers that require them.

Gov banner - for the outline and icon, I went with:
- Blue https: 
  - Icon: blue 50
  - Outline: blue 60v
- Green dot gov: 
  - Icon: green 40v
  - Outline: green 50v


thisisdano:
May as well update the PNGs as well. Any reason why not?

maya:
Don't really have an efficient way of doing that? Do you want to take it on?

thisisdano:
I updated the original PR description to also note that we removed the PNGs

Issue #2866, opened 2018-11-02
Compile errors when using dart Sass


thisisdano:
While everything has been compiling OK with our standard gulp Sass — we're seeing errors in the stricter and more official Dart Sass.

in `elements/_inputs.scss`

```
Error: `4` is not a valid spacing unit token. Valid spacing unit tokens: 1px, 2px, 05, 1, 105, 2, 205, 3, neg-1px, neg-2px, neg-05, neg-1, neg-105, neg-2, neg-205, neg-3, 4, 5, 6, 7, 8, 9, 10, 15, neg-4, neg-5, neg-6, neg-7, neg-8, neg-9, neg-10, neg-15, card, card-lg, mobile, mobile-lg, tablet, tablet-lg, desktop, desktop-lg, widescreen, 0, auto
        on line 941 of /Users/danielwilliams/code/uswds/uswds/src/stylesheets/core/_functions.scss
        from line 48 of /Users/danielwilliams/code/uswds/uswds/src/stylesheets/uswds.scss
```

and in `components/_banner.scss`

```
Error: $string: rgba(0, 0, 0, 0) is not a string for `quote'
        on line 122 of /Users/danielwilliams/code/uswds/uswds/src/stylesheets/core/_functions.scss
        from line 54 of /Users/danielwilliams/code/uswds/uswds/src/stylesheets/uswds.scss
```

PR #3009, opened 2019-03-29
Allow values of `.5` and `-.5` without leading zeros

thisisdano:
External contribution: originally in https://github.com/uswds/uswds/pull/3008
Thanks @dmethvin-gov!

[Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/external-allow-leading-zeros/)

- - -

## Description

The [USDS website](https://github.com/usds/website) uses Jekyll and its build process generates numbers like `.5` rather than `0.5`. Those are valid in CSS so they should be accepted.

I hope this is the right branch for the PR, it would be good to get this into final 2.0 so Jekyll sites will work.

## Additional information

Related to #2866 and #2867, should fix https://github.com/uswds/uswds-sandbox/issues/18.

I attempted to run `npm test` but got cascading errors once it reached "fractal component". The same errors occur on 2.0/beta7 so I don't think this change caused them.


thisisdano:
This all looks good to me, works as expected in testing, and should go into the 2.0 release.

PR #3010, opened 2019-04-01
Move aria search to form. Remove styling on aria role.

fskreuz:
## Description

PR removes `<div role="search">` from the search component and moves `role="search"` to the enclosing form instead. Reasoning behind the change:

- `<form role="search">` is enough to describe a search form.
- Most examples I've seen on `role="search"` put it on `<form>`.
    - https://www.w3.org/TR/wai-aria-practices/examples/landmarks/search.html
    - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role
- The same appearance can be achieved without the extra `<div>`.

### DIFF
--- a/src/components/search/search--header.njk
+++ b/src/components/search/search--header.njk
@@ -1,7 +1,5 @@
-<form class="usa-search usa-search--small {% if search_js %} js-search-form{% endif %}">
-  <div role="search">
-    <label class="usa-sr-only" for="{{ id_prefix }}search-field-small">Search small</label>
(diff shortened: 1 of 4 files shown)


maya:
Hi @fskreuz, thanks for opening this PR. The reason the search role is put on the div and not the form is to improve the accessibility of it as detailed here: http://adrianroselli.com/2015/08/where-to-put-your-search-role.html and https://www.a11ymatters.com/pattern/accessible-search/#6--where-to-add-rolesearch

At this point, we will leave the role as is but thank you for your contribution and let us know if you find anything else.

PR #3012, opened 2019-04-03
Update outline() and overflow-[xy]() mixins

### LINKED ISSUE #3011: A few mixin errors
There are a few issues I am getting with some of the mixins in `2.0`

### Outline

**When I use** `@include u-outline('05');`
I am getting the error — 
```
`05` is not a valid outline value.
```

**When I use** `@include u-outline('1px');`
I am getting the error — 
```
`1px` is not a valid outline value.
```

**When I use** `@include u-outline(1);`
I am getting the error — 
```
1 is not a valid outline width. Accepted values: 1px, 2px, 0, 05
```

---

### Overflow

When I use `@include u-overflow-x('hidden');`
I am getting the error — 
```
o mixin named u-overflow-x
```

thisisdano:
- fixed a bug in the `u-outline()` mixin where it wasn't accepting strings properly
- added `u-overflow-x()` and `u-overflow-y()` mixins to match utilities

- - -

Fixes #3011

### DIFF
--- a/src/stylesheets/core/_properties.scss
+++ b/src/stylesheets/core/_properties.scss
@@ -501,7 +501,10 @@ $system-properties: (
     standard: map-collect(
       map-get($system-spacing, 'smaller'),
       map-get($partial-values, 'zero-zero'),
-      ('05': spacing-multiple(0.5))
+      (
+        '05': spacing-multiple(0.5),
+        1: spacing-multiple(1)
+      )
     ),
     extended: (),
   ),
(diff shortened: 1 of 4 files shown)


src/stylesheets/core/_properties.scss
```diff
@@ -501,7 +501,9 @@ $system-properties: (
     standard: map-collect(
       map-get($system-spacing, 'smaller'),
       map-get($partial-values, 'zero-zero'),
-      ('05': spacing-multiple(0.5))
+      (
+        '05': spacing-multiple(0.5),
```
thisisdano: wrote this in expanded format to better accept new custom values if we decide to accept, say, `1` unit

PR #3013, opened 2019-04-04
Support “grid-offset-none” for each breakpoint

robertromore:
Currently unable to use responsive class names with "grid-offset-none", such as "desktop:grid-offset-none".

### DIFF
--- a/src/stylesheets/core/_layout-grid.scss
+++ b/src/stylesheets/core/_layout-grid.scss
@@ -146,6 +146,10 @@ $namespace-grid: ns('grid');
         }
       }
     }
+    .#{$mq-key}\:#{$namespace-grid}offset-none {
+      $props: append-important($grid-global, none);
+      @include grid-offset($props);
+    }
   }
 }
 /* stylelint-enable */


src/stylesheets/core/_layout-grid.scss
```diff
@@ -146,6 +146,10 @@ $namespace-grid: ns('grid');
         }
       }
     }
+    .#{$mq-key}\:#{$namespace-grid}offset-none {
```
thisisdano: You'd need to wrap this in a `@include at-media($mq-key) {}` wrapper here to not only output the class, but the proper media query as well.

thisisdano:
Thanks for bringing this up and submitting a pull request. I can confirm that the `offset-none` responsive variants weren't outputting. This fix is close, but is missing the `@include at-media($mq-key) {}` wrapper around the `offset-none` code. As is, it'll output the proper classes, but not scope them to the proper media query.

For simplicity's sake, I've opened a new internal PR to fix this, which is essentially your fix plus the media query.

I appreciate your contribution.

- - -
Closed in favor of https://github.com/uswds/uswds/pull/3014

PR #3022, opened 2019-04-06
Update SVG hover color to primary-darker, was primary-dark

maya:
To confirm: link hover state color is primary-darker and not primary-dark?

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/update-svg-hover-color/components/detail/links.html)


thisisdano:
Now that you bring it up, I don't think there's a good reason why it's `darker` and not `dark` — the primary button uses `dark` for hover so I think we might want to use `dark` instead.

maya:
OK, will close this and plan to change the link hover color instead. Glad I checked!

PR #3041, opened 2019-04-15
Update Merriweather font to 2.005

maya:
- Updates Merriweather font to 2.005 (using latest files in master)
- Updates font face declaration names

Lots of good, new stuff from 2.002 to 2.005 (we were at 2.001): https://github.com/SorkinType/Merriweather/#changelog

@thisisdano Is cap height still the same? (I think it is)

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/mb-update-merriweather/components/detail/typesetting.html)

/cc @EbenSorkin


EbenSorkin:
I know that Marc at Google altered the Win vertical metrics a bit. The TTFAutohinting may have also been updated/changed. Either thing could make specific pt em or px rendering alter from A to B. I'll be interested in hearing about any bugs you find or suggestions/wishes. FYI - In theory we are making widths and optical sizes next.

thisisdano:
Thanks for the update @EbenSorkin — we like the revisions!

@maya I updated the fractal site with a new normalization viewer and updated the normalization values a bit to balance the faces.

viewer: https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/mb-update-merriweather/components/detail/normalization-test.html

thisisdano:
We did not subset this new version of Merriweather. Perhaps we should? 
@EbenSorkin how might you propose creating a Latin subset with all relevant opentype features?

We will also need to update the license with the proper version and links. 
@EbenSorkin will this release be tagged for a GitHub release like [2.001](https://github.com/SorkinType/Merriweather/releases/tag/v2.001)?

maya:
Updated license info here in 6a7c8a6. Removed the add'l formats line in daa4903 bc all formats were taken directly from the Merriweather repo.

Correct, this version is not subsetted, it's taken directly from the files in GitHub. @thisisdano let me know if you will be subsetting. Otherwise, the info in the license is not accurate (says it was subsetted).

@EbenSorkin mentioned he would publish the release on GitHub at some point.

maya:
Unrelated to this, but should we move "Type spacing" to the "Test" section as well (since normalization is there)?

EbenSorkin:
Dan, is there a specific subset of targeted glyphs in a set you would like to see? Is there a link to it?

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

maya:
I don't know if we should follow this as a model, but here is the subset of Source Sans Pro that was created for us: https://github.com/miguelsousa/source-sans-pro-subset

EbenSorkin:
I can extract the glyph list from their font if that's useful. But I am curious, what was the context for that subset of their font? Was it requested by GSA? Was there a target size for the data footprint of the font? It seems like understanding that stuff would be the ideal thing.

maya:
The goal behind the subsetting of their font was to reduce the fonts' footprint/size and increase character support. Here is the context: #802 #848 #2934. It was not requested by GSA but offered by @miguelsousa.

We'll have to follow up regarding any target sizes but don't believe we have anything set.

EbenSorkin:
We don't have any extensive legal text to cut out. There may be languages
that you don't care about but I tend to like the idea that if someone is
using an auto translator service they can keep the design while swapping
languages even if that's Mongolian, Hungarian or Basque.

But on the other hand supporting cyrillic isn't a priority and we cut that
out we can probably drop 1/4 of the file size right there.

The single largest file size benefit available would probably be in using
the variable font version of the family. Also, because we are likely to
offer widths and optical sizes this year the typographic utility range and
power of the family will increase in a very big way. If you have built
support for the variable font weight axis then adding two more axis will be
technically simple. Taking advantage of the new typographic options will be
of course be more complex.

-e.

thisisdano:
@EbenSorkin it needn't happen for this PR, but I think the best way forward is to try and match the glyph list and features @miguelsousa used for his SSP subset. Does that seem possible?

We want to move in the direction of variable font support, but it'll take a sec for browser support reasons. (It'll make a big difference when it happens, though.)

EbenSorkin:
I'll examine what that subset is like then and let you know.

-e.

thisisdano:
@EbenSorkin Any update on the subset front? I'd like to get your new files in our next release which would mean having subsetted files by this Friday. Does that seem reasonable?

EbenSorkin:
I currently am assembling the team that will make Optical variations and
widths for Merriweather and getting the sample masters approved. I won't be
able to sub set this for probably a few weeks.

As a side note - I was never able to sort out the intended purpose other
than to hear that Adobe did it. I get that Adobe can be wise about things
at times and I get that it makes it smaller and that could be better. But
as the glyph sets of the projects differ it seems potentially problematic
to just go with the adobe subset list. Or it may be fine. In any event I
would still appreciate understanding the intent behind doing it a bit
better before I do it.

-e.

thisisdano:
[_time passes_]

Circling back on this. I would love to include the latest and greatest version of Merriweather in subsequent USWDS releases. Currently, the `woff2` files come in at ~100 KB for each style. This compares to ~30 KB for Public Sans, and ~20 KB for both Source Sans Pro and Roboto Mono. 

How could we get Merriweather in the ballpark of a ~30-40 KB performance budget? I'd like to this that the SSP subset and feature set is a reasonable starting point, but I'm open to other suggestions.

EbenSorkin:
One way would be to remove Cyrillic if you don't use it. Another way would
be to get rid of small caps. Subsetting Latin to selectively remove support
for some languages can help too. Similarly if you don't use them some
number styles could also be removed.

It is painful to say but some/all kerning could be removed as well.

I should note that there is a second round of variable development coming
with widths and optical sizes which is I think going to make Merriweather
3-5x more useful. IDK what the total file size would be for a final build
but it could also be subset. Might the browser set you need to support
going to prohibit variable?

I'd love to chat at some point. Is that possible?

-e.

thisisdano:
Don't remove the kerning! :)

And yeah, I'd be happy to set up some time to talk about the details. (I tried to reach out via the email address at http://sorkintype.com/contact.html, but my message bounced back...)

> One way would be to remove Cyrillic if you don't use it. Another way would be to get rid of small caps. Subsetting Latin to selectively remove support for some languages can help too. Similarly if you don't use them some number styles could also be removed.

I would be tentatively inclined to pursue any of these solutions. Especially Cyrillic and small caps. Perhaps looking at the Public Sans feature set would be useful?

EbenSorkin:
Sure.

I need to meet my obligation to the new Variable project but I can
experiment with a custom build after that.

-e.

EbenSorkin:
Also are you interested in both a variable and a conventional set of web
fonts ( TTF and WOFF no?) or just the conventional?

Is there any budget for this?

-e.

thisisdano:
Sadly, we can't yet support variable fonts, just TTF, WOFF, and WOFF2. And also, sadly, there is no budget.

EbenSorkin:
I’ll see what I can do then when this latest variable push is done. I do want to suggest that at that time you may want to try the optical axis and let me know if there maybe should be a new standard which is optically slightly bigger and maybe a bit narrower for efficiency.

PR #3044, opened 2019-04-18
[WIP] Add responsive headings

maya:
- Added setting: `$theme-responsive-headings` and set to `true` for testing purposes, may change to `false` if this goes into production.
- Added responsive style for `display`, `h1`, `h2`, `h3`. Left `h4`-`h6` mixins alone, but can update the mixins to match the previous ones.
  - The mixins seem kind of repetitive and wondered if there's a better solution out there. 

[😎 Preview](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/add-responsive-headings/components/detail/typesetting.html)

/cc @hursey013

### DIFF
--- a/src/components/01-type/typesetting.njk
+++ b/src/components/01-type/typesetting.njk
@@ -8,6 +8,7 @@
 
   <h6 class="usa-heading-alt">Spacing</h6>
   <section class="example-spacing usa-prose margin-top-1 font-sans-sm measure-3 bg-primary-lighter">
+    <p class="usa-display">Display text</p>
     <h1>Page heading</h1>
     <p class="usa-intro">Great Smoky Mountains National Park straddles the border of North Carolina and Tennessee.</p>
 
--- a/src/stylesheets/core/mixins/_typography.scss
+++ b/src/stylesheets/core/mixins/_typography.scss
@@ -87,11 +87,27 @@ Sets:
 }
 
 @mixin display {
-  @include typeset(
-    'heading',
-    $theme-display-font-size,
-    $theme-heading-line-height
-  );
+  @if $theme-responsive-headings == true {
+    @include typeset(
+      'heading',
+      $theme-display-mobile-font-size,
+      $theme-heading-line-height
+    );
+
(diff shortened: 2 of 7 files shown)


hursey013:
I'll defer to the team on whether you all think it's a good idea, but I love this!

I think my only real feedback would be on the necessity of the `$theme-responsive-headings` boolean. Since all of the size for mobile are included in the settings by default, I would just expect those to take precedent and would not (myself) have a need for a global on/off toggle.

maya:
Thanks @hursey013! I'm using the boolean, bc I don't know if most users would necessarily want responsive headings and may end up setting this to false by default. Therefore you'd need to set the setting to true in order to "activate" it. But I still wanted to have those responsive heading defaults in the variables.

I'm also wondering if there's any way to make these mixins more DRY, b/c I feel like they are doing the same thing (e.g. with a loop or Sass map)? Any ideas? Or do folks like it this way?

thisisdano:
I like the idea of not tying this to a bool setting and maybe just having `$theme-[type]-mobile-font-size` and `$theme-[type]-font-size` — the mixin could check for `false` (or equivalence) and output accordingly. The question is what our default should be — and if we want to go down the path of providing more granular font style controls in settings. (I tend to think yes.)

hursey013:
I guess there is another aspect to consider with this. If I have an `h2` that is visually styled as say an `h5` with something like `.font-heading-sm` I would want that heading to also be responsive, but that could get really messy since there is no direct mapping from `.font-heading-sm` to heading sizes. There would almost need to be utility classes for headings like `.font-h1` that could then have the same responsive scale applied to it as a normal `h1`. Does that make sense?

maya:
Closing this as this is stale and I'm no longer on the project. The team may need to research this further, e.g. not sure if it's validated that the headings are too large on mobile devices. In some cases, people may even prefer text (like body text) to be larger on mobile.

mejiaj:
This might be useful in the future:
https://www.stevefenton.co.uk/2020/04/css-clamp-the-goldilocks-of-css-math-functions/

PR #3072, opened 2019-06-20
Add @imports to SCSS

turnerhayes:
## Description

I added `@import`s to the SCSS files and updated the image url paths such that the individual SCSS files can be used on their own.

## Context

I'm working on React components for USWDS components. While I could import the full SCSS file (or the compiled CSS), importing only the files needed for the component, it would help Webpack do tree-shaking to minimize the amount of CSS.

# Tests

`npm run test:visual` fails, but I tried it in the base repo and it seems to fail in exactly the same way, so this seems like a net no-diff.


thisisdano:
Fascinating and potentially useful. I will merge your work into an internal branch for a bit more testing. I'll ping you with questions on that branch later.

thisisdano:
Merging into `ext-component-imports` for internal testing.

PR #3075, opened 2019-06-27
[Banner / Nav Primary] Prevent FOUC caused by hidden elements when page loads

w00fz:
## Description

When loading a page where a banner/navigation are rendered, the hidden content is showing until the JS kicks in and adds the `hidden` attribute which hides it.
This is causing the hidden elements to flash on the page for a moment and gives the impression of [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) does. 

## Additional information

* This issue is particularly more evident in Chrome
* The issue can be also reproduced on https://designsystem.digital.gov/
    1. Load site with Chrome
    2. Open Dev Tools
    3. Hold down left mouse button on the Refresh button and select "Empty Cache and Hard Reload" (this is in order to get a non-cached version which makes the FOUC more evident)
        empty-reload

### DIFF
--- a/src/components/08-navigation/_nav-primary.njk
+++ b/src/components/08-navigation/_nav-primary.njk
@@ -4,7 +4,7 @@
   <li class="usa-nav__primary-item">
     {% if link.links -%}
     <button class="usa-accordion__button usa-nav__link{% if link.is_current %}  usa-current{% endif %}" aria-expanded="false" aria-controls="{{ nav.id_prefix }}{{ link.id }}"><span>{{ link.text }}</span></button>
-    <{% if nav.mega %}div{% else %}ul{% endif %} id="{{ nav.id_prefix }}{{ link.id }}" class="usa-nav__submenu{% if nav.mega %} usa-megamenu{% endif %}">
+    <{% if nav.mega %}div{% else %}ul{% endif %} id="{{ nav.id_prefix }}{{ link.id }}" class="usa-nav__submenu{% if nav.mega %} usa-megamenu{% endif %}" hidden>
       {%- if nav.id_prefix == "basic-" or nav.id_prefix == "extended-" -%}
         {%- for child in link.links -%}
           {%- if loop.index < 4 -%}
--- a/src/components/banner/banner.njk
+++ b/src/components/banner/banner.njk
@@ -10,7 +10,7 @@
           <p class="usa-banner__header-action" aria-hidden="true">Here’s how you know</p>
         </div>
         <button class="usa-accordion__button usa-banner__button"
-          aria-expanded="false" aria-controls="gov-banner">
+          aria-expanded="false" aria-controls="gov-banner" hidden>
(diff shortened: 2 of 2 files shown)


thisisdano:
I believe the reason our component behaves as it does is so any content that is shows/hidden by javascript defaults to visible if js is blocked or does not load. In our current implementation, this does result in some FOUCy stuff, but we consider that a reasonable cost of our desire to make site content accessible regardless of javascript. 

Given our constraint, might there be another way to minimize FOUC?

w00fz:
Thanks for your reply @thisisdano ,

I was actually under the impression that the reasoning for it was to make the banner/accordion more accessible in case of failing JS, however, I then realized that there are some implementations where the JS is kind of vital.

1. For instance, none of the accessibility aria work if JS fails because you do need JS to toggle the expanded/collapsed ([reference](https://github.com/uswds/uswds/blob/develop/src/js/utils/toggle.js)).

2. Another issue I noticed is that without JS you wouldn't be able to navigate the site because the dropdowns cannot be expanded via CSS and heavily rely on the JS ([reference](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/release-2.0.3/components/preview/header--extended.html)).

3. Most form validations also don't work ([reference](https://github.com/uswds/uswds/blob/develop/src/js/utils/validate-input.js), [example](https://components.designsystem.digital.gov/components/detail/validation.html)).

Unfortunately, I can't think or propose any better solution to prevent FOUC on the banner/accordion, the structure itself is not helping either as with CSS3 there's only so much you can target. 

If the trigger and the panel were siblings, a `:focus` plus a `~` target could have helped to make the toggling JS-free. But of course, that would get quite complicated and would imply rethinking the templating.

Having JS not available nowadays is very rare, less rare is having it to break but thankfully there are unit tests.

It really comes down to the project determining how much is compromisable and how far the JS-free should be pushed.

I would personally give more priority to no flashing content over the possibility of JS failing and not being able to see that content or prevent navigating to other pages: the JS failing is something you can (and should) catch and fix, the FOUC is there always for everyone.

thisisdano:
Thanks for responding — I appreciate it. I do think that there are some reasonable pros and cons to consider in this instance. There is a case to be made that the content of the gov banner is not critical content and needn't be subject to the functional-without-js requirement. And if we see that the FOUC has a real usability effect then we might move toward hidden-by-default for this component. 

- As for the accordions, they are open by default (with proper ARIA markup) even if toggle.js is not present.
- Yes, dropdown menus will also fail without js, _but_ the parent menu item link will work — perhaps it is good and reasonable guidance to assure that no pages are _only_ navigable via dropdown menus.
- And yeah... form validation will fail, but that may be a 🤷‍♀what-can-you-do situation.

We could do some more work to determine just how relevant the no-js situation actually is. My understanding is that no-js is more of an issue in government, as there are more cases where agency and mission context might force no-js with its web clients.

thisisdano:
This is still an annoying problem, but we won't be solving it in this way.

PR #3084, opened 2019-08-09
Add @imports to SCSS

thisisdano:
Originally submitted by @turnerhayes in https://github.com/uswds/uswds/pull/3072

- - -

## Description

I added `@import`s to the SCSS files and updated the image url paths such that the individual SCSS files can be used on their own.

## Context

I'm working on React components for USWDS components. While I could import the full SCSS file (or the compiled CSS), importing only the files needed for the component, it would help Webpack do tree-shaking to minimize the amount of CSS.

# Tests

`npm run test:visual` fails, but I tried it in the base repo and it seems to fail in exactly the same way, so this seems like a net no-diff.


config/gulp/sass.js
```diff
@@ -95,6 +91,10 @@ gulp.task('sass', gulp.series('copy-vendor-sass',
           .on('error', sass.logError),
       )
       .pipe(postcss(pluginsProcess))
+      .pipe(replace(
+        /\buswds @version\b/g,
+        `uswds v${pkg.version}`,
+      ))
```
thisisdano: OK, moving the string replace to the end makes sense.

src/stylesheets/core/mixins/_all.scss
```diff
@@ -11,7 +11,7 @@
 @import 'utilities/cursor';
 @import 'utilities/display';
 @import 'utilities/flex';
-@import 'utilities/flex-direction';
+// @import 'utilities/flex-direction';
```
thisisdano: What's the point of removing these?

thisisdano:
So — in reviewing this PR, we decided to make some changes to the approach. We had four fundamental priorities:

1. Use a component-based approach. 
2. Reduce the time it takes to build the CSS.
3. Reduce the amount of final CSS output.
3. Make it simple to minimize the USWDS footprint.

To do this, we decided to implement a solution that was less focused on the _internal_ dependencies of each stylesheet file and more focussed on the style dependencies of the component markup itself. That is, less on the functions and mixins that don't emit code directly and more on the actual code required to fully style a particular component. 

For example, `stylesheets/components/search` requires the following to render as from USWDS Sass:
```scss
@import "../core/base";
@import "../core/functions";
@import "../core/mixins/border-box-sizing";
@import "../core/mixins/add-background-svg";
@import "../core/mixins/clearfix";
@import "../core/mixins/typography";
@import "../core/mixins/at-media";
@import "../core/mixins/utilities/padding";
@import "../core/mixins/screen-reader";
@import "../settings/settings-components";
```

But that code alone is not sufficient for `components/search` to render. This is the markup for the basic search component:
```html
<form class="usa-search">
  <div role="search">
    <label class="usa-sr-only" for="search-field">Search</label>
    <input class="usa-input" id="search-field" type="search" name="search">
    <button class="usa-button" type="submit">
      <span class="usa-search__submit-text">Search</span>
    </button>
  </div>
</form>
```

This markup also needs:
- `usa-sr-only`
- `usa-input`
- `usa-button`
- _in addition to_ the `usa-search` source styles

We decided to focus this PR on providing a way to track the _component_ dependencies rather than the _Sass_ dependencies.

## Introducing packages
To do this, we're expanding the idea of **packages** that already existed in USWDS 2.0. The USWDS 2.0 version of packages allowed users to include only parts of the USWDS codebase by using `@import`ing specific files within `stylesheets/packages`. (See [the theme stylesheet](https://github.com/uswds/uswds/blob/develop/src/stylesheets/theme/styles.scss) for an example of this.) The initial version of packages allowed importing only the fonts, the components, the utilities, or the layout grid. **Packages in USWDS 2.1 allows a much more granular and modular component-based implementation of packages.**

In USWDS 2.1 we consider a **package** to be a discrete unit of functionality — typically a component, as defined on [the USWDS components pages](https://designsystem.digital.gov/components/). These packages are kept in `stylesheets/packages` and they point to the stylesheet dependencies and source code. For example, the `usa-search` package looks like this:

```scss
// dependencies
@import '../base/accessibility';
@import '../elements/form-controls/global';
@import '../elements/form-controls/text-input';
@import 'usa-button';

// src
@import '../components/search';
```

Now, if users want to use the `usa-search` component, they can do so by `@import`ing `packages/usa-search` in their stylesheet.

## `required` and `global` package dependencies
USWDS packages attempt to improve the performance of all this importing by requiring users `@import` certain non–code emitting internal Sass required of every component once. (This is Sass like mixins, settings, and functions.) These are collected in the `required` package.

`packages/required`:
```scss
@import '../settings/settings-general';
@import '../settings/settings-typography';
@import '../settings/settings-color';
@import '../settings/settings-spacing';
@import '../settings/settings-utilities';
@import '../settings/settings-components';
@import '../core/functions';
@import '../core/system-tokens';
@import '../core/variables';
@import '../core/properties';
@import '../core/mixins/all';
@import '../core/placeholders/all';
```

There's also a set of basic code-emitting global styles necessary for the best performance of components. These are collected in the `global` package. 

`packages/global`:
```scss
@import '../lib/normalize';
@import '../global/font-face';
@import '../global/focus';
@import '../global/sizing';
@import '../global/typography';
```

## Using packages
Use packages as you would any USWDS `import`: import `required`, `global` and any individual functional packages after importing your project's theme settings, and before importing your project's custom styles. For example (assuming `'./node_modules/uswds/dist/scss'` in your `sass.includePaths`) the following file would import only the USWDS code necessary to render the banner and footer components:

```scss
// styles.scss
// import your project theme settings
@import 'uswds-theme-general';
@import 'uswds-theme-typography';
@import 'uswds-theme-spacing';
@import 'uswds-theme-color';
@import 'uswds-theme-utilities';

// then any packages
@import 'packages/required';
@import 'packages/global';
@import 'packages/usa-banner';
@import 'packages/usa-footer';

// then any custom Sass
@import 'my-custom-sass';
```

The current method of using `@import uswds` to import everything has not changed.

## Reducing footprint with packages
Using packages to subset your CSS can result in a dramatic reduction of your overall USWDS footprint. Using packages can also reduce the build time of your project.

| package | .css | .min | .zip
| --- | --- | --- | ---
**default USWDS** | **359** | **271** | **40 KB**
form-controls | 18 | 15 |  3
form-templates | 60 | 47 | 6
typography | 17 | 15 | 3
validation | 32 | 28 | 5
usa-accordion | 13 | 12 | 2
usa-alert | 13 | 12 | 2
usa-banner | 50 | 38 | 5
usa-button | 19 | 16 | 3
usa-checklist | 11 | 9 | 2
usa-footer | 59 | 46 | 6
usa-header | 113 | 63 | 9
usa-media-block | 10 | 9 | 2
usa-search | 29 | 25 | 4
usa-sidenav | 12 | 10 | 2
usa-skipnav | 10 | 9 | 2
usa-table | 11 |  9 | 2
usa-tag | 10 | 9 | 2
usa-header + usa-footer | 162 | 72 | 10
usa-banner + usa-footer | 99 | 55 | 7

## Other changes in this PR
- Reorganized a lot of the source files to match the new model
- Small change to the markup of the slim footer to prevent USWDS utilities from appearing in production components.

turnerhayes:
That's great, I'm so glad you were able to do a more sane approach to my quick-and-dirty hackish thing! Thanks for the update!

Issue #3124, opened 2019-10-03
Footer issues


thisisdano:
See https://designsystem.digital.gov/components/footer/

<img width="945" alt="Screen Shot 2019-10-03 at 12 31 04 PM" src="https://user-images.githubusercontent.com/11464021/66157958-b80c7980-e5d9-11e9-8893-dd10a9dd5348.png">

- [ ] We explicitly add a `mobile-lg:grid-gap-0` to the contact info grid, but that seems to be a mistake.

- - -

- [ ] If utilities aren't set to use important `usa-footer__contact-info.grid-row` will be overridden CSS to `display: inline-block` — we need to constrain this behavior to the slim footer only.

- - -

- [ ] `.usa-footer__logo-img` needs an explicit `width: 100%`set

- - -

- [ ] Big footer open carets are too big after svg refresh...

<img width="370" alt="Screen Shot 2019-10-03 at 2 19 08 PM" src="https://user-images.githubusercontent.com/11464021/66165091-d4fc7900-e5e8-11e9-881d-1625277e9290.png">

- - -

- [ ] Big footer should be able to accept nav items that don't contain sub-items, just as the primary nav can be a combination of simple links and links that trigger menus.

- - -

- [ ] Use `auto` for default footer link spacing `@desktop` to improve horizontal spacing

<img width="1020" alt="Screen Shot 2019-10-03 at 3 35 58 PM" src="https://user-images.githubusercontent.com/11464021/66168928-86081100-e5f3-11e9-8d6e-332d13e72e74.png">

mejiaj:
> If utilities aren't set to use important usa-footer__contact-info.grid-row will be overridden CSS to display: inline-block — we need to constrain this behavior to the slim footer only.

Just to confirm this means:
```css
.usa-footer__contact-info {
    display: inline-block;
    line-height: 1.1;
}
```

Should be:
```css
.usa-footer__contact-info {
 ... all current styles minus `display: inline-block;`
}

.usa-footer--slim .usa-footer__contact-info {
  display: inline-block;
}
```

--- 

>.usa-footer__logo-img needs an explicit width: 100%set

What issue is this causing and where? I assume this is an IE11 issue?

thisisdano:
Fixed in above PRs

PR #3143, opened 2019-10-31
develop+3124: Update footers.

mejiaj:
## Description

fixes following issues from #3124:

- [x] Remove `mobile-lg:grid-gap-0` from `usa-footer__address > .grid-row.grid-gap` on 
`.usa-footer--slim`

image

---

- [x] Set `display: inline-block` to `.usa-footer__contact-info` on `.usa-footer--slim` vs default

image

_We might not even need `display: inline-block` to be set at all._

---

- [x] Fix big carets on collapsible items in `.usa-footer--big`

image

This was a big change, I ended up using a pseudoelement (instead of tweaking the background sizes/positioning) to get the sizing right. This approach was tested and worked in IE11 as well. 

I chose to go this route because:
1. It saves us from nudging the spacing on collapsed vs expanded.
2. Keeps assets untouched

Now it depends on `.usa-footer__primary-content--collapsible` and `.usa-footer__primary-content--collapsible.hidden`. That way the arrow isn't added just because it's from the `--big` variation.

---

- [x] Use auto for default footer link spacing @desktop to improve horizontal spacing

image

---

### TODO:
- [ ]  Big footer should be able to accept nav items that don't contain sub-items, just as the primary nav can be a combination of simple links and links that trigger menus.

This seems like it's own task? Could use more detail so we don't have unintended consequences.

- [ ] .usa-footer__logo-img needs an explicit width: 100%set

Wasn't seeing any issues with or without this.

---

- [] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

### DIFF
--- a/src/stylesheets/components/_footer.scss
+++ b/src/stylesheets/components/_footer.scss
@@ -276,18 +276,12 @@
 
   .usa-footer__primary-link {
     @include h4;
-    @include add-background-svg('arrow-down');
-    background-position: units(2) center;
-    background-size: units(1.5);
     line-height: line-height('heading', 2);
     margin: 0;
-    padding-left: units(4);
 
     @include at-media('mobile-lg') {
       @include u-padding-y(0);
-      background: none;
       margin-bottom: units(1);
-      padding-left: 0;
 
       &:hover {
         cursor: auto;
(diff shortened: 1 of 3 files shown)


src/stylesheets/components/_footer.scss
```diff
@@ -296,19 +291,45 @@
   }
 
   .usa-footer__primary-content--collapsible {
+    .usa-footer__primary-link {
+      cursor: pointer;
```
thisisdano: House style is to display properties in alpha order

src/stylesheets/components/_footer.scss
```diff
@@ -296,19 +291,45 @@
...
+      display: flex;
+      align-items: center;
+      justify-content: flex-start;
+
+      // Arrow for collapsible content.
+      &::before {
```
thisisdano: I like this approach. It's simpler and more predictable

thisisdano:
⚠️ Small point, but important: we can't use `+`s in branch names because that character is incompatible with our hosting/preview integration.

thisisdano:
Merging into an internal branch to finish this up and test.

PR #3238, opened 2019-11-25
Update list of sites

afeijoo:
- Remove archived and inactive sites
- Update URLs


thisisdano:
This should be live in a couple minutes

PR #3249, opened 2019-11-27
Remove "Who is Using" from Fractal docs

thisisdano:
This is only useful on the main website, so we're removing it from the Fractal build. This prevents it from appearing on components.designsystem.digital.gov and allows the canonical list to live on designsystem.digital.gov

### DIFF
--- a/fractal.js
+++ b/fractal.js
@@ -32,9 +32,6 @@ components.engine(
   })
 );
 
-const docs = fractal.docs;
-docs.set("path", "docs");
-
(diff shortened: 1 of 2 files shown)


fractal.js
```diff
@@ -32,9 +32,6 @@ components.engine(
   })
 );
 
-const docs = fractal.docs;
```
thisisdano: Removed since the list is the only docs element. We don't need to build docs if we're not displaying this element.

PR #3253, opened 2019-12-06
Prevent subcomponent type styles from overriding the primary component

thisisdano:
**Now components reliably respect their font settings.** Setting values like `$theme-footer-font-family` should set the font face for the entire component, but some CSS specificity quirks were overriding these values in some instances. Now, setting a component's font face works as expected, with no secret overrides.

### DIFF
--- a/src/stylesheets/components/_banner.scss
+++ b/src/stylesheets/components/_banner.scss
@@ -7,6 +7,10 @@
     font-size: font-size($theme-banner-font-family, "3xs");
     padding-bottom: units(0);
   }
+
+  .usa-accordion {
+    @include typeset($theme-banner-font-family);
+  }
 }
 
 .usa-banner__content {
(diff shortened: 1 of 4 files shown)


src/stylesheets/components/_banner.scss
```diff
@@ -7,6 +7,10 @@
     font-size: font-size($theme-banner-font-family, "3xs");
     padding-bottom: units(0);
   }
+
+  .usa-accordion {
+    @include typeset($theme-banner-font-family);
```
mejiaj: @thisisdano, this is under the assumption that `usa-accordion` font could be overridden right?

thisisdano: yup — this can be set with the `$theme-accordion-font-family` theme variable — but in this instance we don't want that value to render, instead we want to render the value of `$theme-banner-font-family`

because those values happened to be the same in earlier testing, we didn't catch this

PR #3258, opened 2019-12-11
Allow non-token values in theme color settings

### LINKED ISSUE #3209: Allow non-token values in theme color token assignments
As much as we want to share a common palette across projects and agencies, there are real instances (like colors specified in legislation) where teams and projects need to use custom colors. 

1) USWDS should allow custom colors in theme color token assignments
2) USWDS should publish its color grade guidance so teams could understand how their custom color might translate into the standard USWDS color grade system

thisisdano:
**Allow non-token values in theme color settings.** While USWDS promotes and encourages using our system tokens in theme color settings, agencies have a real need to occasionally use non-token colors. This includes instances where certain colors are mandated by law and cannot be easily changed. Now, teams can add non-token colors to theme color settings like `$theme-color-primary: #f00`. This new non-token value of `'primary'` will apply anywhere the `'primary'` token is used: functions, mixins, settings, and utilities. Using non-token values will throw a warning in the compile process, but this, like all compile warnings, can be disabled by setting `$theme-show-compile-warnings: false`.

Fixes #3209 

### Success criteria
- [ ] Any valid color can be added a theme palette color variable
- [ ] Non-token color overrides compile with a useful warning
- [ ] Compile warning may be disabled with `$theme-show-compile-warnings: false`
- [ ] Non-color, non-token assignments result in a useful error
- [ ] Assigned color compiles and outputs properly 
- [ ] Assigned color works in mixins
- [ ] Assigned color works in functions
- [ ] Assigned color works in utilities
- [ ] Assigned color works as a setting (ex: `$theme-link-color`)
- [ ] New code is organized, efficient, and readable
- [ ] Passes all CI tests


src/stylesheets/core/_functions.scss
```diff
@@ -123,7 +123,7 @@ Leaves bools as is
 
   @if type-of($value) == "color" {
     @error 'Only use quoted color tokens in USWDS functions and mixins. '
-      + 'See v2.designsystem.digital.gov/style-tokens/color '
+      + 'See designsystem.digital.gov/design-tokens/color '
```
thisisdano: Fixed this url as well!

src/stylesheets/core/_functions.scss
```diff
@@ -900,28 +900,61 @@ Derive a color from a color shortcode
...
+      @if index($flags, no-warn) {
+        @return $value;
+      }
+
+      @if $theme-show-compile-warnings {
+        @warn 'Override: `#{$value}` is not a USWDS color token.';
+      }
+
+      @return $value;
+    }
```
thisisdano: The color function can now include a number of flags like `color(value, flag-1, flag-2, ...)`. `override` and `set-theme` flags allow passing a color. `no-warn` suppresses the warning.

src/stylesheets/core/_functions.scss
```diff
@@ -900,28 +900,61 @@ Derive a color from a color shortcode
...
-  @if map-has-key($system-color-shortcodes, $shortcode) {
-    $our-color: map-get($system-color-shortcodes, $shortcode);
-    @if $our-color == false {
-      @error '`#{$shortcode}` is a color that does not exist '
-        + 'or is set to false.';
+
+  // False values may be passed through when setting theme colors
+  @if $value == false {
+    @if index($flags, set-theme) {
+      @return $value;
```
thisisdano: Rewrote this to restrict the `false` pass-through to `set-theme` only — this allows the `color()` function to be used when setting the theme token (since theme tokens can be `false`).

src/stylesheets/core/_functions.scss
```diff
@@ -900,28 +900,61 @@ Derive a color from a color shortcode
...
+  @if not index($flags, set-theme) {
+    @if map-has-key($project-color-shortcodes, $value) {
+      $our-color: (map-get($project-color-shortcodes, $value));
+      @if $our-color == false {
+        @error '`#{$value}` is a color that does not exist '
+          + 'or is set to false.';
+      }
+      @return $our-color;
+    }
+  }
```
thisisdano: Improves error handling. We can't check against the theme colors when we're setting them, so  setting the `set-theme` flag bypasses this check

src/stylesheets/core/_functions.scss
```diff
@@ -900,28 +900,61 @@ Derive a color from a color shortcode
...
+
+  @error '`#{$value}` is not a valid USWDS color token. '
+      + 'See designsystem.digital.gov/design-tokens/color '
+      + 'for more information.';
```
thisisdano: Improves error message

src/stylesheets/core/_functions.scss
thisisdano: Definitely

thisisdano:
I'll update asap

src/stylesheets/core/_functions.scss
thisisdano: Updated in every instance I could find

rhukster:
Greatly looking forward to this!

JayDarnell:
Was this solution reversed at some point? I'm really struggling to figure out how to update $theme-link-color to a hex value in my **@use "uswds-core" with** declaration. I was able to update $theme-color-base-ink directly in this manner but cannot update $theme-link-color.

```
Error: "Only use quoted color tokens in USWDS functions and mixins. See designsystem.digital.gov/design-tokens/color for more information."

46 │   $value: general.smart-quote($value);
```

thisisdano:
Jay, you need to use a token for `$theme-link-color` but you _could_ assign a theme color token like `theme-color-accent-warm` to `$theme-link-color` and then assign a custom hex to the value of `$theme-color-accent-warm`.

JayDarnell:
Thanks for taking the time to respond @thisisdano. It really appreciate it. I haven't worked with USWDS in years and I'm finding a lot has changed. That suggestion got me what I needed.

thisisdano:
Awesome! I'm so glad it worked

PR #3261, opened 2019-12-12
Promote better deprecation communication and compatibility

### LINKED ISSUE #3254: Add deprecated vars and alias them to their replacements
Users should have less of a necessary burden to update deprecated variables and settings. They _should_ — but their sites should not break if they do not.

A solution could be to maintain a small file that maps the deprecated var to the new var. Something like `_deprecated-settings.scss`. This file could include:

```
$deprecated-var-1: $new-var-1;
$deprecated-var-2: $new-var-2;
...
```

This might go some small way to assuring that variable names aren't serious breaking changes.

thisisdano:
**We're handling deprecations more gracefully.**  Occasionally the design system will deprecate variables or functionality. Now we'll display a deprecation message in the terminal when compiling USWDS Sass to better communicate these changes. We're also improving backward compatibility by supporting deprecated variables, functions, and mixins throughout the major version cycle. This way, we can continue to improve how our code is structured while minimizing the effects of this restructuring on your projects.

This is how the deprecation warning prints in the terminal:

```
✉ USWDS Notifications
2.4.0: If your component settings aren't working as expected, make
sure you're importing the components settings in your Sass entry
point (often styles.scss) with `@import "uswds-theme-components"`.
A bug in 2.0 omitted that import.
2.2.0: We changed the names of some settings.
$theme-navigation-width → $theme-header-min-width
$theme-megamenu-logo-text-width → $theme-header-logo-text-width
2.0.2: We changed the names of some settings and mixins.
$theme-title-font-size → $theme-display-font-size
@include title → @include display
@include typeset-title → @include typeset-display
These are notifications from the USWDS team, not necessarily a
problem with your code.

Disable notifications using $theme-show-notifications: false
```

### Success criteria
- [ ] Deprecated or changed variables, mixins, and functions continue to work throughout the major version
- [ ] Accurate warning shows on compile, with the changed functionality and its replacement
- [ ] Warning may be disabled

Fixes #3254

### DIFF
--- a/src/stylesheets/core/_deprecated.scss
+++ b/src/stylesheets/core/_deprecated.scss
@@ -1,0 +1,31 @@
+/*  deprecated.scss
+    ---
+    Occasionally the design system will deprecate
+    old variables or functionality. If we replace
+    the old functionality with something new, this is a
+    place to connect the old functionality to the
+    new functionality, in the service of better
+    continuity and backwards compatibility within a
+    major release cycle.
+
+    Note the USWDS version where we deprecated the
+    old functionality in a comment.
+
+    This file should started fresh at each
+    major version.
+*/
+
+// Deprecated in 2.2.0
+$theme-navigation-width: $theme-header-min-width;
+$theme-megamenu-logo-text-width: $theme-header-logo-text-width;
+
+// Deprecated in 2.0.2
+$theme-title-font-size: $theme-display-font-size;
+
+@mixin title {
+  @include display;
+}
+
+@mixin typeset-title {
+  @include typeset-display;
+}
--- a/src/stylesheets/core/_deprecated.scss
+++ b/src/stylesheets/core/_deprecated.scss
@@ -17,24 +17,14 @@
     major version.
 */
 
-@if $theme-show-deprecation-warnings {
-  @warn "\a"
-    + "--------------------------------------------------------------------\a"
-    + "Deprecated in USWDS 2.x\a"
-    + "--------------------------------------------------------------------\a"
-    + "2.2.0\a"
-    + "$theme-navigation-width \2192  $theme-header-min-width\a"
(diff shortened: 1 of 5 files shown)


src/stylesheets/core/_deprecated.scss
```diff
@@ -0,0 +1,52 @@
...
+    Add the deprecation to the warning.
+
+    This file should started fresh at each
+    major version.
+*/
+
+@if $theme-show-deprecation-warnings {
+  @warn "\a"
+    + "--------------------------------------------------------------------\a"
+    + "Deprecated in USWDS 2.x\a"
```
thisisdano: Maybe this should be more like `Notifications` and less specifically about deprecations

mejiaj:
@thisisdano on a `watch` task this would run every SCSS update right?

Do you have any thoughts on possibly moving the changelog message as a gulp task and maybe we can take advantage of:

```js
const log = require('fancy-log');
const c = require('ansi-colors');
```

thisisdano:
It would yeah — I don't think I know how to make run only once. I do worry that it's too "heavy", but I also think that's important to get a notification in the build process in addition to release notes, etc. Perhaps there is a better way? 

For me, the downside to a gulp-based solution is it's too tied to `uswds-gulp` — if we can build it into the Sass itself, it should work regardless of how a user compiles.

The upside it it looks less like an error and could maybe have more sophisticated logic. Even so, I'm inclined to keep it in Sass.

thisisdano:
Perhaps this needs more time to bake and we should wait for a later release

thisisdano:
On second thought, I think this is better to get in right now. Then we can work to improve it. 
@mejiaj Anything you consider a blocking issue here?

mejiaj:
No sir, seems good to go. I'll see if I can prototype something on my end and see how/if we can improve. But as of now, you're right. It's good info for users.

Issue #3270, opened 2019-12-19
Seemingly contradictory CSS in banner on mobile


kmonahan:
## Description

In the gov banner (components/_banner.scss), lines 196 and 197 absolutely position the close button that's visible below the tablet breakpoint at top: 0, bottom: 0, right: 0. However, lines 213 and 214 set the position to absolute again and change the right and top positioning. This also only applies below the tablet breakpoint, because the position value is changed to static at the tablet breakpoint. I think the first one is correct positioning, because the close button then fits into the right padding in .usa-banner__header--expanded.

The relevant portion of my generated CSS (we're importing USWDS components into our sass file), looks like this:
```
.usa-banner__button[aria-expanded=true] {
    background-image: none; }
    @media all and (max-width: 39.99em) {
      .usa-banner__button[aria-expanded=true]::after {
        background-image: url("../images/close-blue-60v-alt.svg"), linear-gradient(transparent, transparent);
        background-repeat: no-repeat;
        background-position: center center;
        background-repeat: no-repeat;
        background-size: 1rem;
        content: '';
        display: inline-block;
        height: 3rem;
        width: 3rem;
        /* stylelint-disable block-closing-brace-newline-after, at-rule-empty-line-before */
        margin-left: 0;
        /* stylelint-enable */ }
      .usa-banner__button[aria-expanded=true]::after {
        position: absolute ;
        bottom: 0;
        top: 0;
        position: absolute ;
        right: 0;
        background-color: #dcdee0;
        bottom: 0;
        height: auto; } }
    @media all and (min-width: 40em) {
      .usa-banner__button[aria-expanded=true] {
        height: auto;
        padding: 0;
        position: relative; }
        .usa-banner__button[aria-expanded=true]::after {
          background-image: url("../images/angle-arrow-up-primary.svg"), linear-gradient(transparent, transparent);
          background-repeat: no-repeat;
          background-position: center center;
          background-repeat: no-repeat;
          background-size: 0.5rem;
          content: '';
          display: inline-block;
          height: 0.5rem;
          width: 0.5rem;
          /* stylelint-disable block-closing-brace-newline-after, at-rule-empty-line-before */
          margin-left: 2px;
          /* stylelint-enable */ }
        .usa-banner__button[aria-expanded=true]:hover::after {
          background-image: url("../images/angle-arrow-up-primary-hover.svg"), linear-gradient(transparent, transparent);
          background-repeat: no-repeat; } }
[...]

PR #3271, opened 2019-12-19
[UI component]: Remove additional positioning rules on banner close button

kmonahan:
Addresses #3270 by removing the extra positioning rules. Also removed the bottom: 0 declaration in line 199 because u-pin("y") already sets bottom: 0.


mejiaj:
@thisisdano I haven't run into any issues with this in fractal. Orange is just to highlight the specific element.

image

PR #3291, opened 2020-01-27
Replace `css-mqpacker` with `postcss-sort-media-queries`.

### LINKED ISSUE #3262: Build process should not require media query packing
Media query packing is an extra requirement to the build process, necessary to reorder the output from the utilities builder to give it the proper CSS source order. Our current mq packing solution is deprecated (see #3079) and complicates the user build process.

- Is there another post-processing tool that could serve the same purpose as our current mq packer?
- Could we rebuild the utilities builder so such a requirement is no longer necessary?
- Can any solution we implement be as fast or faster than the current solution?

mejiaj:
## Description

Fixes #3262 by replacing deprecated `css-mqpacker` with `postcss-sort-media-queries`, which is maintained.

## Additional information

* Also updated docs and test files


README.md
```diff
@@ -222,7 +222,7 @@ The design system requires both **autoprefixing** and **media query packing** to
 '> 2%','Last 2 versions', 'IE 11'
 ```
 
-- **media query packing** collects all similar media queries together at the end of the compiled stylesheet. USWDS and `uswds-gulp` use [`css-mqpacker`](https://github.com/hail2u/node-css-mqpacker). This is necessary to provide the proper cascade and specificity for utility classes. If you are not implementing utility classes, media query packing may not be necessary — but it is still a good idea.
+- **media query packing** collects all similar media queries together at the end of the compiled stylesheet. USWDS and `uswds-gulp` use [`postcss-sort-media-queries`](https://github.com/solversgroup/postcss-sort-media-queries). This is necessary to provide the proper cascade and specificity for utility classes. If you are not implementing utility classes, media query packing may not be necessary — but it is still a good idea.
```
maya: Just noticed this link redirects to: https://github.com/yunusga/postcss-sort-media-queries

thisisdano: Good catch  — I'll fix this in the develop branch

Issue #3285, opened 2020-01-17
Mutli-line strings are breaking webpack sass compiles.


mejiaj:
Sass loader breaks when using latest `2.4.0` SASS changelog.

mejiaj:
@thisisdano I took another look at the webpack issue from [here](https://github.com/greenca6/uswds-tester/tree/uswds-2.4.0-bug).

I definitely see the benefits of having it in SASS, but unfortunately we'd have to use or write a separate loader that would handle this properly. 

An alternative I've mocked up can be found on `feature/USWDS-3285`. This moves the message to `changelog.js` and is currently shown on `gulp build` it looks like this:

<img width="563" alt="image" src="https://user-images.githubusercontent.com/3385219/72646168-19fbaf80-393b-11ea-8407-88ba6cd0e7ee.png">

I'm sure we can keep improving this, but I just wanted to provide this as an option.

Another alternative is having a `CHANGELOG.md` file in the root directory.

thisisdano:
I really like where this is going

sawyerh:
Was just coming here to report the same. I've been struggling to get a Sass file to compile using either `sass` or the `@csstools/postcss-sass` packages, due to what I think is the same issue. Seeing the following error:

```
> postcss styles/**/[!_]*.scss -d build --ext css

Error: Expected ".
   ╷
14 │ --------------------------------------------------------------------
   │                                                                     ^
   ╵
  node_modules/uswds/dist/scss/core/_notifications.scss 14:69  @import
  node_modules/uswds/dist/scss/packages/_required.scss 21:9    @import
  node_modules/uswds/dist/scss/uswds.scss 5:9                  @import
  styles/app.scss#sass 1:9                                     root stylesheet
    at Object._newRenderError (/Users/sawyer/Dev/Sites/mass/uswds-mayflower/node_modules/sass/sass.dart.js:10899:19)
```

sawyerh:
To recreate the issue from my previous comment, you can:

1. Visit
2. Enter the following into the Sass panel:
```
/*  notifications.scss
    ---
    Adds a notification at the top of each USWDS
    compile. Use this file for important notifications
    and updates to the design system.

    This file should started fresh at each
    major version.

*/

/* stylelint-disable string-no-newline  */
$uswds-notifications: "\a
--------------------------------------------------------------------
\2709  USWDS Notifications
--------------------------------------------------------------------
2.4.0: If your component settings aren't working as expected, make
[...]

PR #3304, opened 2020-02-11
Use Dart Sass for compiling

thisisdano:
**Use canonical Sass compiler for compiling.** We're now using Dart Sass to compile the system Sass. Dart Sass is the canonical Sass compiler, so using it will keep us at feature parity and help prevent inadvertent inconsistencies.

- - -

We've been using node-sass (libsass) as our primary compiler, but we should use the canonical Sass compiler (Dart Sass) to assure that anything we build works with the official compiler. This would have alerted us to issues like #3285 and should also assure that we can use all the most current Sass directives when we choose to upgrade.

This also fixes the irritating and undocumented `4 is not a valid flex value` error!


mejiaj:
LGTM after conflicts are resolved.

PR #3327, opened 2020-02-25
USWDS - Search: Use consistent search markup.

mejiaj:
## Description

Search styles depend on `div[role="search"]` being a child of `.usa-search`. Otherwise it breaks to next line on small screens.

## Additional information
image

image


mejiaj:
@thisisdano, looks like axe is complaining about the search component page where the variations are listed on the same page.

I've tried adding them to their own iframe, but then axe doesn't test them. We could:
1. Separate into separate components and ignore the combined one.
2. Ignore the combined one and use the `search-header` component to test the accessibility of search

Happy to hear any other suggestions though.

thisisdano:
I think it is a best practice to put the `role=search` in the form element, as we changed it in `2.5.0`, so I'd prefer to change the stylesheet to style properly whether or not the `role` is in the `< 2.5.0` markup or the `2.5.0+` markup, to support both better markup and backward compatibility.

I'll push this change.

src/components/search/search.njk
```diff
@@ -14,22 +14,26 @@
   <h2>Big</h2>
 
   <form class="usa-search usa-search--big" role="search">
-    <label class="usa-sr-only" for="search-field-big">Search</label>
-    <input class="usa-input" id="search-field-big" type="search" name="search">
-    <button class="usa-button" type="submit">
-      <span class="usa-search__submit-text">Search</span>
-    </button>
+    <div>
```
mejiaj: Looks like this in fractal for me:
image

We need to add the role to the `div`, remove the div, or update the styling.

src/stylesheets/components/_search.scss
```diff
@@ -11,9 +11,11 @@
   @include typeset($theme-search-font-family);
   position: relative;
 
-  // if role is in the <form> element (>=2.5.0)...
+  // if role is in the <form> element (>=2.6.0)...
   &[role="search"],
-  // ...or if the component has a separate <div> (<2.5.0)
+  // if there is a (=2.5.[0,1])...
+  &[role="search"] div,
```
mejiaj: Should we be extra-cautious and do `&[role="search"] > div`?

thisisdano: Yes

thisisdano: I'll do that

Issue #3349, opened 2020-03-16
Create a character limit component.


wbyoko:
## User Story

As a front-end developer, I need a way to help users know how much text they can enter upfront when there is a limit on the number of characters.

## Examples
* [GOV.UK](https://design-system.service.gov.uk/components/character-count/)
* [Shopify](https://polaris.shopify.com/components/forms/text-field) *(set 'showCharacterCount' to true)*
* [Material](https://material.io/components/text-fields/#anatomy) *(in 'Assistive elements' - 'Character counter')*

## Checklist
- [x] Create low-fidelity design
- [x] Review & discuss low-fi design (designer + developers)
- [X] SKIPPED Get user feedback on low-fi designs (optional - if team feels is needed)
- [ ] Code
- [ ] Conduct usability test with working code (optional - if team feels is needed)

Ben-Judy-Flexion:
**Design:**
![Character limit (wireframes) State 1_ default](https://user-images.githubusercontent.com/56978395/76870425-ae996580-6837-11ea-8de2-01143095f35a.png)
![Character limit (wireframes) State 2_ typing - under character limit](https://user-images.githubusercontent.com/56978395/76870426-af31fc00-6837-11ea-92c0-eb9e6b644ff0.png)
![Character limit (wireframes) State 3_ typing - close to character limit](https://user-images.githubusercontent.com/56978395/76870428-af31fc00-6837-11ea-9073-142dfed7cd92.png)
![Character limit (wireframes) State 4_ typing - at character limit](https://user-images.githubusercontent.com/56978395/76870429-afca9280-6837-11ea-8f4c-a7c9865f59bf.png)

**_Figma source for design:_**
https://www.figma.com/file/jLPd8N0c0BHH6Bd3vFDFaw/Form-Component-Wireframes?node-id=0%3A1

wbyoko:
@Ben-Judy-Flexion should there be a comment at every empty state? or should there be a message?

Ben-Judy-Flexion:
Great question, @wbyoko. I updated the design to include guidance around this. Let me know what you think. (Let's iterate on this together if you think there is a better approach.)

Also, I found this article helpful:
http://www.breck-mckye.com/blog/2012/05/character-count-design-some-guidelines/

jaredcunha:
If you want to follow GDS, they do bold red only after you pass the limit. The bold works so it's not a color only indicator. 

[...]

PR #3370, opened 2020-03-20
USWDS-3349: Character count component

wbyoko:
## Description

#3349. Character Count component, tests, styles, fractal example

[Current implementation](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/10x-forms/3349-character-count/components/preview/character-count.html)

## Additional information

Screenshot 2020-03-20 15 08 06
Screenshot 2020-03-20 15 08 36
Screenshot 2020-03-20 15 08 49


jaredcunha:
Reposting comment left here: http://localhost:4000/components/form-controls/#character-count


I think this is why i thinking it's also worth making the text bold after the character limit is surpassed (turning bold and red at the same time)… the warning state is triggered after the first character is entered, which as a user, I would find distracting. I think this would be my only sticking point. 
Screen Recording 2020-03-23 at 02 28 PM

Tested in:
- Chrome
- Firefox
- Sarfari
- IE11
- Edge
- iOS
- VoiceOver

Ben-Judy-Flexion:
Oh, that is a really good point. For fields with a small character limit, it does make more sense this way.

jaredcunha:
@Ben-Judy-Flexion I'd side towards keeping the behavior consistent. I did notice that the switch to bold stands a lot out much more than the color change. The color change, once the text is bold, was pretty difficult to perceive.

Ben-Judy-Flexion:
I'm all for consistency. (No strong reason to have it behave differently for long textareas vs. short text input fields.) And I'm okay with simply doing a normal -> bold font-weight change, no color change. I think the color is a nice enhancement but not something that drastically improves the overall experience.

wbyoko:
@Ben-Judy-Flexion @jaredcunha  we could also update the guidance to suggest that it is only used for larger text fields.

Should any one really be using the character count component for a 10 letter input?

jaredcunha:
@wbyoko I think that makes sense, that it should only be used for textareas. That would, in a way, align with [when a limit is already implied](https://github.com/uswds/uswds-site/pull/885/files#diff-a7e2a6dce886a7172bb8071b07043b7dR26) by the nature of the single line input.

thisisdano:
One process comment: It's useful to have a link to the fractal page in the issue description, like [Preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/10x-forms/3349-character-count/components/preview/character-count.html) →

thisisdano:
Also, consider making the character count hint a `block` or `inline-block` element with 0.5 units of top padding, so the text isn't nearly occluded by the focus highlight.

**With padding:**
Screen Shot 2020-03-24 at 9 57 47 AM

**Without padding:**
Screen Shot 2020-03-24 at 9 58 01 AM

jaredcunha:
Where were we at with regard to having a warning state vs the text changing bold and red only after the limit is passed, following the GDS example. Personally, I find it more clear that I've passed the limit because the color change on its own was difficult to perceive. 

If we do want a warning state, how do we know we've hit the right number to offer the warning? Is the same warning sufficient on a 250 char max as a 10,000 char max essay?

spec/unit/character-count/invalid-template-no-message.spec.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('character count component without message', () => {
+  const { body } = document;
+
+  afterEach(() => {
+    body.textContent = '';
+    CharacterCount.off(body);
+  });
+
+  it('should throw an error when a character count compoent is created with no message element', () => {
```
thisisdano, line 16: spelling: `component`

spec/unit/character-count/invalid-template-no-wrapper.spec.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('character count input without wrapping element', () => {
+  const { body } = document;
+
+  afterEach(() => {
+    body.textContent = '';
+    CharacterCount.off(body);
+  });
+
+  it('should throw an error when a character count compoent is created with no wrapping class', () => {
```
thisisdano, line 16: spelling: `component`

src/stylesheets/elements/form-controls/_character-count.scss
```diff
@@ -0,0 +1,13 @@
...
+  .usa-character-count__message {
+    display: inline-block;
+    padding-top: units(0.5);
+  }
+}
+
+.usa-character-count--invalid {
+  .usa-character-count__message {
+    font-weight: font-weight("bold");
+    color: color("error-dark");
```
thisisdano, line 11: Use alpha order for declarations

src/components/07-form/controls/character-count.njk
```diff
@@ -0,0 +1,35 @@
...
+<form class="usa-form" novalidate>
+  <div class="usa-character-count" data-module="usa-character-count">
+    <div class="usa-form-group">
+      <label class="usa-label" for="with-hint-textarea">
+        Textarea
+      </label>
+      <span id="with-hint-textarea-hint" class="usa-hint">
+        This is a textarea with a character counter.
+      </span>
+      <textarea class="usa-textarea usa-character-count__input" id="with-hint-textarea" maxlength="50" name="with-hint-textarea" rows="5" aria-describedby="with-hint-textarea-info with-hint-textarea-hint"></textarea>
```
jaredcunha, line 27: The following is not a super strongly-held opinion, but thought I'd bring up for discussion.

The `usa-character-count__input` class does not have any CSS properties. It's mainly a JS hook. So wondering if we should use the BEM naming here or use some other convention. Brad Frost discusses class prefixes [his post about css naming architecture](https://bradfrost.com/blog/post/css-architecture-for-design-systems/).  Just trying to think if we cut the number class names the component needs in order to work perhaps.

Another suggestion, if we do keep this convention, is to consider renaming as `usa-character-count__field`, because it's more generic, considering we already have established `usa-input` and `usa-textarea`. Using `__input` on a `<textarea>` feels a little counterintuitive considering that.

src/components/07-form/controls/character-count.njk
```diff
@@ -0,0 +1,35 @@
+<form class="usa-form">
+  <div class="usa-character-count" data-module="usa-character-count">
```
jaredcunha, line 2: Out of curiousity, what is `data-module` doing? Haven't seen this used elsewhere in the design system.

wbyoko: Good catch. `data-module` isn't being used at this point. I'll push up an update to remove this attribute from the examples.

src/components/07-form/controls/character-count.njk
wbyoko: I like the idea of making it more generic and using `usa-character-count__field`, I'll update that as well.

We could look into extending `usa-input` or `usa-textarea` onto `usa-character-count__field`.

wbyoko: @thisisdano would you have an feeling of if/how we should address the issue of needing multiple classes on an element within a component? something like `class="usa-hint usa-character-count__message"`.

Some options for changing the current setup include: 

Styling based off of current USWDS classes and being nested within a `class="usa-character-count"` element being the key selector.

updating the BEM classes to include the styling for `usa-textarea`, `usa-hint`, `usa-input`, etc.. Using `@extend %block-input-general; @extend %block-input-styles;` gets us most of the way there but there would still be more duplication needed.

PR #3373, opened 2020-03-26
In banner template, clarify that it's .gov + HTTPS that make sites official

h-m-f-t:
This small change aims to clarify that it's both things described in the banner _combined_ that make a site official (.gov and HTTPS). There is some public confusion about this: https://krebsonsecurity.com/2020/03/us-government-sites-give-bad-security-advice/

I opted to not say "Combined with .gov _or .mil_" because the header reads "The .gov means it’s official."

As has been [said before](https://github.com/uswds/uswds-site/issues/315#issuecomment-311774919), the broader issue in addressing potential confusion is in downstreaming this change, but this is a fair start.

### DIFF
--- a/spec/unit/banner/template.html
+++ b/spec/unit/banner/template.html
@@ -30,7 +30,7 @@
                     <div class="usa-media-block__body">
                         <p>
                             <strong>The site is secure.</strong>
-                            <br> The <strong>https://</strong> ensures that you are connecting to the official website and that any information you provide is encrypted and transmitted securely.
+                            <br> Combined with .gov, the <strong>https://</strong> ensures that you are connecting to the official website and that any information you provide is encrypted and transmitted securely.
                         </p>
                     </div>
                 </div>


thisisdano:
Thanks @h-m-f-t — This feels reasonable to me. Gonna loop some others in, but I'm very much in favor of improving the accuracy and legibility of our banner message.

konklone:
FWIW, I'm also a :+1: on this change. It preserves the overall message while further minimizing any risk of confusion.

I want to be on record as saying that, though I've voiced skepticism about the value of this banner specifically due to the difficulty in getting changes deployed downstream, I have not had any issue with the text of the banner once it was updated to the current version. I do not think the current text risks misleading users in a serious way about "HTTPS = completely safe at all times on the general web", because it is presented in combination with .gov, and is only displayed specifically if the user clicks on a prompt saying "Here's how you know" why you're on "an official website of the U.S. government". 

Nonetheless, this helps address any incidental risk of confusion, and reads nicely and organically. 

Maybe the one minor wording change I'd suggest is adding "When" beforehand, to read "When combined with .gov...". But reads well either way.

IanLee1521:
I also personally prefer the "When combined with ..." wording suggested by @konklone .

thisisdano:
Closed in favor of https://github.com/uswds/uswds/pull/3524

PR #3376, opened 2020-03-27
Update to Sass module system

thisisdano:
[Preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/use-sass-modules/) → 

Updates the USWDS source code to Sass modules. This is a pretty big change, so we'll have to figure out what to do with it.

This is working but not yet ready for review or merge.

Get ready for something like

```
@use "uswds" as * with (
  $theme-show-compile-warnings: false,
  $theme-color-base: "green-50v",
  ...
)
```

I thinking this is something we may need to seed as an RC version or some kind of beta before we can fully release it. There're enough little quirks that this kind of change could qualify for a major version... 😬


thisisdano:
Still testing, but it looks like this also seems to work with the existing, non-Module `@imports` flow, like:

```
$theme-show-notifications: false;
@import "uswds"; // internally, all built with Modules
```

If this does hold up to more rigorous testing, that would be very 💯

mejiaj:
> Still testing, but it looks like this also seems to work with the existing, non-Module `@imports` flow, like:
> 
> ```
> $theme-show-notifications: false;
> @import "uswds"; // internally, all built with Modules
> ```
> 
> If this does hold up to more rigorous testing, that would be very 💯

Still a ways off until support for `@import` is dropped, but wanted to share this. 

>One year after this deprecation goes into effect (1 October 2022 at latest), we will drop support for @import and most global functions entirely. This will involve a major version release for all implementations.

https://sass-lang.com/blog/the-module-system-is-launched#future-plans

thisisdano:
Yeah, all the more reason to get this 99.9% of the way there. It'll be very simple to remove the `@import` when necessary

thisisdano:
Current having trouble finding an easy way to natively integrate this with jekyll, as the ruby-based Sass compilers use libsass and can't parse the new Sass modules. So this work is effectively on ice until this can be resolved.

thisisdano:
Putting this in carbonite until support improves.

thisisdano:
I've set up a `uswds-2.10-beta` branch for the Beta. We can merge this work in there, then use that branch as the canonical branch.

PR #3446, opened 2020-05-04
USWDS - Accordion: Add expand/collapse all to accordion.

### LINKED ISSUE #3435: Accordion needs  "expand all'' functionality
- [x] **@thisisdano New issue is created #3654 with original issue linked. I'd like to close this ticket and use the new ticket to track further work.**

## Accordion feature request:  Accordion component really needs some kind of expand all functionality.

http://designsystem.digital.gov/components/accordion

Accordions facilitate scanning a page, but they need to be used with caution because they break search and wreak havoc with off-line uses of content.

Example of accordion breaking search:
- Search engine returns a hit for term which is hidden in an accordion.  The context might not display in the search engine UI or following the link lands the user at a seemingly random place on the page.
- The in-browser find-on-the-page feature (typically Ctrl-F) will not display content in a closed accordion.

Examples of accordion breaking off-line uses:
- Printing the page typically does not include content inside accordions
- Saving the page or copying to word processor document may not render accordion content

## Description

My expectation is that the USWDS accordion component would include a promient feature to expand all accordion content all at once.  Of course, there should be an easy way to reset (collapse) accordion content as well.

As one example, that does a good job with this, please see the WCAG Quick Reference.
https://www.w3.org/WAI/WCAG21/quickref/

1. Use the USWDS Accordion Component as-is.
2. If you have more than one accordion on the page, you have just created a problem for your users!

- Discussion 4/24/2020 on GSA Section508 listserv

### LINKED ISSUE #3654: USWDS-Accordion: add "expand all'' functionality
Accordion: add "expand all'' functionality

#3435 (original issue)
#3446 (Prototype)

**Tasks to complete**

- [ ] Add `expand all` and `collapse all` functionality 
- [ ] Is this a retrofit or a new feature?
- [ ] Define and text printing expectations, what should users see (or not) when printing
- [ ] Update guidance, make clear how collapsing affects searchability and printing

mejiaj:
[Component Preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-accordion-all/components/preview/accordion--expand-collapse-all.html)

[All accordions on 1 page →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-accordion-all/components/preview/accordion-controls.html)

## Description
Resolves #3654. Adds expand/collapse all controls to **individual** accordions. If you have multiple separate accordions, you'll need to add the controls to each of those accordions.

### New settings

| Settings    | Description     |
| :---------- | -------------- |
| $theme-accordion-controls-alignment | The alignment of the controls button group. |

### Additional information
- Added JS for expand/collapse all per accordion component
- Added unit test for new controls
- Added new $theme-accordion-controls-alignment settings

---

### TODO

- [x] Update accordion unit test
- [x] Cleanup JS
- [x] Test accessibility
  - Do we need aria-live to announce the Toggle All button text updates?
  - How do we let the screen reader know that all items in this particular accordion have been opened or closed?

---

Adds a button to expand/collapse all accordion items. Related to issue #3435.


thisisdano:
I think this is going to need both `expand all` and `close all` buttons as a button group. I don't think it's worth the effort to try to find a smart way to determine the desired state of the button if, say, some panels are open and some are closed.

thisisdano:
This _could_ be helpful here: https://inclusive-components.design/collapsible-sections/

sarnikbixal (approved):

sarnikbixal (approved):
Ugly polyfill but if it works looks good.

fafnirical:
Should this functionality be applied to _all_ accordions, or only multiselectable accordions? I'd expect "Expand All" and "Collapse All" not to make sense on accordions that can't normally have multiple sections open, but maybe this is more related to searchability?

thisisdano:
I'm strongly inclined to deprecate non-multiselect accordions. It probably requires further research, but they feel like a product of an older time and no longer a best practice

thisisdano:
But you're right @fafnirical — it seems odd to have expand all apply to single-select accordions, but I also think it's the right thing to do here

thisisdano:
@mejiaj Let's get our accessibility Qs answered here before we go any further

thisisdano (changes requested):
Let's address accessibility Qs before merging

mejiaj:
@thisisdano Added a test page w/ all accordions. You can find that [here →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-accordion-all/components/preview/accordion-controls.html)

src/js/components/accordion.js
```diff
@@ -53,6 +54,33 @@ const toggleButton = (button, expanded) => {
...
+ * @param {HTMLElement} button
+ */
+const toggleText = (element) => {
+  let controlBtn = element;
+
+  const originalString = controlBtn.textContent.trim();
+  const startText = controlBtn.getAttribute("data-label--expanded");
+  const endText = controlBtn.getAttribute("data-label--collapsed");
+
+  controlBtn.textContent = (originalString === startText) ? endText : startText;
```
sarnikbixal, line 68: Can you just use the expanded attribute to set text content based? Or do we need the data-label attribute? This check seems unnecessary.

mejiaj: For the original string? I felt like it was more obvious this way what texts are being toggled. Happy to revisit.

mgifford:
Ok looking at the test page here which was updated last week https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-accordion-all/components/preview/layout--content.html

This generally works with my keyboard-only testing. Accessibility Insights also likes it (which I like better than the HTML_CodeSniffer)

But I notice that there are 5 expandable options. If I expand one, any other one I expand, it automatically collapses the one I had previously expanded. That could be convenient, but also annoying.  What If I want to compare what is under tabs 1, 3 & 5? 

Fortunately I can expand all and then collapse individual tabs without any additional impacts.  I do think that auto-closing tabs when I open a new one isn't a good pattern.  Might be a nice option for some instances, but shouldn't be the default.

Issue #3464, opened 2020-05-12
Radio and checkboxes legend formatting issue


maya:
Radio and checkbox code example uses a screen reader–only class, but in most cases, we need to have a visible legend on those form controls. When you remove the class, the legend is too close to the controls.

## Without `.usa-sr-only` class
<img width="287" alt="Screen Shot 2020-05-11 at 7 34 38 PM" src="https://user-images.githubusercontent.com/5249443/81632200-8bfe7580-93be-11ea-8d1d-f71027006290.png">

## Without `.usa-sr-only` class and with `.usa-legend` class
<img width="318" alt="Screen Shot 2020-05-11 at 7 34 08 PM" src="https://user-images.githubusercontent.com/5249443/81632202-8d2fa280-93be-11ea-8f32-dc7310599a87.png">

thisisdano:
I think this legend formatting should look just like a standard label, so it fits in with the rest of a form. @mejiaj can you integrate this into your PR?

PR #3465, opened 2020-05-12
USWDS-3464 - Checkboxes & Radios: Add spacing to first checkbox/radio after visible legend.

mejiaj:
**Fixed spacing between legend and checkboxes/radio boxes.** Now there's appropriate spacing between the legend and either radio boxes or checkboxes.

## Description

Resolves: https://github.com/uswds/uswds/issues/3464

- Add margin-top to 1st checkbox/radio button after visible legend
- Update checkbox template

## Additional information

~[Preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-legend-margins/components/detail/checkboxes.html)~

[Checkbox preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-legend-margins/components/preview/checkboxes.html)

[Radio preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-legend-margins/components/preview/radio-buttons.html)


thisisdano:
I'm not seeing the proper spacing on the legend when I add `usa-legend` to the radio buttons here: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-legend-margins/components/preview/radio-buttons.html

Screen Shot 2020-05-18 at 3 05 35 PM

Screen Shot 2020-05-18 at 3 05 45 PM

Can it be 2units for each?

A separate issue is that legend looks honkingly large here. It should probably be smaller. `lg` or even `md` might be appropriate.

maya:
In the original issue, I included both with `usa-legend` and without so that form questions would look consistent. For example, if a radio group was on the same page as a text field, I'd expect the questions to be the same size.

Since USWDS doesn't have a form template, you're not able to see multiple form controls on the same page to check for this consistency.

thisisdano:
It seems like we should adjust the formatting of `usa-legend` to have the same look as a `usa-label` for a standard field. The use case for the legend is for introducing fieldsets like radio buttons and checkboxes, and since we're overriding the default formatting (that weird outline) for these elements, it doesn't make sense to have a legend that looks different from other form elements.  The following seems like the most common usage:

- - -

How long have you worked at ExampleCorp?
- [ ] Under a year
- [ ] 1-2 years
- [ ] 3-5 years
- [ ] 5-10 years
- [ ] Over 10 years

- - -

So, @mejiaj, let's change the formatting of `usa-legend` to look the same as a conventional `usa-label` and add an example that uses a visible legend for both checkbox and radio buttons. I tend to think we should change the default example to show use of `usa-legend` without the `add-sr-only` hiding, which feels more like an edge case.

thisisdano:
Hm. I spoke too soon. I guess this is complicated by the fact that we use `legend` a couple different ways: in _this_ case we'd want it to look like a label, and in our form templates, we want it to look more like a header (https://designsystem.digital.gov/components/form-templates/). Since we currently have no form templates that use radio buttons or checkboxes, this never came up.

I guess this reveals a bit of a quirk in markup usage, where fieldsets are used a couple valid ways that require different visual styles. 

@afomi — How are you using fieldsets and checkboxes/radios in touchpoints?

afomi:
Hi @thisisdano,

Touchpoints uses radio buttons primarily to support likert options (1-5 ratings). Radios are used in the basic, conventional way: select 1 of N options.

Checkboxes are not fully implemented yet.

Radio button markup can be viewed here: https://github.com/GSA/touchpoints/blob/develop/app/views/components/forms/question_types/_radio_buttons.html.erb

Happy to test/dogfood any USWDS changes for these components.

cc @ryanwoldatwork

thisisdano:
Thanks @afomi — I see you're using the hidden `legend`, then including a `label` after for the actual label text. 

```
<legend class="usa-sr-only"><%= question.answer_field %></legend>
<label class="usa-label">
  <%= t question_number.to_s %>.
  <%= question.text %>
</label>
```

In practice, how does the content of the `legend` differ from that of the `label`?

maya:
@thisisdano `label` is not the appropriate element for radio buttons, `legend` is.

mejiaj:
The radio and checkbox spacing doesn't seem consistent with the other input elements.

Dropdowns, text, number, textarea and other inputs all have their margins set via `margin-top`. Radio and checkbox use `margin-bottom` instead.

This is alongside offering a variation to legend for regular label-sized text.

thisisdano:
- Adjusted styling to match comments in https://github.com/uswds/uswds/issues/3464
- Made default version of legend match label, with a `usa-label--large` variant to mimic the old (not recommended) styling.
- Simplified and synced the checkbox and radio markup, and added a `disabled` element to each

thisisdano:
Screen Shot 2020-09-11 at 9 51 57 AM

Screen Shot 2020-09-11 at 9 52 13 AM

thisisdano (approved):

PR #3468, opened 2020-05-13
USWDS-3073 - Mobile nav: Fix `ESC` key not closing on IE11.

### LINKED ISSUE #3073: Can not close "mobile menu" using Escape key in IE11
## Description

IE 11 users are unable to close the "mobile menu" by pressing the Escape key.
This presents an accessibility issue.

## Steps to reproduce the issue (using IE 11)

1. View a [page](https://federalist-proxy.app.cloud.gov/preview/uswds/uswds/develop/components/preview/header--basic.html) with a collapsible nav area.
2. Reduce browser width enough to show the menu button.
3. Toggle the menu open.
4. Try pressing Escape key to close the menu.

## Additional information

There's a lot going on to help cause this issue!

It revolves around the inconsistency of IE11 returning "Esc" instead of "Escape" for it's KeyboardEvent.key. The issue is compounded in the fact that the keyboardevent-key-polyfill (used in receptor/keymap) is not wired up to actually run 'polyfill()'. And, even if it was, the published version of the polyfill doesn't actually fix the issue in IE11.

You can fix by doing **any** of the following...

* [uswds/navigation](https://github.com/uswds/uswds/blob/develop/src/js/components/navigation.js) -- below, could fix the one current instance
````js
    ...
    if (trapContainer) {
      navigation.focusTrap = FocusTrap(trapContainer, {
        Esc: onMenuClose, // Fix for IE11
        Escape: onMenuClose,
      });
    }
    ...
````
* [uswds/focus-trap](https://github.com/uswds/uswds/blob/develop/src/js/utils/focus-trap.js) -- below fixes the current instance and some potential future use
````js
...
module.exports = (context, additionalKeyBindings = {}) => {
  const tabEventHandler = tabHandler(context);

  // Fix for IE11
  const { Esc, Escape } = additionalKeyBindings
  if (Escape && !Esc) additionalKeyBindings.Esc = Escape
  ...
````

### Ideally, you'll want to fix the dependencies

**Both** of the following would be required.

* [receptor/keymap](https://github.com/shawnbot/receptor/blob/master/src/keymap.js) -- needs to run polyfill()
````js
import keyboardeventKeyPolyfill from 'keyboardevent-key-polyfill'
keyboardeventKeyPolyfill.polyfill() // actually run the polyfill
...
````

* [keyboardevent-key-polyfill](https://github.com/cvan/keyboardevent-key-polyfill/blob/master/index.js) -- has updated source code that would fix this.
  * But, it didn't actually get published to NPM.
  * The published code is actually 1.0.2 and does not fix this issue.

### PR DESCRIPTION
**Pull request by mejiaj**
**Allows mobile menus to be close with the `esc` key in IE11.** Now, the escape key will properly close mobile menus in IE11. Thanks @joncasey!

## Description

Fixes #3073 based on guidance in ticket.

## Additional information

Tested and working in IE11 on browserstack.

### DIFF
--- a/src/js/utils/focus-trap.js
+++ b/src/js/utils/focus-trap.js
@@ -38,9 +38,10 @@ const tabHandler = context => {
 
 module.exports = (context, additionalKeyBindings = {}) => {
   const tabEventHandler = tabHandler(context);
+  const bindings = additionalKeyBindings;
+  const { Esc, Escape } = bindings;
 
-  const { Esc, Escape } = additionalKeyBindings;
-  if (Escape && !Esc) additionalKeyBindings.Esc = Escape;
+  if (Escape && !Esc) bindings.Esc = Escape;
 
   //  TODO: In the future, loop over additional keybindings and pass an array
   // of functions, if necessary, to the map keys. Then people implementing
--- a/src/js/utils/focus-trap.js
+++ b/src/js/utils/focus-trap.js
@@ -39,6 +39,9 @@ const tabHandler = context => {
 module.exports = (context, additionalKeyBindings = {}) => {
   const tabEventHandler = tabHandler(context);
 
+  const { Esc, Escape } = additionalKeyBindings;
+  if (Escape && !Esc) additionalKeyBindings.Esc = Escape;
+
   //  TODO: In the future, loop over additional keybindings and pass an array
   // of functions, if necessary, to the map keys. Then people implementing
   // the focus trap could pass callbacks to fire when tabbing

### CODE REVIEW

**Comment by thisisdano** [comment]
@mejiaj Do you have any insight into the dependencies issue referenced in the original issue?

**Comment by mejiaj** [comment]
@thisisdano, a little. We're using `keymap` from `receptor` to make sure keys behave the same across all browsers.

The polyfill is used in our dependency [here](https://github.com/shawnbot/receptor/blob/master/src/keymap.js#L1) but it's never actually run.

And the update never got published, it seems.

---

According to the guidance of the polyfill it should be:
```js
require('keyboardevent-key-polyfill').polyfill();
```

instead of what receptor has:
```js
require('keyboardevent-key-polyfill');
```
https://github.com/cvan/keyboardevent-key-polyfill

joncasey:
@mejiaj , is summing it up well.
The fix applied already is plenty sufficient.
Trying to fix the dependencies becomes much harder. As, none of them seem to be maintained any more.
Another link http://uswds.glitch.me/ie11/escape-key-fix/ for reference.
It's my glitch, when I was actively using USWDS on a project.
May, or may not, help to explain it better?

Issue #785, opened 2015-10-05
Guidance on breadcrumbs


joewiz:
Does the framework currently provide guidance on breadcrumbs? 

I have been looking for guidance like http://getbootstrap.com/components/#breadcrumbs but couldn't find anything here.

joewiz:
I'm also interested in pagination - both numbered and prev/next.

carodew:
@joewiz These are things that we're considering adding to the design standards in a future release, but we don't have them included at the moment.

joewiz:
Okay, I'll follow developments.

Drakenhart:
Our project has been testing pagination options from both a visual design as well as a 508 compliance side. @carodew would it be helpful to share the data we've discovered here?

carodew:
@Drakenhart Sorry for the delayed reply. Yes, we'd love to hear more about your data. Any information you can provide about the context you used pagination in (what kind of page, etc) (even screenshots) and who you tested with would be extra helpful. Thanks!

Drakenhart:
@carodew I understand. The Holidays complicate online communication. I will see what I can manage.

For context:
We have need for multiple-page table-based lists. Having a single massive page that scrolls more then a screen height only works for screen readers, it's too much information for the user base (all at once).

So as to not hijack this issue further from breadcrumbs, should I open a new one?

Drakenhart:
Side note- on the breadcrumbs topic.

The common design pattern is to use breadcrumbs as a secondary form of supportive navigation. Also the best practices, as I understand them, is that one does not use H-tags in a breadcrumb.

I was curious about the choice, as it is very unique. Where did the inspiration for it come from and why do it this way?

carodew:
@Drakenhart It's probably best to discuss the multi-page tables as a separate discussion. Of course, the length of the scrollable table really depends on the use case. ;)

Drakenhart:
@carodew understood and agreed. I'll create a new item. :)

jbruce-gsa:
[...]

PR #3488, opened 2020-05-27
Add breadcrumb component

thisisdano:
[Preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/add-breadcrumb-component/components/preview/breadcrumb--default.html) → 
Documentation TK 

## Overview
Breadcrumb presents a navigable hierarchy of the current page.

```
nav.usa-breadcrumb
└─ol.usa-breadcrumb__list
│ └─ li.usa-breadcrumb__list-item
│    └─ a.usa-breadcrumb__link
│ └─ li.usa-breadcrumb__list-item
│    └─ a.usa-breadcrumb__link
...
  └─ li.usa-breadcrumb__list-item
```

The links begin with a `Home` link and proceeds with links down the site hierarchy, ending with the full text of the current page title.

We're using a `nav` element to encapsulate our breadcrumbs, and a `ol` to collect the links together. The ordered list provides a legible reading experience when styles do not load, and helps readers using a screenreader understand the page hierarchy because screenreaders will announce which element in the list they're in (ie, reading `3 of 4` indicates that the reader is on the current page's parent).

**Truncation and wrapping:** The default breadcrumb displays all the breadcrumb items in a single line at wide widths, truncating the last element of the line if there is any overrun. The breadcrumb will wrap when using `usa-breadcrumb--wrap`.

**Narrow widths:** At narrow widths (defined by `$theme-breadcrumb-min-width`), the breadcrumb shows the current page's parent to sighted users. (Screenreaders see the full trail.) The parent is preceded by a `←`.

breadcrumbs-at-different-widths

> **Top:** Breadcrumb at wide witdhs show the full trail.
> **Bottom:** Narrow-width breadcrumb shows only the current page parent


thisisdano:
Note that the current prototype uses a couple nonstandard settings values for testing purposes. We'll revert the values of `$theme-breadcrumb-icon-color` and `$theme-breadcrumb-link-colors` to the documented defaults before releasing this component.

thisisdano:
Some open questions and comments for feedback:

### Initial functionality
- By default, breadcumb trails would be constrained to a single line, but wrapping is an option in settings.
- Bt default, overflow is truncated.
- Below a breakpoint setting, breadcrumb displays the immediate parent of the current page only, with a back arrow.
- Icons in the breadcrumb are replaceable SVGs.
- Potentially, a progressive JavaScript enhancement would automatically display the breadcrumb trail as parent-only (the mobile default; see PR description, above) if the trail truncated beyond the current page.

### Questions for comment
- **Wrapping and truncation:** How do you feel about the proposed truncation behavior? Is a JavaScript enhancement too much for a breadcrumb component? Any uses cases you’ve seen for wrapping? Nielsen Norman [recommends against wrapping](https://www.nngroup.com/articles/breadcrumbs/) but truncation has it own problems.
- **SVG icons:** Does a settings-customizable SVG provide enough out-of-the-box usefulness and flexibility?
- **Anything else?** What are we missing? Any features we should add or remove from this component?

Thanks for taking look, and we look forward to your feedback! 🚀

afeijoo:
Feedback: Use an ordered list and include a recommendation in the guidance to use `BreadcrumbList` Schema.org markup: https://schema.org/BreadcrumbList

pglevy:
Piggybacking on Aimee's comment, it looks like your intention was to use an ordered list, but the markup has an unordered list. Also, in the description, you say, "The _unordered_ list provides..."

pglevy:
In the narrow widths description, "the breadcrumb shows on the current page's parent to sighted users" should be "the breadcrumb shows the current page's parent to sighted users" (removed "on").

thisisdano:
I was unclear — I did intend to use an **unordered** list but I keep writing it incorrectly as `ol`. :( 

Opinions really vary on the proper markup for a breadcrumb, but the consensus I've seen is that an ordered list is no more accurate, semantically, than an unordered list. I don't have a strong opinion about it, but an ordered list seems more _opinionated_ and I don't think the additional opinion adds anything to the comprehensibility of the markup.

dawnpm:
Hi there, I agree that they should be an UNordered list.

I also encourage you to add the schema.org markup to the templates - users can always remove it if they don't want it, but it'll save a lot of people a lot of work in adding them post-facto.

afeijoo:
Guidelines for breadcrumb markup:
 
* Google - https://developers.google.com/search/docs/data-types/breadcrumb#html
* Bing - https://www.bing.com/webmaster/help/markup-breadcrumbs-72419f3f

Also, based on my testing, Google's Structured Data Tool at https://search.google.com/structured-data/testing-tool/u/0/ requires a value for the `position` field and their examples use `ol`.

dawnpm:
Would the USWDS templates be able to support the content team managing the content of the breadcrumbs list in the head, and then inserting the values in the template body, if that component is used? That would be ideal, but also seems highly dependent on what publishing platform / framework the site is built on.

joewiz:
This sounds and looks great (along with the whole [roadmap for 2020](https://designsystem.digital.gov/about/product-roadmap/)! 

This PR addresses https://github.com/uswds/uswds/issues/785 - which I mention because one reply referenced the [WAI-ARIA Authoring Practices breadcrumbs examples](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html), another document worth referencing. I think the only recommendation there not reflected here is:

> The set of links is structured using an **ordered** list.

src/stylesheets/settings/_settings-general.scss
```diff
@@ -99,3 +99,10 @@ Icons
 */
 
 $theme-icon-image-size: 2 !default;
+
+/*
+----------------------------------------
+Easing
+----------------------------------------
+*/
+$theme-easing: 0.2s ease-in-out !default;
```
mejiaj, line 108: This doesn't exist in `_uswds-theme-general.scss` yet.

Issue #3491, opened 2020-05-28
USWDS - Buttons: Increase contrast of Accent Cool hover to increase accessibility.


mejiaj:
## Description

WAVE Browser plugin is flagging contrast issues on the following buttons:
- usa-button--accent-cool (hover)

## Steps to reproduce the issue

1. Get [WAVE plugin](https://wave.webaim.org/extension/)
2. Run the plugin on the [Buttons page](https://designsystem.digital.gov/components/button/)
3. Click `View Details` on Report
![image](https://user-images.githubusercontent.com/3385219/83186554-4e098d00-a0f2-11ea-8258-adb9a24b3f49.png)

4. You should see the following:
![image](https://user-images.githubusercontent.com/3385219/83186497-36ca9f80-a0f2-11ea-8847-6d151bf4ca24.png)

## Additional information

You can also inspect the buttons in firefox click on the text color and see the same error.
![image](https://user-images.githubusercontent.com/3385219/83186860-cb350200-a0f2-11ea-9b35-2f1a76f647a7.png)

![image](https://user-images.githubusercontent.com/3385219/83186835-bf494000-a0f2-11ea-9b79-f42e3c9f37d5.png)

Axe browser plugin only shows the warning for `accent-cool hover` instead of the disabled state. See note below.

---

Note from @pglevy regarding disabled:
>Form elements/buttons that are marked up as disabled are not subject to the usual text contrast guidelines because the state is programmatically understood by assistive tech. See https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html

https://gsa-tts.slack.com/archives/C050HRGN7/p1590695196365400?thread_ts=1590694097.359800&cid=C050HRGN7

PR #3492, opened 2020-05-28
USWDS-3491: Fix accent-cool contrast issues with a more dependable contrast tool

mejiaj:
##  Description

Fixes https://github.com/uswds/uswds/issues/3491

Darkened `accent-cool-dark` and `accent-cool-darker` so that
`.usa-button--accent-cool`'s hover state meets WCAG AAA contrast standards.

This makes it consistent with its sibling buttons.

## Additional information

image

[Preview 👁](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-button-accent-cool-a11y/components/detail/buttons--accent-cool.html)

---

Is it possible to check for contrast issues with [axe-tester](https://github.com/uswds/uswds/blob/jm-button-accent-cool-a11y/spec/axe-tester.js#L14)?


thisisdano:
I think we need a smarter way for the button code to calculate whether it needs white or black text — I'm hesitant to just change the colors and call it a day...

mejiaj:
Something like this?

https://css-tricks.com/programming-sass-to-create-accessible-color-combinations/

thisisdano:
Added a couple new mixins, plus some other goodies.

**set-text-from-bg**(`$bg-token`, `$dark-token`, `$light-token`): for a given background color, use either the dark or light token for the text color, whichever assures proper contrast. prefer the light text color, and use `back` and `white` as the dark and light defaults. Sets _only_ the text color.

**set-text-and-bg**(`$bg-token`, `$dark-token`, `$light-token`): for a given background color, use either the dark or light token for the text color, whichever assures proper contrast. prefer the light text color, and use `back` and `white` as the dark and light defaults. Sets _both_ the background color and text color.

PLUS, **we can now do token grade checking in Sass**. Mostly. The relative luminance function I found is very very close to the output I get in our luminance tool, but not exactly the same. I don't knwo why not. So the checker has about 4 false positives. But if you turn the setting `$theme-check-colors` to `true` you can see it run. It's pretty slow and resource intensive, so you wouldn't want to do a luminance check to often. (It's why I built another function — **decompose**() — to convert a color token into a list of form `family`, `grade`, `variant`... it's a LOT faster.)

mejiaj:
Just to confirm the above:
```sass
@include set-text-from-bg("primary-darker", "ink", "primary")
```
With this would I expect to get an error? 

---

```sass
 @include set-text-and-bg("primary-darker", "ink", "red-40")
```

The difference between `red-40` and `primary-darker` is 40. And `primary-darker` to `ink` is 10, shouldn't it set the greatest difference? In this case red-40. Or throw an error that neither are acceptable?

---

Variable name for doc's needs updating:

>But if you turn the setting $theme-check-colors to true you can see it run. 

`$theme-check-colors → $theme-test-colors`

Do you get these in your results when set to `true`?
```zsh
WARNING: green-warm-70v (#4b4e10) lum: 0.071 is not in the range 0.05, 0.07
    src/stylesheets/core/_functions.scss 1118:9   test-colors()
    src/stylesheets/core/_variables.scss 610:16   @import
    src/stylesheets/packages/_required.scss 14:9  @import
    src/stylesheets/uswds.scss 5:9                root stylesheet

WARNING: green-10v (#c5ee93) lum: 0.745 is not in the range 0.75, 0.82
    src/stylesheets/core/_functions.scss 1118:9   test-colors()
    src/stylesheets/core/_variables.scss 610:16   @import
    src/stylesheets/packages/_required.scss 14:9  @import
    src/stylesheets/uswds.scss 5:9                root stylesheet

WARNING: gray-cool-10 (#dfe1e2) lum: 0.744 is not in the range 0.75, 0.82
    src/stylesheets/core/_functions.scss 1118:9   test-colors()
    src/stylesheets/core/_variables.scss 610:16   @import
    src/stylesheets/packages/_required.scss 14:9  @import
    src/stylesheets/uswds.scss 5:9                root stylesheet
```

How would we expect users to use `decompose()` to run their own tests?

thisisdano:
> `@include set-text-from-bg("primary-darker", "ink", "primary")`
This is most likely an error, if the difference between the background and neither light nor dark is 50 or more

>  @include set-text-and-bg("primary-darker", "ink", "red-40")
This should throw an error unless the difference is 50 or more. I don't want it just to use the biggest magic number, just the magic number that works.

Now we could rewrite it in the form (background, first choice, second choice), so there's more control over which color "wins" if both are sufficient. Or we could use the combination with the greatest magic number. And we could potentially pass in an optional third prop, like `level` which would default to `AA`, but could be `AALarge` or `AAA`...

As for decompose() — I'd probably consider it mostly an internal function.

And yes, I get some false positives, as the lum result from Sass is a little different from the lum I get in other checkers.

PR #3493, opened 2020-05-29
USWDS-3490 Combo Box Mouseover Selection

### LINKED ISSUE #3490: Coordinate mouse and keyboard selection in the Combo Box component
As a user I would expect that hovering over a list item in the combo box would focus that item so that subsequent keyboard events move from that hovered-now-focused item. Currently hover does not focus the list item.

wbyoko:
Adds hover focus for combo box list items.

closes #3490 

[Current Implementation](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/10x-forms/3490-mouseover-selection/components/preview/combo-box.html)

### DIFF
--- a/spec/unit/combo-box/combo-box-change-event.spec.js
+++ b/spec/unit/combo-box/combo-box-change-event.spec.js
@@ -102,6 +102,18 @@ EVENTS.keydownArrowDown = el => {
   el.dispatchEvent(evt);
 };
 
+/**
+ * send a keydown Tab event
+ * @param {HTMLElement} el the element to sent the event to
+ */
+EVENTS.keydownTab = el => {
+  const evt = new KeyboardEvent("keydown", {
+    bubbles: true,
+    key: "Tab"
+  });
+  el.dispatchEvent(evt);
+};
+
 describe("combo box component change event dispatch", () => {
   const { body } = document;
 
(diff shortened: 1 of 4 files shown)


thisisdano:
@Ben-Judy-Flexion Does the hover indicator need to have both a fill and a focus outline? Is there actually "focus" on hover/mouseover?

Ben-Judy-Flexion:
I dug into this and found this article, which is super helpful:
https://zellwk.com/blog/style-hover-focus-active-states/

Basically, we want to thoughtfully style :hover, :focus, and :active differently.

**For mouse users:**
- When the user hovers over an element, background-color (and/or color) changes. They get feedback.
- When the user clicks on an element, focus outline shows. They get feedback.

**For keyboard users:**
- When the user tabs into an element, focus outline shows. They get feedback.
- When they interact with the element, background-color (and/or color) changes. They get feedback.

So in the case of the combo box list items, we probably do **not** want both the outline and the background color change on :hover. (Besides, visually I think it's a bit much.) The background color change should suffice on :hover, but they get the outline on click.

Ben-Judy-Flexion:
The current implementation looks good to me!

PR #3497, opened 2020-06-02
Fix Unhandled rejection Parsing error

hursey013:
## Description

While attempting to pull some sass variables into javascript using https://github.com/jgranstrom/sass-extract, I encountered the following error:
```
Unhandled rejection Parsing error: Please check validity of the block starting from line #16

14 | @each $face, $stack in $project-font-stacks {
15 |   @if $stack {
16*|     [class*="#{ns("utility")}font-#{$face}-"] {
17 |       font-family: #{$stack}#{$if-important};
18 |     }
```
It looks like using two sets of double quotes (line 16) causes the parser to have an error. This PR switches the innermost to single quotes. This resolves the error and allows the sass to be interpreted correctly.

### DIFF
--- a/src/stylesheets/utilities/utility-fonts.scss
+++ b/src/stylesheets/utilities/utility-fonts.scss
@@ -13,7 +13,7 @@ $if-important: "";
 
 @each $face, $stack in $project-font-stacks {
   @if $stack {
-    [class*="#{ns("utility")}font-#{$face}-"] {
+    [class*="#{ns('utility')}font-#{$face}-"] {
       font-family: #{$stack}#{$if-important};
     }
   }


thisisdano:
Thanks @hursey013 — I'll get this merged, but I'm really curious to learn about how you're using sass-extract.

hursey013:
Awesome, thanks @thisisdano! No real use case right now, just toying around. I think one potential use could be allowing variables to be referenced within JS - if you needed to grab a hex value for a particular color that is defined within the sass theme files for example. That way you avoid duplication and the sass theme files can be the canonical source. I'm thinking something similar to the the scenario outlined here: https://github.com/adamgruber/sass-extract-js#nice-how-do-i-use-it

Is something like this possible currently?

Issue #3360, opened 2020-03-16
Create a document upload component.


wbyoko:
## User Story
As a front-end developer, I need a way to help users select a file in a way that is compact, convenient, and accessible.

## Examples
* [Carbon](https://www.carbondesignsystem.com/components/file-uploader/code/)
* [Lightning](https://www.lightningdesignsystem.com/components/file-selector/)

## Checklist
- [x] Create low-fidelity design
- [x] Review & discuss low-fi design (designer + developers)
- [x] Get user feedback on low-fi designs (optional - if team feels is needed)
- [x] Code
- [x] Conduct usability test with working code (optional - if team feels is needed)
- [x] Create hi-fi design assets (Adobe XD, Sketch, Figma)
- [x] Review & approve hi-fi design assets
- [x] Publish design assets (update downloadable files for Adobe XD, Sketch, Figma)

Ben-Judy-Flexion:
**Design:**
_The following design should be considered "low fidelity" -- not matching the visual style of USWDS. Let @Ben-Judy-Flexion know if high fidelity designs are needed._
![Document upload (wireframes) State 1_ default](https://user-images.githubusercontent.com/56978395/76873528-dbe81280-683b-11ea-9e4e-1219a8fecd44.png)
![Document upload (wireframes) State 2_ add file](https://user-images.githubusercontent.com/56978395/76873529-dbe81280-683b-11ea-8b56-88105d3287fe.png)
![Document upload (wireframes) State 3_ confirmation](https://user-images.githubusercontent.com/56978395/76873531-dc80a900-683b-11ea-9476-cdd8d872dd5e.png)

**_Figma source for design:_**
https://www.figma.com/file/jLPd8N0c0BHH6Bd3vFDFaw/Form-Component-Wireframes?node-id=0%3A1

jaredcunha:
@thisisdano I'm about to start working on this component. Before we start modeling this component, I'd like to get an idea of what degree of complexity you prefer. We can build some simliar to [Polaris Drop Zone](https://polaris.shopify.com/components/actions/drop-zone), which has a richer interaction to

thisisdano:
There are certainly similarities to the interaction in those two examples. Looks like they both have a drop target and an upload button. The Polaris example has (to my mind) a more intuitive UI, though perhaps it's too "big" for our most common use cases. That said, what _are_ our most common use cases?

[...]

PR #3523, opened 2020-06-25
USWDS #3360: file input

jaredcunha:
## Description

Creates file input

## Additional information

* Issue: https://github.com/uswds/uswds/issues/3360
Screen Shot 2020-06-25 at 6 50 42 PM


jaredcunha:
@thisisdano This is ready for your review. I have written tests to check that the component builds properly. I wanted to write a test that checks that previews are generated when you attach files, but that requires `DataTransfer`, which is not included in `jsdom`, and there's not a clear workaround for that.

jaredcunha:
Since drag and drop is not supported natively on File Inputs in IE or Edge, the "drag" text is removed.  I'd love to get my hands on computer with Edge using Chromium to see if that behavior has changed.

Screen Shot 2020-06-25 at 7 45 50 PM

thisisdano:
This seems to be working really well. I see how its functionality mimics that of the non-enhanced file input, and preserves something of the default ambiguity about choose/replace and remove. I wonder if there are some worthwhile further enhancements to change the behavior and reduce that ambiguity:

1. **If a replacement is cancelled, don't clear the file cache** (not sure if this is actually an improvement): This was the first issue I'd noticed, since it isn't immediately clear how to remove a file — and the only working way is what feels like a hack: to replace/choose a new file, and then cancel the action. This is not really an intuitive remove flow. And it could also be a problem if the user wanted to replace, but decided against replacing, and was surprised when cancelling the action removed their selected files. 
1. **Replace `Replace?` with `Replace file` or `Replace files`:** This is the only improvement that might be a good immediate fix, since if you select it, it will definitely either replace or remove the current files. It's not really a question you can actually answer with "no" if selected.
1. **Allow individual file removal** with a red `Remove` action (or an `x`): This could allow explicit removal without the ambiguity of the replace/cancel action. 

There's so much to like here, and I don't think we'll want to hold up this component over these issues, but it may be important to keep them in mind as future enhancements. Perhaps only the `Replace?` text issue is one we'll want to fix right now. Do you agree that phrasing the action as a statement is clearer?

jaredcunha:
@thisisdano That's great! I have some responses. 

**1. If a replacement is cancelled, don't clear the file cache.** This is actually how the browser behaves with default file inputs. It seems that hitting cancel to remove a file only happens in Chrome. It doesn't happen Safari or Firefox. The component is no different from a standard `<input type="file">`. 

**2. Replace Replace? with Replace file or Replace files.** This was my nitpicky-ness of having that top line say something like "2 files selected | replace files?" The "files" appearing twice felt a little repetitious and allowing more space between the two sides on narrower screens. Definitely not a hill I'm willing to die on. Screenshot show iPhone SE width (320px). 
Screen Shot 2020-06-30 at 9 05 09 AM

**3. Allow individual file removal.** This one would fundamentally change how the file input is handled. Right now, it's no different from `<input type="file">`, except that its display is larger to increase the drop area (in which the drop action on the input is supported natively by the browser) and it improves how selected files are displayed. Support for individually removing files introduces a new set of controls and forces us to rethink some of the interactions as well since we'd be going beyond what is natively supported by `<input type="file">`. Not that it couldn't be done. The component is set up in a way that if it was refactored to handle that, it can be done in a way that is backward compatible with previous implementations because JavaScript is constructing all of the parts.

thisisdano:
@jaredcunha Can we change the text to remove the question mark (and add `file`)? So `Change file` or `Change files` (if multiple)? That keeps it simple and direct, I think.

jaredcunha:
@thisisdano Done

PR #3543, opened 2020-07-09
Set a max-width on the banner content to prevent long lines

thisisdano:
The banner content was set up to be in a centered container. It should be in a left-aligned container with a max-width of "desktop" to prevent long lines.

### DIFF
--- a/src/stylesheets/components/_banner.scss
+++ b/src/stylesheets/components/_banner.scss
@@ -1,3 +1,4 @@
+$banner-guidance-measure: 3;
 $banner-icon-colors: get-link-tokens-from-bg(
   $theme-banner-background-color,
   $theme-banner-link-color
@@ -51,9 +52,8 @@ $banner-icon-close: (
 }
 
 .usa-banner__content {
+  @include grid-container($theme-banner-max-width);
   @include add-responsive-site-margins;
-  @include u-margin-x("auto");
-  @include u-maxw("desktop");
   background-color: color("transparent");
   font-size: font-size($theme-banner-font-family, 4);
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


src/stylesheets/components/_banner.scss
```diff
@@ -51,8 +51,8 @@ $banner-icon-close: (
 }
 
 .usa-banner__content {
-  @include grid-container($theme-banner-max-width);
   @include add-responsive-site-margins;
+  @include u-maxw("desktop");
```
mejiaj, line 55: Shouldn't it be `@include u-maxw($theme-banner-max-width);`?

We have:
```scss
$theme-banner-max-width: "desktop" !default;
```
In `_settings-components.scss`

thisisdano: Not in this case, exactly, since valid values for that setting can include widths wider than desktop, or even ‘none’, and the content region needs to be constrained

mejiaj: Wouldn't we want to center it on desktop+ to have it align properly to the text above? 

Here's what I see on 1300px
image

I know it's a weird issue, but I want to make sure I understand the intention.

thisisdano: Ah yes you are right! :facepalm: One sec.

mejiaj: I think `@include u-margin-x("auto");` would be enough on `.usa-banner__content`.

I've pushed a commit so you can pull and test.

#### 1300px
image

Looks the same on smaller sizes too. Let me know if this works or if you intended something else.

thisisdano: OK — I have what I think is the proper solution here. We leave everything as it was, but we add a `measure` token to the `usa-banner__guidance` element. `measure-3` seems to be the best option. Pushed the change... take a look.

PR #3630, opened 2020-09-16
Update step indicator

thisisdano:
- No counters by default, enable with `--counters`
- Use `--center` for centered variant
- Remove counter size from public settings


thisisdano:
Circle is degraded, but passes tests locally.

PR #3652, opened 2020-09-22
Add alpha/beta CI workflows

thisisdano:
Adds alpha and beta jobs to the CI setup, and attaches the alpha deploy job to the `v2.10-beta--main` branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,9 +71,6 @@ jobs:
     <<: *container
     steps:
       - checkout
-      - restore_cache:
-          keys:
-            - v2-uswds-dependencies-{{ checksum "package-lock.json" }}
       - run: npm install --ignore-scripts
       - run: npm rebuild node-sass
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,9 +20,6 @@ jobs:
     <<: *container
     steps:
       - checkout
-      - run:
-          name: Install codeclimate
-          command: sudo npm install -g codeclimate-test-reporter
       - restore_cache:
           keys:
             - v2-uswds-dependencies-{{ checksum "package-lock.json" }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -65,6 +52,42 @@ jobs:
...
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v2-uswds-dependencies-{{ checksum "package-lock.json" }}
+      - run: npm install --ignore-scripts
+      - run: npm rebuild node-sass
+      - run:
+          name: Build the USWDS package
+          command: npm run release
```
thisisdano, line 66: @mejiaj Is the `release` command here the right one? Are we deploying with a fake hash key in the package every release?

mejiaj: No, we shouldn't be. Does this task re-run creating the ZIP?

We should run the hash task once when we generate the ZIP we'll use to attach to GH release.

thisisdano: I'm going to remove it from the job

sarnikbixal (approved):

.circleci/config.yml
thisisdano: Changed this to `build`

.circleci/config.yml
```diff
@@ -58,13 +45,46 @@ jobs:
...
       - run:
           name: Publish to NPM
           command: |
             npm config set "//registry.npmjs.org/:_authToken=$NPM_TOKEN"
             npm publish
 
+  deploy-alpha:
+    <<: *container
+    steps:
+      - checkout
```
thisisdano, line 58: Note that I'm not restoring the `node_modules` cache here as we're installing in the next step

.circleci/config.yml
```diff
@@ -76,3 +96,9 @@ workflows:
           filters:
             branches:
               only: master
+      - deploy-alpha:
```
thisisdano, line 99: This uses the `deploy-alpha` job for now, but we'll update this to `deploy-beta` when we release the beta

mejiaj (approved):

PR #3757, opened 2020-10-09
USWDS - PL: Migrate Card.

mejiaj:
## Description

Migrated:
- USA Card component w/ variations
- Card compare
- Card tests
- Moved Card `scss` 

### New menu items
image
Added 2 new folders (`Compare` & `Test`) to port over the Card comparisons and Tests.

The Card test needs some work. In creating the test it showed some limitations with the default template. For example, you can't pass in classes to the card container.

We'll need to revisit once more components are ported over.


chazchumley:
@mejiaj this looks really good.  nice job.

PR #3765, opened 2020-10-13
Build Patternlab on Circle

thisisdano:
Updates Circle CI and some gulpfile stuff to allow PatternLab to build on Circle.

- Uses a PHP/Node docker image
- Builds Chrome from an Orb (not useful yet, but will be when tests are re-integrated)
- Removes nonfunctional tests from the CI tests _for now_. We'll add these back at the end stage of the migration

So now Circle will install the project, build Patternlab, and run the available checks. This lets us at least check the PL build, and sets the stage for re-integrating test later.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ references:
     docker:
       - image: misterio92/ci-php-node
         environment:
-          CHROME_PATH: /usr/local/bin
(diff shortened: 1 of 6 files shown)


chazchumley:
@thisisdano looks good.  Thanks for tackling that.

PR #3810, opened 2020-11-04
Try: Implement banner toggle as substituted noscript markup

### LINKED ISSUE #3092: Banner open on page load is distracting
Like our accordions, the USWDS gov banner is opened and closed with javascript. The content in these components is accessible whether or not javascript loads because the components start open by default until the javascript loads to close them. 

As a result, even with a fast connection and performant JS, there is a brief moment (usually < 1s) when the banner is open on page load. Slower connections and larger JS files can make this open period more pronounced — staying open for seconds at a time. (For a good example, try loading https://digital.gov/ — especially on a slower connection.)

banner-fouc

This can be very distracting. Site content that unexpectedly disappears is confusing to site users (_"huh? what was that? was that important?"_) as well as aesthetically uncomfortable, even if you know what to expect (_"i sure wish the content didn't jump all around every time I loaded the page!"_).

The gov banner is one of our most important and most-used components, appearing on every load of every USWDS page. While our general accessibility guidance is to make all site content accessible regardless of the presence of javascript, this is negatively affecting the experience of _all USWDS sites and pages_. 

We need a better solution here. Whatever the solution, it should result in
- no flashing text regardless of bandwidth conditions
- no flashing text regardless of the presence of javascript

There are likely a few ways to approach this issue! The first that cone to mind are:
- We could change our no-js policy for this component
- We could change our no-js policy for all components
- We could use render-blocking js for this component
- We could try to better use caching and preloading for js
- We could rewrite the component to remove the js

Related PRs: 
- https://github.com/uswds/uswds/pull/3075
- https://github.com/GSA/digitalgov.gov/pull/1258

- [x] JA - resolve [zendesk ticket](https://searchusa.zendesk.com/agent/tickets/20791)
- [ ] JA - reply to zenkdesk when complete

aduth:
Closes #3092
Related: #3809

## Description

This pull request aims to explore an approach to #3092 based on @eddietejeda suggestion at https://github.com/uswds/uswds/issues/3092#issuecomment-533674613 to use a `noscript` container to show banner contents for no-JavaScript environments. In JavaScript environments, the container is replaced with its children's markup. This approach avoids the need for multiple copies of the banner content.

## Additional information

In its current form, the pull request primarily serves as a proof-of-concept.

Further work needed:

- Security audit (`outerHTML` substitution)
- Browser testing (I've read [some indication](https://stackoverflow.com/a/620955) that the contents of a `noscript` tag will be empty in some browsers in JavaScript-capable environments?)

An advantage to this over #3809 is that it can be considered an enhancement to opt-in to, and will not conflict with existing banner markup.

---

### DIFF
--- a/src/components/banner/banner.njk
+++ b/src/components/banner/banner.njk
@@ -23,32 +23,34 @@
       </div>
     </header>
     <div class="usa-banner__content usa-accordion__content" id="gov-banner">
-      <div class="grid-row grid-gap-lg">
-        <div class="usa-banner__guidance tablet:grid-col-6">
-          <img class="usa-banner__icon usa-media-block__img" src="{{ uswds.path }}/img/icon-dot-gov.svg" role="img" alt="Dot gov">
-          <div class="usa-media-block__body">
-            <p>
-              <strong>
-                {{ domain.heading | safe | replace(tld_text, domain.tld) -}}
-              </strong>
-              <br/>
-              {{ domain.text | safe | replace(tld_text, domain.tld) }}
-            </p>
+      <noscript>
+        <div class="grid-row grid-gap-lg">
+          <div class="usa-banner__guidance tablet:grid-col-6">
+            <img class="usa-banner__icon usa-media-block__img" src="{{ uswds.path }}/img/icon-dot-gov.svg" role="img" alt="Dot gov">
+            <div class="usa-media-block__body">
+              <p>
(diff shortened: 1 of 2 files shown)


aduth:
Since this was intended as a proof-of-concept, I'm going to go ahead and preemptively close the pull request. The proposed changes remain for future reference if needed.

PR #3846, opened 2020-11-23
Explicitly set visited state of unstyled button

andycochran:
## Description

This PR explicitly sets the `:visited` state of unstyled buttons when they are also inverse and outline. 

Note: This solves a problem that's specific to `<a class="usa-button">` elements that include an `href` value (i.e. buttons that have a `visited` state). 

When buttons have all three of these classes… 

`.usa-button--outline.usa-button--inverse.usa-button--unstyled` 

…[their ruleset](https://github.com/uswds/uswds/blob/44a9e68cf397fc1a8c9f46361dfe2c470a5c7309/src/stylesheets/elements/_buttons.scss#L124-L138) includes the [`button-unstyled` mixin](https://github.com/uswds/uswds/blob/50f6ffd6b0b6afa1b499daa36b6f8ee0312b1123/src/stylesheets/core/mixins/_button-unstyled.scss), which in term includes the [`typeset-link` mixin](https://github.com/uswds/uswds/blob/de5536bf830594d3576c904bf772b311d535a770/src/stylesheets/core/mixins/_typography.scss#L59), which makes the button look like an anchor by setting all 4 anchor pseudo classes (including a purple :visited state). So, when these 3 classes are used on a dark background, the `:visited` state is not inversed and the color is not accessible. 

This proposed change overrides the mixin's `:visited` state with the inverse color so that an outline, inverse, unstyled button has the inverse text color and is accessible on a dark background.

### DIFF
--- a/src/stylesheets/elements/_buttons.scss
+++ b/src/stylesheets/elements/_buttons.scss
@@ -138,6 +138,10 @@ $button-stroke: inset 0 0 0 units($theme-button-stroke-width);
       @include button-unstyled;
       color: color($button-inverse-color);
 
+      &:visited {
+        color: color($button-inverse-color);
+      }
+
       &:hover,
       &.usa-button--hover {
         color: color($button-inverse-hover-color);


thisisdano (approved):
Excellent — thank you!

Issue #3848, opened 2020-11-24
File Input's accept attribute does not support MIME types or multiple file extensions when dragging and dropping file(s)


ConnorDY:
### Describe the bug
If using a MIME type with the `accept` attribute (such as `application/pdf`), the proper files are accepted when clicking the `input` to select the files. However, the same files when dragged and dropped give the "This is not a valid file type"  error message.

There is also an issue when multiple file extensions are specified when dragging and dropping. For example, if I set the `accept` attribute to be `.pdf,.txt`, I can select files with both extensions when using the file browser, but I can't drop any files of these types when dragging and dropping.

This appears to happen because the USWDS code checks if the value of `accept` is included in the name of the file:
```javascript
const acceptedFiles = fileInputEl.getAttribute("accept");

// ... skipping some code here ...

for (let i = 0; i < e.dataTransfer.files.length; i += 1) {
  const file = e.dataTransfer.files[i];
  if (allFilesAllowed) {
    allFilesAllowed = file.name.indexOf(acceptedFiles);
    if (allFilesAllowed < 0) {
      break;
    }
  }
}
```

### Steps to reproduce the bug
1. Use one or more MIME types or file extensions with the `accept` attribute on the `input` element.
2. Attempt to drag and drop one or more valid files of that MIME type or file extension
3. See "This is not a valid file type" error

### Expected behavior
Dragging and dropping should accept the same files that are allowed when clicking the `input` element to browse for files.

### Screenshots
![image](https://user-images.githubusercontent.com/945062/100109847-42e1b080-2e3a-11eb-917b-ffea6176a544.png)

### System setup
- **USWDS version:** 2.9.0
- **Device:** N/A
- **Operating system:** N/A
- **Browser and version:** N/A

### Additional context
N/A

ConnorDY:
My recommendation for fixing this is splitting the value of the `accept` attribute:
```javascript
const acceptedFiles = fileInputEl.getAttribute("accept").split(',');
```

And then checking against `file.type` and `file.name` like so:
```javascript
for (let j = 0; j < acceptedFiles.length; j += 1) {
  const fileType = acceptedFiles[j];
  allFilesAllowed = file.name.indexOf(fileType) > 0 || file.type === fileType;
  if (!allFilesAllowed) break;
}
```

Let me know if y'all would like me to make a PR for this or if your team can tackle this.

aduth:
[...]

PR #3851, opened 2020-11-24
[FILE INPUT] fix multiple fix extension bug

jaredcunha:
## Description

Should fix issue #3848 

## Additional information

* Switched file input example to `multiple` to ensure script works when selecting multiple files of different type 
* Tested mix of accepted and not accepted files
Screen Shot 2020-11-24 at 1 53 28 PM


ConnorDY (approved):
This looks great to me and solves the issues I described.

However, I think a separate issue should be created for what was brought up here:
https://github.com/uswds/uswds/issues/3848#issuecomment-733091993

src/js/components/file-input.js
```diff
@@ -142,27 +142,29 @@ const removeOldPreviews = (dropTarget, instructions) => {
...
     for (let i = 0; i < e.dataTransfer.files.length; i += 1) {
       const file = e.dataTransfer.files[i];
       if (allFilesAllowed) {
-        allFilesAllowed = file.name.indexOf(acceptedFiles);
-        if (allFilesAllowed < 0) {
-          break;
+        for (let j = 0; j < acceptedFiles.length; j += 1) {
+          const fileType = acceptedFiles[j];
+          allFilesAllowed = file.name.indexOf(fileType) > 0 || file.type === fileType;
+          if (allFilesAllowed) break;
```
ConnorDY, line 161: This only breaks out of the file types/extensions loop. Might be worth adding an `else break;` after the the `if (allFilesAllowed) { /* ... */ }` block.

src/js/components/file-input.js
```diff
@@ -142,27 +142,29 @@ const removeOldPreviews = (dropTarget, instructions) => {
...
         }
```
ConnorDY, line 162: There's actually more that needs to be done here. This fails if the first file type isn't the same as the first type listed in the `accept`.

ConnorDY (changes requested):

src/js/components/file-input.js
ConnorDY, line 162: There's actually a little more that needs to be done here. This actually doesn't work if the first file isn't the same type as the first file type/extension from the `accept` attribute.

ConnorDY (changes requested):

src/js/components/file-input.js
ConnorDY: I was wrong about this. Please disregard.

ConnorDY (approved):

src/js/components/file-input.js
jaredcunha: managed to work on my end - attribute is `accept=".pdf,.txt"`
Screen Shot 2020-11-24 at 2 10 57 PM
Screen Shot 2020-11-24 at 2 11 30 PM

jaredcunha:
> However, I think a separate issue should be created for what was brought up here:
#3848 (comment)

Managed to push a fix that I think might support the wildcard

thisisdano (approved):

jaredcunha:
@thisisdano - checking back in on this saw. Saw this was flagged
Screen Shot 2020-12-11 at 10 03 47 AM

Thoughts about dismissing? Wildcards would only have one `*`. Would just need to update the replace method to `replace(/\*/g, "")`

src/js/components/file-input.js
thisisdano: Done

PR #3868, opened 2020-12-03
Update _typography.scss

astiefvater:
## Description

 Re-order link styles to enable hovering on visited links. Link, Visited, Hover, Active, Focus


thisisdano (approved):
Great catch — thanks!

PR #3892, opened 2020-12-14
Add more resilience to decorative SVGs

thisisdano:
### **Use the `aria-hidden="true"` property on decorative images (``) whose source is an SVG.**

Adding `role="img"` fixes a webkit bug that causes images with an SVG source be be announced as a group, not an image. However, adding `role="img"` and `alt=""` result in an object that's selectable by AT but not announced (at least in our testing on VoiceOver and Chrome). Adding `aria-hidden="true"` results in a properly hidden element. 

Adding `aria-hidden="true"` only would be equally effective (it renders `role="img"` moot) but we choose to add all three to be as consistent as possible with ``, even if it is redundant. Thus, each of the following can always be true for ``:
- Use `alt=""` for decorative image
- Use `role="img"` for images with an SVG source
- Use `aria-hidden="true"` for decorative images with an SVG source

### **Add `focusable="false"` to `<svg>` elements meant to be images (ie not text/data).** 

This prevents an IE11 bug that could allow the `Tab` key to navigate into the SVG. 

For decorative SVGs, use `role="img"` and `aria-hidden="true"` as above. Using both allows us the following consistent rules for the `<svg>` element:

- Use `role="img"` and `focusable="false"` for SVGs meant to be images (ie don't contain text/data)
- Use `aria-hidden="true"` for decorative SVGs
- Use `aria-label="[accessible name]"` or `aria-labelledby="[id list]"` for SVGs that convey information and need an accessible description

Source: https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html

### DIFF
--- a/src/components/banner/banner.njk
+++ b/src/components/banner/banner.njk
@@ -14,7 +14,7 @@
         </div>
         <div class="grid-col-fill tablet:grid-col-auto">
           <p class="usa-banner__header-text">{{ banner.text }}</p>
-          <p class="usa-banner__header-action" alt="" aria-hidden="true">{{ banner.action }}</p>
+          <p class="usa-banner__header-action" aria-hidden="true">{{ banner.action }}</p>
         </div>
         <button class="usa-accordion__button usa-banner__button"
           aria-expanded="false" aria-controls="gov-banner">
--- a/src/components/banner/banner.njk
+++ b/src/components/banner/banner.njk
@@ -37,7 +37,7 @@
           </div>
         </div>
         <div class="usa-banner__guidance tablet:grid-col-6">
-          <img class="usa-banner__icon usa-media-block__img" src="{{ uswds.path }}/img/icon-https.svg" role="img" aria-hidden="true">
+          <img class="usa-banner__icon usa-media-block__img" src="{{ uswds.path }}/img/icon-https.svg" role="img" alt="" aria-hidden="true">
           <div class="usa-media-block__body">
             <p>
               <strong>
--- a/src/components/banner/banner.njk
+++ b/src/components/banner/banner.njk
@@ -14,7 +14,7 @@
         </div>
         <div class="grid-col-fill tablet:grid-col-auto">
           <p class="usa-banner__header-text">{{ banner.text }}</p>
-          <p class="usa-banner__header-action" aria-hidden="true">{{ banner.action }}</p>
+          <p class="usa-banner__header-action" alt="" aria-hidden="true">{{ banner.action }}</p>
         </div>
         <button class="usa-accordion__button usa-banner__button"
           aria-expanded="false" aria-controls="gov-banner">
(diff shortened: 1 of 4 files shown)


src/components/icons/icons.config.yml
```diff
@@ -350,8 +350,10 @@ context:
         meta: "occupancy"
       - name: remove_red_eye
         meta: "visibility"
+      - name: remove
+        meta: "minus collapse"
       - name: report
-        meta: "exlamation danger stop"
+        meta: "exclamation danger stop"
```
jaredcunha, line 356: Doh!

jaredcunha (commented):
This looks good to me. Do you want to update the SVGs in https://github.com/uswds/uswds/blob/develop/src/components/icons/icons--sizes.njk to include `focusable="false"`?

thisisdano:
Mmm, yes I can do that.

src/components/banner/banner.njk
```diff
@@ -1,7 +1,7 @@
...
 {% set lock %}
-<span class="icon-lock"><svg xmlns="http://www.w3.org/2000/svg" width="52" height="64" viewBox="0 0 52 64" class="usa-banner__lock-image" role="img" aria-labelledby="banner-lock-title banner-lock-description"><title id="banner-lock-title">Lock</title><desc id="banner-lock-description">A locked padlock</desc><path fill="#000000" fill-rule="evenodd" d="M26 0c10.493 0 19 8.507 19 19v9h3a4 4 0 0 1 4 4v28a4 4 0 0 1-4 4H4a4 4 0 0 1-4-4V32a4 4 0 0 1 4-4h3v-9C7 8.507 15.507 0 26 0zm0 8c-5.979 0-10.843 4.77-10.996 10.712L15 19v9h22v-9c0-6.075-4.925-11-11-11z"/></svg></span>
+<span class="icon-lock"><svg xmlns="http://www.w3.org/2000/svg" width="52" height="64" viewBox="0 0 52 64" class="usa-banner__lock-image" role="img" aria-labelledby="banner-lock-title banner-lock-description" focusable="false"><title id="banner-lock-title">Lock</title><desc id="banner-lock-description">A locked padlock</desc><path fill="#000000" fill-rule="evenodd" d="M26 0c10.493 0 19 8.507 19 19v9h3a4 4 0 0 1 4 4v28a4 4 0 0 1-4 4H4a4 4 0 0 1-4-4V32a4 4 0 0 1 4-4h3v-9C7 8.507 15.507 0 26 0zm0 8c-5.979 0-10.843 4.77-10.996 10.712L15 19v9h22v-9c0-6.075-4.925-11-11-11z"/></svg></span>
```
mejiaj, line 4: This still is being read in Win10 Narrator. It says:

>Lock a lock padlock

Maybe we should ignore this as well with `aria-hidden="true"` since the word lock is in front of the icon?

mejiaj (changes requested):
I think we should move the `focusable` attribute to `<span class="icon-lock">`.

#### Sidenote 
I'm not sure if it's a blink/webkit thing, but I had trouble tabbing to the banner contents on the latest chromium version of Edge.

jaredcunha:
Oh, there's also the icon list too!

src/components/banner/banner.njk
thisisdano: Currently, this is supposed to read. Let's leave it readable for now

mejiaj (approved):

PR #3899, opened 2020-12-16
Final-stage bug fixes

thisisdano:
- Use `"none"` as a border radius option addition to `0`
- Don't show visited state in summary box links
- Remove margin from the first child of a process list item
- Display collection properly in prose scope
- Add top margin to summary box when it follows another element
- Properly calculate Process List connector position in prose scope
- Prettier ignore .md files

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,3 +6,4 @@
 *.png
 *.ico
 *.gif
+*.md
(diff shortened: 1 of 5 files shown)


mitchmoccia (approved):
You could do this in your sleep Dan haha... Brilliant updates!

Issue #3904, opened 2020-12-17
Footer logo name should not be a heading - a11y bug


maya:
`<h3 class="usa-footer__logo-heading">Name of Agency</h3>` should be a `p`. There's nothing nested below it, it's not a heading.

thisisdano:
Done in `2.10.1`

PR #3923, opened 2021-01-07
USWDS - Footer:  Change the markup tag for logo-heading from an h3 to a p.

carwin:
## Description

As mentioned in #3904, the heading element for `.usa-footer__logo-heading` contains no siblings and effectively creates an empty section. While I wasn't able to find any documentation in the HTML spec saying this was specifically not allowed, it still seems like a good practice to swap to a `<p>` tag.

## Additional Information

This change likely needs some input about whether the class name should be changed since it is no longer a heading.


thisisdano:
Moving into an internal branch before final testing/merge.

Issue #3869, opened 2020-12-03
Identifier component - wrong link name


maya:
Should be Budget and Performance rather than Performance reports per https://digital.gov/resources/required-web-content-and-links/

afeijoo:
The link text on digital.gov is _suggested_ link text. 

We are working with the Federal Web Council and digital.gov team to update the guidance on https://digital.gov/resources/required-web-content-and-links/.

thisisdano:
Blocked by https://github.com/uswds/uswds/issues/3744

afeijoo:
We are updating the guidance in https://github.com/uswds/uswds/issues/3744. The link name in the identifier component will remain as-is.

PR #3924, opened 2021-01-07
USWDS - Identifier: Update Performance reports link text.

carwin:
## Description

According to #3869, https://digital.gov/resources/required-web-content-and-links/ requires the link text to read `Budget and Performance` rather than `Performance reports`.


thisisdano:
We're going to work on harmonizing the language with digital.gov. Going to close this PR until we've worked it out with them.

PR #3942, opened 2021-01-12
Alert: Utilize $theme-alert-icon-size variable for setting the alert icon size

sawyerh:
## Description

- Update Alert component so that its background size is set by the `$theme-alert-icon-size` variable, rather than being hard-coded.

## Additional information

This makes the behavior of the design system match with what is in the documentation:

image

The Alert background size was increased in https://github.com/uswds/uswds/commit/92c7be5bc54fabedbc23bddf70aeeede6f0bee65 which causes issues for projects that have themed the design system and are utilizing their own icon files.

### DIFF
--- a/src/stylesheets/theme/_uswds-theme-components.scss
+++ b/src/stylesheets/theme/_uswds-theme-components.scss
@@ -25,7 +25,7 @@ $theme-accordion-font-family: "body";
 // Alert
 $theme-alert-bar-width: 1;
 $theme-alert-font-family: "ui";
-$theme-alert-icon-size: 4;
+$theme-alert-icon-size: 5;
 $theme-alert-padding-x: 2.5;
 
 // Banner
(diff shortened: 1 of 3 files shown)


src/stylesheets/theme/_uswds-theme-components.scss
```diff
@@ -25,7 +25,7 @@ $theme-accordion-font-family: "body";
 // Alert
 $theme-alert-bar-width: 1;
 $theme-alert-font-family: "ui";
-$theme-alert-icon-size: 4;
+$theme-alert-icon-size: 5;
```
sawyerh, line 28: Check me on this one. This variable _is_ used in one spot, which I imagine would be affected by this change:

https://github.com/uswds/uswds/blob/480fa3464b428ecb0aafd4f27df237b2a4ff0dd9/src/stylesheets/core/mixins/_alert-status-styles.scss#L18

thisisdano (approved):
Thanks, this is a great change and a good bugfix!

PR #3943, opened 2021-01-14
found a hotfix for babel transpiler issue with svg4everybody, this is…

sarnikbixal:
Babel compilers that add 'strict-mode' to all files will have issues transpiling this, unless users add special loaders. root object will become undefined using the polyfill. This hotfix set's the root object = window if it comes in undefined. This will work for those who are using default create-react-app webpacks and unable to eject them to add special loader logic.

- [x ] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).

### DIFF
--- a/src/js/polyfills/svg4everybody.js
+++ b/src/js/polyfills/svg4everybody.js
@@ -1,5 +1,6 @@
 /* eslint-disable */
 !function(root, factory) {
+  if(root === undefined) root = window; //*  THIS IS A HOTFIX FOR BABEL TRANSPILERS ADDING USING STRICT MODE CAUSING ROOT TO BE UNDEFINED. 
   "function" == typeof define && define.amd ? // AMD. Register as an anonymous module unless amdModuleId is set
   define([], function() {
       return root.svg4everybody = factory();


mejiaj (commented):
I'd prefer not to modify 3rd party code. Is there no way users can just ignore the compiled `uswds.min.js` file or we [exclude](https://babeljs.io/docs/en/options#exclude) the polyfill from transpiling?

sarnikbixal:
> I'd prefer not to modify 3rd party code. Is there no way users can just ignore the compiled `uswds.min.js` file or we [exclude](https://babeljs.io/docs/en/options#exclude) the polyfill from transpiling?

Not without making it a cjs (module) file or adding some special exclude logic in our NPM package. People who are using custom webpacking can add exlcude parameters in their webpack config (it's different for every version and tool). But people using something like create-react-app will not be able to add that unless they eject the webpack and modify themselves.

mejiaj:
Okay, just wanted to make sure we explored that option. Based on the lack of updates and numbers for IE11 dropping I'm good moving forward.

mejiaj (approved):

Issue #3935, opened 2021-01-12
Add table sorting


debjudy:
Code: https://github.com/uswds/uswds/pull/3950
Documentation: https://github.com/uswds/uswds-site/pull/1152

TODO
- [ ] First code review (@thisisdano)
- [x] Check heading order (@thisisdano)
- [x] First guidance review (@thisisdano)
- [x] Second code review (@mejiaj)
- [x] Further dev review (@scottqueen-bixal)
- [x] Merge code into `uswds-2.11.0`
- [ ] Second guidance review (@katieekline)
- [ ] Further content review (_optional_)
- [x] Link guidance to `uswds-2.11.0` (@thisisdano)
- [x] Update guidance to new formatting (@thisisdano)
- [x] Merge guidance into `uswds-2.11.0-docs`
- [ ] Update design assets if necessary (@thisisdano)

- - -

Ability to reorder table based on one of the columns
Assumptions 
* "sort by" drop down
* include in guidance when is it appropriate to make a sortable table
* Will use JavaScript
Note: for massive datasets, this might get harder.

**DESIGN ASSETS**
- [x] 10% - Collect and analyze examples; draft research insights and list of features for team/client feedback
- [x] 20% - Draft user story & clarify the problem(s) / user needs
- [x] 30% - Low-fi design review
- [x] 40% - Get community feedback on initial, Hi-fi designs
- [x] 50% - Hi-fi Sketch assets review
- [x] 60% - Hi-fi Adobe XD assets review
- [x] ~80% - Sketch and XD assets review w/ Dan + merged into master working files~
- [x] ~100% - PR approved by Dan~
----
**GUIDANCE**
- [x] 10% - Summarize insights from research (desk + interviews)
- [x] 20% - Draft guidance review no. 1 (minimum 2 bullets each section)
- [x] 40% - Get community feedback on draft guidance
- [x] 50% - Draft guidance review no. 2 (accessibility focus)
- [x] ~60% - Draft guidance review no. 3 (w/ Dan)~
- [x] ~80% - Add component to research report and slide deck~
- [x] ~100% - PR approved by Dan~
----
**CODE**
- [x] 10% - Code review no. 1 (Federalist preview) - interaction design focus
- [x] 20% - Code review no. 2 (Federalist preview) - visual design focus
- [x] 40% - Get community feedback or conduct tests with users, round 1
- [x] 60% - Code review no. 3 (Federalist preview) - accessibility focus
[...]

Issue #1643, opened 2016-12-21
[Component] Sortable table


shawnbot:
As mentioned in #1079, it would be really nice to offer a sortable table feature, or at least styles that indicate the sorted status of a table's column(s). To aid accessibility, we should offer the [`aria-sort` attribute](https://www.w3.org/TR/wai-aria/states_and_properties#aria-sort) as the styling mechanism. Specifically, we can probably do _all_ of the styling in CSS with generated content along the lines of:

```scss
[usa-sort] {
  background-color: $sortable-background;
  position: relative;

  &[aria-sort] {
    background-color: $sortable-background-sorted;

    &:after {
      position: absolute;
    }
  }

  &[aria-sort=ascending]:after {
    content: "▼";
  }

  &[aria-sort=descending]:after {
    content: "▲";
  }
}

```

As in all things a11y, we have to be super conscious of contrast and ensure that the sorted column header is sufficiently distinguishable from the others.

Functionally, sortable tables can be tricky to implement because we have to accommodate different types of sortable data: namely, that numbers should be sorted numerically and strings alphanumerically. Ideally, there would be a way to control whether each column is sortable and how it should be sorted. So maybe sorting is enabled via an optional attribute on the header, e.g. `usa-sort="num"` to enable numeric sorting or `usa-sort="alpha"` for alphanumeric.

And, because numbers can be formatted in all sorts of ways that are hard to parse as numbers for the sake of sorting in JavaScript (e.g. large negative values like `$(1,000,000)`), we should provide a way to pass a "raw" representation of the sortable value for each cell:

```html
<table usa-sortable="true">
  <thead>
    <tr>
      <th usa-sort="alpha">Product</th>
      <th usa-sort="num" aria-sort="ascending">Price</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Widget B</td>
      <td usa-sort-value="1234.56">$1,234.56</td>
    </tr>
    <tr>
      <td>Widget A</td>
      <td usa-sort-value="2016.33">$2,016.33</td>
    </tr>
  </tbody>
</table>
```

All of the behavior should be implemented with event delegation and keep state in DOM attributes so that tables can be modified by other code and sorting updated via user interactions.

thisisdano:
Thanks for contributing to USWDS. Unfortunately, this issue is not on our current [product roadmap](https://designsystem.digital.gov/whats-new/product-roadmap/).

sknep:
[...]

Issue #1079, opened 2016-03-25
Suggestions for tables


DruidSmith:
Suggest adding standard controls/icons for tables to allow sorting/filtering based on column heading; also standard icons/controls for downloading table data or visualizing the table as a graph

maya:
Hey thanks @DruidSmith, I've added this as a feature request and we'll see how this fits with our product roadmap https://github.com/18F/web-design-standards/wiki/Product-Roadmap and component prioritization: https://github.com/18F/web-design-standards/wiki/Component-prioritization:-The-factors-that-guide-our-prioritization-process

jhfrench:
[DataTables](https://github.com/DataTables/DataTables) is a well-written [open-source, free](https://github.com/DataTables/DataTables/blob/master/license.txt), plugin for progressively-enhancing tables to add filtering, sorting, pagination and even optional exports to PDF, CSV and clipboard. A lot of thought has gone into it.

I don't see jQuery in the Design Standards project, which is fine. But perhaps the good play here is to create a "Design Standards" [theme for DataTables](https://datatables.net/manual/styling/). That way a project that needs the enhancement(s) can include jQuery, DataTables, and the relevant theme CSS.

I'll donate the time to create the theme if it is wanted...just let me know!

jhfrench:
The visualization can be a little more tricky. I'd love to see [Highcharts](http://www.highcharts.com/) (which is "pure" JavaScript) adopted for this purpose. I know they are fleshing-out the DOM-reading widget of that software...perhaps they allow distribution of their product in exchange for making that part of the product better? [Highcharts is free for non-commercial organizations](http://www.highcharts.com/products/highcharts#non-commercial)...perhaps US Gov qualifies?

mebates:
+1 for table sorting standards.

shawnbot:
FYI, I fleshed out a proposal for the sortable table in #1643. We've ditched jQuery recently, so DataTables as a specific solution seems less appropriate. A solid stepping stone to offering our own component would be well-researched 3rd party library recommendations (as we've done with Politespace for zip code masking on the [address form](https://standards.usa.gov/form-templates/#address-form)), and I would happily suggest DataTables for folks who use jQuery elsewhere. I'd also like to recommend vanilla JS solutions, though; if you know of any, feel free to mention them in #1643.

[...]

PR #3950, opened 2021-01-19
Add table sorting

sknep:
# Adds table sorting to existing table component.

Component / Fractal Federalist Link: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/accelerator/3935-sortable-table/components/detail/table--sortable.html

USWDS Site Federalist Link: https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/accelerator/3935-table-sorting/components/table/

Addresses https://github.com/uswds/uswds/issues/3935
See also: #1643 , #1079

## Draft usage documentation 
The following will be added to [existing table page here](https://designsystem.digital.gov/components/table/) (subject to change):

### Guidance

**Enable sort where useful.** Add row sorting to individual columns of long tables where the data can be logically ordered either alphabetically or numerically. To activate row sorting, add the `data-sortable` attribute to the table header element (`<th>`) of any column with sortable data, and insert an element with the `aria-live=”polite”` attribute and the class “.usa-table__announcement-region” immediately following the table. 

**Set a default sort column and direction.** To sort a table’s rows by a specific sortable column on load, add the attribute `aria-sort` equal to a sort direction such as “ascending” or “descending” to that column header. For example, 

`<th data-sortable aria-sort="ascending" scope=”col”>`. 

**Provide raw values for cells with formatted number content.** If you have formatted your cell content for display (such as using percent, currency, or comma formatting) or if your cell content is non-numeric but should be sorted in a numeric order (such as months, days of the week, or dates), then provide a numeric-sortable value in a `data-sort-value` attribute on each table cell. For example: 

Numbers without currency or comma formatting: 

`<td data-sort-value="132773.54"> $132,773.54 </td>`

Percentages or fractions, converted to decimal:  

`<td data-sort-value="0.943"> 94.3% </td>`

Months, weekdays, or other orderable text: 

`<td data-sort-value="2”> February </td>`

Dates, in Unix timestamp: 

`<td data-sort-value="327092400"> Aug. 21, 1959 </td>`

**Don’t use row sorting with merged cells.** Sorting will not work properly if your table contains `colspan` or `rowspan` attributes on the cells.

**Don’t use row sorting with the mobile stacked variants.** With these variants, the column headers at the top of the table do not appear at narrow widths, and are instead moved into the cell content in each row as prefixing headers.

### Accessibility 

**Add an aria-live region to the page when enabling row sorting**. An `aria-live` region immediately following the `<table>` element automatically announces when the sort state changes for visitors using screen readers, but it must be added to the HTML document before load:

```
<div class="usa-sr-only usa-table__announcement-region" aria-live="polite">
  <!-- this content will change when sort changes -->
</div>
```

**Enabling row sorting automatically adds `aria-label` attributes** to the sortable column headers and their toggle sort buttons via JavaScript. These labels are updated to reflect each column’s current sort state (ascending, descending, or unsorted) whenever sort changes. You don’t need to apply these to sortable headers yourself.
### Implementation

(these will be added to the existing table)

Variable | Description | Default
-- | -- | --
`$theme-table-sorted-border-color` | Defines a border color for the outline of the column that is currently sorted | `"accent-cool-darker"`
`$theme-table-sorted-header-background-color` | Defines a background color for the header of the column that is currently sorted | `"accent-cool-light"`
`$theme-table-sorted-background-color` | Defines a background color for the cells of the column that is currently sorted | `"accent-cool-lighter"`
`$theme-table-sorted-stripe-background-color` | Defines a background color for the alternating horizontal stripes applied to cells of the column that is currently sorted | `"blue-cool-10v"`
`$theme-table-sorted-icon-color` | Defines a fill color for the activated toggle sort button in a sortable column header | `"ink"`
`$theme-table-unsorted-icon-color` | Defines a fill color for the unactivated toggle sort button in a sortable column header | `"base"`

### Other


jaredcunha:
Though we discussed, I'm adding some other ideas worth exploring here. 
- Deque has a nice accessible sortable table: https://dequeuniversity.com/library/aria/table-sortable I was able to review it using Voice Over and the audible feedback and instructions were quite nice.
- Exploring `role="grid"` and `aria-readonly="true"`, which are also used in the Deque example for improving keyboard navigation around cells.
- There's a visually hidden live region below the table that updates when tables are sorted.

sknep:
Thanks Jared! I added a "todo" section to my regular comment based on your suggestions.

src/js/components/table.js
```diff
@@ -0,0 +1,220 @@
...
+const behavior = require("../utils/behavior");
+const { CLICK } = require("../events");
+const { prefix: PREFIX } = require("../config");
+
+const TABLE = `.${PREFIX}-table`;
+const SORTED = "aria-sort";
+const ASCENDING = "ascending";
+const DESCENDING = "descending";
+const SORT_OVERRIDE = "data-sort-value";
+const ICON_PATH = "../../dist/img/sprite.svg";
```
sknep, line 11: @jaredcunha Is it a bad idea to hardcode the icons path in here? i want to dynamically change the button icon based on sort state, but not sure how to do that and still have the icons be anything but background images.

jaredcunha: I don't think we should have the path in the JS. It's going to be hard to change if someone is using USWDS JS out of the box and decides on a different image path. There's a Sass setting for this, but not a way to update in JS that I'm aware of.

src/js/components/table.js
```diff
@@ -8,10 +8,20 @@ const SORTED = "aria-sort";
...
+      <path d="M19,15l-1.41-1.41L13,18.17V2H11v16.17l-4.59-4.59L5,15l7,7L19,15z"/>
+    </g>
+    <g id="ascending">
+      <path transform="rotate(180, 12, 12)" d="M19,15l-1.41-1.41L13,18.17V2H11v16.17l-4.59-4.59L5,15l7,7L19,15z"/>
+    </g>
+    <g id="unsorted" >
+      <polygon points="15.17 15 13 17.17 13 6.83 15.17 9 16.58 7.59 12 3 7.41 7.59 8.83 9 11 6.83 11 17.17 8.83 15 7.42 16.41 12 21 16.59 16.41 15.17 15"/>
+    </g>
+  </svg>
+`;
```
sknep, line 24: @jaredcunha I've included the src for all three icons in this svg, and I use some CSS to control the fill color based on which group/node within the SVG should be active

src/js/components/table.js
```diff
@@ -0,0 +1,230 @@
...
+  const allRows = [].slice.call(tbody.querySelectorAll('tr'));
+  const allHeaders = [].slice.call(header.parentNode.children);
+  const thisHeaderIndex = allHeaders.indexOf(header);
+  allRows.sort(
+    compareFunction(
+      thisHeaderIndex,
+      !ascending
+    )
+  ).forEach(tr => {
+    [].slice.call(tr.children).forEach(td => td.removeAttribute("data-sort-active"));
```
mejiaj, line 121: Would `[...tr.children]` also work in this case?

Does babel not cover either of those cases?

src/js/components/table.js
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Compares the values of two row array items at the given index, then sorts by the given direction
+ * @param {number} index
+ * @param {string} direction
+ * @return {boolean}
+ */
+const compareFunction = (index, direction) => (a, b) => ((v1, v2) =>
+    // if neither value is empty, and if both values are already numbers, compare numerically. Otherwise, compare alphabetically based on current user locale
+    v1 !== '' && v2 !== '' && !Number.isNaN(Number(v1)) && !Number.isNaN(Number(v2)) ? v1 - v2 : v1.toString().localeCompare(v2, navigator.language, {numeric: true, ignorePunctuation: true})
```
mejiaj, line 52: Can we clean this up or format it in a way that's more readable?

mejiaj (changes requested):
Overall looks good. Don't forget to run prettier on the SCSS.

@scottqueen-bixal or @sarnikbixal can you take a look?

src/js/components/table.js
sknep: Unfortunately not. In IE11, this fails with "Invalid attempt to spread non-iterable instance"

src/js/components/table.js
sknep: Made some clarifying changes -- what do you think? https://github.com/uswds/uswds/pull/3950/commits/540972fdc9e9abe550bbf25181a1bd77971a1991

mejiaj: Yeah that and running prettier via `npx prettier --write ./src/js/**/*.js` might help.

src/js/components/table.js
```diff
@@ -0,0 +1,241 @@
...
+ * Compares the values of two row array items at the given index, then sorts by the given direction
+ * @param {number} index
+ * @param {string} direction
+ * @return {boolean}
+ */
+const compareFunction = (index, isAscending) => (thisRow, nextRow) => {
+
+  // get values to compare from data attribute or cell content
+  let value1 = getCellValue(isAscending ? thisRow : nextRow, index);
+  let value2 = getCellValue(isAscending ? nextRow : thisRow, index);
```
mejiaj, line 54: Do We should use `const` if these don't get reassigned.

src/js/components/table.js
```diff
@@ -0,0 +1,241 @@
...
+
+  // if neither value is empty, and if both values are already numbers, compare numerically
+  if (value1 !== '' && value2 !== '' && !Number.isNaN(Number(value1)) && !Number.isNaN(Number(value2))) {
+    return value1 - value2;
+  } else {
```
mejiaj, line 59: Do we need this `else`?

src/js/components/table.js
```diff
@@ -0,0 +1,241 @@
...
+ * @return {boolean}
+ */
+const compareFunction = (index, isAscending) => (thisRow, nextRow) => {
+
+  // get values to compare from data attribute or cell content
+  let value1 = getCellValue(isAscending ? thisRow : nextRow, index);
+  let value2 = getCellValue(isAscending ? nextRow : thisRow, index);
+
+  // if neither value is empty, and if both values are already numbers, compare numerically
+  if (value1 !== '' && value2 !== '' && !Number.isNaN(Number(value1)) && !Number.isNaN(Number(value2))) {
```
mejiaj, line 57: If we're checking if it's not empty couldn't we do something like:
```js
  if (value1 && value2 && !Number.isNaN(Number(value1)) && !Number.isNaN(Number(value2)))
```

And get the same result?

sknep: Your suggestions, as always, make the code better :) Thanks James.

mejiaj (approved):
Ran prettier on table.js. Otherwise LGTM, thank you!

mejiaj:
> 
> @mejiaj What command did you use to do this? the only one I see is `npm run prettier:scss`

Just modified the `prettier:scss` command to target the table.js. We should create a `prettier:js` command.

```bash
npx prettier --write ./src/js/components/table.js
```

src/components/table/table--sortable.njk
```diff
@@ -0,0 +1,227 @@
...
+      <th data-sortable scope="col" role="columnheader">
+        Estimated population at time of admission
+      </th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <th scope="row" role="rowheader">Hawaii</th>
+      <td data-sort-value="50">50th</td>
+      <td data-sort-value="331844400">Jun. 27, 1959</td>
```
jaredcunha, line 33: where are these sort values coming from?

src/js/components/table.js
jaredcunha: I like this approach!

src/components/table/table--sortable.njk
sknep: this number is a Unix timestamp

jaredcunha: Would it be worth in the guidance adding indicating supporting additional formats like YYYY:MM:DD? Thinking of support for tools where Unix timestamps might be as readily available.

sknep: I've deferred parsing dates from strings because they're finicky and would probably have weird issues across timezones unless we included a library like Moment.JS or one of its successors (which I am loathe to do).

sarnikbixal (approved):

scottqueen-bixal (approved):
I spent some time this morning building with this component in a React application and pulling data from https://www.ftc.gov/developer/api/v0/endpoints/do-not-call-dnc-reported-calls-data-api. 

**Implementation**
- I was able to get up and running within just a few minutes following the guidance from the documentation.

**table.js**
- `slice.call()` is good for those NodeLists
-  readability is good/clear
-  great comments

**spec**
- these assertions are deprecated, can we update? ie. `equal()` -> `strictEqual()` etc.

Everything works well for me, nice job

thisisdano:
@scottqueen-bixal @mejiaj Updated those deprecated assertions in this PR. 

Also, updated settings to use `default` colors for text. This will pull the color from either `$theme-text-color` or `$theme-text-reverse-color` based on background colors, unless specifically overridden in the setting.

Issue #3805, opened 2020-10-29
secondary link examples mis-positioned on Safari


femmebot:
Screenshot taken from [page template layout example](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/release-2.9.0/components/preview/layout--landing.html) on Safari

<img width="1426" alt="Screen Shot 2020-10-29 at 2 50 31 PM" src="https://user-images.githubusercontent.com/471514/97619430-96f1a480-19f6-11eb-9e3f-7d2053d5c28b.png">

thisisdano:
Oh hi Phoebe! 👋 I hadn't noticed this was your issue until I was drafting the release notes for the fix!

PR #3953, opened 2021-01-20
USWDS - Header: Fix header secondary links positioning on Safari.

akanelson:
Addresses #3805 

## Description

Fixes secondary link examples mis-positioned on Safari.


thisisdano (approved):

thisisdano:
Thanks for pointing out this bug! Your fix works, but I think I'd like to implement it in a more resilient manner, like using flex to lay out this container....  I'll link to the new PR when it's ready.

thisisdano:
New fix PR here: https://github.com/uswds/uswds/pull/3960

PR #3964, opened 2021-01-28
USWDS - Library: Test different template engine.

### LINKED ISSUE #3985: USWDS - Library: Remove PHP dependency in beta library so it compiles with Federalist.
Federalist fails with PHP twig. We should use a node twig engine so Federalist can build previews properly.

mejiaj:
## Preview
[Preview link →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-library--develop-check-engine/?p=all)

## Description
Closes #3985. This uses `engine-twig` to remove PHP dependency and get templates compiling properly on Federalist.

---

To get up and running:
```bash
npm install
npx patch-package
npx gulp watch
```

#### Namespacing doesn't work so we need to use relative paths.

Example:
```twig
❌ {% include "@components/usa-card/usa-card.twig" %}

✔ {% include "./components/usa-card/usa-card.twig" %}
```

---

#### Generated variants via YML don't work*

Example:
```bash
//  Works
usa-button-group.twig 

// Doesn't work without patch
usa-button-group~segmented.yml
```

_* This works with a patch 31931c7_

---

#### HTML in YML renders as string

```yml
# Given
items:
  - '<button class="usa-button">Map</button>'
  - '<button class="usa-button usa-button--outline">Satellite</button>'
  - '<button class="usa-button usa-button--outline">Hybrid</button>'
```

Results in:
image

You can get around with `autoescape false` in twig template
```twig
    <li class="usa-button-group__item">
      {% autoescape false %}
        {{ item }}
      {% endautoescape %}
    </li>
```

Or `|raw` filter

```twig
{{ domain.text|raw }}
```
---

#### Global variables don't work as expected
Setting a variable like `img_path` in `src/styleguide/data/data.yml` renders as empty string in some templates like `usa-nav__primary.twig`.

---


mejiaj:
No issues found reverting back to `engine-twig → engine-twig-php` with all current template changes.

mejiaj:
Need to fix watch task. `node-engine` doesn't compile twig files on change like php version, but it does build without errors and updates the twig code example.

Getting a compile requires stopping the task and restarting it every template change.

mejiaj:
CircleCI is failing because the testing scripts were commented out in [gulpfile.js:4](https://github.com/uswds/uswds/blob/jm-library--develop-check-engine/gulpfile.js#L4) because of their Fractal dependency.

package.json
```diff
@@ -23,6 +23,7 @@
     "prepare": "gulp",
     "preversion": "npm test",
     "release": "gulp release",
+    "federalist": "npm install && npx patch-package && npx gulp",
```
gscottqueen, line 26: consider adding the patch as a `postinstall` https://docs.npmjs.com/cli/v6/using-npm/scripts#npm-install
which would prevent us from having to include in the chain command here and as well for running after install manually when getting started locally

```suggestion
    "postinstall": "npx patch-package",
    "federalist": "npm install && npx gulp",
```

sarnikbixal (commented):
Only question I have is because of the Twing plugin is it required to change the pathing from "patternlab" to "build" ? Or can we just change the setting to point to patternlab rather than build - just less places we have to change that and if one was missed we don't have to search for it. Other than that looks like a terrible amount of work and I'm glad you guys got the raw end of the stick and not me :P. Good Job.

gscottqueen:
@mejiaj I was able to pull this feature locally and get running just fine.  I really like the documentation that you have included here with "Key Differences", wondering if that should make its way into an update for the README for posterity/reference 🤔

package.json
mejiaj: Added it, thank you!

mejiaj:
@sarnikbixal no file name requirements. I liked the more generic sounding `build` and Federalist was already looking for it anyways. Keeping it generic will allow us to test other libraries more easily in the future too.

And I appreciate the kind words! Glad we were able to come up with a good solution.

mejiaj:
> 
> 
> @mejiaj I was able to pull this feature locally and get running just fine. I really like the documentation that you have included here with "Key Differences", wondering if that should make its way into an update for the README for posterity/reference 🤔

@gscottqueen great idea, updated README.

PR #3980, opened 2021-02-05
gscottqueen-gsq-library--develop-check-engine

mejiaj:
## Description
Replaces pr #3977. Use async function for gulp task.

## Additional information

* Screenshot images
* Notes


package.json
```diff
@@ -34,7 +35,15 @@
     "test:visual:update": " node ./spec/visual-regression-tester.js test --updateGolden",
     "version": "gulp build release",
     "watch": "NODE_ENV=development nswatch",
-    "prettier:scss": "npx prettier --write ./src/stylesheets/**/*.scss"
+    "prettier:scss": "npx prettier --write ./src/stylesheets/**/*.scss",
+    "patternlab": "patternlab",
+    "pl:dev": "node ./node_modules/@pattern-lab/uikit-workshop/build-tools.js",
```
mejiaj, line 40: Do we need this `pl:dev` task?

package.json
```diff
@@ -34,7 +35,15 @@
...
+    "build:uikit": "cross-env-shell PL_CONFIG_PATH='${INIT_CWD}/.patternlabrc.js' npm run build --prefix node_modules/@pattern-lab/uikit-workshop -- --patternlabrc '$PL_CONFIG_PATH'",
```
mejiaj, line 41: Same with this `build:uikit` — what's this for?

package.json
```diff
@@ -34,7 +35,15 @@
...
+    "pl:build": "patternlab build --config ./patternlab-config.json",
+    "pl:help": "patternlab --help",
+    "pl:install": "patternlab install --config ./patternlab-config.json",
+    "pl:serve": "patternlab serve --config ./patternlab-config.json -v",
+    "pl:version": "patternlab --version"
```
mejiaj, line 46: Let's only include the tasks we need to build/watch the library.

gscottqueen: you probably don't need any of them since they are just alias for some patternlab CLI tasks.

from the docs
```
		build|compile [options]   Build the PatternLab. Optionally (re-)build only the patterns
		export                    Export a PatternLab patterns into a compressed format
		init [options]            Initialize a PatternLab project from scratch or import an edition and/or starterkit
	        install|add [options]     Installs Pattern Lab related modules like starterkits or plugins
		serve|browse [options]    Starts a server to inspect files in browser
```

but agree, we could/should only include the one we are leveraging if it helps to reduce confusion "pl:build"

patternlab-config.json
```diff
@@ -1,6 +1,6 @@
 {
   "cacheBust": true,
-  "cleanPublic": true,
+  "cleanPublic": false,
```
gscottqueen, line 3: for the local build test we are running incremental builds, that is the fastest way to recompile with the watch task - we likely want a "cleanPublic: true" for the default build

package.json
```diff
@@ -113,6 +124,7 @@
     "mocha": "^8.2.1",
     "node-notifier": "^9.0.0",
     "normalize.css": "^8.0.1",
+    "npm-force-resolutions": "0.0.3",
```
gscottqueen, line 127: this was pulled a bit quicker than I was expecting - but we may not need these forced resolutions.  I'll run a check and see if we can take them out of the package.

patternlab-config.json
mejiaj: I'd like to try incremental builds for performance and then revisit if/when it starts complicating things.

package.json
mejiaj: We can merge once this is verified.

gscottqueen: This was originally placed in to get around an error that is no longer present when we rebuild with the CLI.  We should remove
```suggestion
```
which would also include removing
```json
  "resolutions": {
    "chokidar": "^3.5.1",
    "webpack": "^5.20.2"
  }
```
and the preinstall
```json
  "preinstall": "npx npm-force-resolutions",
```
once those are gone, we would want to recompile and include the update `package-lock.json` 👍 

Thanks @mejiaj

Issue #3936, opened 2021-01-12
Add modal component


debjudy:
Code: https://github.com/uswds/uswds/pull/3981
Documentation: https://github.com/uswds/uswds-site/pull/1155

TODO
- [x] First code review (@thisisdano)
- [x] Check heading order (@thisisdano)
- [ ] First guidance review (@thisisdano)
- [x] Second code review (@mejiaj)
- [x] Further dev review (_optional_)
- [x] Merge code into `uswds-2.11.0`
- [ ] Second guidance review (@katieekline)
- [ ] Further content review (_optional_)
- [x] Link guidance to `uswds-2.11.0` (@thisisdano)
- [x] Update guidance to new formatting (@thisisdano)
- [ ] Merge guidance into `uswds-2.11.0-docs`
- [ ] Update design assets if necessary (@thisisdano)

- - -

* How to display a lightbox/modal
* When is it appropriate
* How to make accessible

https://cfpb.github.io/design-manual-archive/page-components/modals.html

Talking about the hierarchy of content "interruption" - modal = STOP and read, banner = important, featured content = group of related content to highlight. "users consume important info and then get back into what they need to do". There's often a lot of usability/accessibility issues with modals when done incorrectly

DESIGN ASSETS
- [x] 10% - Collect and analyze examples;  draft research insights and list of features for team/client feedback
- [x] 20% - Draft user story & clarify the problem(s) / user needs
- [x] 30% - Low-fi design review
- [x] 40% - Get community feedback on initial, Hi-fi designs
- [x] 50% - Hi-fi Sketch assets review
- [x] 60% - Hi-fi Adobe XD assets review
- [x] ~80% - Sketch and XD assets review w/ Dan + merged into master working files~
- [x] ~100% - PR approved by Dan~

GUIDANCE
- [x] 10% - Summarize insights from research (desk + interviews)
- [x] 20% - Draft guidance review no. 1 (minimum 2 bullets each section)
- [x] 40% - Get community feedback on draft guidance
- [x] 50% - Draft guidance review no. 2 (accessibility focus)
- [x] 60% - ~Draft guidance review no. 3 (w/ Dan)~
- [x] 80% - ~Add component to research report and slide deck~
- [x] 100% - ~PR approved by Dan~

CODE
- [x] 10% - Code review no. 1 (Federalist preview) - interaction design focus
- [x] 20% - Code review no. 2 (Federalist preview) - visual design focus
[...]

PR #3981, opened 2021-02-08
USWDS - Modal: Create modal component

jaredcunha:
## Description

This will add an accessible modal window to the design system
Issue: https://github.com/uswds/uswds/issues/3936

##  Screenshots
Default size
Screen Shot 2021-02-11 at 10 32 08 AM

Large size
Screen Shot 2021-02-11 at 11 05 10 AM

### Additional features
- Adds utility to capture scrollbar width
   -  This utility is added to `navigation.js` to fix content shifting when mobile navigation is opened.
- Adds feature to `focus-trap.js` to trap focus if focus is set to a window, instead of content within the window

### Accessibility features:
- Focus set to modal window when opened
- Focus returned to opening element (includes support for multiple openers)
- Traps focus to open modal
- Hides all other content when modal is open
- Reads title of modal when opened
- Press escape key to close, unless disabled with `force-action` data attribute

Other to-do's:

- [x] Write unit tests
- [x] Build size variants
- [x] Incorporate theme settings
- [x] Remove test modal from the documentation page (test for collisions with focus trapping on mobile nav)

Testing conducted:
- [x] IE11
- [x] Safari
- [x] Mobile Safari
- [x] Chrome
- [x] Chrome for iOS
- [x] VoiceOver
- [x] VoiceOver on iOS
- [x] Edge
- [x] Firefox
- [x] NVDA


jaredcunha:
Two size variants, default:
Screen Shot 2021-02-11 at 10 30 52 AM

Large:
Screen Shot 2021-02-11 at 10 32 08 AM

src/js/components/modal.js
```diff
@@ -0,0 +1,265 @@
...
+        originalOpener = this.getAttribute("id");
+      }
+      targetModal.setAttribute("data-opener", originalOpener);
+    }
+
+    // This basically stops the propagation if the element
+    // is inside the modal and not a close button or
+    // element inside a close button
+    if (clickedElement.closest(`.${MODAL_CLASSNAME}`)) {
+      if (clickedElement.hasAttribute(CLOSER_ATTRIBUTE) ||
```
jaredcunha, line 100: Had to revert back to writing the if statement like this. VoiceOver didn't seem to like: 

```js
    if (
      clickedElement.closest(`.${MODAL_CLASSNAME}`) &&
      (!clickedElement.hasAttribute(CLOSER_ATTRIBUTE) ||
        !clickedElement.closest(`[${CLOSER_ATTRIBUTE}]`))
    ) {
      event.stopPropagation();
      return false;
    }
  ```

**Comment by maya** [comment]
Would you consider removing the focus outline offset on the close button? I think it was more meant for the regular `usa-button`s.

**Comment by maya** [comment]
USWDS has been historically very minimal with animations/transitions b/c of motion sensitivities. There are two things with transitions on them currently here: the background and modal.

I think you should re-use the same background transition and opacity that the mobile header nav uses. I'd like to see an option with no transitions on the modal itself.

**Comment by jaredcunha** [comment]
I think it should remain a consistent indicator that keyboard users can follow when tabbing through the page. The focus indicator is used on mobile navigation, accordion, form controls, main navigation, and other interactive components.

**Comment by maya** [comment]
I'm not suggesting removing the focus indicator, that would be sacrilege! I'm only talking about the offset, which is only used on `usa-button` and not on the other elements you mentioned.

**Comment by jaredcunha** [comment]
Oh - I missed the offset part. I think we can do that.

**Comment by maya** [comment]
Isn't the focus supposed to be on the close button when you open the modal? I can't close it without tabbing through the entire modal.

**Comment by jaredcunha** [comment]
Modal uses the same transition setting as the main navigation, the `$project-easing` variable, which I think we can add as a token that folks can adjust. The opacity on the overlay is a little darker to create a little more contrast since the content is in the middle of the page.

> Isn't the focus supposed to be on the close button when you open the modal? I can't close it without tabbing through the entire modal. This is how the mobile nav the nav works.

Focus is set to the modal window initially, which on the screen reader won't read the title and the close button together. Focus doesn't have to go directly to the close button, as long as it goes to the window, and focus is trapped within that window while it's open, that is sufficient. If you want the close button to have focus when the window is opened, you can add a `data-focus` attribute to it. But after some research and a lot of discussions, we felt the best experience was reading the window and a quicker path the primary actions was a better experience.

**Comment by maya** [comment]
[WCAG 2.0 37 Success Criterion](https://www.w3.org/TR/WCAG20-TECHS/C27.html): C27: Making the DOM order match the visual order is a rule about having the same visual order as it is in the markup and the way it is now, the close button is after the content and doesn't comply with this rule.

**Comment by maya** [comment]
Consider equal spacing around the modal content as in this screenshot. It looks more polished and visually balanced:

Screen Shot 2021-02-19 at 6 39 57 PM

**Comment by maya** [comment]
As long as we can set `$project-easing` to 0/none that should work. I still encourage you to consider motion sensitivities as large objects that animate can trigger that for some people.

**Comment by jaredcunha** [comment]
> WCAG 2.0 Success Criterion 27: Making the DOM order match the visual order is a rule about having the same visual order as it is in the markup and the way it is now, the close button is after the content and doesn't meet the success criterion.

I believe there are _some_ exceptions to this, and [these modal demos from a Paciello Group developer](https://scottaohara.github.io/accessible_modal_window/tests/general.html) behave similarly.

**Comment by jaredcunha** [comment]
Additional background on leaving the close bottom of the modal code:

Changing the positioning of the close button to last is not a C27 failure or violation because it doesn't change the meaning of the content itself.  It makes sense to not have the close button first so that user knows what they are closing. In a weird Voice Over on Safari and Mobile Safari, the indicator moves to the first piece of content in the modal despite `focus` on the modal window. When the first item is the heading, the heading is read. When the first item is the close button, the close button is read. So I think it does make more sense that it is the heading.

See description of [F1 - Failure of Success Criterion 1.3.2 due to changing the meaning of content by positioning information with CSS](https://www.w3.org/WAI/WCAG21/Techniques/failures/F1#description)

Also applicable here is [G57 - Ordering the content in a meaningful sequence](https://www.w3.org/WAI/WCAG21/Techniques/general/G57)

> The objective of this technique is to ensure that the order of content presented to assistive technologies allows the user to make sense of the content. Some techniques permit the content to be rendered visually in a meaningful sequence even if this is different from the order in which the content is encoded in the underlying source file.

The second sentence in the quote describes the exception in C27 mentioned above.

File: src/stylesheets/components/_modal.scss [hunk 2]
```diff
@@ -0,0 +1,175 @@
...
+  &.is-visible {
+    visibility: visible;
+    opacity: opacity(100);
+    position: fixed;
+    z-index: z-index($theme-modal-z-index);
+  }
+}
+
+.usa-modal-overlay {
+  background: rgba(0, 0, 0, 0.5);
```
mejiaj, line 41:
```suggest
  background: color("black-transparent-50");
```

src/stylesheets/components/_modal.scss
```diff
@@ -0,0 +1,175 @@
...
+  width: 100%;
+  padding-top: units(4);
+}
+
+.usa-modal__main {
+  margin: 0 auto;
+  padding: units(1) units(4) units(4);
+}
+
+.usa-modal__content {
```
mejiaj, line 107: I think this is supposed to be `usa-modal__content--reverse`?

src/stylesheets/components/_modal.scss
```diff
@@ -0,0 +1,175 @@
...
+  flex-direction: column-reverse;
+}
+
+.usa-modal__header {
+  display: flex;
+  width: 100%;
+}
+
+// Prevents JS not recognizing as part of close button.
+.usa-modal-wrapper *[data-close-modal] > .usa-icon,
```
mejiaj, line 118: Is the `*` necessary here?

src/stylesheets/components/_modal.scss
jaredcunha: That might have been a loose end that fell through the cracks when going back and forth on where to place the close button. Should be safe to take two properties and move them to the selector on line 97. Then remove any instance of `usa-modal__content--reverse` because it won't be necessary.

src/stylesheets/components/_modal.scss
jaredcunha: Probably not. If you remove it and want to test, you'd have to do it using VoiceOver and Safari and attempt to close the modal using the VoiceOver controls (`ctrl` +`cmd` + `space`), not the `enter` key.

src/stylesheets/components/_modal.scss
jaredcunha: Had no idea that token existed!

src/stylesheets/components/_modal.scss
```diff
@@ -0,0 +1,175 @@
...
+  }
+}
+
+.usa-modal {
+  @include u-radius($theme-modal-border-radius);
+  @include u-shadow($theme-modal-box-shadow);
+  background: white;
+  display: inline-block;
+  margin: units(2.5) auto;
+  max-width: $theme-modal-default-max-width;
```
mejiaj, line 87: This seems to stay at 442.25px.

To recreate update `$theme-modal-default-max-width` to 700px. Might also cause some confusion with the `usa-modal--lg` variant that overrides the max-width anyways.

src/js/components/modal.js
```diff
@@ -0,0 +1,265 @@
...
+  let clickedElement = event.target;
+  const { body } = document;
+  const safeActive = !isActive();
+  const modalId = clickedElement
+    ? clickedElement.getAttribute("aria-controls")
+    : document.querySelector(".usa-modal-wrapper.is-visible");
+  const targetModal = safeActive
+    ? document.getElementById(modalId)
+    : document.querySelector(".usa-modal-wrapper.is-visible");
+  const openFocusEl = targetModal.querySelector(INITIAL_FOCUS)
```
mejiaj, line 66: Modal stops responding to clicks on the slightest resize.  

https://user-images.githubusercontent.com/3385219/109984362-4ef6cd00-7cc9-11eb-8fe0-a7dddf571aae.mp4

<details>
  <summary>❗ Console error</summary>

```
Uncaught TypeError: can't access property "querySelector", i is null
    L modal.js:66
    exports index.js:8
    exports index.js:4
    exports index.js:3
    simulateClickEffect browser-sync-client.js:10
    setTimeout handler*8558/e.simulateClickEffect/< browser-sync-client.js:10
    __tryOrSetError browser-sync-client.js:10
    next browser-sync-client.js:10
    _next browser-sync-client.js:10
    next browser-sync-client.js:10
modal.js:66:22
```

</details>

---

Reproduced in:
- Firefox 87.0b4 (64-bit)
- Chrome 89.0.4389.72

src/stylesheets/components/_modal.scss
mejiaj: I wonder if it'd be beneficial to provide settings for `usa-modal-overlay` and `usa-modal` to allow users to re-theme it for dark mode.

src/components/modal/modal.njk
```diff
@@ -0,0 +1,84 @@
...
+          </li>
+          <li class="usa-button-group__item">
+            <button type="button" class="usa-button usa-button--unstyled padding-105 text-center" data-close-modal>Sign out</button>
+          </li>
+        </ul> 
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
```
mejiaj, line 84: We should create settings for this component so we can use it like a real component. See [Search](https://github.com/uswds/uswds/blob/accelerator-3936/modal/src/components/search/search.njk) for example.

src/js/components/modal.js
jaredcunha: I see what's happening here. The modal uses the same body class and logic (`safeActive`) as the navigation when open: `usa-js-mobile-nav--active`. When you resize the window, the navigation js removes that class, so the modal thinks it should be opening a window instead of closing it, so it errors out.

I think a fix for this might just to create a new selector `usa-js-modal--active` and use that for `ACTIVE_CLASS`. It'd need a CSS selctor, like https://github.com/uswds/uswds/blob/781b9d00abf70eec9930333016b6999d9eca8a47/src/stylesheets/components/_navigation.scss#L349

jaredcunha: Here's an example of the modal successfully closing when `usa-js-mobile-nav--active` is added it back to the `<body>`:

https://share.getcloudapp.com/2NuEo1Ak

So I think what is described above should do the trick and eliminate the scripts from colliding.

jaredcunha:
@thisisdano @mejiaj - When the modals are open, might want to allow the focus indicator to display on the modal itself, in accordance with [Criterion 2.4.7](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html). It would involve removing [line 93 in the modal CSS](https://github.com/uswds/uswds/pull/3981/files#diff-6ea828aaa6c7117d3db5427db2a32b4a328037f3150583620ca3ad68fbdeef21R93). 

It'd look like this when the user opens the modal:
Screen Shot 2021-03-08 at 10 40 09 AM

src/js/components/modal.js
mejiaj: Thank you! Added this in a21ce77.

src/stylesheets/components/_modal.scss
```diff
@@ -0,0 +1,176 @@
...
+  padding-top: units(4);
+  width: 100%;
+}
+
+.usa-modal__main {
+  margin: 0 auto;
+  padding: units(1) units(4) units(4);
+}
+
+.usa-modal__header {
```
mejiaj, line 109: Is the intention that `usa-modal__heading` live inside here or is this separate?

src/stylesheets/components/_modal.scss
scottqueen-bixal: @mejiaj, tested this locally and it still seems to be able to close the modal without the `*` in front of `[data-close-modal]`

https://user-images.githubusercontent.com/37077057/110541909-63d2c680-80f6-11eb-9ea0-7e863b09d90d.mov

src/components/modal/modal.njk
mejiaj: Added in c6bcc42.

src/stylesheets/components/_modal.scss
thisisdano: The way we usually use this is that a `header` is a section and a `heading` is an `h`-level element. We currently have a heading, but it looks like the header is not yet implemented. I'm going to remove this code

thisisdano (approved):

src/components/test/kitchen-sink.njk
```diff
@@ -2,6 +2,8 @@
 
 {% block body %}
 
+  {% render '@modal' %}
```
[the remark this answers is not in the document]
thisisdano: I didn't remove it, rather just moved it to a more conventional location. It'll be useful to keep it these I think.

PR #3995, opened 2021-02-12
Typography: Fix links in `.usa-prose`

maya:
## Description

Fixes https://gsa-tts.slack.com/archives/C3F14AHSQ/p1613078830023900

Before you hit Submit, make sure you’ve done whichever of these applies to you:

- [ ] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [ ] Run `npm test` and make sure the tests for the files you have changed have passed.
- [ ] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [ ] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #3996, opened 2021-02-12
Fix link styling in usa-prose

thisisdano:
Internal clone of https://github.com/uswds/uswds/pull/3995
Removes nonfunctional link styling from `usa-prose`.

- - -

## Description

Fixes https://gsa-tts.slack.com/archives/C3F14AHSQ/p1613078830023900


thisisdano:
@maya this makes sense, but would it be better to encourage the use of the `$theme-global-link-styles: true` setting instead — which applies theme styles to `a`? I definitely see that the `. usa-prose > a` rule is pretty much ineffective, but I have a small worry about the specificity combined with the greater scope  of `. usa-prose a`.

maya:
Perhaps that's the way to go. I was initially concerned with applying that link styles globally, but maybe it won't be an issue. Might as well remove the link mixin from `usa-prose` then.

src/stylesheets/packages/_uswds-components.scss
```diff
@@ -14,10 +14,10 @@
 @import "../elements/layout-grid";
 @import "../elements/table";
 @import "../elements/tags";
-@import "../elements/typography/content";
 @import "../elements/typography/links";
 @import "../elements/typography/list";
 @import "../elements/typography/prose";
+@import "../elements/typography/content";
```
thisisdano, line 20: Why did this need to move in the import order?

maya: If links are removed from usa-prose you wouldn't need to. When they were displaying there was a [color a11y issue](https://app.circleci.com/pipelines/github/uswds/uswds/2047/workflows/b615ec35-0bd8-412f-b54c-5ddfc3507abf/jobs/6431) for the link on dark background, bc of a specificity issue.

src/stylesheets/elements/typography/_prose.scss
```diff
@@ -1,6 +1,10 @@
 .usa-prose {
   @include typeset($theme-prose-font-family);
+
   & > {
-    @include usa-content-styles;
+    @include usa-paragraph-style;
+    @include usa-headings-styles;
+    @include usa-list-styles;
+    @include usa-table-styles;
```
maya, line 8: @thisisdano you may want to remove the link style mixin from usa-content-styles and then re-add it here.

thisisdano: I think I will leave link styles in `usa-content-styles` if only because link styles are currently included when `$theme-global-content-styles: true` and I want to keep that conceptually consistent (ie `$theme-global-content-styles` means use `usa-content-styles`)

PR #4006, opened 2021-02-19
Js consolidate components

sarnikbixal:
Moved all JS files inside of their corresponding folders. Updated password.js to be its own unexposed component utilized by other templates.


mejiaj (approved):
```
 require("../usa-date-picker/date-picker");
```

We should look at LOE and implications of something like:
```
 require("../usa-date-picker");
```

Issue #3965, opened 2021-02-01
Summary Box Text Overflow/Responsiveness


ascreven:
### Describe the bug
Text in the summary box component is overflowing on some viewports when summary is inside of of grid columns

### Steps to reproduce the bug
Add grid-row and grid-cols outside of summary box and paragraphs inside box text

```
<div class="grid-row">
  <div class="grid-col-2">
    <div class="usa-summary-box" role="complementary">
      <div class="usa-summary-box__body">
        <h3 class="usa-summary-box__heading">Key information</h3>
        <div class="usa-summary-box__text">
          <p>
            Morbi nec semper nulla. In hac habitasse platea dictumst. Donec bibendum volutpat varius. Lorem ipsum dolor
            sit amet, consectetur adipiscing elit. Proin sollicitudin egestas fermentum. Sed ornare lectus et urna
            maximus vulputate. Nunc et consectetur lacus, varius vulputate arcu. Cras porta tempus magna sed dictum.
            Nullam rutrum dui sed mi bibendum gravida. Sed congue aliquet enim a sodales. Fusce ut lacus ipsum.
            Vestibulum ac imperdiet leo, sed convallis dui. Aenean pulvinar, ex eu lobortis pharetra, dolor massa
            molestie massa, et volutpat lorem neque accumsan purus. Integer volutpat eros a tempus pulvinar. Nam
            consectetur eget erat vitae dictum.</p>
          <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum facilisis maximus nulla a fermentum.
            Duis finibus ligula id ex ornare, vel rhoncus odio posuere. Vivamus iaculis lectus in orci laoreet
            fringilla. Cras sodales euismod dignissim. Sed dignissim, libero et tincidunt commodo, elit tortor
            sollicitudin neque, semper rutrum sem nunc eu magna. Proin cursus volutpat elit, semper scelerisque arcu
            vulputate sed. Maecenas in tellus a velit iaculis aliquam venenatis finibus eros. Aliquam auctor neque
            tempor, faucibus dolor ac, placerat lectus. Praesent vel massa libero. Pellentesque vel finibus velit, eget
            lobortis lorem. Integer vel lorem faucibus, interdum ante sit amet, imperdiet urna. Donec dignissim mauris
            lacus, et molestie lectus accumsan eu. Sed vel blandit diam. Pellentesque mollis ipsum a interdum sodales.
          </p>
        </div>
      </div>
    </div>
[...]

PR #4039, opened 2021-02-25
Summary box - Remove table-cell formatting

mitchmoccia:
This [issue was raised](https://github.com/uswds/uswds/issues/3965) when testing on smaller viewports. The `.usa-summary-box__body` had `display: table-cell` which was causing text to overflow outside of the summary box on some screens. The heading was also overflowing, so we added the following code to the class `.usa-summary-box__heading`:

```
@include at-media-max("mobile-lg") {
  word-wrap: break-word;
}
```

The use of at-media-max was employed to prevent `word-wrap: break-word;` from happening on larger screens with longer headings.

Here is the result:
image

Federalist link:
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/accelerator/3965-summary-box-text-overflow-fix/components/detail/summary-box.html


thisisdano:
I've updated this PR only to address the `table-cell` formatting, which is unnecessary. The change to `word-wrap` is a potential problem more severe than text overflow at extremely narrow-width situations

thisisdano (approved):

Issue #4053, opened 2021-02-26
File Input: Format validation is limited to drag-and-drop


aduth:
The File Input component only validates accepted formats on drag-and-drop interactions. While most browsers will filter files in a file picker dialog when an `accept` attribute is present, most will still allow the user to override this filter and select any file. If the expectation is that the File Picker will validate any files, it should run the validation on all changes, not only drop events.

### Steps to reproduce the bug

1. Navigate to https://designsystem.digital.gov/components/form-controls/#section-form-controls-08-file-input
2. In the "Input accepts only specific file types" (PDF, TXT) example, click the link to "choose from folder"
3. Change the filter to allow any file to be selected
4. Choose any file

**Expected behavior:** I am warned that the file is not a valid type.

**Actual behavior:** The file is selected, and no warnings are shown.

### Screenshots

Overriding filter format in macOS Chrome:

![image](https://user-images.githubusercontent.com/1779930/109360514-5fd0ba00-7855-11eb-8428-7119a3cc94b8.png)

After selecting an invalid file:

![image](https://user-images.githubusercontent.com/1779930/109360551-6d863f80-7855-11eb-8fa6-498f558859ba.png)

### System setup

- **USWDS version:** v2.10.1
- **Operating system:** macOS, Windows 10
- **Browser and version:** Chrome 88.0.4324.192, Internet Explorer 11

### Additional context

Existing validation occurs here:

https://github.com/uswds/uswds/blob/1f9bb1cca8afdd1d00481f00bf32a38169790058/src/js/components/file-input.js#L356-L363

Validation should probably occur during `onchange` as well:

https://github.com/uswds/uswds/blob/1f9bb1cca8afdd1d00481f00bf32a38169790058/src/js/components/file-input.js#L366-L368

The existing `preventInvalidFiles` may need to be adapted, because it assumes a drop event where files are assigned on the [`event.dataTransfer` property](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/files):

https://github.com/uswds/uswds/blob/1f9bb1cca8afdd1d00481f00bf32a38169790058/src/js/components/file-input.js#L210

mejiaj:
Moved to `2.12.0` release due to ongoing development.

PR #4054, opened 2021-02-26
Replace String#includes with browser-compatible indexOf

aduth:
This pull request seeks to replace code which would be incompatible if run in Internet Explorer with a browser-compatible variant.

## Additional information

`String#includes` is a feature of ECMAScript 2015 and isn't supported in Internet Explorer ([see browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#browser_compatibility)).

The intention of the code is to test whether the MIME type of an uploaded file matches an `accept` parameter, which can be in the format of either e.g. `image/png` or `image/*` ([see specification](https://html.spec.whatwg.org/multipage/input.html#file-upload-state-(type=file))). Therefore, it should be enough to use a `String#indexOf` to check to see that the MIME type of the file starts with the `accept` parameter type sans at most one `*`.

Note that it does not currently appear that this code is run in Internet Explorer due to a combination of the facts that only drop events are validated (#4053) and drag-and-drop doesn't appear to work in Internet Explorer (possibly a bug?).

An alternative here is to configure Babel to provide polyfills for these, by one of the following:

1. [`@babel/polyfill`](https://babeljs.io/docs/en/babel-polyfill/)
2. [`core-js`](https://github.com/zloirock/core-js)
3. [`@babel/preset-env`'s `useBuiltIns`](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) option
   - This is most promising since the preset is already being used ([source](https://github.com/uswds/uswds/blob/1f9bb1cca8afdd1d00481f00bf32a38169790058/config/gulp/javascript.js#L27))

### DIFF
--- a/src/js/components/file-input.js
+++ b/src/js/components/file-input.js
@@ -214,7 +214,7 @@ const preventInvalidFiles = (e, fileInputEl, instructions, dropTarget) => {
           const fileType = acceptedFiles[j];
           allFilesAllowed =
             file.name.indexOf(fileType) > 0 ||
-            file.type.includes(fileType.replace(/\*/g, ""));
+            file.type.indexOf(fileType.replace("*", "")) === 0;
           if (allFilesAllowed) break;
         }
       } else break;


mejiaj:
Thank you for this PR! We'll be extending this in PR #4149, so closing this for now.

PR #4059, opened 2021-03-01
Use theme reverse colors across reversed components

thisisdano:
[Preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/dw-use-proper-color-tokens/components/detail/buttons--outline-inverse.html) →

This PR replaces any hardcoded instance of colors that should more correctly refer to theme-able color settings. Specifically, to using reverse text and link colors.


src/stylesheets/elements/typography/_content.scss
```diff
@@ -45,14 +45,14 @@
   }
```
mejiaj, line 45: Should the `<p>` and `<span>` selector in `usa-dark-background` get the reverse text color?

It's currently:
```scss
  p,
  span {
    color: color("white");
  }
```

mejiaj (changes requested):

mejiaj (approved):

src/stylesheets/elements/typography/_content.scss
thisisdano: Made this change. The `span` seems unusual, but we'll leave it for now.

PR #4060, opened 2021-03-02
Add a white variant of the expand icon for dark bgs

thisisdano:
Adds a `expand_more--white.svg` icon to `usa-icons-bg` so there's an option available for projects that use the top nav with a dark background


thisisdano:
Oops, I messed this up — this should be built from a different base.

Issue #2835, opened 2018-10-08
Vertically center search when there are no secondary links


stphnwlkr:
With the search form now displayed, if there are no secondary links, the search bar sits below the midline in the desktop view.  If possible, have the search form vertically centered in the header area (aligned with usa-logo). The manual override  .usa-nav-secondary { bottom: 4.8rem;}, is fairly easy to implement, but it should be required (IMO).

![search_field](https://user-images.githubusercontent.com/1405187/46630887-a6ea7300-cb13-11e8-9041-9206ca454240.png)

It is even more pronounced on the demo page

![nav_issue2](https://user-images.githubusercontent.com/1405187/46633006-70642680-cb1a-11e8-91d1-205039339e0f.png)

mejiaj:
We could probably target it with the `.usa-nav__secondary:only-of-type` selector.

annepetersen:
Hi there: this is one of many submitted PRs/issues that have languished here. USWDS is currently understaffed, so I'm closing items that aren’t as high on the priority list right now, including this one, unfortunately.

PR #4065, opened 2021-03-03
USWDS - Navigation - vertically center search when there are no secondary links.

jeffffeffff:
For https://github.com/uswds/uswds/issues/2835

## Description

When `.usa-search` exists inside of `.usa-nav__secondary` alone, it was not vertically centered. This styling fix makes the assumption that `.usa-search` should vertically center by way of absolute positioning by default and then reset back to bottom: 0 when it's the sibling of `.usa-nav__secondary-links`. Another approach might be the target the vertical centering for `.nav__secondary form:only-child` assuming the case where the search form is the only form that should appear in this part of the DOM, but that seems both prescriptive and less clear in intent.

The current layout of the header and navigation for complex variants looks like it would be prohibitive to use flex or other nice auto-centering techniques, there's a lot of absolute positioning going on.

### DIFF
--- a/src/stylesheets/components/_navigation.scss
+++ b/src/stylesheets/components/_navigation.scss
@@ -212,11 +212,19 @@ $nav-link-arrow-icon-size: 2;
   }
 
   .usa-search {
-    margin-top: units(2);
     width: 100%;
+    bottom: units("05");
 
     @include at-media($theme-header-min-width) {
       margin-left: 0;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


mejiaj (approved):
Not a fan of magic numbers, but all things considered, it works. We've talked about auditing and updating this component.

thisisdano:
I'm concerned that this will have unintended downstream consequences

PR #4069, opened 2021-03-04
docs: update links to markdown section

mchelen-gov:
current link https://github.com/uswds/uswds#download does not match section id https://github.com/uswds/uswds#download-and-install


thisisdano (approved):
Thanks!

Issue #3998, opened 2021-02-16
Research testing options, omit Fractal


jAigret-Bix:
LOE: 1 week 

Currently tests via `gulp test` don't run in new pattern library because of missing fractal dependency via [fractal.config.js](https://github.com/uswds/uswds/blob/develop/fractal.config.js).

1. Uncomment [line 4 of gulpfile.js](https://github.com/uswds/uswds/blob/library--develop/gulpfile.js#L4) in `library--develop` branch
2. Run `gulp test`
3. View error 
```
[13:06:45] Using gulpfile ~/web/uswds/gulpfile.js
[13:06:45] Starting 'test'...

✖ ERROR: Error: Cannot find module '../fractal.config'
…
```
---

- [ ] Research  headless testing framework
- [ ] Discover what works w PL, How can you spin up a PL version in memory. 
- [ ] Do we need to build our own? what would that entail? 

**acceptance criteria**
- [ ] solution to unit testing
- [ ] create unit testing plan and timeline
- [ ] create issues for next steps and implementation

scottqueen-bixal:
Picking this up, will take a look into our testing suite and see how we can configure it to integrate it with our patternlab workshop.

sarnikbixal:
After researching options for testing I feel we should go with Cypress.io for spinning up a headless browser and axe testing.

I looked at both selenium and cypress and although selenium is more powerful and has more features, it has a steeper learning curve and the extra features do not really pertain to our case. Of those being able to spin up headless IE, and multiple tabs. 

Cypress has been gaining support to add IE testing in the future and I think that since we currently don't do headless testing on IE, we lose no functionality. 

Cypress also has an axe-core library that is widely used and seems easy to implement and import our config settings.

Next steps would be to setup a cypress build configuration then modify our tests with cypress syntax (very easy to read).

For CircleCI we will need to run our pattern-lab instance on the server, then tear it down. Previously Fractal ran a local instance on a built in node server. Since we have Node.js on our CircleCI we should have no problem running a local pattern lab instance on a port of choice for the headless browser to attach its-self too.

We can still use mocha-chai for the unit testing of components, this is mainly for axe and regression testing.

Sources:
https://www.cypress.io/
https://www.npmjs.com/package/cypress-axe

scottqueen-bixal:
yes, this was closed with,  https://github.com/uswds/uswds/pull/4071

jAigret-Bix:
This issue is complete with #4071 and can be closed @thisisdano

PR #4071, opened 2021-03-04
USWDS | PATTERN LAB - migrate testing suite

scottqueen-bixal:
## USWDS | PATTERN LAB - migrate testing suite

closes https://github.com/uswds/uswds/issues/3998

## Description
Currently, tests via `gulp test` don't run in the new pattern library because of missing fractal dependencies. 

We should expect that this new work runs our expected test which includes: 

- Sass linting
- Sass unit tests
- Javascript linting
- Javascript unit tests
- Accessibility tests
- Snyk security tests
- Typescript tests

> Although there are some changes here to how we call these tests, running `npm run test` still runs these as expected.

This update works to: 
- provide similar fractal utilities from PatternLab
- update Unit test references to any generated templates
- move our test specs into "mono-component" organization
- integrate our testing tasks into `gulp`
- update mocha config references
- update `CONTRIBUTING.md` notes

## Additional information

- [Fractal API](https://fractal.build/api/.) provided easy access to connect with servers and build files. Pattern Lab has not yet reached the same level of maturity from its [core usage](https://github.com/pattern-lab/patternlab-node/blob/master/packages/core/README.md#usage)
- The headless chrome constructor adds an unnecessary level of complexity for the working test suite, this has been simplified in our a11y spec.
- The visual regression spec and dependencies have been removed since it is not being utilized. 


.codeclimate.yml
```diff
@@ -8,5 +8,5 @@ plugins:
 exclude_patterns:
   - config/**/*.js
   - dist/**/*.js
-  - spec/**/*
+  - src/patterns/**/**/*.js"
```
mejiaj, line 11: Don't we still want codeclimate to run on the actual JS files?

For example: `src/patterns/components/usa-accordion/accordion.js`

CONTRIBUTING.md
```diff
@@ -90,7 +91,7 @@ Chrome v59 or higher (v60 if you're on Windows).
 
 If you want to run a single test file, run `npm run mocha ${path/to/spec-file}`,
 substituting the actual path to the spec. Only javascript files can be executed by the `mocha` runner,
-and only those js files in the `spec` directory ending with a `.spec.js`.
+and only those js files in the `src/patterns/` directory ending with a `.spec.js`.
```
mejiaj, line 94: Thank you for updating the docs!

Should this be `src/patterns/test/`?

.codeclimate.yml
scottqueen-bixal: good catch, yeah should probably be `src/patterns/**/*spec.js`

CONTRIBUTING.md
scottqueen-bixal: This could probably be confusing because of the way `patternlab-config.json` is set up right now.

```JSON
  "styleGuideExcludes": [
    "compare",
    "test"
  ],
```

maybe the way to resolve this is to update that config to say,

```JSON
  "styleGuideExcludes": [
    "compare",
    "test-patterns"
  ],
```

then our contrib file would state
```
and only those js files in the `src/patterns/*` directory ending with a `.spec.js`.
```

mejiaj: I agree, it is confusing to call it test. Let's rename it to `test-patterns` and we can merge.

package.json
```diff
@@ -79,11 +97,14 @@
   },
   "devDependencies": {
     "@babel/preset-env": "^7.12.10",
+    "@frctl/fractal": "^1.5.4",
```
mejiaj, line 100: Is it safe to remove this now?

scottqueen-bixal: yeah, let's leap off that bridge.  We can probably remove the `njk` dependency too.

mejiaj (approved):

PR #4077, opened 2021-03-10
Button: Include explicit state classes in unstyled hover, active overrides

aduth:
## Description

The unstyled button mixin customizes the hover and active appearance of a button, but only applies this with the browser-native `:hover` and `:active` modifiers, and not with USWDS' explicit `usa-button--hover` and `usa-button--active` modifier classes.

## Additional information

Before these changes, if one would apply `usa-button--hover` to an unstyled button, the overrides in these styles would not be applied, most notably the background color. You can test this by adding `usa-button--hover` class to the "Unstyled button" example in the Buttons component page.

Before|After
---|---
before|after

---

### DIFF
--- a/src/stylesheets/core/mixins/_button-unstyled.scss
+++ b/src/stylesheets/core/mixins/_button-unstyled.scss
@@ -26,4 +26,12 @@
   &.usa-button--disabled {
     color: color("disabled");
   }
+
+  &.usa-button--hover {
+    color: color($theme-link-hover-color);
+  }
+
+  &.usa-button--active {
+    color: color($theme-link-active-color);
+  }
 }
--- a/src/stylesheets/core/mixins/_button-unstyled.scss
+++ b/src/stylesheets/core/mixins/_button-unstyled.scss
@@ -13,7 +13,9 @@
   &:hover,
   &.usa-button--hover,
   &:active,
-  &.usa-button--active {
+  &.usa-button--active,
+  &:disabled,
+  &.usa-button--disabled {
     @include no-knockout-font-smoothing;
     background-color: transparent;
     box-shadow: none;
--- a/src/stylesheets/core/mixins/_button-unstyled.scss
+++ b/src/stylesheets/core/mixins/_button-unstyled.scss
@@ -11,7 +11,9 @@
   text-align: left;
 
   &:hover,
-  &:active {
+  &.usa-button--hover,
+  &:active,
+  &.usa-button--active {
     @include no-knockout-font-smoothing;
     background-color: transparent;
     box-shadow: none;
--- a/src/stylesheets/core/mixins/_button-unstyled.scss
+++ b/src/stylesheets/core/mixins/_button-unstyled.scss
@@ -21,4 +21,9 @@
     box-shadow: none;
     text-decoration: underline;
   }
+
+  &:disabled,
+  &.usa-button--disabled {
+    color: color("disabled");
+  }
 }
--- a/src/stylesheets/core/mixins/_button-unstyled.scss
+++ b/src/stylesheets/core/mixins/_button-unstyled.scss
@@ -12,8 +12,20 @@
 
   &:hover,
   &.usa-button--hover,
+  &:disabled:hover,
+  &:disabled.usa-button--hover,
+  &.usa-button--disabled:hover,
+  &.usa-button--disabled.usa-button--hover,
   &:active,
   &.usa-button--active,
+  &:disabled:active,
+  &:disabled.usa-button--active,
+  &.usa-button--disabled:active,
+  &.usa-button--disabled.usa-button--active,
+  &:disabled:focus,
+  &:disabled.usa-focus,
+  &.usa-button--disabled:focus,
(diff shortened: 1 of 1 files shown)


aduth:
I just noticed this also happens with disabled unstyled buttons, regardless if disabling is applied by browser-native `disabled` attribute or using the `.usa-button--disabled` modifier class.

**Screenshot:**

disabled button

src/stylesheets/core/mixins/_button-unstyled.scss
```diff
@@ -11,7 +11,9 @@
   text-align: left;
 
   &:hover,
-  &:active {
+  &.usa-button--hover,
+  &:active,
+  &.usa-button--active {
```
thisisdano, line 16: Let's add that `disabled` override as well...

```suggestion
  &.usa-button--hover,
  &:active,
  &.usa-button--active,
  &:disabled,
  &.usa-button--disabled {
```

thisisdano (changes requested):

src/stylesheets/core/mixins/_button-unstyled.scss
aduth: > Let's add that `disabled` override as well...

Makes sense to me 👍 

One thing I notice is that we'll likely want to override the default [white text color](https://github.com/uswds/uswds/blob/66b4487f1fb8eb754fb325acbaaaf3a587bbcfa7/src/stylesheets/core/mixins/_button-disabled.scss#L4) of a disabled button, as otherwise the button will be invisible 😅  The outline button is a similar example ([source](https://github.com/uswds/uswds/blob/66b4487f1fb8eb754fb325acbaaaf3a587bbcfa7/src/stylesheets/elements/_buttons.scss#L219)), though it applies the disabled text color using `:disabled` (expected) and a custom [`usa-button--outlined-disabled`](https://github.com/uswds/uswds/blob/66b4487f1fb8eb754fb325acbaaaf3a587bbcfa7/src/stylesheets/elements/_buttons.scss#L216) (not so expected).

Here, I'd propose something like:

```scss
&:disabled,
&.usa-button--disabled {
  color: color("disabled");
}
```

aduth:
Another thing I'm noticing is that we rely on the `typeset-link` mixin to apply text color for states, but that only applies those styles using `:hover`, `:active`, and `:focus`.

aduth:
> Another thing I'm noticing is that we rely on the `typeset-link` mixin to apply text color for states, but that only applies those styles using `:hover`, `:active`, and `:focus`.

Extended for hover and active states in 9cb66ff79 . `usa-focus` works well as-is.

aduth:
Today I encountered another conflict where a button can be both `:active` and `:disabled`, which also conflicts with styles between the unstyled button and default disabled button styles in `_button-disabled.scss`:

https://github.com/uswds/uswds/blob/0f1f4d0a12bb3981a52cc882c78b79931e6b004c/src/stylesheets/core/mixins/_button-disabled.scss#L7-L16

https://github.com/uswds/uswds/blob/0f1f4d0a12bb3981a52cc882c78b79931e6b004c/src/stylesheets/elements/_buttons.scss#L49-L51

The selector being revised here could be expanded to include those additional combination of states, to ensure that the background color is being reset correctly to `transparent`:

```
  &.usa-button:disabled:hover,
  &.usa-button:disabled.usa-button--hover,
  &.usa-button:disabled:active,
  &.usa-button:disabled.usa-button--active,
  &.usa-button:disabled:focus,
  &.usa-button:disabled.usa-focus
```

Context: https://github.com/18F/identity-idp/pull/4881

mejiaj (approved):

mejiaj:
The only outstanding thing is this [comment](https://github.com/uswds/uswds/pull/4077#issuecomment-813648888) about disabled states in `uswds/src/stylesheets/elements/_buttons.scss`.

aduth:
> The only outstanding thing is this [comment](https://github.com/uswds/uswds/pull/4077#issuecomment-813648888) about disabled states in `uswds/src/stylesheets/elements/_buttons.scss`.

Oops, I'd meant to revisit this sooner, though I'm glad I'd at least left a comment about the finding 😅 

I pushed a change for this in 8a4a326. Overall I wish it wouldn't be necessary to have such a large grouping of selectors, though there's a lot of different states to account for. This should now account for the [disabled hover, active, and focus styles](https://github.com/uswds/uswds/blob/0f1f4d0a12bb3981a52cc882c78b79931e6b004c/src/stylesheets/core/mixins/_button-disabled.scss#L7-L12) for the mixin's usage in both [`.usa-button:disabled`](https://github.com/uswds/uswds/blob/0f1f4d0a12bb3981a52cc882c78b79931e6b004c/src/stylesheets/elements/_buttons.scss#L49-L51) and [`.usa-button--disabled`](https://github.com/uswds/uswds/blob/0f1f4d0a12bb3981a52cc882c78b79931e6b004c/src/stylesheets/elements/_buttons.scss#L193-L195).

thisisdano (approved):

PR #4078, opened 2021-03-10
Footer: Switch to matchMedia for viewport-specific JavaScript

aduth:
Changes from responding to resize event with conditions on current viewport size, to event-based matchMedia.

Why:

- Fewer runtime dependencies, reducing bundle size ([~2kb](https://bundlephobia.com/result?p=lodash.debounce@4.0.8)) and maintenance overhead.
- Faster response time of viewport changes, since viewport-specific logic is run immediately, without a debounce delay.
- Simpler tests, due to the absence of timer scheduling accounting for debounce.
- Less computationally-intensive, since resize events are no longer handled, regardless whether they're debounced.

Browser support is good, supporting IE10 and newer ([reference](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener)).

---

### DIFF
--- a/spec/unit/footer/footer.spec.js
+++ b/spec/unit/footer/footer.spec.js
@@ -1,11 +1,11 @@
 const assert = require("assert");
 const fs = require("fs");
 const path = require("path");
+const { default: matchMediaPolyfill } = require("mq-polyfill");
 const behavior = require("../../../src/js/components/footer");
 
 const TEMPLATE = fs.readFileSync(path.join(__dirname, "/template.html"));
 
-const { DEBOUNCE_RATE } = behavior;
 const HIDDEN = "hidden";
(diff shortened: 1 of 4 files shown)


spec/unit/footer/footer.spec.js
```diff
@@ -24,9 +24,9 @@ const resizeTo = width =>
   new Promise(resolve => {
```
aduth, line 24: These tests could be simplified a lot more. This doesn't need to return a promise anymore, since it resolves immediately. That said, the tests are currently written assuming this to return a promise, so more changes would be needed to remove the promise.

thisisdano (commented):
This looks good to me — passing to @mejiaj and @scottqueen-bixal for another review. Thanks!

mejiaj (approved):

PR #4079, opened 2021-03-11
Add better support for custom icon colors

### LINKED ISSUE #4064: Provide better support for color icons that adapt to theme settings
### Describe the bug
Any attempt to change `$theme-text-color` to anything other than 'ink' leads to an error. 

From @aduth 
> It does seem like there might be a bug here, in that [this code](https://github.com/uswds/uswds/blob/aead326413aee55858faa17fc689610eff9d3063/src/stylesheets/core/mixins/_alert-status-styles.scss#L10) expects to find icons of various alert types (warning, etc) in [the img/alerts folder](https://github.com/uswds/uswds/tree/350a02ea0d1b427d1a8205bd779e02f8fd0b01b5/src/img/alerts) ending in `-white`, but there’s only the error icon there.

1. Change the value of `$theme-text-color` to anything other than `ink`
2. Look at the info alert, it will no longer have an icon and there will be an error in the console

This may affect more than just the alert.

### Expected behavior
In addition to updating the text color, the system should be able to display icons that match the theme color in relevant components.

### Additional context
We want to use our color icon mixins to place these icons into an alert.

thisisdano:
Fixes #4064 

- Support automatic icon color in alerts
- Fix color icon mixin to better support IE variants
- Uses new icon set in thew banner

### DIFF
--- a/src/stylesheets/components/_alerts.scss
+++ b/src/stylesheets/components/_alerts.scss
@@ -113,11 +113,9 @@ $alert-padding-left: units($theme-alert-bar-width);
 }
 
 .usa-alert--validation {
-  background-size: units(3);
-  background-position: $alert-padding-left units($theme-alert-padding-x);
-
-  .usa-alert__body {
-    padding-left: units(5);
+  &:before {
+    background-position: 0 units(1);
+    mask-position: 0 units(1);
   }
 
   .usa-checklist {
--- a/src/stylesheets/components/_alerts.scss
+++ b/src/stylesheets/components/_alerts.scss
@@ -8,7 +8,7 @@ $alert-icon-optical-padding: calc(
   $this-icon-object: map-merge(
     $icon-object,
     (
-      "name": "usa-icons/#{$name}",
+      "name": $name,
       "color": $color,
       "height": $theme-alert-icon-size,
     )
@@ -45,7 +45,7 @@ $alert-icons: (
 );
 
 $icon-object: (
-  "name": "usa-icons/ICON_NAME",
+  "name": "ICON_NAME",
   "svg-height": 40,
   "svg-width": 40,
   "height": $theme-icon-image-size,
--- a/src/stylesheets/components/_alerts.scss
+++ b/src/stylesheets/components/_alerts.scss
@@ -1,3 +1,31 @@
+$alert-icon-optical-factor: calc(#{units($theme-alert-icon-size)} / 5);
+$alert-icon-optical-padding: calc(
+  #{units($theme-alert-padding-x)} + #{units($theme-alert-bar-width)} - #{$alert-icon-optical-factor}
+);
+
+@mixin add-alert-icon($name, $color) {
+  $this-icon-object: map-merge(
+    $icon-object,
+    (
+      "name": "usa-icons/#{$name}",
+      "color": $color,
+      "height": $theme-alert-icon-size,
+    )
+  );
+  &:after {
+    @include add-color-icon($this-icon-object, "base-lightest");
+    // Position centered in the alert
+    bottom: 0;
+    content: "";
+    display: block;
+    // bar + padding - optical spacing value
+    left: $alert-icon-optical-padding;
+    position: absolute;
+    height: auto;
+    top: 0;
+  }
+}
+
 // Alert variables ---------- //
 
 $alert-icons: (
@@ -8,13 +36,24 @@ $alert-icons: (
   emergency: "error",
 );
 
-$alert-padding-left: units($theme-alert-padding-x) +
-  units($theme-alert-bar-width);
+$icon-object: (
+  "name": "usa-icons/ICON_NAME",
+  "svg-height": 40,
+  "svg-width": 40,
+  "height": $theme-icon-image-size,
+  "color": "ink",
+);
+
+$alert-padding-left: units($theme-alert-bar-width);
 
 .usa-alert {
   @include typeset($theme-alert-font-family);
   @include border-box-sizing;
-  background-color: color("base-lightest");
+  @include set-text-and-bg(
+    "base-lightest",
+    $theme-alert-text-color,
+    $theme-alert-text-reverse-color
+  );
   background-position: $alert-padding-left units($theme-alert-padding-x - 1);
   background-repeat: no-repeat;
   background-size: units($theme-alert-icon-size);
@@ -29,7 +68,7 @@ $alert-padding-left: units($theme-alert-padding-x) +
   }
 
   // TODO: why is this not simply a border?
-  &::before {
+  &:before {
     background-color: color("base-light");
     content: "";
     height: 100%;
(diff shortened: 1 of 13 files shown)


src/stylesheets/components/_alerts.scss
```diff
@@ -8,36 +44,32 @@ $alert-icons: (
...
-  &::before {
-    background-color: color("base-light");
-    content: "";
-    height: 100%;
-    left: 0;
-    position: absolute;
-    top: 0;
-    width: units($theme-alert-bar-width);
-  }
+  border-left: units($theme-alert-bar-width) solid color("base-light");
```
thisisdano, line 72: Change alert bar to a border for simplicity

src/stylesheets/components/_alerts.scss
```diff
@@ -1,3 +1,39 @@
+$alert-slim-icon-size: units(3);
```
thisisdano, line 1: We'll be adding this icon with a `:before` block and the mask technique instead of a background image on the alert itself, so some vars and calculations need to change.

src/stylesheets/components/_alerts.scss
```diff
@@ -49,14 +81,14 @@ $alert-padding-left: units($theme-alert-padding-x) +
...
+@each $name, $icon in $alert-icons {
+  .usa-alert--#{$name} {
+    @include alert-status-styles($name, $icon);
+  }
 }
 
 .usa-alert__body {
-  display: table-cell;
-  vertical-align: top;
+  @include u-padding-x($theme-alert-padding-x);
```
thisisdano, line 91: Remove the weird `table`-related styling

src/stylesheets/core/mixins/_alert-status-styles.scss
```diff
@@ -2,23 +2,20 @@
   $bgcolor: if($name != "emergency", "#{$name}-lighter", $name);
   $banner-text-color-token: get-color-token-from-bg(
     $bgcolor,
-    $context: "Alert text"
+    $theme-alert-text-reverse-color,
```
thisisdano, line 5: Use theme colors

src/stylesheets/core/mixins/_icon.scss
```diff
@@ -116,15 +116,16 @@
...
+  $filename-ie11: if(
+    $ie11-variant == "white",
+    "usa-icons-bg/#{$filename-base}--white.svg",
+    "usa-icons/#{$filename-base}.svg"
   );
 
-  $image-props: url("#{$path}/#{$filename}") no-repeat center / #{$width} #{$height};
+  $mask-props: url("#{$path}/#{$filename-base}.svg") no-repeat center / #{$width}
+    #{$height};
+  $image-props: url("#{$path}/#{$filename-ie11}") no-repeat center / #{$width} #{$height};
```
thisisdano, line 128: Fix this code so it pulls the right `--white` variant from `usa-icons-bg` for IE support

src/stylesheets/settings/_settings-components.scss
```diff
@@ -27,6 +27,7 @@ $theme-alert-bar-width: 1 !default;
 $theme-alert-font-family: "ui" !default;
 $theme-alert-icon-size: 5 !default;
 $theme-alert-padding-x: 2.5 !default;
+$theme-alert-padding-y: 2 !default;
```
thisisdano, line 30: Add explicit `padding-y` setting instead of calculating it from `padding-x`, which isn't expected.

src/stylesheets/components/_alerts.scss
```diff
@@ -1,3 +1,39 @@
...
+    @include add-color-icon($this-icon-object, $bgcolor);
+    // Position centered in the alert
+    bottom: 0;
+    content: "";
+    display: block;
+    // padding - optical spacing value
+    left: $alert-icon-optical-padding;
+    position: absolute;
+    height: auto;
+    top: 0;
```
thisisdano, line 26: This changes the icon positioning a bit, so the icon is vertically centered on the alert.

mejiaj: IE11 doesn't support `calc()` on generated content.

Maybe we should try a `left` value before this one or something like:
```
// Some left value for ie11
left: 1rem;

// Calc for modern browsers
@supports (left: $alert-icon-optical-padding) {
  left: $alert-icon-optical-padding;
}
```

image

src/stylesheets/components/_alerts.scss
mejiaj: We might need a `$alert-slim-icon-padding`. The icon on slim variation stays the same but the padding varies when the icon size is updated.

For example, change `$theme-alert-icon-size` to `6`.

image

src/stylesheets/settings/_settings-components.scss
```diff
@@ -27,6 +27,7 @@ $theme-alert-bar-width: 1 !default;
...
 $theme-alert-text-color: default !default;
 $theme-alert-text-reverse-color: default !default;
```
mejiaj, line 32: Should we deprecate these?

We could use the defaults instead
```
$theme-text-color
$theme-text-reverse-color
```

src/stylesheets/components/_alerts.scss
mejiaj: We should move this up before `position: relative`.

src/stylesheets/components/_alerts.scss
```diff
@@ -74,25 +106,13 @@ $alert-padding-left: units($theme-alert-padding-x) +
...
 .usa-alert--slim {
   @include alert-slim-styles;
 }
 
-.usa-alert--no-heading {
-  background-position: $alert-padding-left
-    calc(#{units($theme-alert-padding-x)} - #{units(0.5)});
-}
-
 .usa-alert--validation {
```
mejiaj, line 116: This variant needs the latest changes.

image
[Source](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/dw-color-icons/components/detail/validation.html)

mejiaj (changes requested):
Noticed a few issues, 1 on IE11 and another with the validation variant.

src/stylesheets/components/_alerts.scss
thisisdano: It doesn't look like the calc() is working for the content `padding-left` either

thisisdano: Just removed the calc()s and used Sass math instead

src/stylesheets/settings/_settings-components.scss
thisisdano: They're maybe a little edge-y but if you wanted a different color for alerts or reverse alerts it would be helpful. I'd prefer not to deprecate them one release after introducing them, I guess. 😬

src/stylesheets/components/_alerts.scss
thisisdano: DONE

mejiaj (approved):

PR #4080, opened 2021-03-11
Checkbox, Radio: Prevent stray A circumflex (Â)

andycochran:
## Description

The `Â` character can show up where a non-breaking space should be when `<meta charset="utf-8">` isn't set properly. The non-breaking space character is byte 0xA0 in ISO-8859-1. When encoded to UTF-8 it'd be 0xC2,0xA0, which, if incorrectly viewed as ISO-8859-1 displays as `Â`. 

This bug is manifesting in our app's radio buttons and checkboxes: 

image

…since `content: "\a0";` is used to render checkboxes and radios: 

https://github.com/uswds/uswds/blob/a683079b1409a00345162a81ec90cef7d966d17f/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss#L40-L48

This change refactors the checkboxes and radios to use a regular space (and `white-space: pre;` so it does not collapse). This prevent stray A circumflex (Â) characters from appearing next to checkboxes and radio buttons.


thisisdano (approved):
Nice. This is a smart fix.

PR #4092, opened 2021-03-15
2.11.0 bugfixes

thisisdano:
## Alert icon position
The icon position in site alert regressed when we implemented color icon in alerts. This updates the padding and margin in the various site alert variants.

This also reverts the icon alignment to top and reduces its size to be more like the previous icon size.

Before:
Screen Shot 2021-03-15 at 3 45 19 PM

After:
Screen Shot 2021-03-15 at 3 45 01 PM

## Sortable table button hover
Fixed the hover state on the sortable indicator to fill the icon with the header text color, not the sorted header text color

## Input prefix and suffix
- Set an explicit white background for inputs on non-white backgrounds
- Give more padding around prefix or suffix to accommodate dynamically colored backgrounds
- **Note:** Input suffix and prefix doesn't support dynamic background colors, like the background color set by autofill values in webkit
- Added `--error` and ``--[size]` variants to `input-group` just as we have them for `input`

Example of dynamic color background (autofill in chrome) and prefix:
Screen Shot 2021-03-16 at 2 22 21 PM

## Banner
Fixed an alignment issue with the `Here's how you know` button. _Note the button has moved down slightly to the text baseline._

Before:
Screen Shot 2021-03-16 at 2 25 41 PM

After:
Screen Shot 2021-03-16 at 2 25 51 PM

### DIFF
--- a/src/components/input-prefix-suffix/input-prefix-suffix.njk
+++ b/src/components/input-prefix-suffix/input-prefix-suffix.njk
@@ -9,6 +9,16 @@
   <input type="text" id="example-input-prefix" class="usa-input" pattern="[0-9]*" inputmode="numeric">
 </div>
 
+<label class="usa-label" for="example-input-prefix-error">Credit card number (Error)</label>
+<div class="usa-input-group usa-input-group--error">
+  <div class="usa-input-prefix" aria-hidden="true">
+    <svg aria-hidden="true" role="img" focusable="false" class="usa-icon">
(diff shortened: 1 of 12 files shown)


mejiaj (approved):

mejiaj:
Emergency icon disappears on IE11

image

thisisdano:
I do see the icon in IE11, not sure what's up...

Screen Shot 2021-03-16 at 9 14 54 AM

mejiaj:
Checked again and am seeing it in IE11. Sorry about that.

PR #4115, opened 2021-03-22
Buttons: Use "inherit" for font smoothing reset

aduth:
## Description

The unstyled button resets font smoothing which may have been applied by e.g. disabled button styling. However, the reset is opinionated and assumes a specific default font smoothing. This can cause some inconsistency between the appearance of links and unstyled buttons if the page's default font smoothing does not match the opinionated reset styling. Instead, reset using the "inherit" keyword to inherit what would be the normal font smoothing behavior absent smoothing applied by the "add-knockout-font-smoothing" mixin.

## Additional information

See CodePen demo: https://codepen.io/aduth/pen/WNReKGb

In the demo, it should be expected that links and unstyled buttons always appear visually identical. However, you can see in "Before" "Webkit anti-aliased" content, the unstyled button appears slightly heavier font weight than the link.

image

The difference is subtle to notice, only appears in Webkit/Blink-based browsers (Chrome, Safari), and may also depend on the pixel density of the display being used.

These revisions may change the behavior of an unstyled link appearing in dark background content area. Dark background content applies font smoothing to its content ([source](https://github.com/uswds/uswds/blob/1908d1391bc59410624ca1934cc70b7404e8f443/src/stylesheets/elements/typography/_content.scss#L39)). Previously, since the reset styles were opinionated, the smoothing would bypass the smoothing applied by the dark background. Now it would inherit these as the default appearance. This may be desirable, but it's worth clarifying.

---

### DIFF
--- a/src/stylesheets/core/mixins/_add-knockout-font-smoothing.scss
+++ b/src/stylesheets/core/mixins/_add-knockout-font-smoothing.scss
@@ -4,6 +4,6 @@
 }
 
 @mixin no-knockout-font-smoothing {
-  -moz-osx-font-smoothing: auto;
-  -webkit-font-smoothing: subpixel-antialiased;
+  -moz-osx-font-smoothing: inherit;
+  -webkit-font-smoothing: inherit;
 }


mejiaj:
I'm not able to reproduce on Win10 machine. Tested in Chrome v89.0.4389.90 on a 1440p display.

aduth:
> I'm not able to reproduce on Win10 machine. Tested in Chrome v89.0.4389.90 on a 1440p display.

Would you happen to have access to a macOS machine? Based on the MDN documentation, it would seem to only be a feature supported on Mac:

>**WebKit** implements a similar property, but with different values: **`-webkit-font-smoothing`**. It only works on Mac OS X/macOS.

https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth

mejiaj:
I don't at the moment, unfortunately. Thanks for pointing that out!

mejiaj:
@scottqueen-bixal can you take a look at this and confirm?

scottqueen-bixal:
This is appearing as expected for me on `Version 1.22.71 Chromium: 89.0.4389.114 (Official Build) (x86_64)`

scottqueen-bixal (approved):

thisisdano (approved):

PR #4132, opened 2021-03-26
Button: Remove pointer-events styling from disabled buttons

### LINKED ISSUE #4122: Button: Disabled button pointer events styling may be unnecessary or is at least inconsistent
### Describe the bug

Disabled buttons currently apply a `pointer-events: none` style that prevents pointer interactions (mouse clicks, hover) from taking effect:

https://github.com/uswds/uswds/blob/3dc296ec56cd621fe52d918701fd94621d96a198/src/stylesheets/core/mixins/_button-disabled.scss#L5

This can stop click handlers from being triggered if a user clicks on the button using a mouse, but importantly it does nothing to prevent a user from "clicking" the button when using the keyboard.

Instead, I'd expect one or the other of the following:

- If the intention is to prevent click interactions, this should be consistent for clicks which occur by keyboard presses, which may require JavaScript.
- Alternatively, for use-cases where someone may want to give the visual appearance of a disabled button without preventing interactions, the `pointer-events: none;` provides an inconsistent experience and should be removed.
   - (Anecdotally, this is the use-case in our project)

Note that this is only relevant for disabled buttons which don't already use [a `disabled` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled), where click, hover, _and keyboard focus_ interactions would already be disabled, and should be the typical implementation where these behaviors are desired.

1. Go to https://codepen.io/aduth/pen/NWdGQQG
2. Click on the button. Observe that there is no alert shown.
3. Focus the button by clicking into the preview frame and tabbing to the button.
4. "Click" the button by pressing the <kbd>Space</kbd> key. Observe that an alert is shown.

### Expected behavior

The alert is consistently either never shown or always shown, regardless if the the click occurs by mouse or by keyboard.

### System setup

- **USWDS version:** 2.11.1
- **Operating system:** macOS Catalina
- **Browser and version:** Chrome 89

aduth:
Closes #4122

## Description

Removes `pointer-events: none` styling from disabled buttons.

## Additional information

See #4122 for additional context.

Developers who wish to disable interactivity on disabled buttons should apply [a `disabled` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled), as this is its intended purpose. At worst, using `pointer-events: none;` could give developers a false sense of confidence that their buttons are non-interactive, when in-fact this is not true (see #4122).

### DIFF
--- a/src/stylesheets/core/mixins/_button-disabled.scss
+++ b/src/stylesheets/core/mixins/_button-disabled.scss
@@ -2,7 +2,6 @@
   @include add-knockout-font-smoothing;
   background-color: color("disabled");
   color: color("white");
-  pointer-events: none;
 
   &:hover,
   &.usa-button--hover,
--- a/src/stylesheets/elements/_buttons.scss
+++ b/src/stylesheets/elements/_buttons.scss
@@ -200,7 +200,6 @@ $button-stroke: inset 0 0 0 units($theme-button-stroke-width);
 .usa-button--outline-inverse:disabled,
 .usa-button--outline-inverse:disabled {
   background-color: color("transparent");
-  pointer-events: none;
 
   &:hover,
   &.usa-button--hover,


mejiaj:
I agree, it gives developers a false sense of confidence. But I think we should only remove this if we have JS in place to ensure they're truly disabled.

And/or we could also update the [button guidance](https://designsystem.digital.gov/components/button/) to recommend using `disabled` and note that the class is purely for styling purposes.

thisisdano (approved):

PR #4140, opened 2021-04-01
USWDS - LIBRARY |  Consolidate gulp tasks

scottqueen-bixal:
## Description

This works to improve our gulp tasks within this library beta by doing the following actions:

1. Migrate all task partials into one directory
  - all gulp tasks are now located in the `gulp-task` directory
  - their utils are in `gulp-task/utils`
2. Compose all tasks to the recommended gulp v4 format
  - this mostly requires exporting them and importing them into the `gulpfile.js`
  - some tasks are exposed globally, while others are localized in series, parallel, and watch tasks
3. Deduplicate any tasks that were cloned during the early stages of the library beta development
  - the main culprit here was `compile` which seemed to take from `build`, we don't need both - `compile` was removed

---


gulp-tasks/release.js
```diff
@@ -77,24 +78,25 @@ gulp.task("zip-archives", (done) => {
   zip.on("close", (code) => {
     if (code === 0) {
       createHash(`dist/${dutil.dirName}.zip`);
+      const stats = fs.statSync(`./dist/${dutil.dirName}.zip`);
```
scottqueen-bixal, line 80: including this here to output zip size, we could port this over into production sooner if we think it is of value

gulpfile.js
```diff
@@ -148,24 +114,82 @@ function watchFiles() {
...
+  lintSass,
+  sassTests,
+  unitTests,
+  serve,
+  a11y,
+  exitServer
+);
+
+// building
+exports.buildSprite = buildSprite;
```
mejiaj, line 169: Runs without errors, but I don't see a sprite.svg file generated.

```bash
➜ npx gulp buildSprite
[13:11:31] Using gulpfile ~/web/uswds/gulpfile.js
[13:11:31] Starting 'buildSprite'...
[13:11:31] Finished 'buildSprite' after 22 ms

```

gulpfile.js
```diff
@@ -1,36 +1,33 @@
-// Old gulp tasks
-// Patch these in until ported into new syntax
-// typecheck
-require("./config/gulp/javascript");
-
 // Todo: convert release tasks.
```
mejiaj, line 1: We don't need this anymore.

gulpfile.js
```diff
@@ -148,24 +114,82 @@ function watchFiles() {
...
+
+// flags
+exports.noTest = noTest;
+exports.noCleanup = noCleanup;
+
+// Clean all directories.
+exports.clean = parallel(cleanCSS, cleanFonts, cleanImages, cleanJS, cleanSass);
+
+// Lint Sass
+exports.lintSass = parallel(lintSass);
```
mejiaj, line 130: Do we need `parallel` for a single task?

```bash
uswds on  gsq-library-consolidate-gulp-tasks [$!] via ⬢ v14.16.0
❯ npx gulp lintSass
[13:43:17] Using gulpfile ~/web/uswds/gulpfile.js
[13:43:17] Starting 'lintSass'...
[13:43:17] Starting 'lintSass'...
[13:43:24] Finished 'lintSass' after 7.41 s
[13:43:24] Finished 'lintSass' after 7.41 s
```

gulpfile.js
```diff
@@ -148,24 +114,82 @@ function watchFiles() {
...
+
+// Lint JavaScript
+exports.lintJS = parallel(lintJS);
```
mejiaj, line 133: Same here with parallel.

mejiaj (changes requested):
A lot of great work here, thank you. Added some notes.

We should discuss how we can clean up `gulpfile.js`. Right now there's too much going on.

.nycrc
```diff
@@ -0,0 +1,8 @@
+{
+  "all": true,
+  "include": ["src/patterns/components/**/*.js"],
+  "exclude": [
+    "**/*.spec.js",
+    "**/test/**"
+  ]
+}
```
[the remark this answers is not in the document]
mejiaj: Nevermind, cover test seems to require it.

gulp-tasks/build.js
```diff
@@ -0,0 +1,29 @@
...
+    dutil.logIntroduction();
+    dutil.logMessage("build", "Creating distribution directories.");
+    done();
+  },
+  cleanDist,
+  buildSprite,
+  copyDocs,
+  parallel(compileSass, compileJS),
+  parallel(copyImages, copyFonts, copySass)
+);
```
mejiaj, line 29: Build task is separated out in a way that makes more sense. Added comments as well.

**Note** there isn't a `gulp build` task it's just the default when you run `gulp`.

gulp-tasks/clean.js
```diff
@@ -1,33 +1,22 @@
...
+    dutil.logMessage(
+      "clean-dist",
+      "Skipping cleaning up the distribution directories."
+    );
+    return done();
   }
+  dutil.logMessage("clean-dist", "Removing distribution directories.");
+
+  return del("dist");
 };
```
mejiaj, line 22: The previous `cleanCSS, cleanFonts, cleanImages, cleanJS` weren't used anywhere. The only real clean task existed elsewhere; which I moved here.

gulp-tasks/copy.js
```diff
@@ -1,31 +1,52 @@
...
       .pipe(dest('dist/scss'));
   },
 
+  /**
+   * Todo: remove once issue below is resolved
+   * Patternlab doesn't move CSS from `dist/` to `build/`.
+   * Alternative is run `rebuildPL` task, but that takes 11s on average
+   * https://github.com/pattern-lab/patternlab-node/issues/1310
+   */
+  copyCSSToPL() {
```
mejiaj, line 42: Created `copyCSSToPL()` because of a pattern lab issue with compiles. This just copies the compiled `dist/` directory and moves it into `build`.  It gets called in `watch.js`.

gulp-tasks/lint.js
```diff
@@ -1,35 +1,67 @@
...
+        formatter: ignoreStylelintIgnoreWarnings,
+        console: true,
+      },
+    ],
+    syntax: "scss",
+  };
+
+  return src(
+    "./src/patterns/stylesheets/**/*.scss",
+    { since: lastRun(lintSass) }
```
mejiaj, line 42: Had to move `lintSass()` to it's own function to be able to use [lastRun](https://gulpjs.com/docs/en/api/lastrun/). Reduces wait times for linting from an average of `11s → 3s`

gulp-tasks/serve.js
```diff
@@ -0,0 +1,41 @@
...
+  serve(done) {
+    // See https://browsersync.io/docs/options for more options.
+    server.init(serverOptions);
+    done();
+  },
+  server,
+  buildPL,
+  rebuildPL,
+  exitServer
+}
```
mejiaj, line 41: These tasks were previously in the gulpfile. Moved them to their own file for consistency.

gulp-tasks/svg-sprite.js
```diff
@@ -27,28 +28,28 @@ const config = {
...
 
-gulp.task(
-  "svg-sprite",
-  gulp.series("build-sprite", "rename-sprite", "clean-sprite")
-);
+exports.buildSprite = series(
+  buildSprite,
+  renameSprite,
+  cleanSprite
+)
```
mejiaj, line 55: None of these tasks are run on their own so I've combined them in `exports.buildSprite` and exported it.

gulp-tasks/watch.js
```diff
@@ -0,0 +1,70 @@
...
+      (done) => done())
+  );
+}
+
+exports.watch = series(
+  build,
+  buildPL,
+  serve,
+  watchFiles
+);
```
mejiaj, line 70: These were also previously in the gulpfile.js and I've moved them for consistency.

gulpfile.js
```diff
@@ -110,62 +60,26 @@ exports.test = series(
...
 
 // Default Task
-exports.default = series(
-  parallel(cleanCSS, cleanFonts, cleanImages, cleanJS, cleanSass),
-  parallel(copyVendor),
-  parallel(lintSass, compileSass, lintJS, compileJS, compileSprite),
-  parallel(copyFonts, copyImages, copySass, copyStyleguide),
-  buildPL
-);
+exports.default = build;
```
mejiaj, line 85: This is where we should define the tasks are actually useful to us.

package.json
```diff
@@ -143,7 +139,6 @@
     "postcss-csso": "^5.0.0",
     "postcss-discard-comments": "^4.0.2",
     "prettier": "^2.2.0",
-    "resemblejs": "^3.2.5",
```
mejiaj, line 146: Was a remnant of visual regression testing.

package.json
```diff
@@ -32,21 +32,18 @@
...
     "preversion": "npm test",
     "release": "gulp release",
     "federalist": "npm install && npx gulp",
     "start": "gulp watch",
-    "test": "snyk test && gulp typecheck && gulp test",
-    "test:ci": "gulp typecheck && gulp test",
+    "test": "snyk test && gulp test",
+    "test:ci": "gulp test",
     "test:a11y": "gulp a11y",
-    "test:ci:temp": "gulp typecheck && gulp lintJS && gulp lintSass",
```
mejiaj, line 49: Seemed like a duplicate so I removed this test.

src/patterns/utils/test/a11y.js
```diff
@@ -21,6 +21,7 @@ function launchChrome() {
     .then((newChrome) => {
       chrome = newChrome;
       chromeHost = chrome.host;
+      // Todo: make this dynamic
       serverUrl = "http://localhost:3333/";
```
mejiaj, line 25: Fractal used to have this method we could use to handle this URL dynamically in our tests. Patternlab doesn't have one.

scottqueen-bixal: There are not any good options that I came accross in the browsersync API to get the current running port.

I was able to find util called portscanner, but the best we can get there is to return the first active port. We would need some way to confirm that the node server running at the returned port is our Browsersync instance.

There are some HTTP methods we can call like `reload` that would give us a 200 response if the request is made to a browsersync server.  

something like

```
http.request(
          `http://${server.instance.utils.devIp()}:${port}/__browser_sync__?method=reload`,
          (res) => ... 
```

This could work, but we would have trouble knowing if this was our specific instance of multiple browsersync servers running.

tbh this magic number might be the best approach.

gulpfile.js
```diff
@@ -1,105 +1,55 @@
...
-  server.init({
-    server: ["./build/"],
-    notify: false,
-    open: false,
-    port: 3333,
-  });
-  done();
-}
-
 exports.a11y = series(serve, a11y, exitServer);
```
scottqueen-bixal, line 47: this spins up a separate instance of browsersync server - if we already have one running - we probably don't need to spin up a new one if w have one

PR #4155, opened 2021-04-15
USWDS | HEADER - Menu Toggle Background

gscottqueen:
## Description
The overlay for the mobile header was darkened to match the Modal overlay, but they're set up differently. So we tweak the value in `_header.scss`  to match.

## Additional information
Screen Shot 2021-04-15 at 5 35 34 AMBefore you hit Submit, make sure you’ve done whichever of these applies to you:


src/stylesheets/components/_header.scss
```diff
@@ -125,7 +125,7 @@ $z-index-overlay: 400;
   @include u-pin("all");
   position: fixed;
   background: color("black");
```
mejiaj, line 127:
```suggestion
  background: color("black-transparent-70");
```

Let's use the same background color as modal overlay.

src/stylesheets/components/_header.scss
```diff
@@ -125,7 +125,7 @@ $z-index-overlay: 400;
...
-  opacity: opacity(0);
+  opacity: opacity(70);
```
mejiaj, line 128:
```suggestion
  opacity: opacity(0);
```

Let's revert this change because our background color is semi-transparent.

Which means we'll have to update `line:134` below.

```scss
&.is-visible {
  opacity: opacity(100);
  visibility: visible;
}
```

mejiaj (changes requested):
Added suggestions to make it more like Modal overlay.

There's also this line in [`navigation.js:38`](https://github.com/uswds/uswds/blob/gsq-lightens-background-menu-toggle/src/js/components/navigation.js#L38) that toggles _all_ menus and overlays. Might be a rare case, but do you think it's worth updating this selector so it **only** toggles it's own menu + overlay and not all of them?

src/stylesheets/components/_header.scss
scottqueen-bixal: updated

src/stylesheets/components/_header.scss
scottqueen-bixal: updated

scottqueen-bixal:
> There's also this line in [`navigation.js:38`](https://github.com/uswds/uswds/blob/gsq-lightens-background-menu-toggle/src/js/components/navigation.js#L38) that toggles _all_ menus and overlays. Might be a rare case, but do you think it's worth updating this selector so it **only** toggles its own menu + overlay and not all of them?

The code is working as expected. This is an edge case where most folks won't have multiple navigation headers/toggles that would create this multiple overlaying of the `overlays`.  

We could as an option traverse the dom from the target (currently `button`), and get specific for the `NAV` selector toggle. but going after the `overlay` as a sibling might not be what we want.  

example:

```js
const firstOverlay = body.querySelectorAll(OVERLAY)[0];
const activeNAV = active.target.closest(".usa-header").querySelector(NAV);
```

styles are good to go, we can discuss a solution for the specificity needs on the toggle if needed.

mejiaj (approved):

mejiaj:
> > There's also this line in [`navigation.js:38`](https://github.com/uswds/uswds/blob/gsq-lightens-background-menu-toggle/src/js/components/navigation.js#L38) that toggles _all_ menus and overlays. Might be a rare case, but do you think it's worth updating this selector so it **only** toggles its own menu + overlay and not all of them?
> 
> The code is working as expected. This is an edge case where most folks won't have multiple navigation headers/toggles that would create this multiple overlaying of the `overlays`.
> 

I agree, multiple nav's and overlays seems like an edge case. We could even handle this in the template.

PR #4178, opened 2021-04-30
Radio / Checkbox: Add support for themeable tile input background color

aduth:
## Description

Allows for a theme to customize the default background color of the tile [radio button](https://designsystem.digital.gov/components/radio-buttons/) or [checkbox](https://designsystem.digital.gov/components/checkbox/) components.

## Additional information

As implemented, this is entirely backward compatible and preserves the existing behavior of assigning a transparent background color, while allowing downstream projects to override this to customize the default background color of the tile inputs.

This improves consistency with selected input background color, which is already customizable. It also better supports assignment of background color in cases of e.g. desiring a white tile input background when within a container with a colored background, rather than default transparent.

For example, adding a background color (`bg-secondary-lighter` in this example) to the [component preview](https://designsystem.digital.gov/components/radio-buttons/) reveals that the tile inputs can be hard to distinguish:

image

With this change, one could override the background to improve the legibility. In this example with `'white'`:

image

An alternative or complementary improvement could be to change the default to always apply a white background color to tile inputs if it's naturally assumed to be shown with a white background (in current examples leveraging the container's lack of background color).

---

### DIFF
--- a/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
+++ b/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
@@ -127,6 +127,7 @@
 // Checkboxes and radios with tap-friendly targets
 .usa-checkbox__input--tile + .usa-checkbox__label,
 .usa-radio__input--tile + .usa-radio__label {
+  background-color: color($theme-input-tile-background-color);
   border: units($theme-input-tile-border-width) solid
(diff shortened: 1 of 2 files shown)


mejiaj:
Thanks for this! Closing it in favor of #4199. Pretty much the same except I'm using `set-text-and-bg()` to set an accessible color on the background.

Issue #4098, opened 2021-03-17
Add success and error state color utilities


maya:
**Is your feature request related to a problem? Please describe.**
There are no utilities for state colors, specifically: error and success. It's needed for custom UIs that need states, form iconography, etc.

**Describe the solution you'd like**
Add success and error state color tokens for text and background color:

- `.text-success`
- `.text-error`
- `.bg-success`
- `.bg-error`

If you prefer to make these a setting to turn these on, that's fine too. e.g. looks like there are emergency color utilities now but it's not something I'd likely use (it's so specific) and would be fine if it was off.

**Describe alternatives you've considered**

- I've added these manually to multiple projects in my custom Sass
- Using red and green color tokens but the colors are different from states

**Additional context**

- There are a lot of state colors, but I think error and success is good enough
- Note: Add documentation that the error state color could be alarmist and only use when intended.

PR #4182, opened 2021-05-01
Add basic state palette to color utilities

maya:
Trying to address #4098, but I don't think I know enough about how to add in color utilities. Maybe someone can help get this working? Or if there's an easier way to do this in my theme, let me know.


mejiaj:
Closing in favor of #4198.

PR #4183, opened 2021-05-01
Update `success-dark` to vivid variant

maya:
Needed for light text on a `success` background, for example: status tags. The current one is drab and doesn't look like a state color and this is more similar to the `success` token.

[Brought up in Slack in 2019](https://gsa-tts.slack.com/archives/C3F14AHSQ/p1576013330201600?thread_ts=1576012784.201200&cid=C3F14AHSQ)


mejiaj (changes requested):
Value also needs to be updated on `src/stylesheets/theme/_uswds-theme-color.scss`.

maya:
@mejiaj both changes done in 6186fcc.

mejiaj (approved):

thisisdano (approved):

PR #4187, opened 2021-05-03
USWDS - Buttons: Added button type to non-form buttons

### LINKED ISSUE #3563: Buttons which do not submit a form should be assigned type
## Description

Most (all?) component examples which incorporate `button` element markup do not currently assign a `type` attribute.

Example (Accordion):

https://github.com/uswds/uswds/blob/f7acaf7e09840f4f84b1398c4d30a977cd7fc1e5/src/components/06-accordion/accordion.njk#L7-L9

In absence of an assigned `type`, the default value is `submit`, which inherits the semantics of form submission.

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type

Arguably, many of these buttons are not intended to be interpreted as submit buttons. Instead, the use-case of `type="button"` applies instead.

From the above reference:

>`button`: The button has no default behavior, and does nothing when pressed by default. It can have client-side scripts listen to the element's events, which are triggered when the events occur.

The default behavior of submit buttons would cause their placement within a `form` element to trigger the form submission upon click. In the case of USWDS, there is no known user-facing issues by omitting `type` due to the fact that the default behavior of buttons is prevented.

Example (Accordion):

https://github.com/uswds/uswds/blob/f7acaf7e09840f4f84b1398c4d30a977cd7fc1e5/src/js/components/accordion.js#L70-L72

However:

- Semantically, any technology which crawls the markup may infer these as submit buttons. It's not clear to me if there are specific examples of this.
- `event.preventDefault()` would not be needed if these were assigned `type="type"`, which could serve as a simplification

mahoneycm:
## Description
 
Fixes #3563. Adds `type="button` to all non-form button elements 
 
## Additional information

Example (Accordion):

**BEFORE**

```
 <button class="usa-accordion__button" 
   aria-expanded="{{ item.expanded | default('false') }}" 
   aria-controls="{{ accordion.id_prefix }}{{ item.id }}"> 

```

**AFTER**

```
<button class="usa-accordion__button"
          type="button"
          aria-expanded="{{ item.expanded | default('false') }}"
          aria-controls="{{ accordion.id_prefix }}{{ item.id }}">
```


src/components/card/card--standard.njk
```diff
@@ -8,7 +8,7 @@
         <p>Lorem ipsum dolor sit amet consectetur adipisicing elit. Facilis earum tenetur quo cupiditate, eaque qui officia recusandae. Excepturi nobis dolores molestias! Quas quisquam a officia eos.</p>
       </div>
       <div class="usa-card__footer">
-        <button class="usa-button">Visit Florida Keys</button>
+        <button class="usa-button" type="button">Visit Florida Keys</button>
       </div>
     </div>
   </li>
```
mejiaj, line 14: Missing type on line 47, 66, and 85.

mejiaj (changes requested):
Thanks for helping out with this, noticed 2 templates with missing `type`. One is marked in review and the other is in  `src/components/test/sidenav-button.njk:38`.

mejiaj (approved):

mejiaj:
Based on guidance in original issue we could remove the `e.preventDefault()` line.
https://github.com/uswds/uswds/blob/f7acaf7e09840f4f84b1398c4d30a977cd7fc1e5/src/js/components/accordion.js#L70-L72

We also need to decide if we're going to add `type` to buttons and button group NJK templates.

amyleadem:
Closing in favor of PR https://github.com/uswds/uswds/pull/4695, which uses current 3.0 component structure.

PR #4196, opened 2021-05-13
[4194] Adds component JS files to package.json exports

### LINKED ISSUE #4194: Unable to import component-specific JS dependencies
### Describe the bug

Although undocumented, prior to Webpack 5, a project could import an individual component's JS using an `import` statement like:

```js
import tooltip from "uswds/src/js/components/tooltip";

tooltip.on();
```

When running Webpack 5, the following error now occurs:

```
Module not found: Error: Package path ./src/js/components/tooltip is not exported from 
package /node_modules/uswds (see exports field in /node_modules/uswds/package.json)
```

This appears to be caused by the `exports` field in the USWDS `package.json`. [More details of how Webpack uses this field is here](https://webpack.js.org/guides/package-exports/). Webpack [began supporting](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#major-changes-new-nodejs-ecosystem-features) this field in version 5.

It's important to be able to selectively import only the JS a project needs, rather than importing all of the USWDS JS. This helps reduce the JS bundle size. For instance, if the only JS-dependent component used by a project is the tooltip, they should have a clear way to only import the Tooltip component's JS.

1. Clone [this repo](https://github.com/sawyerh/uswds-webpack-example)
2. Follow the readme instructions

### System setup
- **USWDS version:** 2.11.2
- Webpack: 5.37.0

### Additional context

- [A similar issue was reported in Slack](https://gsa-tts.slack.com/archives/C3F14AHSQ/p1619454492072000).
- An alternative might be for USWDS to provide an index file that better supports tree-shaking, so that a project could do something like this:
   ```js
   import { tooltip } from "uswds"
   ```

kmonahan:
Resolves #4194 . Adds individual components to package.json exports so that they can be imported with Webpack 5.

e.g.
```
import accordion from 'uswds/src/js/components/accordion';
accordion.on(document.body);
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,12 @@
   ],
   "module": "./dist/js/uswds.min.js",
   "exports": {
-    "import": "./dist/js/uswds.min.js",
-    "default": "./dist/js/uswds.min.js"
+    ".": {
+      "import": "./dist/js/uswds.min.js",
+      "default": "./dist/js/uswds.min.js"
+    },
+    "./src/js/components": "./src/js/components/index.js",
+    "./src/js/components/*": "./src/js/components/*.js"
   },
   "types": "./index.d.ts",
   "main": "dist/js/uswds.min.js",


thisisdano (approved):

scottqueen-bixal:
good solve, closing in favor of https://github.com/uswds/uswds/pull/4206

PR #4200, opened 2021-05-19
USWDS - date-picker: Resolve chrome non-unique empty id warnings.

seanpclark:
## Description

The date picker component sets an id to "". 
Chrome gives a warning as it treats multiple empty ids as non-unique.

## Additional information

Visit a following page in Chrome withe more than 1 date picker, and see console warnings (ex. https://designsystem.digital.gov/components/date-range-picker/)

### DIFF
--- a/src/js/components/date-picker.js
+++ b/src/js/components/date-picker.js
@@ -877,7 +877,7 @@ const enhanceDatePicker = (el) => {
     "usa-sr-only",
     DATE_PICKER_INTERNAL_INPUT_CLASS
   );
-  internalInputEl.id = "";
+  internalInputEl.removeAttribute("id");
   internalInputEl.required = false;
 
   datePickerEl.appendChild(calendarWrapper);


mejiaj:
Thanks for submitting this! We'll take a look as soon as we can.

mejiaj (commented):
Confirming this change removes the empty ID. 

But here's some strange thing I noticed. There's an empty `name` attribute on the newly generated input.
image

```html
<input 
  class="usa-input usa-date-picker__external-input" 
  id="event-date-start" 
  name="" 
  type="text" 
  aria-describedby="event-date-start-label event-date-start-hint"
>
```

This doesn't seem practical because if we click the label it points to the hidden input, not the enhanced one.

That `name` value gets intentionally deleted on line:862
https://github.com/uswds/uswds/blob/976948a2d54868cbfb632050df94da4c1b68bb0e/src/js/components/date-picker.js#L861-L862

Also, notice how the `internalInputEl` gets cloned and then it's type gets set to text. The original is text, so this seems unnecessary.

mejiaj:
Closing in favor of #4218.

PR #4205, opened 2021-05-24
POAM | CVE-2020-7729

scottqueen-bixal:
# CVE-2020-7729

## Description

Updates package to resolve the alert.

## Additional information

We should reconsider the value of the examples directory here.  

Are folks using it? 
Could be outdated recommendations for node dependencies here? (ie, node-sass)

We might consider removing the directory and revisit the need for including an updated "recommended approach" dir in future `library` updates.

### DIFF
--- a/examples/grunt/package.json
+++ b/examples/grunt/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "dependencies": {
-    "uswds": "^2.12.0"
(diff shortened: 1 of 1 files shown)


scottqueen-bixal:
closing in favor of https://github.com/uswds/uswds/pull/4216

PR #4209, opened 2021-05-26
Update Sass slash as division

### LINKED ISSUE #4204: Dart Sass deprecation warnings: slash as division
We're [trying to move to Dart Sass](https://github.com/CDCgov/prime-simplereport/pull/1681), but getting warnings on `watch-css` and `compile-scss` about some deprecations in Dart Sass:

See: https://sass-lang.com/documentation/breaking-changes/slash-div

Related #4129 
/cc @scottqueen-bixal @mejiaj

maya:
Fixes #4204

### DIFF
--- a/src/stylesheets/components/_alerts.scss
+++ b/src/stylesheets/components/_alerts.scss
@@ -1,5 +1,7 @@
+@use "sass:math";
+
 $alert-slim-icon-size: units(3);
-$alert-icon-optical-factor: units($theme-alert-icon-size) / 6;
+$alert-icon-optical-factor: math.div(units($theme-alert-icon-size), 6);
 $alert-icon-optical-padding: units($theme-alert-padding-x) -
   $alert-icon-optical-factor;
 
--- a/src/stylesheets/components/_banner.scss
+++ b/src/stylesheets/components/_banner.scss
@@ -1,3 +1,5 @@
+@use "sass:math";
+
 $banner-context: "Banner";
 $banner-guidance-measure: 3;
 $banner-icon-colors: get-link-tokens-from-bg(
@@ -92,7 +94,7 @@ $banner-icon-close: (
 .usa-banner__lock-image {
   $lock-h: 64; // unitless height of svg
   $lock-w: 52; // unitless width of svg
-  $lock-aspect: $lock-w / $lock-h;
+  $lock-aspect: math.div($lock-w, $lock-h);
   $icon-height: 1.5ex; // height of the lock icon; use ex for resilience
 
   height: $icon-height;
--- a/src/stylesheets/components/_breadcrumb.scss
+++ b/src/stylesheets/components/_breadcrumb.scss
@@ -1,3 +1,5 @@
+@use "sass:math";
+
 //
 // General Settings
 // =====================================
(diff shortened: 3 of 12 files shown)


thisisdano:
ooh, thanks. we were just about to get to this

thisisdano:
If we do this, it will be a breaking change for anyone who does something like compiles with the built-in Jekyll compiler. I guess we've known this day was coming for some time, but we're pushing against the requirements of modern Sass and the limitations of the Sass ecosystem

maya:
Jekyll 4.0 and up uses `sass`. Assuming you'll want to do some testing/research.

maya:
Also, could consider publishing as a release candidate or beta release and asking folks to test it 🔬 

What about offering the opposite of https://sass-lang.com/documentation/breaking-changes/slash-div#automatic-migration if it causes any issues

thisisdano:
@maya I found that trying to use modern Sass syntax with Jekyll 4.2.0 resulted in a compile error. Have you successfully used module syntax with native Jekyll?

maya:
It looks like Jekyll uses LibSass in their Jekyll-sass-converter plugin, so it would not work with their out-of-box Sass processing. Is that what you're using for your Jekyll projects?

thisisdano:
It's not what we're using, but it is somewhat common

maya:
@thisisdano Another developer I chatted with gave some good advice:

> we’re in this situation and have opted to have a period of time where we’re offering both options before introducing the breaking change to cut over to `@use` entirely. So - for our core package, people can continue bringing `core.scss` into their project or use the new module version `core.mod.scss`. This does introduce duplication of partials, but since our styles are pretty established at this point, I’m not worried about maintaining it for the deprecation period

there are some similarities with where USWDS is at: 
1. USWDS styles are quite established
2. minimal component growth on the roadmap

You may want to start socializing folks move over to dart sooner than later. 

If folks want to stay in the Ruby ecosystem, I found https://www.npmjs.com/package/dart-sass-rails

You could also tell users of deprecated Sass implementations to pin their versions to avoid breaking changes if they don't want to change.

aduth:
There's an interesting cross-compatible solution mentioned at https://github.com/sass/sass/issues/2565#issuecomment-866326336. It's not perfect, since it would leave `@use` statements in the generated CSS for non-Dart SASS, but as mentioned in the comment, it shouldn't break anything to have in the output.

I tried playing with some ideas to detect if the current compiler is Dart, and while it does seem possible to detect, there's still no obvious way to bring in Dart's math methods, since conditional imports aren't supported.

```scss
@if function-exists("module-functions") {
  // Dart SASS
} @else {
  // Not Dart SASS
}
```

aduth:
This was originally superseded by #4314, but if USWDS v3 is aiming to target Sass modules, it would be a good idea to revisit the native Sass math function, as I would assume it's going to perform much better than the polyfilled version of #4314.

@mejiaj, will this be incorporated in #4539 ?

mejiaj:
@aduth, yes we'll be using the native `sass.math()` in beta.

mejiaj:
Closing this since we're using SASS modules in [USWDS 3](https://github.com/uswds/uswds/releases/tag/v3.0.0). 

Thanks @maya for the contribution and nudge to get on this!

PR #4233, opened 2021-06-16
Optimize initializer script

aduth:
## Description

Optimizes initializer script load and runtime performance, reducing size by 55%, from 869 bytes (520 bytes gzipped) to 362 bytes (231 bytes gzipped), and avoiding delays in removing the document loading class.

## Additional information

The intended optimizations here are to:

- Avoid Browserify for the initializer script. Browserify, as with most bundlers, has a static cost in the form of a runtime loader for resolving modules internally. You can see this on the [first line of the unminified script](https://unpkg.com/uswds@2.11.2/dist/js/uswds-init.js). While some bundlers do better at this than others, since it's intended to be a small standalone script, it doesn't seem necessary to use a bundler at all.
- Use events to verify USWDS script load instead of timers. Since we can assume that the global `uswdsPresent` will be assigned as a result of loading the USWDS script, we can monitor for `load` events on page assets (images, scripts, styles, etc) to respond to this immediately. This avoids any delays that could happen if the script finishes loading between the 100ms polling that exists today, and may be less intensive overall especially for slow page loads, where the code only runs for each asset load and not every 100ms regardless.

A few additional ideas were explored for more size savings, but not included here. Since `revertClass` is safely idempotent, and since most `load` events would happen only at the start of a page session, we could also consider to:

- Avoid clearing timeout for fallback class revert after verified loaded: -22 bytes (-13 bytes gzipped)
   - `revertClass` would be a noop after the fallback delay if the class had already been removed previously
- Avoid removing load event listener after verified loaded: -46 bytes (-16 bytes gzipped)
   - This would cause `revertClass` to be called for every `load` event after `uswdsPresent` is assigned, but similar to the above note, each of these would be a noop anyways.

---


src/js/uswds-init.js
```diff
@@ -1,35 +1,18 @@
...
-}
-
-// Based on https://www.filamentgroup.com/lab/enhancing-optimistically.html
-if ("querySelector" in window.document && "addEventListener" in window) {
-  addLoadingClass();
+(() => {
+  document.documentElement.classList.add("usa-js-loading");
+  function revertClass() {
+    document.documentElement.classList.remove("usa-js-loading");
+  }
```
aduth, line 5: A few of the changes aren't strictly "best practices", but are implemented to take advantage of how gzip works to [optimize repeated strings](https://en.wikipedia.org/wiki/Deflate#Duplicate_string_elimination).

This alternative might be more conventional, but would increase size ever-so-slightly (+4 bytes gzipped):

```suggestion
  const loadingClass = "usa-js-loading";
  document.documentElement.classList.add(loadingClass);
  function revertClass() {
    document.documentElement.classList.remove(loadingClass);
  }
```

mejiaj (commented):
Great work, thanks for this!

Added some comments and will let the team chime in with their thoughts.

config/gulp/javascript.js
```diff
@@ -9,38 +9,30 @@ const rename = require("gulp-rename");
 const source = require("vinyl-source-stream");
 const sourcemaps = require("gulp-sourcemaps");
 const uglify = require("gulp-uglify");
+const babel = require("gulp-babel");
```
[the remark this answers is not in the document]
aduth: > Some concerns about including `gulp-babel` since we're trying to get away from dependencies that don't update regularly.
> 
> Looking at their github, the last release was a year ago.

`gulp-babel` doesn't really do a whole lot, since it delegates most of the work to the underlying `@babel/core` peer dependency. The way I see it, a main risk for it becoming outdated would be if it keeps up with a future 8.0.0 release ([since it's pinned to `^7.0.0`](https://github.com/babel/gulp-babel/blob/46db7eefad83aee00de102128ba2745b687d3136/package.json#L47)).

Since gulp works with plain Node streams, it could be substituted with a stream. Babel doesn't come with a stream-based API, but we could emulate one with a little wrapper:

```diff
diff --git a/config/gulp/javascript.js b/config/gulp/javascript.js
index b549fc027..8edb6ce99 100644
--- a/config/gulp/javascript.js
+++ b/config/gulp/javascript.js
@@ -1,2 +1,3 @@
 /* eslint-disable arrow-body-style */
+const { Transform } = require("stream");
 const buffer = require("vinyl-buffer");
@@ -11,3 +12,3 @@ const sourcemaps = require("gulp-sourcemaps");
 const uglify = require("gulp-uglify");
-const babel = require("gulp-babel");
+const babel = require("@babel/core");
 const merge = require("merge-stream");
@@ -32,3 +33,17 @@ gulp.task(task, () => {
     stream
-      .pipe(babel({ presets: ["@babel/preset-env"] }))
+      .pipe(
+        new Transform({
+          objectMode: true,
+          async transform(file, _encoding, callback) {
+            const { code } = await babel.transformAsync(
+              file.contents.toString(),
+              { presets: ["@babel/preset-env"] }
+            );
+            // eslint-disable-next-line no-param-reassign
+            file.contents = Buffer.from(code);
+            this.push(file);
+            callback();
+          },
+        })
+      )
       .pipe(rename({ basename }))
```

This is basically a reimplementation of what `gulp-babel` does anyways.

Another option: We could write the `uswds-init.js` in ES5-safe JavaScript, then skip Babel for that file altogether, then revert back to `babelify` for the remaining entrypoint.

src/js/uswds-init.js
aduth: > I'd lean towards the more convential style, but no strong opinions.

While it's outside the scope of this pull request, one thing I notice is that if ever the project migrates from Uglify to [Terser](https://github.com/terser/terser), Terser seems to compress the modified version to inline those strings anyways, so we'd get the best of both worlds: Smallest built file size, while keeping best practices in the source.

In the meantime, the few extra bytes aren't make or break, so I'll just apply the revision 👍

PR #4256, opened 2021-07-08
Date Picker / Combo Box: Use mouseover event instead of mousemove

### LINKED ISSUE #4184: Uncaught TypeError: t.target.closest is not a function. In the header
### Describe the bug
Mouse rollover causes Uncaught TypeError.  Nav header dropdown will through the error when mousing over the dropdown and mousing out

1. Go to 'https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/release-2.11.2/components/preview/header--basic-mega.html'
2. Mouse over the dropdown several times to trigger the error
4. See Uncaught TypeError:

### Expected behavior
That no error will occur when mousing over the dropdown nav

### Screenshots
Add screenshots to help provide context, if applicable.

### System setup
- **USWDS version: 2.11.2**
- **Device: PC**
- **Operating system: Windows 10 **
- **Browser and version: Firefox Version 87.0**

### Additional context
Add any other context about the problem.

aduth:
## Description

This pull request seeks to improve performance of USWDS JavaScript by limiting the amount of event handling that occurs in response to mouse events.

## Additional information

USWDS component scripts typically use event delegation for handling UI events, meaning that the actual event listener is attached to the document, and not to individual elements. A side-effect of this is that events are bound regardless of the presence of any particular components in the page.

The Date Picker and Combo Box components each bind events for [`mousemove`](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousemove_event) in order to apply a focus effect when the cursor is placed over items within the element. `mousemove` can be a very noisy event, since it is emitted for every pixel movement ([see demo](https://www.w3schools.com/jquery/tryit.asp?filename=tryjquery_event_mouseenter_mouseover)). This can be problematic in combination with event delegation, where every movement of the cursor anywhere in the body is handled multiple times for each of the components, regardless whether any instances of that component exist in the page.

Truthfully, I'm not sure I understand the goal in programmatically focusing the component elements in response to cursor movement, but the changes here don't aim to change the behavior, but instead serve as a refactoring of the existing implementation.

Future improvements could consider...

* Not using JavaScript to respond to mouse events at all, instead leaning primarily on CSS styling (i.e. `:hover`)
* Apply event handlers at the component root instead of the document, so that the events are limited to only when instances of the component are present on the page, and only when cursor events occur inside the element.

This may indirectly fix #4184, based on debugging at https://github.com/uswds/uswds/issues/4184#issuecomment-832201956 and https://github.com/uswds/uswds/issues/4184#issuecomment-872394018.

---

### DIFF
--- a/spec/unit/date-picker/date-picker-month-selection.spec.js
+++ b/spec/unit/date-picker/date-picker-month-selection.spec.js
@@ -137,7 +137,7 @@ describe("date picker component month selection", () => {
   });
 
   it("should update the focus when moving over a non-selected valid month", () => {
-    EVENTS.mousemove(
+    EVENTS.mouseover(
       getCalendarEl().querySelector(
         '.usa-date-picker__calendar__month[data-label="October"]'
       )
--- a/spec/unit/date-picker/date-picker-mousemove.spec.js
+++ b/spec/unit/date-picker/date-picker-mousemove.spec.js
@@ -41,7 +41,7 @@ describe("date picker component mouse move selection", () => {
       "focuses correct date"
     );
 
-    EVENTS.mousemove(
+    EVENTS.mouseover(
       getCalendarEl().querySelector(
         '.usa-date-picker__calendar__date[data-day="26"]'
       )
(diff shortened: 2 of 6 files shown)


mejiaj (approved):
Thanks for submitting this.

I agree with you; in the near future we shouldn't depend on JS for styling things that `:hover` should cover. Event handlers on component instead of document is something we want to work towards to make our components more modular.

thisisdano (approved):

PR #4258, opened 2021-07-09
Library beta 1 finalize

thisisdano:


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
 references:
   container: &container
     docker:
-      - image: circleci/node:12.18.4-browsers
+      - image: circleci/node:14.16.0-browsers
```
thisisdano, line 9: Use node version that matches our `.nvm`

package.json
```diff
@@ -1,9 +1,10 @@
 {
   "name": "uswds",
-  "version": "2.11.2",
+  "version": "2.14.0-beta.1",
```
thisisdano, line 3: Used `2.14.0-beta.1` here because it wouldn't pass with something like `library-beta.1` — it's just as accurate as `2.10.0-beta.0` I guess

package.json
```diff
@@ -130,7 +130,7 @@
     "handlebars-helpers": "^0.10.0",
     "jsdom": "^16.6.0",
     "jsdom-global": "^3.0.2",
-    "mocha": "^8.4.0",
+    "mocha": "^6.2.0",
```
thisisdano, line 133: Downgraded this because it was giving me `npm install` errors with the most current version on `npm` v7 — now that I've downgraded to `npm` v6 this may not be necessary, but I've left it

mejiaj (approved):

PR #4269, opened 2021-07-16
Modal: Fix modal aria-hidden for non-modal content after script load

aduth:
## Description

Accounts for cases where page content changes after initial script load, including:

- If USWDS script is loaded early in page content, all server-rendered content to follow will not be counted as non-modal content ([demo](https://codepen.io/aduth/pen/PommWEK))
- If any JavaScript alters the page after load, this will not be counted as non-modal content ([demo](https://codepen.io/aduth/pen/GRmJaZe))

## Additional information

Because `nonModals` was assigned only once at script load, it would not include anything from the above scenarios.

Notably, this means that users navigating page content using assistive technology can often escape the modal, which is not the intended behavior.

The newly-added test cases will fail on the `develop` branch, because this issue is also present in the tests.

---

### DIFF
--- a/src/js/components/modal.js
+++ b/src/js/components/modal.js
@@ -12,20 +12,19 @@ const WRAPPER_CLASSNAME = `${MODAL_CLASSNAME}-wrapper`;
 const OPENER_ATTRIBUTE = "data-open-modal";
 const CLOSER_ATTRIBUTE = "data-close-modal";
 const FORCE_ACTION_ATTRIBUTE = "data-force-action";
+const NON_MODAL_HIDDEN_ATTRIBUTE = `data-modal-hidden`;
 const MODAL = `.${MODAL_CLASSNAME}`;
 const INITIAL_FOCUS = `.${WRAPPER_CLASSNAME} *[data-focus]`;
 const CLOSE_BUTTON = `${WRAPPER_CLASSNAME} *[${CLOSER_ATTRIBUTE}]`;
 const OPENERS = `*[${OPENER_ATTRIBUTE}][aria-controls]`;
 const CLOSERS = `${CLOSE_BUTTON}, .${OVERLAY_CLASSNAME}:not([${FORCE_ACTION_ATTRIBUTE}])`;
+const NON_MODALS = `body > *:not(.${WRAPPER_CLASSNAME}):not([aria-hidden])`;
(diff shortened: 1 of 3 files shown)


src/js/components/modal.js
```diff
@@ -12,21 +12,20 @@ const WRAPPER_CLASSNAME = `${MODAL_CLASSNAME}-wrapper`;
 const OPENER_ATTRIBUTE = "data-open-modal";
 const CLOSER_ATTRIBUTE = "data-close-modal";
 const FORCE_ACTION_ATTRIBUTE = "data-force-action";
+const NON_MODAL_HIDDEN_ATTRIBUTE = `data-modal-hidden`;
 const MODAL = `.${MODAL_CLASSNAME}`;
 const INITIAL_FOCUS = `.${WRAPPER_CLASSNAME} *[data-focus]`;
 const CLOSE_BUTTON = `${WRAPPER_CLASSNAME} *[${CLOSER_ATTRIBUTE}]`;
 const OPENERS = `*[${OPENER_ATTRIBUTE}][aria-controls]`;
 const CLOSERS = `${CLOSE_BUTTON}, .${OVERLAY_CLASSNAME}:not([${FORCE_ACTION_ATTRIBUTE}])`;
+const NON_MODALS = `body > *:not(.${WRAPPER_CLASSNAME}):not([aria-hidden])`;
```
mejiaj, line 21:
```suggestion
const NON_MODALS = `body > *:not(.${WRAPPER_CLASSNAME}):not([aria-hidden]):not(script)`;
```
Not that it matters, but for clarity we might want to exclude script tags too.

mejiaj (commented):
Minor comment about the `NON_MODALS` selector. Otherwise LGTM, thanks for submitting this!

src/js/components/modal.js
aduth: I don't disagree with the premise, but I might worry it opens a can of worms as far as what other elements we might want to exclude (`style`, `link`, etc). Is there a particular scenario which would be improved by excluding them?

mejiaj: Good point, we don't need it for functionality.

mejiaj (approved):

thisisdano (approved):

PR #4278, opened 2021-07-31
"Alt text" is bad alt text

mgifford:
## Have good examples for alt text

This is just an example, but good examples matter. Let's make sure that we have good alternative text in this.

`Alt text`

Is never appropriate.


mejiaj (approved):
Very true, I appreciate you keeping us honest and even more for submitting the fix.

thisisdano (approved):
Thanks for these

PR #4280, opened 2021-07-31
Let's minimize default images

mgifford:
## Lets improve performance

Several PNG & GIF files could be minimized to improve performance and reduce the CO2 impact of sites using the USWDS.

Most of these I just ran through TinyPNG. 

The more performance is the norm for libraries like the USWDS, the easier it will be for individual sites to find those places where they have added bloat to the site.


mejiaj (approved):
Thanks for submitting this!

thisisdano (approved):

thisisdano:
This should also be a wake-up call to do a serious tidy on these directories as we move forward, as many of these images are no longer in use or suitable for modern USWDS production sites.

PR #4282, opened 2021-08-02
[USWDS] Modal | return early if no modal

scottqueen-bixal:
## Description

~We seem to need a delay before handling the event, this gives us a bit of space and allows the test to successfully pass.~

setting an early return so we don't query null items

### DIFF
--- a/spec/unit/navigation/navigation.spec.js
+++ b/spec/unit/navigation/navigation.spec.js
@@ -135,8 +135,10 @@ describe("navigation toggle", () => {
 
   it("collapses dropdowns when the Escape key is hit", () => {
     accordionButton.click();
-    EVENTS.escape(accordionButton);
-    assert.strictEqual(accordionButton.getAttribute("aria-expanded"), "false");
+    setTimeout(() => {
+      EVENTS.escape(accordionButton);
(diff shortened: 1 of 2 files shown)


mejiaj (changes requested):
LGTM, can you please add a comment as to why we're adding this. In case Mocha adds a better way to prevent these types of conflicts in the future.

spec/unit/navigation/navigation.spec.js
```diff
@@ -135,8 +135,10 @@ describe("navigation toggle", () => {
 
   it("collapses dropdowns when the Escape key is hit", () => {
     accordionButton.click();
-    EVENTS.escape(accordionButton);
-    assert.strictEqual(accordionButton.getAttribute("aria-expanded"), "false");
+    setTimeout(() => {
```
aduth, line 138: Mocha won't wait for this timeout unless you add `done` as an argument and call it in the callback, which can cause some strange side effects in subsequent tests.

https://mochajs.org/#asynchronous-code

scottqueen-bixal: you're right  @aduth, will take another look

mejiaj (approved):

PR #4286, opened 2021-08-11
Checkbox / Radio Button: Improve HTML whitespace sensitivity

aduth:
## Description

This pull request seeks to maintain the existing visual appearance of the Checkbox and Radio Button components, while improving developer ergonomics regarding HTML whitespace sensitivity.

Context: https://github.com/18F/identity-idp/pull/5279#discussion_r686146591

## Additional information

Because the current implementation styles the icon pseudo-element and label text as adjacent inline elements, it suffers from a common issue with unintended whitespace between the elements (see [related blog post](https://css-tricks.com/fighting-the-space-between-inline-block-elements/)). For this reason, the code must be marked up very carefully to avoid additional whitespace within the label class (including newlines), as otherwise the text alignment between label and description is misplaced, seen in the screenshots below:

Before|After
---|---
localhost_3000_components_preview_checkbox--tile (1)|localhost_3000_components_preview_checkbox--tile

There are a few ways this could be addressed, including using Flexbox, though most of these alternative solutions would require backwards-incompatible changes to the component markup, such as the introduction of a new element to wrap the label text. A Flexbox approach can _almost_ work with the existing markup through a combination of `flex-wrap: wrap` and `flex-basis: 100%` on the description element, but unfortunately this implementation fails when the label text is especially long.

Positioning absolutely is mostly straight-forward, with one exception: Because it is no longer in line with the label text, the checkmark icon must be manually positioned to restore the appearance of vertical alignment. The proposed implementation here uses some extensive math to achieve this, tested in the demo samples with default font size and with an adjusted `$theme-body-font-size: "xl";`. Ideally there'd be a simpler solution to this, though I could not come up with one.

---

### DIFF
--- a/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
+++ b/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
@@ -176,7 +176,7 @@
       padding: units(1.5) units(2) units(1.5) units(5);
 
       &:before {
-        left: units(5 - $input-select-margin-right - $theme-input-select-size);
+        left: units(5) - units($input-select-margin-right) - units($theme-input-select-size);
       }
     }
   }
--- a/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
+++ b/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
@@ -174,6 +174,10 @@
       border-radius: radius($theme-input-tile-border-radius);
       margin-top: units(1);
       padding: units(1.5) units(2) units(1.5) units(5);
+
+      &:before {
+        left: units(5 - $input-select-margin-right - $theme-input-select-size);
+      }
     }
   }
 }
@@ -201,7 +205,6 @@
         background-image: none;
         background-color: color("white");
         content: url("#{$theme-image-path}/checkbox-check-print.svg");
-        text-indent: 0;
       }
     }
   }
@@ -215,18 +218,18 @@
   margin-top: units(1.5);
   padding-left: units($input-select-margin-right + $theme-input-select-size);
   position: relative;
-  text-indent: units(-$input-select-margin-right - $theme-input-select-size);
 
   &:before {
     content: " ";
-    display: inline-block;
-    left: units($theme-input-select-border-width);
-    line-height: units($theme-input-select-size);
-    margin-right: units($input-select-margin-right);
-    position: relative;
-    text-indent: 0;
-    vertical-align: middle\0; // Target IE 11 and below to vertically center inputs
-    white-space: pre;
+    display: block;
+    left: 0;
+    margin-left: units($theme-input-select-border-width);
+    margin-top: (
+      line-height($theme-form-font-family, $theme-input-line-height) *
+      fs($theme-form-font-family, $theme-body-font-size) -
+      units($theme-input-select-size)
+    ) / 2;
+    position: absolute;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
```diff
@@ -215,18 +218,18 @@
...
-    position: relative;
-    text-indent: 0;
-    vertical-align: middle\0; // Target IE 11 and below to vertically center inputs
-    white-space: pre;
+    display: block;
+    left: 0;
+    margin-left: units($theme-input-select-border-width);
+    margin-top: (
+      line-height($theme-form-font-family, $theme-input-line-height) *
+      font-size($theme-form-font-family, $theme-body-font-size) -
```
aduth, line 229: This may need a rethink, since it assumes font styles inherited from `.usa-fieldset`, which are likely not guaranteed.

https://github.com/uswds/uswds/blob/f3bdc643dac135ef51ae75e91cde96b0a4d8c7b3/src/stylesheets/elements/form-controls/_global.scss#L32-L35

aduth: This is a situation where it would be nice to have the [CSS4 `lh` unit](https://www.w3.org/TR/css-values-4/#font-relative-lengths), since essentially what we want is `margin-top: calc((1lh - #{units($theme-input-select-size)}) / 2);`.

A couple alternatives to consider:

- Maybe the checkbox should explicitly set font expectations, which might be desirable anyways. i.e. Add the `%block-input-general` placeholder to `usa-checkbox`.
- Another I'd hoped could work is to chop up the elements so that the natural height of the label is only the label text excluding the description, so that a standard vertical centering approach could work. I've got this close to working in a [CodePen prototype](https://codepen.io/aduth/pen/GRmzQaO), but positioning the description absolutely means that it won't "push down" content that occurs after the checkbox.

mejiaj: @aduth have you tried Flexbox? Seems similar to the [media object pattern [solved by flexbox - philipwalton.github.io]](https://philipwalton.github.io/solved-by-flexbox/demos/media-object/)

aduth: > * Maybe the checkbox should explicitly set font expectations, which might be desirable anyways. i.e. Add the `%block-input-general` placeholder to `usa-checkbox`.

I went this direction in 9e958ae, based on there being some precedent of this applying at the component level [[1]](https://github.com/uswds/uswds/blob/6218825ad5b5054057bc8d3dd0f7255296ddeaa9/src/stylesheets/elements/form-controls/_combo-box.scss#L21)[[2]](https://github.com/uswds/uswds/blob/6218825ad5b5054057bc8d3dd0f7255296ddeaa9/src/stylesheets/elements/form-controls/_dropdown.scss#L2)[[3]](https://github.com/uswds/uswds/blob/6218825ad5b5054057bc8d3dd0f7255296ddeaa9/src/stylesheets/elements/form-controls/_range-input.scss#L30)[[4]](https://github.com/uswds/uswds/blob/6218825ad5b5054057bc8d3dd0f7255296ddeaa9/src/stylesheets/elements/form-controls/_text-input.scss#L3), where presumably it's expected that the checkbox and radio could (or should) inherit from the input-related typography settings.

thisisdano (approved):

PR #4287, opened 2021-08-12
[USWDS] COMBOBOX | Fix textContent combobox

scottqueen-bixal:
## Manually creates `li` and includes `option.text` as the safer `li.textContent`

includes a new test to check for potentially rendered `innerHtml` of list child elements.

### DIFF
--- a/spec/unit/combo-box/combo-box.spec.js
+++ b/spec/unit/combo-box/combo-box.spec.js
@@ -127,7 +127,7 @@ describe("combo box component", () => {
 
   it("should set up the list items for accessibility", () => {
     let i = 0;
-    let len = list.children.length;
+    const len = list.children.length;
     EVENTS.click(input);
 
     assert.strictEqual(
@@ -199,7 +199,7 @@ describe("combo box component", () => {
     assert.ok(!list.hidden, "should display the option list");
     assert.strictEqual(
       list.children.length,
-      43,
+      44,
       "should filter the item by the string being present in the option"
     );
   });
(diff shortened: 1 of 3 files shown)


mejiaj (approved):
Looks good, seems to resolve the text issue.

thisisdano (approved):
This look good to me. Thanks! 
@aduth any comments?

aduth:
Looks good to me 👍

spec/unit/combo-box/combo-box.spec.js
```diff
@@ -500,4 +500,20 @@ describe("combo box component", () => {
...
+    assert.ok(!list.hidden, "should display the option list");
+
+    // eslint-disable-next-line no-plusplus
+    for (let i = 0; i < list.children.length; i++) {
+      assert.strictEqual(
+        list.children[i].children.innerHTML,
+        undefined,
+        "should not contain child HTML"
+      );
+    }
```
aduth, line 517: `children.innerHTML` will always be `undefined`, because [`children` is an array-like `HTMLCollection`](https://developer.mozilla.org/en-US/docs/Web/API/Element/children).

You may need a nested loop to check each list item's children. As an alternative to checking `innerHTML`, you could check that each of the children is a text node:

```suggestion
  list.children.forEach((listItem) => {
    listItem.children.forEach((childNode) => {
      assert.strictEqual(childNode.nodeType, Node.TEXT_NODE);
    });
  });
```

scottqueen-bixal: TIL!! this change worked for us locally - committing the suggestion

aduth: Hm, something's not quite right with this actually. That assertion should be checking `childNode.nodeType`. In debugging, it looks like the assertion's not being called, because `Array.from(listItem.children)` is empty, even though that's valid in the real-world 🤷  Might need to revert back to plain-ol' `for` loop?

scottqueen-bixal: oof, this thing is trickier than I expected, taking another gander

aduth: Ah, I see the issue. `children` doesn't include the text nodes ([source](https://developer.mozilla.org/en-US/docs/Web/API/Element/children)). Works after switching to `childNodes`.

```js
    Array.from(list.children).forEach((listItem) => {
      Array.from(listItem.childNodes).forEach((childNode) => {
        assert.strictEqual(childNode.nodeType, Node.TEXT_NODE);
      })
    });
```

scottqueen-bixal: confirming that [nodeType](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#node_type_constants)  is returning  `3`

Screen Shot 2021-08-12 at 5 05 08 PM

PR #4295, opened 2021-08-20
USWDS - Library: Add storybook.

### LINKED ISSUE #4289: Add yaml loader to SB instance

mejiaj:
## Preview link
[USA Banner →
](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-add-storybook/?path=/story/components-banner--default)

## Description

Initial scaffolding for Storybook with webpack 5. Closes #4289.

## Additional information

- Removes `patterns` directory in a5a5a51a for simplicity
- Updates unit tests so they continue to work (still PL dependency) 42849e46

**Commands**

```bash
npm run start:storybook
npm run build:storybook
```


scottqueen-bixal:
got `yaml-loader` working over here https://github.com/uswds/uswds/tree/v2.10-beta-main-webcomponent-storybook. I was able to set up simple conditionals on the component attribute that renders the expected content from partial files exported from a single index.

.storybook/main.js
```diff
@@ -0,0 +1,48 @@
...
+        use: ['style-loader', 'css-loader', 'sass-loader'],
+        include: path.resolve(__dirname, '../'),
+      },
+      {
+        test: /.(jpe?g|png|svg)$/,
+        loader: "url-loader",
+        include: path.join(__dirname, "src/img")
+      },
+      {
+        test: /\.twig$/,
```
scottqueen-bixal, line 35: I wonder if we are going to be able to set up lit-html (web components) with this storybook base, or if it is better to go the other way?

src/stories/Button.stories.ts
```diff
@@ -0,0 +1,44 @@
+import { Story, Meta } from '@storybook/html';
```
scottqueen-bixal, line 1: probably can remove these default storybook files?

src/utils/test/patternlab-utils.js
```diff
@@ -27,24 +27,26 @@ async function getComponents() {
...
         const builtComponentDir = PL_BUILD_PATTERNS_DIR + file;
         fs.readdir(builtComponentDir, (err, dir) => {
-          dir.forEach((component) => {
-            if (component.includes(".rendered.html")) {
-              const c = file + "/" + component;
-              components.push(c);
-            }
-            if (index === directories.length - 1) {
-              resolve(components);
+          // eslint-disable-next-line no-restricted-syntax
```
scottqueen-bixal, line 30: what benefit or problem is this refactor solving?

.storybook/main.js
scottqueen-bixal: was able to use a conditional `resourceQuery` and call separate loaders for the scss files. Maybe there is another way to do this, but I think it at least gives us a POC on how we can build web-components and the twig templates in the same storybook project 👍 

https://github.com/uswds/uswds/blob/jm-add-storybook-lithtml-test/.storybook/main.js#L26-L43

```js
        "oneOf": [
          {
            "resourceQuery": /wc/, // foo.scss?wc
            "use": [{
              loader: 'lit-scss-loader',
                options: {
                  minify: true,
                },
              },
              "sass-loader",
            ],
          },
          {
            "use": ['style-loader', 'css-loader', 'sass-loader'],
            "include": path.resolve(__dirname, '../'),
          }
        ],
      },
```

mejiaj: This is great! I think this will be the best route, for now.

src/stories/Button.stories.ts
mejiaj: Yes, I think the initial non-USA components are from initializing Storybook.

src/utils/test/patternlab-utils.js
mejiaj: This file was reporting ESLint errors, this is just to get around that. Right now we're only using the compiled PL files to run our unit tests. They'll be removed moving forward.

mejiaj:
I've moved banner content to it's own `content` directory. I've also added an index.js file for exporting the content.

https://github.com/uswds/uswds/blob/7e886ae61b829de4562c7fe10d8bdda468e804f5/src/components/usa-banner/content/index.js#L1-L4

mejiaj:
This adds two new plugins for a11y.

[a11y addon [github.com]](https://github.com/storybookjs/storybook/tree/main/addons/a11y) for a11y alerts during development

image


[chanzuckerberg/axe-storybook-testing [github.com]](https://github.com/chanzuckerberg/axe-storybook-testing) for testing a11y in CI.

```bash
➜ npm run test:a11y

// Returns
info => Output directory: /home/jmejia/web/uswds/storybook-static
[chromium] accessibility
  Components/Banner
    ✓ Default (181ms)
    ✓ Default Spanish (85ms)
    ✓ Mil (96ms)
    ✓ Mil Spanish (81ms)

4 passing (1s)
0 failing
0 pending
```

.storybook/preview.js
```diff
@@ -0,0 +1,11 @@
+import "../dist/js/uswds.min.js";
+
```
mejiaj, line 2: It'd be best to add the main SCSS entry point here via:
`import "../src/stylesheets/uswds.scss";`

However this would require using the SCSS and PostCSS addons. The latter hasn't been updated and is stuck on `2.0.0`. This conflicts with the newest versions of Autoprefixer, even when forcing the latest PostCSS in Storybook.

When this is updated we should try again:
```js
module.exports = {
  core: {
    builder: "webpack5",
  },
  stories: ["../src/**/*.stories.mdx", "../src/**/*.stories.@(js|jsx|ts|tsx)"],
  addons: [
    {
      name: "@storybook/preset-scss",
      options: {
        cssLoaderOptions: {
          sourceMap: true,
          esModule: false,
        },
        sassLoaderOptions: {
          sourceMap: true,
        },
      },
    },
    {
      name: "@storybook/addon-postcss",
      options: {
        postcssLoaderOptions: {
          implementation: require("postcss"),
        },
      },
    },
    "@storybook/addon-links",
    "@storybook/addon-essentials",
    "@storybook/addon-a11y",
  ],
  ...
```

scottqueen-bixal:
I am noticing a bug when I switch between canvas and docs.

https://user-images.githubusercontent.com/37077057/134047331-9e1e1137-cee0-433e-bc4d-12a227e64485.mov

Consider that this may have something to do with how the autogenerated docs are wrapped, or it could be that there is no clean-up when switching between the two tabs. (might impact why we are seeing similar behavior in switching between stories of the same component)

I am not sure how important docs are for these initial HTML components since we are only using the canvas currently

see https://storybook.js.org/docs/react/writing-docs/docs-page#with-null-to-remove-docs
```
    docs: { 
      page: null 
    } 
```

mejiaj:
Good point, I don't think we have any definite strategy for using the docs right now.

mejiaj:
@scottqueen-bixal let's keep the docs for now and work towards a fix in the future.

scottqueen-bixal (approved):

PR #4304, opened 2021-08-31
adding images with better titles & dark mode support

mgifford:
I wanted to upload a few SVG files to evaluate if this is a better approach to SVG files for the USWDS. There are a lot more to take on, but I wanted to start and see what we could standardize on. 

This would include dark mode support by default. There is more that should be done for accessibility, but at least it is more consistent.

## Additional information

Looking at these SVG files I think there's more room for making them more consistent still, but this is a good step to makign them support dark mode out of the box & also providing more meaningful accessibility.

There were some things in the existing SVGs that really shouldn't be baked into a basic shape file. 

I think this is a good basic format. Can be minimized afterwards.

```
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" width="64" height="39" viewBox="0 0 64 39" role="img">
  <title>Down angle arrow</title>
  <style>
    path {
      fill: #005EA2;
   }
   @media (prefers-color-scheme: dark) {
     path {
        fill: #58b4ff;
      }
    }
  </style>
  <g>
    <path d="M63.36 4.004L60.155.741c-.427-.436-.919-.654-1.475-.654-.555 0-1.047.218-1.475.654L32 26.4 6.798.74C6.37.306 5.878.088 5.323.088c-.556 0-1.048.218-1.476.654L.642 4.005C.213 4.44 0 4.941 0 5.507c0 .565.214 1.066.642 1.501l29.883 30.427c.427.435.92.652 1.475.652.556 0 1.047-.217 1.474-.652L63.361 7.008c.427-.435.639-.936.639-1.501 0-.566-.212-1.067-.64-1.503z" />
  </g>
</svg>
```

I don't have an opinion if `<?xml version="1.0" encoding="UTF-8”?>` is needed or not. It works with or without. I like smaller file sizes. I've left it. 

Is `fill-rule=“evenodd"` used? Again this seems to be inconsistently applied. Filename should describe edge use cases.  I've removed it.

I really see no reason why any of these images should have `aria-hidden="true" focusable="false”` applied. I've removed it.

I don't know what the point of `data-prefix="fas" data-icon="angle-double-left”`  is. I think I left them, but...

I don't know why `class="svg-inline--fa fa-angle-double-left fa-w-14"` is defined in some. I've removed it. 

I think a lot of these we could probably just do with CSS.  Might be space to eliminate some of the variations.

My assumption was that for dark mode, we'd just flip the colors and choose one that is essentially the opposite in your color chart.


mejiaj:
@mgifford why not let `currentColor` set the fill based on the text color?

Before [2.10.1](https://github.com/uswds/uswds/releases/tag/v2.10.1) USWDS used font awesome icons. That's probably where those `fa-` prefixed classes come from. Similar case with the data attributes. I don't think they're used for anything.

mgifford:
@mejiaj so like this here:

  <style>
    path {
      fill: currentColor;
   }
  </style>
  
  I think this should work according to https://hiddedevries.nl/en/blog/2018-12-24-making-single-color-svg-icons-work-in-dark-mode
  
  I just haven't tested it yet. It would simplify it.

mgifford:
I'm going to close this and start a new one.

PR #4313, opened 2021-09-03
USWDS - File Input: Creates unique ID per upload. Fixes #4144

### LINKED ISSUE #4144: File Input: Loader spins indefinitely after selecting files with the same name in multiple inputs
### Describe the bug

When using multiple File Input components on the same page or form, subsequent selections of files with the same file name will cause those input forms to display the filename, but the loading icon remains permanently visible on the page.

This can be reproduced on the [USWDS File input component demo page](https://designsystem.digital.gov/components/file-input/):

1. Go to https://designsystem.digital.gov/components/file-input/
2. Upload any text file to the first file input component in the form.
3. Upload the same file to the second file input component in the form.
4. The second file input will display the name of the newly uploaded file but, but the loader spins forever.

### Expected behavior

The inputs should not remain in a loading state once the files are selected, regardless of a non-unique file names between inputs on the same page.

### Screenshots

Screen Shot 2021-04-06 at 9 55 17 AM
Screen Shot 2021-04-06 at 9 54 17 AM

### System setup
- **USWDS version:** Found locally on version 2.9.0, and on the USWDS demo website
- **Device:** MacBook Pro
- **Operating system:** macOS 11.2.3
- **Browser and version:** Safari 14, Edge 89

### Additional context

Discovered while developing https://github.com/raft-tech/TANF-app/pull/705

mahoneycm:
## Description

#4144 pointed out that if a file with the same name is uploaded in two separate file input fields the preview spinner would spin indefinitely. After some research, I found this is because the two documents would share the same ID and the function that would replace the spinner would _only_ select the first ID on the page. By implementing a function that assigns each upload an individual ID, an image preview properly loads.

The function used uses the date and time down to the second to assign an ID.

## Steps to reproduce error

(As provided in #4144)

1. Go to https://designsystem.digital.gov/components/file-input/
2. Upload any text file to the first file input component in the form.
3. Upload the same file to the second file input component in the form.
4. The second file input will display the name of the newly uploaded file but, but the loader spins forever.

## Additional information

Code to create unique ID based on upload time. Here we devide by 1000 and use mathfloor to get rid of the milliseconds.
```
const createUniqueID = (name) => `${name}-${Math.floor(Date.now().toString() / 1000)}`;
```

Partnered with the existing makeSafeForID function:
```
const imageId = createUniqueID(makeSafeForID(fileName));
```

Example Output:
image

## Screenshot

image

### DIFF
--- a/src/js/components/file-input.js
+++ b/src/js/components/file-input.js
@@ -106,7 +106,6 @@ const makeSafeForID = (name) => name.replace(/[^a-z0-9]/g, replaceName);
 
 const createUniqueID = (name) => `${name}-${Math.floor(Date.now().toString() / 1000)}`;
 
-
 /**
  * Builds full file input comonent
  * @param {HTMLElement} fileInputEl - original file input on page
--- a/src/js/components/file-input.js
+++ b/src/js/components/file-input.js
@@ -104,6 +104,9 @@ const replaceName = (s) => {
  */
 const makeSafeForID = (name) => name.replace(/[^a-z0-9]/g, replaceName);
 
+const createUniqueID = (name) => `${name}-${Math.floor(Date.now().toString() / 1000)}`;
+
+
 /**
  * Builds full file input comonent
  * @param {HTMLElement} fileInputEl - original file input on page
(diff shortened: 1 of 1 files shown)


src/js/components/file-input.js
```diff
@@ -104,6 +104,9 @@ const replaceName = (s) => {
  */
 const makeSafeForID = (name) => name.replace(/[^a-z0-9]/g, replaceName);
 
+// Creates a unique ID based on upload time. Here we devide by 1000 and use mathfloor to get rid of the milliseconds.
```
mejiaj, line 107:
```suggestion
// Takes a generated safe ID and creates a unique ID.
```

src/js/components/file-input.js
```diff
@@ -221,7 +224,8 @@ const handleChange = (e, fileInputEl, instructions, dropTarget) => {
 
     // Starts with a loading image while preview is created
     reader.onloadstart = function createLoadingImage() {
-      const imageId = makeSafeForID(fileName);
+      const imageId = createUniqueID(makeSafeForID(fileName));
+      console.log(imageId);
```
mejiaj, line 228: Remove test code `console.log()`.

mejiaj (changes requested):
Nice work! Just some minor comments.

scottqueen-bixal (approved):

PR #4314, opened 2021-09-03
Build: Replace SASS division with function polyfill

### LINKED ISSUE #4204: Dart Sass deprecation warnings: slash as division

aduth:
Fixes #4204
Previously: https://github.com/uswds/uswds/pull/4209#issuecomment-905728491

## Description

This is an attempt to resolve SASS division deprecation warnings and errors in projects which use recent versions of Dart Sass. As an alternative to #4209, it is intended to be compatible across Dart Sass, LibSass, and Ruby Sass. It does this by avoiding the division operation altogether, generating an equivalent result with only multiplication, addition, and subtraction. This could serve as a temporary solution until support for LibSass and Ruby Sass are dropped.


src/stylesheets/core/_functions.scss
```diff
@@ -387,10 +387,13 @@ the desired final units (currently rem)
 */
 
 @function spacing-multiple($unit) {
-  $grid-to-rem: ($system-spacing-grid-base * $unit) / $root-font-size-equiv *
-    1rem;
-
-  @return $grid-to-rem;
+  @return 1rem * divide(
+    divide(
```
mejiaj, line 391: Was there a reason for the extra division instead of something like:

```scss
  @return 1rem * divide(
    $system-spacing-grid-base * $unit,
    $root-font-size-equiv
  );
```

mejiaj (commented):
Fantastic, thanks for your help on this. Minor comment on placement on function and question on `spacing-multiple()`.

Otherwise LGTM.

src/stylesheets/core/_functions.scss
aduth: Good point! I can't recall exactly why I'd implemented it this way, though I remember having to do some trickery to convert units correctly. Testing the flattened version seems to produce the same output, so I'll simplify it with your suggestion.

src/stylesheets/core/_functions.scss
```diff
@@ -925,6 +927,47 @@ functions and mixins.
   @return $value;
 }
 
+/*
+----------------------------------------
+divide()
```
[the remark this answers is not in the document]
aduth: Interesting that I wasn't seeing these messages locally, though it makes sense. I moved up the function in 214463cad to occur before the first usage in the file.

src/stylesheets/core/_functions.scss
aduth: Simplified in 7f7ee9857.

src/stylesheets/core/_system-tokens.scss
```diff
@@ -142,18 +142,18 @@ Layout grid widths
...
+  3: divide(100% * 3, 12),
+  4: divide(100% * 4, 12),
+  5: divide(100% * 5, 12),
+  6: divide(100% * 6, 12),
+  7: divide(100% * 7, 12),
+  8: divide(100% * 8, 12),
+  9: divide(100% * 9, 12),
+  10: divide(100% * 10, 12),
+  11: divide(100% * 11, 12),
+  12: divide(100% * 12, 12),
```
mejiaj, line 156: Let's just confirm this gives the expected results.

aduth: A small difference in that there's no rounding for the final precision value at 10 decimal places (e.g. `16.6666666667%` vs. `16.6666666666%`):

https://www.sassmeister.com/gist/aduth/bcfcde6eeeec8272e56c4321a6678f48

aduth: It _might_ be possible to emulate the rounding: Do one extra level of recursion past `$precision`, then `+ 1` the last digit if that value is `>= 5`. 🤔

aduth: Here's a version with rounding: https://www.sassmeister.com/gist/aduth/95bacec332c6390bacaff9fe5ccc82b9

aduth: Pushed the rounding changes in dad2a78.

aduth:
A couple additional thoughts as I've been thinking about this again lately:

- Depending on what your schedule is for dropping legacy SASS support, this might not be worth it if old SASS support can be dropped outright. The advantage here is supporting both legacy SASS compilers as well as the Dart SASS compiler.
- I expect the division function here would perform slower than a native division operation. Depending how much a concern this is, and how long it might need to linger around for, it could be worth changing some of the operations to some equivalent operation, such as using multiplication.
- In testing, there's a significant difference in compile time between legacy SASS compilers and Dart SASS, around 90% faster:
   - `time dsass src/stylesheets/uswds.scss out.css # 1.18s user 0.14s system 101% cpu 1.295 total`
   - `time sass src/stylesheets/uswds.scss out.css # 18.07s user 0.40s system 99% cpu 18.582 total`
- This might not be the only deprecated feature that needs to be resolved before being compatible with Dart SASS. Specifically, there's also a warning about interpolation. This no longer logs when running Dart SASS, but I expect it's because the new behavior is taking effect:
   - >DEPRECATION WARNING on line 55 of src/stylesheets/components/_icon-list.scss: #{} interpolation near operators will be simplified in a future version of Sass.

aduth:
>This might not be the only deprecated feature that needs to be resolved before being compatible with Dart SASS. Specifically, there's also a warning about interpolation. This no longer logs when running Dart SASS, but I expect it's because the new behavior is taking effect:

See also: #4380

sean-usds:
@aduth thanks for fixing this! Just checking to see if/when this will get merged. Looking to clean up build warnings in a USDS/CDC project

thisisdano:
Will be released this week in `USWDS 2.13.0`

Issue #4300, opened 2021-08-30
Depending on how modal is initialized in React, there's different strange behavior


Swimburger:
### Describe the bug
We are trying to create a reusable Modal component using React, but integrating the JavaScript from USWDS has been challenging.
We're following the suggestions from [USWDS docs ](https://designsystem.digital.gov/documentation/developers/) on importing the components and using the `.on`/`.off` functions.

To demonstrate the different approaches we have taken, each with its own quirks, I have provided [a sample for you](https://github.com/Swimburger/uswds-sample/blob/master/src/App.js).
(I wanted to use JSBin or JSFiddle, but since the components aren't available without a bundler I have to create a sample with a bundler, so I'm providing a sample which you'll have to clone and run locally)

I'm not sure if these are bugs or not as there's no guidance on how to use JavaScript for specific components. I'm just guessing and experimenting until I get the components to work.
**So my question is, how should we initialize the buttons and modals?
And shouldn't modal 1 and 3 work?**

## Modal 1

For this modal implementation, I have created a ModalButton and Modal1 component.
The button component will run `modal.on(modalButtonElement)` after added to DOM and `modal.off(modalButtonElement)` before removal from DOM.
I think this button works fine I think, and I think the issue lays with the Modal1 component.

The modal component also runs `modal.on(modalElement)` after added to DOM and `modal.off(modalElement)` before removal from DOM.
After `modal.on` invocation, the `.usa-modal` element should be removed from the DOM by USWDS JS and added to the document.body wrapped in a couple of wrapper div-elements, but this doesn't happen.
When you click on the button to show the modal, there are some classes and attributes added to different places in the DOM tree, but no modal is ever shown because the modal was never initialized properly, or so I assume...

### Steps to reproduce the bug
1. Run the sample
2. Navigate to /modal1
3. Open elements tab in devtools
4. Observe nothing happens, although some attributes are added to button, modal, and body.

### Expected behavior
I would expect the modal to behave as it does in the styleguide.

## Modal 2

For this modal implementation, I have created a Modal2 component, but no dedicated ModalButton component as it is not needed.
The modal component runs `modal.on()` after added to DOM and `modal.off()` before removal from DOM.
Since I'm not passing a specific DOM element to on/off, it will use the entire DOM tree including the button and modal DOM.
This works completely as expected if you only have one modal, but if you have multiple modal, the on/off code will be run multiple times which breaks the functionality.

[...]

Issue #4055, opened 2021-03-01
Allow for dynamic initialization of tooltip component


jakerella:
**Is your feature request related to a problem? Please describe.**

I dynamically add elements to my page, some of which need a tooltip. Adding the tooltip classes and attributes does not allow the USWDS tooltips to be visible (although the browser default UI is) because the JS code has already initialized the components that were present at page load.

**Describe the solution you'd like**

I would like the ability to "trigger", in some way, the initialization of a tooltip (or multiple) given a root node. It looks like the code already does this through the `init` method of the component, so I just need a way to hit that. I would like to NOT modify the USWDS codebase to do this.

**Describe alternatives you've considered**

I've poked around at the JS code, but since nothing is exposed on the global context, I don't see a way to re-execute the tooltip init method for a given (sub)root node.

jakerella:
I have created a short term workaround for this. It only works for "top" tooltips, but could be made to work with other positions. The key was to trigger the adding and removing on classes myself.

In my html, I manually created the tooltip body at the same time that I create the tooltip trigger element:

```
<span class="usa-tooltip__body usa-tooltip__body--top" id="tooltip-[some id]" role="tooltip" aria-hidden="true">My tooltip text</span>`
```

In my stylesheet, I changed to styles on the tooltip body. The `display` was set to `none`, but that was unnecessary as the `opacity` is also `0` (note that this is bad for non-visual users). The `bottom` was reset because I ended up using `top` instead.

```
.usa-tooltip__body {
    display: block;
    bottom: auto;
}
```

In my JS, I added a `mouseover` and `mouseout` handler for the table I was adding elements to dynamically

```
someParentElem.addEventListener('mouseover', (e) => {
    let tooltipTrigger = findTrigger(e.target)  // helper method to get the correct parent
    // now find the tooltip described by the trigger and add the "is-visible" class
    someParentElem.querySelector('#' + tooltipTrigger.getAttribute('aria-describedby')).classList.add('is-visible')
})
someParentElem.addEventListener('mouseout', (e) => {
    // do the reverse of above
})
```

Last thing was to set the top and left of each tooltip body after I created them. Not too difficult, but also not a perfect solution:

```
Array.from(row.querySelectorAll('.usa-tooltip')).forEach((trigger) => {
    const triggerPos = trigger.getBoundingClientRect()
    const tip = row.querySelector('#' + trigger.getAttribute('aria-describedby'))
[...]

Issue #3753, opened 2020-10-07
Date Picker component does not initialize properly if created after page load


ConnorDY:
### Describe the bug
If you create the DOM structure for a Date Picker component anytime after the page has finished loading, the component will not initialize.

### Steps to reproduce the bug
1. Clone this repo: https://github.com/ConnorDY/uswds-date-picker-bug
2. Install dependencies (`npm ci`)
3. Run the app (`npm start`)
4. Click the button to remove the component from the DOM, then click the button again to re-add it.
5. Notice that the input shows up, but the calendar icon/button never gets added. Typing into the input causes an error.

### Expected behavior
The Date Picker component should initialize fine when created anytime after the page loads.

timmywertz:
I'm also experiencing this issue.

wbyoko:
You may want to look through this article in terms of interacting with JavaScript components inside of React https://reactjs.org/docs/integrating-with-other-libraries.html

This is an example that deals with initializing a USWDS component dynamically with React. https://github.com/flexion/ef-cms/blob/57e5546cbfd9e433f4a183147ad82aac56926673/web-client/src/views/CaseDeadlines/DatePickerComponent.jsx

ConnorDY:
> This is an example that deals with initializing a USWDS component dynamically with React. https://github.com/flexion/ef-cms/blob/57e5546cbfd9e433f4a183147ad82aac56926673/web-client/src/views/CaseDeadlines/DatePickerComponent.jsx

I tried this exact same thing, but calling `on` doesn't seem to make any difference. There are no errors in the console either.
I've updated my reproduction repo to show this.

jkreft-usgs:
we also ran into this issue (not with react directly, but its tied to any type of dynamic javascript) https://github.com/uswds/uswds/issues/3575

ConnorDY:
I mistakenly assumed `on` should be passed the `.usa-date-picker` element, just like the date picker's `setCalendarValue`, `enable`, and `disable` methods. Passing the container element works.

Issue #2890, opened 2018-12-17
USWDS should provide Angular components


duluca:
## Description

Angular being the second most popular web framework, in the same vein as issue #2284, there needs to be Angular components of USDWS designs. Otherwise bringing in the current USWDS implementation results in gross inefficiencies, due to anti-patterns being introduced.

## What's the problem?

Government contracts often include inflexible language, preventing contractors from using anything but this specific USWDS implementation, even if a newer framework being used may implement or exceed the guidelines.

## Alternative Solution

At a minimum, README should be updated to recognize, frameworks like React and Angular may need alternative solutions (i.e. Angular Material) vs this particular implementation, so long as the spirit and general outlines of the guidelines are being adhered to.

danizen:
It is also curious that uswds requires a peer component of react, if it is to support both.   I think the distribution of uswds needs to be split into at least:
* SCSS and minimal javascript
* React components

Possibly more, if this issue is to be worked.

pinkfrogHK:
I implemented USWDS 2.0 component into Angular 7 by npm package, and when I run ng test, the error says uswds is not recognized. I ran gulp for running scss. what is folder structure of the folders or is there any instruction document for Angular set up?

PR #4319, opened 2021-09-14
Add specs and update components to allow JS initialization on document.body and component element itself

Swimburger:
# Add specs and update components to allow JS initialization on `document.body` and component element

## Description

One of the challenges we faced trying to create React components was that we couldn't initialize USWDS components at the root element of the component, and had to use a wrapper element or `document.body`.
Initializing components at `document.body` results in weird issues because the initialization could have already happened before in which case you have a mix of initialized and uninitialized components.
For example, if an accordion is added and initialized it works fine, but if another is added the original does double toggles which results in the accordion not working anymore.

The modal component is the only component that required some extra work to setup and teardown cleanly.

In general, this will make the components more flexible and easy to use, especially in scenarios where components are rendered after initial page load + USWDS init regardless of whether it's with Angular, React, jQuery, vanilla JavaScript, etc.
Fixes:
- #4300
- #3753

Potentially fixes:
- #4055

This would help when creating Angular/React wrappers if USWDS plans to do so (#2890, #2884).

### Code changes
Generally, the approach I took was to update the spec files so they initialized components at `document.body` and also the root component element. This way I could verify that at least all existing tests were still passing for both `document.body` and root component elements.

To get the new tests to work, usually it just meant to not just query the subtree of the passed in element, but also check if the passed in element itself `matches`.

The only component that required more work was the modal, due to its specific behavior.
When a modal is initialized, the modal DOM is wrapped and moved to the document.body, so using the event behaviors, this just wouldn't work in all scenario's.
I fell back to using ```document.querySelectorAll(`[aria-controls="${modalId}"]`)``` to query all openers/closers to attach toggle behavior using event listeners.
I also added some extra logic to clean up on `.off` including moving the original modal DOM back to its original place.

- [x] ~~Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/)~~.
- [x] ~~Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free~~.
- [x] ~~Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.~~

---

This contribution is made on behalf of @MetroStar.

### DIFF
--- a/spec/unit/accordion/accordion.spec.js
+++ b/spec/unit/accordion/accordion.spec.js
@@ -1,6 +1,6 @@
-const Accordion = require("../../../src/js/components/accordion");
 const assert = require("assert");
 const fs = require("fs");
+const Accordion = require("../../../src/js/components/accordion");
 
 const TEMPLATE = fs.readFileSync(`${__dirname}/template.html`);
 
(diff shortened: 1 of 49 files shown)


scottqueen-bixal:
@Swimburger is there a working example that shows some of [these](https://github.com/Swimburger/uswds-sample/blob/master/src/App.js) issues resolved using the proposed work?

scottqueen-bixal:
@Swimburger when we first looked into this I landed in a similar spot with having to handle modal clean up, see https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/main/src/App.js#L34-L38

I've branched this and pulled in your proposed changes (sync'd to our beta dev branch, going on [here](https://github.com/uswds/uswds/pull/4397)) and adjusted the ref to point to the component element, see https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/test/library-develop/src/Modal.js, which seems to be working as expected.

However, I would have also expected that if I still managed the component within the parent component, the clean-up would no longer be needed since it was now included in `modal.off()`. Checkout, https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/test/library-develop-b/src/App.js#L30-L48. This doesn't seem to work for me. If I remove `modal.off` and clean up as I originally had implemented, it does work.  Have I missed something?

Swimburger:
> @Swimburger is there a working example that shows some of [these](https://github.com/Swimburger/uswds-sample/blob/master/src/App.js) issues resolved using the proposed work?

I have updated the sample to pull in the USWDS repo as a submodule, and you do have to then pull the specific PR.
Using the PR code, I got the desired scenarios working. Modal1 works, and the ModalContainer version also works.
Modal2 doesn't work, but it's not supposed to work IMO, so we're good there.

Swimburger:
> 



> @Swimburger when we first looked into this I landed in a similar spot with having to handle modal clean up, see https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/main/src/App.js#L34-L38
> 
> I've branched this and pulled in your proposed changes (sync'd to our beta dev branch, going on [here](https://github.com/uswds/uswds/pull/4397)) and adjusted the ref to point to the component element, see https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/test/library-develop/src/Modal.js, which seems to be working as expected.
> 
> However, I would have also expected that if I still managed the component within the parent component, the clean-up would no longer be needed since it was now included in `modal.off()`. Checkout, https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/test/library-develop-b/src/App.js#L30-L48. This doesn't seem to work for me. If I remove `modal.off` and clean up as I originally had implemented, it does work. Have I missed something?

I pulled this sample, and the cleanup built into `.off` seems to work as expected. No need to cleanup up yourself.

scottqueen-bixal:
> > 
> 
> > @Swimburger when we first looked into this I landed in a similar spot with having to handle modal clean up, see https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/main/src/App.js#L34-L38
> > I've branched this and pulled in your proposed changes (sync'd to our beta dev branch, going on [here](https://github.com/uswds/uswds/pull/4397)) and adjusted the ref to point to the component element, see https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/test/library-develop/src/Modal.js, which seems to be working as expected.
> > However, I would have also expected that if I still managed the component within the parent component, the clean-up would no longer be needed since it was now included in `modal.off()`. Checkout, https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/test/library-develop-b/src/App.js#L30-L48. This doesn't seem to work for me. If I remove `modal.off` and clean up as I originally had implemented, it does work. Have I missed something?
> 
> I pulled this sample, and the cleanup built into `.off` seems to work as expected. No need to cleanup up yourself.

Took another look as well, you're absolutely right, and thanks for pointing to additional working examples. We are working on syncing this with our [beta release branch](https://github.com/uswds/uswds/pull/4397) and hope to get it merged in soon.

Swimburger:
With the spinoff branch merged, what should happen with this PR?

thisisdano:
I think we're going to close this PR, and focus all follow-up work in the beta branch. That does mean that these features won't have a stable release until USWDS 3.0, but we'll try to keep the Beta branch moving and on-track, so folks that want or need this work can try it in Beta and be ready to update to 3.0 when it releases.

Swimburger:
Sounds good, looking forward to 3.0!

jbouder:
Any idea when 3.0 will be released?

thisisdano:
We're aiming for March.

mbucknell:
The issue I am having is somewhat related. I have a Vue application where sometimes my modals get double initialized because apparently the USWDS code hasn't yet initialized the modals when the modal component is "mounted". I can't figure out how to tell when the initial initialization is finished.

Issue #4305, opened 2021-09-01
Add native support for dark mode in SVG images


mgifford:
## Dark mode

Does the USWDS comply with the POUR framework of WCAG if the user has their browser (or operating system) set to dark mode? Not yet, but hopefully, that will change soon. 

I added this PR after looking through some of the SVGs and thinking that there must be a better way.

https://github.com/uswds/uswds/pull/4304

A few other related issues came up when I was looking at these. More standardization would be useful.

Essentially I specified a fill for `prefers-color-scheme: dark` - it would be useful if this were tested to see it worked well with custom style sheets. 

```
  <style>
    path {
      fill: #005EA2;
   }
   @media (prefers-color-scheme: dark) {
     path {
        fill: #58b4ff;
      }
    }
  </style>
```

mgifford:
Also see https://github.com/uswds/uswds/issues/3449 & https://github.com/uswds/uswds/issues/4131

mgifford:
I've created a darkmode version for all of the svg files currently in /src/img

These have been validated and some have been made more consistent.  The sprite at the end will be its own PR as it is a bit of a different beast.

This should be more accessible, however, we really need to re-think how we deal with CSS & SVGs.  This is a start, but I suspect we are going to need to see where this is applied to the themes. I found some odd combinations or inconsistencies.

why are there "alt" versions? What should happen to "white" checkboxes in dark mode? These may require further tweaking, but they have been adjusted to use `currentColor` where it made sense. 

https://jsfiddle.net/mgifford/28w1fr4h/

mgifford:
First attempt to approach this - https://github.com/uswds/uswds/pull/4304

Most recent PR - https://github.com/uswds/uswds/pull/4321

thisisdano:
Dark mode support makes sense, but it's not something we can apply only to SVGs — it will need comprehensive support across the design system, and that's going to be a really big thing. I support this work, but not in isolation.

PR #4321, opened 2021-09-15
Adding accessibility to SVG

mgifford:
## Make .svg images more semantic

Adding both dark mode and more semantic markdown to the SVG files. Titles are included in all files.

Described in more detail here https://github.com/uswds/uswds/issues/4305

Optimization by https://jakearchibald.github.io/svgomg/


mgifford:
@mejiaj would love your feedback on this.

mejiaj:
Thanks for this! I'm assigning it to myself so I can take a close look as soon as I can.

mejiaj:
I like the effort to make the icons more accessible.

Regarding dark mode, the `.usa-icon` class already handles setting color. I don't think we need to include it in the SVG itself.
https://github.com/uswds/uswds/blob/e1ed21783b089a2831504971b3426fbea1b5b752/src/stylesheets/components/_icon.scss#L13-L19

Here's a quick and dirty test to illustrate that.
https://codepen.io/mejiaj/pen/NWgydbr?editors=1111

mgifford:
@mejiaj glad you like the contribution generally. 

With regard to [icon-dot-gov.svg](https://raw.githubusercontent.com/uswds/uswds/develop/src/img/icon-dot-gov.svg) I didn't add that color. 

The fill is defined in the SVG already as `fill="#2378C3"` which nicely meets [light/dark contrast](https://contrast-ratio.com/#%232378c3-on-%23000). I don't think that `.usa-icon {  fill: currentColor; }` is being applied. 

I don't know if the best approach is to remove the `fill="#2378C3"` from the SVG or the `.usa-icon {  fill: currentColor; }` from the CSS, but someone should make that choice. I'm happy to modify the SVG if that is better for the design system. 

I stayed away from the sprite.svg. They are even more powerful, but haven't dug into an accessible way to manage the titles with sprites...

mejiaj (changes requested):
The optimizations and improved titles are great additions, thank you!

I think we should remove the `prefers-color-scheme` styling. The icons already use the current color when using the `usa-icon` class, which is what we expect. But @thisisdano has the final call.

---

Relating to `usa-icon` there's already guidance on making it more accessible here:
https://designsystem.digital.gov/components/icon/#accessibility-icon

Which is what I imagine you'd like to improve with this update.

mgifford:
Let me know which way you want to go @thisisdano I'm happy to help nudge this ahead where I can. I'd love to see full support and documentation of dark mode. 

@mejiaj I think I edited all of the SVGs in that folder except for the sprite.svg file. That one was a bit beyond my SVG skillset, so I left it alone. 

I may be misunderstanding things. This page seems to be all about the sprite.svg file (although it doesn't say this directly). Even the comment about the license, is just about the sprite.

Is the `usa-icon` class commonly used for all of the other SVG files in the folder? Although I suppose that you could override the svg images that I've changed (so that they follow the [direction about usa-icon accessibility](https://designsystem.digital.gov/components/icon/#accessibility-icon) then I think it probably should use a different approach to over-riding the accessibility. 

My view is that you make the icons accessible, people will use them, if folks want to hide them because it will improve the user experience for people with disabilities (removing redundancy and all) then we can give clear instructions on how to do that. I would want to have an example that didn't assume that you're modifying the sprite.svg file. 

Either way there is work to do.

mejiaj:
@mgifford yes, it's about the sprite.svg. It gets built from the SVGs inside the `usa-icons` directory. You don't need to directly modify the sprite. That gets generated.

We've made efforts to move to use less of the old icons and more of those found in `usa-icons`. Any additional icons needed users would pull from the `material-icons` directory and rebuild the sprite.

thisisdano:
I'm going to close this PR — while I do want to work in the direction of better dark mode support, I don't think we can add this to our codebase as-is, separate from other USWDS support for dark mode. I'm concerned that this may have some unintended color consequences. I do want to normalize and compress our SVGs, but I don't want to pursue dark mode compatibility for icons in isolation at the moment

PR #3890, opened 2020-12-14
USWDS - File Input - Add ability to use custom error message when wrong file type is used for file-input.js

hilvitzs:
## Description

Previously there was a hardcoded value for the error message that was displayed inside the `file-input` component when the wrong file type was used that read: `This is not a valid file type.` For the project I am on, designers requested that the message be modified a little bit so this PR adds the ability to use a custom error message with the `data-errormessage` attribute on the `input` with class `usa-file-input`.

## Additional information

Include any of the following (as necessary): 

Here is a screenshot of the error with the updated text:
Screen Shot 2020-12-14 at 11 10 55 AM

I am not sure what documentation change (if any) need to be made but I am happy to make them if someone can point me in the right direction!

Before you hit Submit, make sure you’ve done whichever of these applies to you:

- [x] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [x] Run `npm test` and make sure the tests for the files you have changed have passed.
- [x] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [x] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #4326, opened 2021-09-21
Revert "USWDS - File Input - Add ability to use custom error message when wrong file type is used for file-input.js"

swanprincess386:
Reverts uswds/uswds#3890


mejiaj:
@swanprincess386 reasoning for this?

PR #4364, opened 2021-10-20
USWDS: Checkbox - Checkbox print issue fix. Resolves #3137

### LINKED ISSUE #3137: Selected radio buttons and checkboxes do not print consistently
## Description
We want users to be able to print a copy of a completed form. Rather than generate a new page with the given information, we simply use a custom stylesheet specifically for printing. 

Here is how a radio/check input types render on the page: 
image

And here's how they look when printing:
image

As you can see, the outlines of the checkbox/radio button are highlighted in blue but it is very, very hard to see. Ideally, they would match how they look on the page.

1. Throw some inputs into the DOM, toggle them on
2. Pull up the print dialog.

mahoneycm:
## Description

When a page is sent to print with a selected checkbox, the check does not appear. This was resolved by adding the ASCII code for a checkmark as recommended by @mejiaj 

Resolves #3137 

## Additional information

Print Preview:

image

### DIFF
--- a/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
+++ b/src/stylesheets/elements/form-controls/_checkbox-and-radio.scss
@@ -205,7 +205,8 @@
       @media print {
         background-image: none;
         background-color: color("white");
-        content: url("#{$theme-image-path}/checkbox-check-print.svg");
+        content: "\2714";
+        text-align: center;
       }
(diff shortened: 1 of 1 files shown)


mejiaj (approved):
Works consistently in Safari, Firefox, and Chrome. Thank you!

PR #4397, opened 2021-11-10
gsq-metrostar-library-develop-sync

scottqueen-bixal:
## Description

> ...make the (JS) components more flexible and easy to use, especially in scenarios where components are rendered after initial page load + USWDS init regardless of whether it's with Angular, React, jQuery, vanilla JavaScript, etc.

Additional context included in initial [PR](https://github.com/uswds/uswds/pull/4319).

## Testing

Lots of specs have been updated from the initial [pull request ](https://github.com/uswds/uswds/pull/4319) that pass an array of conditions testing for `document.body`, or an assigned selector

```js
const tests = [
  { name: "document.body", selector: () => document.body },
  { name: "accordion", selector: accordionSelector }
]
```

Some examples are available with code mods
### Additional USWDS examples with code modifications
https://github.com/scottqueen-bixal/uswds-modal-initalizer/blob/test/library-develop/src/Modal.js

You can see in this React example we can now set up a ref and pass it through the initializer for the match script.

```js
  const modalRef = useRef()

  useEffect(() => {
    const modalElement = modalRef.current;
    modal.on(modalElement)
    return () => {
      modal.off(modalElement)
    }
  }, [modalRef])
```

### Initial USWDS code samples from Metrostar

https://github.com/Swimburger/uswds-sample/blob/master/src/App.js

note from contributor (@swimburger)
> I have updated the sample to pull in the USWDS repo as a submodule, and you do have to then pull the specific PR.
Using the PR code, I got the desired scenarios working. Modal1 works and the ModalContainer version also works.
Modal2 doesn't work, but it's not supposed to work IMO, so we're good there.


.eslintrc.yml
```diff
@@ -16,7 +16,9 @@ rules:
   no-unsanitized/method: error
   no-unsanitized/property: error
 overrides:
-  - {
-      files: ["*.spec.js"],
-      rules: { no-unsanitized/method: off, no-unsanitized/property: off },
-    }
+  - { files: ['*.spec.js'],
```
scottqueen-bixal, line 19: check this

Swimburger:
I was going to raise this in a separate issue, but I think when you're doing DOM manipulation, you're supposed to use `.useLayoutEffect` instead of `.useEffect`.
So depending on the USWDS component, it should be one or the other, depending on if there's DOM manipulation.
If only event listeners are attached, `.useEffect` is correct I believe.
Some info on the difference: https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

mejiaj (approved):

thisisdano (approved):

PR #4404, opened 2021-11-24
add-color-icon mixin - webkit support

wendellpiez:
## Description

This addresses a display glitch: icons can be dropped when properties are missing due to partial support of `mask` in webkit browsers (Nov 2021).

The problem is addressed by adding webkit properties back.

An example of a dropped icon (the down arrow after "Here's how you know"):

image

Partial support of `mask` is noted here: https://caniuse.com/?search=mask

### DIFF
--- a/src/stylesheets/core/mixins/_icon.scss
+++ b/src/stylesheets/core/mixins/_icon.scss
@@ -141,6 +141,10 @@
     background: none;
     background-color: if($color == currentColor, $color, color($color));
     mask: $mask-props;
+    // added to work in Chrome/webkit
+    -webkit-mask-image: url("#{$path}/usa-icons/#{$filename-base}.svg");
+    -webkit-mask-repeat: no-repeat;
+    -webkit-mask-position: center;
     @if $color-hover {
       &:hover {
         background-color: color($color-hover);


mejiaj:
Thanks for taking the time to contribute @wendellpiez. We mention Autoprefixer in the [SASS compilation requirements](https://github.com/uswds/uswds#sass-compilation-requirements) for this reason.

>The design system requires autoprefixing to work properly. This is included in the uswds-gulp package.
>
>Autoprefixing uses a service like gulp-autoprefixer to automatically add vendor prefixes to the precompiled stylesheets. Don't add vendor prefixes to your custom styles manually — it is more reliable to use autoprefixing.

Going to close this PR, but please don't hesitate to create issues/PRs in the future. Thanks again!

david-waltermire-nist:
@mejiaj I don't believe this is an adequate fix for this issue. There is a fundamental feature gap in all Webkit-based browsers. 

This bug is caused by a bad response from the `@supports (mask: url(""))` clause in _alerts.scss](https://github.com/uswds/uswds/blob/develop/src/stylesheets/components/_alerts.scss#L11) and incomplete support for webkit-based browsers for this alert feature.

The Autoprefixer solution you propose requires that all users of USWDS facing this issue implement the autoprefixer solution. This assumes they can figure out that autoprefixer is the right thing to do and then figure out how to use it. This is not very user friendly. I believe that USWDS should work consistently on all supported browsers out-of-the-box, which is also consistent with the USWDS design goal "Promote continuity". Please reconsider this PR.

david-waltermire-nist:
@mejiaj Based on what @wendellpiez states above, I don't believe autoprefixing fixes this problem, since this issue is really about incomplete CSS support for masks in Webkit-based browsers. Can you provide an illustration of how to fix this bug using autoprefixing?

wendellpiez:
The problem seems to be that Chrome and hence Edge pass the test (since they have *some nominal support* for masks) but they do not support the masking as expressed (with the shorthand `$maskprops`).

@mejiaj can you point me to a site where this is working live so I can see the difference?

david-waltermire-nist:
@mejiaj The problem is that [webkit-based browsers only support](https://caniuse.com/?search=mask) the `mask-image` part of the mask spec. No prefixing of the CSS attributes is going to change this.

The following illustrates this by prefixing `mask` with `-webkit-mask`:
mask-problem

As illustrated above the size parameters of the `mask` cause the icon to be clipped.

You are asking us, and all other USWDS users, to implement a work around for this issue that could be easily fixed in the USWDS core. Can you provide an explanation as to why this can't be fixed in the core?

mejiaj:
@david-waltermire-nist sorry this seems to have caused so much frustration. The icon displays with the `-webkit` prefix, so mask isn't the issue.

There's an override in this CSS block that's causing the clipping:
```css
.usa-site-alert--slim .usa-alert:before {
    background-size: 1.5rem;
    height: 2.5rem;
    width: 1.5rem;
}
```

The size has been bumped up on the pseudoelement's dimensions, but sizing in mask is still the default `2rem`. If you change it to match your overriding dimensions. It should display properly:

```css
-webkit-mask: url("https://pages.nist.gov/oscal-tools/img/usa-icons/info.svg") no-repeat center / 1.5rem 2.5rem;
```

image


---

You need autoprefixer to get both `mask` and `-webkit-mask` to compile SCSS. If you're using the SCSS we provide an easy way to update the size of the icons via theme settings files.

There's a setting called [`$theme-alert-icon-size` here →](https://github.com/uswds/uswds/blob/eb84a497042aa85d79e8b5d2225a874c210bdb33/src/stylesheets/theme/_uswds-theme-components.scss#L28). In your case you'd update the value to `5` and it'll handle the sizing for you.

If you're curious what the other settings do we've highlighted those in the [Alert guidance](https://designsystem.digital.gov/components/alert/#using-the-alert-component).

Issue #3691, opened 2020-09-29
Form controls: labeling required fields


jAigret-Bix:
USAGov accessibility [test results ](https://docs.google.com/document/d/1g5beRCFgb5M7xYrEpBPuZy39AKYxtCXzT54FHTPiRgg/edit)

See associated bug issue #3788 
See associate issue https://github.com/uswds/uswds/issues/3295

1. Instructions of the form not placed at the top of the page (https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-cues.html) . Place instructions for a form at the top of the page. *For example “All fields are required except optional”*
2. Also, required fields not identified. Users should be able to easily determine which data entry fields are required and which are optional. (https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/H90)

See Zendesk ticket #20401 w question: "will marking optional fields only pass the Trusted Tester process and 508 compliance? Also, if all fields are required, is it safe to leave that out of the instructions?"

[Slack](https://gsa-tts.slack.com/archives/C050HRGN7/p1601400724037300): The form required/optional one is a little bit more tricky, perhaps, since it'll also require updating our guidance. See https://github.com/uswds/uswds/issues/3295

- [x] @allie-shaw  to research and develop recommendations. 
- [x] meet with accessibility testers
- [x] review and capture recommendations 
- [ ] 
- [x] add instructions and templates
- [x] Allie - content audit - check and verify all form controls that we don't have conflicting guidance
- [ ] publish

allie-shaw:
## Recommendations
Do not have one error message at the top (ie: All fields required)
- People don’t read instructions at the top of forms.
- Even if people read instructions, they may forget them. 
- People have to scan the form to determine if the field is required. 

Instead:
- Inputs should be clearly indicated as optional or required. These indicators can be included in the label. 
- If most fields in the form are required, indicate only the optional fields.
- If most fields in the form are optional, indicate only the required fields.

_______________________________________________________
### Carbon Design System
https://www.carbondesignsystem.com/components/form/usage#validation-and-errors
Optional versus required fields
All fields in a form are assumed required, with optional fields being tagged as so. It is recommended that an instruction precedes the form stating either “All fields are required” or “All fields are required unless marked optional”.

### Material Design System
https://material.io/components/text-fields
**Required text indicator**
To indicate that a field is required, display an asterisk (*) next to the label text and mention near the form that asterisks indicate required fields.
- If some fields are required, indicate all required ones
[...]

PR #4417, opened 2021-12-06
Forms: Add required field guidance to Sign In form.

mejiaj:
## Description

**Sign in form includes updated required field guidance.** The sign in form template has required field instructions at the top of the form and required fields marked with an asterisk.

Related to https://github.com/uswds/uswds-site/issues/1354.

## Additional information

  Original guidance was provided in [issue #3691 -  Form controls: labeling required fields](https://github.com/uswds/uswds/issues/3691#issuecomment-710312597).

---

### DIFF
--- a/src/components/templates/sign-in-form.njk
+++ b/src/components/templates/sign-in-form.njk
@@ -1,9 +1,13 @@
 <form class="usa-form">
   <fieldset class="usa-fieldset">
     <legend class="usa-legend usa-legend--large">{{ form.legend }}</legend>
+    <p>
+      Required fields are marked with an asterisk (<abbr title="required" class="usa-hint usa-hint--required">*</abbr>).
+    </p>
     {# Email #}
     <label class="usa-label" for="{{ form.email.input.id }}">
       {{ form.email.label }}
+      <abbr title="required" class="usa-hint usa-hint--required">*</abbr>
     </label>
     <input
       class="usa-input"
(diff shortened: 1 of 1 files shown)


thisisdano:
This is one that I don't yet feel comfortable merging. Though guidance says to label required fields, sign in fields (like name + password) are one of the instances where the additional language and asterisks feel possibly _more_ confusing, since they aren't expected in this type of form. (Single-field forms, like search are another place where the guidance breaks down.)

mgifford:
Ok, so this is ultimately about:
  https://designsystem.digital.gov/templates/form-templates/sign-in-form/

I gotta say, I don't see the big deal about adding a required element to a login form. It's the default for Drupal, so I'm used to it. I'm not even sure I helped to put it there. 
  https://www.drupal.org/user

I've never heard a complaint that this was a usability problem for anyone. We haven't done any user research on it though. 

I figured I'd check with WP, and.....  
https://login.wordpress.org/

No required field. 

I frankly don't see what the problem is of marking (even if just with ARIA) that a single input form is required. There are times I hit submit without entering any data, and frankly it should either work (by displaying everything or nothing) or the user should get a validation error. 

Ultimately though, we need user research to actually determine what users need. Perhaps we'll get results which we can use to provide better guidance. I couldn't find a corresponding example from the UK's GDS. They are usually great leaders on user research. - https://design-system.service.gov.uk/patterns/

mejiaj:
@dct2023 Closing this as stale and will revisit once @jaclinec and @mahoneycm have concrete research. We can create a new issue to scope the changes.

Issue #2836, opened 2018-10-08
2.0 Feature Request - Make dropdown navigation behavior similar to mobile and side nav


stphnwlkr:
I think it would be aesthetically appealing, add consistency, and be more user friendly if the links in the desktop dropdown behaved the same way links in the mobile nav and sidenav  with background color change (subtle or dramatic) and the link spanning the full width of the submenu container.  This is especially helpful when link are of varying length. 

Example
![nav_sample](https://user-images.githubusercontent.com/1405187/46632481-ca63ec80-cb18-11e8-9fa9-cdd21769d885.png)

maya:
via @swirtSJW in #3033:

## Sub-links in header menu have too small a target

When I expand the dropdown in a menu, the submenu items target is only the link text rather than the full width and height of the menu element.   This has some accessibility issues.

I expect that when my mouse is over a menu item in the expanded state (on desktop) I can click and go to the page without having to be certain to click only on the text in the menu item.


## Steps to reproduce the issue

1. Click on a level one menu item to expand it.
2. Try to activate a sub-menu element item by clicking near the text but not on the text.


## Additional information [optional]

* Screen shot images
![image](https://user-images.githubusercontent.com/5752113/56057717-87840b00-5d2d-11e9-8354-5c9bd162c7b7.png)

mejiaj:
@mahoneycm needs PR update.

Issue #3033, opened 2019-04-12
Sub-links in header menu have too small a target


swirtSJW:
## When I expand the dropdown in a menu, the submenu items target is only the link text rather than the full width and height of the menu element.   This has some accessibility issues.

I expect that when my mouse is over a menu item in the expanded state (on desktop) I can click and go to the page without having to be certain to click only on the text in the menu item.

## Steps to reproduce the issue

1. Click on a level one menu item to expand it.
2. Try to activate a sub-menu element item by clicking near the text but not on the text.

## Additional information [optional]

* Screen shot images
![image](https://user-images.githubusercontent.com/5752113/56057717-87840b00-5d2d-11e9-8354-5c9bd162c7b7.png)

maya:
Hi @swirtSJW, thank for opening this issue. This looks like a duplicate of #2836. Could you add your comments to that issue so we can keep the conversation in one place?

maya:
Moved content to #2836, closing duplicate issue.

PR #4427, opened 2021-12-14
Create bigger targets for Header Nav submenu items

mahoneycm:
## Description

Modeled some of the suggestions in #2836 and #3033 which give subnav menu items a larger targeting area and a visual queue that it is selected, similar to the mobile version.

The desired effect was accomplished by setting the submenu item anchor tag to `display: block;` and moving the padding from the parent tag to the anchor itself.

I also set the hover state background color to `primary` from `primary-darker` to give a highlighted effect.

## Additional information

Current Nav menus

image
image

Suggested on #2836 
image

After this fix

image
image
_The padding being added to the anchor tags does make the mega menu larger compared to before_

Fixes https://github.com/uswds/uswds/issues/2836


mahoneycm:
@mejiaj

thisisdano:
I like the block anchors, but let's remove the highlighting for now. Perhaps it's something we can enable with a setting, but I don't just want to introduce it out of nowhere.

thisisdano:
Let's also look at reducing the padding a bit, feels a bit too generous as-is.

src/stylesheets/components/_navigation.scss
```diff
@@ -283,19 +282,15 @@ $nav-link-arrow-icon-size: 2;
...
-      & + * {
-        margin-top: units(1.5);
-      }
-
       a {
         color: color("white");
-        padding: 0;
         line-height: line-height($theme-navigation-font-family, 3);
+        display: block;
+        padding: units(2);
```
mejiaj, line 289:
```suggestion
        padding: units(1);
```

src/stylesheets/components/_navigation.scss
```diff
@@ -283,19 +282,15 @@ $nav-link-arrow-icon-size: 2;
...
 
         &:hover {
-          background-color: transparent;
+          background-color: color(primary);
```
mejiaj, line 292: Let's remove this for now.

mejiaj (commented):
Added some suggestions based on @thisisdano's comments.

mahoneycm:
Requested changes made and pushed! @mejiaj

mejiaj (commented):
We might want to update the megamenu link offset value in `_header.scss`. A value of `units('neg-05')` seems to fix it.

Noticed the following regression when focus is on megamenu links.

**Basic with mega-menu**
image
image

**Extended header with mega-menu**
image

mahoneycm:
> We might want to update the megamenu link offset value in `_header.scss`. A value of `units('neg-05')` seems to fix it.
> 
> Noticed the following regression when focus is on megamenu links.
> 
> **Basic with mega-menu** image image
> 
> **Extended header with mega-menu** image

Fixed in commit [4cea191](https://github.com/uswds/uswds/pull/4427/commits/4cea191cdb2f6ccc9d56e790540a6f37e6d1d5df). Note that I had to add this to `_navigation.scss` to get it to work. Not `_header.scss`

mejiaj (approved):

thisisdano:
@mahoneycm Looks like this just needs some merge conflict resolution and we're good to go

PR #4437, opened 2021-12-21
Optimize luminance function as simple look-up

aduth:
## Description

This pull request seeks to improve the performance of the `luminance` function, which is currently implemented using a combination of math polyfills necessary to compute an fractional exponent value following the [relative luminance formula](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#relativeluminancedef).

## Additional information

Because the [relative luminance formula](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#relativeluminancedef) uses the same calculation for each of the R(ed), G(reen), and B(lue) components of a color, and because RGB component values are guaranteed to be a value between 0 and 255, it is possible to reasonably pre-compute the result for each value between 0 and 255, allowing for a trivial O(1) dictionary lookup.

I calculated the results using the following JavaScript script:

```js
[...new Array(256)].map((v, i) => i / 255).map((v) => v < 0.03928 ? v / 12.92 : Math.pow((v + 0.055) / 1.055, 2.4)).join(', ')
```

Evaluated using Node.js:

```
node -p "[...new Array(256)].map((v, i) => i / 255).map((v) => v < 0.03928 ? v / 12.92 : Math.pow((v + 0.055) / 1.055, 2.4)).join(', ')"
```

### Results

The performance of the previous implementation seemed to worsen exponentially based on the number of times it's called. In a reasonable benchmarking scenario of 20 calls, execution time is reduced by ~94%.

Benchmarking code:

```scss
// example.scss
@import './src/stylesheets/packages/required';

@for $i from 1 to 20 {
  $result: luminance(color('primary'));
}
```

Results, using Ruby Sass:

```
$ time sass example.scss
sass example.scss  28.07s user 0.32s system 99% cpu 28.634 total

$ time sass example.scss
sass example.scss  1.54s user 0.20s system 98% cpu 1.773 total
```

I had attempted to measure results in (native) Dart Sass, but the math of **the previous logic seems to not work at all for me in Dart Sass**, due to a division by zero (see related #4438). Thus, this refactoring could also serve as a bug fix for Dart Sass.

---


aduth:
I forgot to mention: If the plan with USWDS v3 is to drop support for non-Dart SASS (i.e. use SASS module syntax), then this could also be solved by using the new [built-in `math.pow`](https://sass-lang.com/documentation/modules/math#pow), which I assume performs reasonably-well. A look-up would still probably be faster, but at least in that case you wouldn't need to include a huge list of 256 float values 😅 

The main advantage of this implementation is that it works across all compilers.

mejiaj (commented):
Pretty nice! Thanks for this.

I wonder if we should move the `$luminance-values` somewhere else. Either inside the function or to `core/_variables.scss`. Otherwise LGTM!

aduth:
Thanks for the review, @mejiaj !

> I wonder if we should move the `$luminance-values` somewhere else. Either inside the function or to `core/_variables.scss`.

Yeah, I think that could make sense to move it to `_variables.scss`. Maybe under a new section "Luminance", since I don't see an existing section that would be a great fit.

Keeping it outside the function might be a soft preference for me, since while I don't know how SASS optimizes variables within functions, I might worry if it's the sort of thing where the 255-entry list would be allocated/garbage collected for every call.

mejiaj:
>Keeping it outside the function might be a soft preference for me, since while I don't know how SASS optimizes variables within functions, I might worry if it's the sort of thing where the 255-entry list would be allocated/garbage collected for every call.

That's a good point. Moving this to final review 🙂

aduth:
> I had attempted to measure results in (native) Dart Sass, but the math of **the previous logic seems to not work at all for me in Dart Sass**, due to a division by zero (see related #4438). Thus, this refactoring could also serve as a bug fix for Dart Sass.

For what it's worth, digging into this, it might be related to [maximum integer size in Dart](https://api.dart.dev/stable/2.15.1/dart-core/int-class.html), causing the [`factorial` function](https://github.com/uswds/uswds/blob/823b7a69af6fecce27b0c1de9dc950c146f4ddf4/src/stylesheets/core/_functions.scss#L991-L1003) to return `0`, and since the result of that is used as a [denominator in `exp`](https://github.com/uswds/uswds/blob/823b7a69af6fecce27b0c1de9dc950c146f4ddf4/src/stylesheets/core/_functions.scss#L1020), it's a division by zero.

Example script to reproduce:

```scss
@function factorial($value) {
  $result: 1;

  @if $value == 0 {
    @return $result;
  }

  @for $index from 1 through $value {
    $result: $result * $index;
    @debug 'index #{$index}, result #{$result}';
  }

  @return $result;
}

@debug factorial(99);
```

```
example.scss:10 Debug: index 1, result 1
example.scss:10 Debug: index 2, result 2
example.scss:10 Debug: index 3, result 6
...
example.scss:10 Debug: index 65, result -9223372036854775808
example.scss:10 Debug: index 66, result 0
...
example.scss:10 Debug: index 99, result 0
```

thisisdano (approved):

thisisdano:
Thank you!

sawyerh:
@thisisdano @aduth Just wanted to pop in and say THANK YOU THANK YOU THANK YOU. It has always confused me why the CSS build step took so long in my application and this made a noticeable dent.

Issue #3395, opened 2020-04-16
External link icon wraps to next line


maya:
<img width="359" alt="Screen Shot 2020-04-16 at 9 04 12 AM" src="https://user-images.githubusercontent.com/5249443/79479241-546bfb80-7fc1-11ea-84e9-e06a4f4724e0.png">

Here is a fix: https://www.jayfreestone.com/writing/wrapping-and-inline-pseudo-elements/

Will enable you to control color directly in CSS (normal, hover, visited), whereas we don't have a visited right now.

Source: https://components.designsystem.digital.gov/components/detail/links.html

PR #4439, opened 2021-12-22
External links: fix Safari display bug.

### LINKED ISSUE #4434: External Links icon not rendering properly in Safari
The external link icon is not rendering correctly in Safari

mejiaj:
## Previews
[External links →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jm-external-link-icon/components/preview/links.html)

## Description

**Improved display of External links.** Now external links display consistently on all browsers.  Specifically fixes a visual regression in Safari when using mask to set the icon. Fixes #4434.

## Additional information

Previously the external-link mixin was using `display: inline` on the `::after` pseudoelement to prevent the icon from breaking into it's own line on narrow screen sizes #3395. This causes issues in Safari when using masks because the background set doesn't get clipped properly.

image

Setting this to `display: inline-block` fixes the display issue in Safari, but then breaks the flow of the icon on smaller screens.

image
Example from designsystem.digital.gov

**The fix**
I'm using `position: absolute` to properly place the icon and using inline-block to get it to display properly in Safari.

---

I've also removed the dimensions being set via `background-size` and `mask-size` in `add-color-icon()`. This was causing the external link icon to appear squished at certain sizes.

This change affects all components that use `add-color-icon()` (`_alerts.scss:26` for example).

---

### DIFF
--- a/src/stylesheets/core/mixins/_icon.scss
+++ b/src/stylesheets/core/mixins/_icon.scss
@@ -124,8 +124,8 @@
   );
 
   $mask-props: url("#{$path}/usa-icons/#{$filename-base}.svg") no-repeat center /
-    #{$width} #{$height};
-  $image-props: url("#{$path}/#{$filename-ie11}") no-repeat center / #{$width} #{$height};
+    contain;
+  $image-props: url("#{$path}/#{$filename-ie11}") no-repeat center / contain;
 
   // Default background shorthand for browsers that don't support mask or supports.
   background: $image-props;
--- a/src/components/identifier/identifier.njk
+++ b/src/components/identifier/identifier.njk
@@ -69,13 +69,13 @@
     <div class="usa-identifier__container">
       <ul class="usa-identifier__required-links-list">
         <li class="usa-identifier__required-links-item">
-          <a href="{{ required_links.links.about.url[lang] }}" class="usa-identifier__required-link">
+          <a href="{{ required_links.links.about.url[lang] }}" class="usa-identifier__required-link usa-link">
             {{- required_links.links.about.content[lang] }} {{ masthead.parent.shortname -}}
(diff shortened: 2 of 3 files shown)


src/stylesheets/core/mixins/_icon.scss
```diff
@@ -124,8 +124,8 @@
   );
 
   $mask-props: url("#{$path}/usa-icons/#{$filename-base}.svg") no-repeat center /
-    #{$width} #{$height};
-  $image-props: url("#{$path}/#{$filename-ie11}") no-repeat center / #{$width} #{$height};
+    contain;
+  $image-props: url("#{$path}/#{$filename-ie11}") no-repeat center / contain;
```
aduth, line 128: Are the `$width` and `$height` variables unused now after these changes? (i.e. can they be removed?)

Oddly, there's also an assignment a few lines below this line, which I'm not sure would have had any effect? 🤔

mejiaj: `$width` and `$height` are used to get consistent sizing of the icons across all browsers in both background and mask.

It seemed more straightforward and allowed consistent rendering this way. But open to any improvements.

aduth: Ohhh, my brain misread line 134 as `$width:` (variable assignment) instead of `width:` (CSS property). Disregard 🤦

mejiaj: All good! Just getting back from break so the brain fog on this was real heavy. 😅

thisisdano (approved):

thisisdano:
Thank you!

PR #4470, opened 2022-01-19
Fix path syntax

thisisdano:
This updates the path syntax in our new exports to match the existing exports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,8 +28,8 @@
     "./js/*": "./dist/js/*",
     "./css/*": "./dist/css/*",
     "./scss/*": "./dist/scss/*",
-    "./img/*": "./dist/img*",
-    "./fonts/*": "./dist/fonts"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,8 +28,8 @@
     "./js/*": "./dist/js/*",
     "./css/*": "./dist/css/*",
     "./scss/*": "./dist/scss/*",
-    "./img/*": "./dist/img*",
-    "./fonts/*": "./dist/fonts"
+    "./img/*": "./dist/img/*",
```
thisisdano, line 31: Fixed as misplaced `*` here

package.json
```diff
@@ -28,8 +28,8 @@
...
+    "./fonts/*": "./dist/fonts/*"
```
thisisdano, line 32: Matched existing `directory/*` syntax here

mahoneycm (approved):

PR #4476, opened 2022-01-21
[FOOTER] At mobile menus change disclosure button elements from <h4> to <button>

### LINKED ISSUE #4475: [Mobile A11y] Expand/collapse buttons on Footer navigation menu are <h4> s but look and function like buttons
### **Describe the bug**
When accessing the Footer component on a mobile device, the navigation menu turns into a stack of 4 expand/collapse buttons indicating menu headers. These elements look and function like buttons (they can be expanded to reach the sub navigation) but are a series of h4s. A screen reader user would not be aware that these headings were actionable. The elements are also missing aria-expanded attributes. A screen reader users would not know whether the element was expanded or collapsed. 

**A note on keyboard accessibility:** menu buttons are navigable on an iPhone using VoiceOver and can be physically expanded/collapsed as expected. On a laptop using Responsive Device Mode on Chrome and Firefox Dev Tools, these elements were not accessible via keyboard. 

### **Steps to reproduce the bug**
1. Go to the [Footer Component](https://designsystem.digital.gov/components/footer/) on a mobile device. 
2. Turn on your devices' screen reader. 
3. Navigate to the first menu header (labeled here as "Primary Link").
3. Screen reader will indicate that the element is an h4.
4. Screen reader will not indicate that the element is clickable nor whether it is expanded or collapsed.

### **Expected behavior**
I would expect to hear a screen reader to report the elements' name (Primary Link), it's role (button) and whether the element is expanded or collapsed. 

### **Screenshots**

Screen Shot 2022-01-20 at 4 47 28 PM

mherchel:
Resolves #4475.

## Description
This PR fixes some accessibility issues in the big footer variation. At mobile widths the footer menus collapse into menus. However the menu disclosure button is actually a `<h4>` element, which is not navigable by keyboard, and has a host of other accessibility issues (for example, a `space` or `enter` won't always activate this.

This PR fixes the issue by actually swapping out the `<h4>` for a `<button>` at mobile widths (and vice-versa at larger widths). 

In addition to the JavaScript changes, I had to make some changes to the styling so that button styles don't leak through, and to the tests, so the tests dynamically query for the selector instead of storing it as a variable.


mahoneycm:
Great job! Everything works as expected across Chrome, Firefox, and Edge. 

We did notice the focus border is hidden on both sides and the bottom when collapsed

Is there a way we can get at least the top and bottom to appear while collapsed to avoid confusion on which button is being selected? Currently, it appears correctly when not collapsed

It also looks like the first secondary link is pushed right up against the button. Could you add some space between them when in the smallest mobile view for a little bit of separation?

image

Thank you for your contribution to USWDS!

mherchel:
Good catch! Just pushed up a fix. Here's a gif of the new focus states in action:
footer-focus

mherchel:
> It also looks like the first secondary link is pushed right up against the button. Could you add some space between them when in the smallest mobile view for a little bit of separation?

Just realized that I missed that comment. This is now fixed in the commit that I just pushed.
footer-focus2

mahoneycm:
Both updates look great! Loving these gif examples

mahoneycm (approved):
Thanks for your work on this Mike!

amyleadem (changes requested):
Thank you for this contribution @mherchel ! This is a very helpful improvement to the accessibility of our footer. 

On review, I discovered a couple items that need   to be addressed prior to merge: 
**Issue 1 - aria-expanded attribute doesn't always accurately reflect current state**
Items affected: footer__primary-link 
Window width: < mobile-lg
 Issue: When an expanded item is collapsed via the user opening a second item, the aria-expanded attribute on the now-collapsed item remains set to "true".
 Expected: Collapsing an expanded item should update the aria-expanded attribute to false.
 
** Issue 2 - Cursor displays with a pointer on non-clickable headers**
   Items: footer__primary-link 
   Window width: > mobile-lg
    Issue: Cursor displays as a pointer despite being a non-actionable item. 
     Expected: Cursor should be set to auto on non-actionable items
 
**Issue 3 - Inaccessible link items at transition width**
Item: footer list--unstyled
Window width: 480px
Issue: The user cannot access secondary link list items at exactly 480px. The footer structure changes but the list items remain hidden with no mechanism for exposing the links. 
image 
Expected: The link list items should be accessible to the user at all window widths.

Thanks again for all your work on this. Please let me know if you have any questions.

mherchel:
Thanks for the review! I'll be able to get back into this next week at the earliest. Thanks again,

thisisdano:
Any update on this one?

mherchel:
Not yet, although it's still on my `@todo` list. I should have some time this upcoming sprint though :)

mherchel:
Alrighty... 

### Issue 1 - aria-expanded attribute doesn't always accurately reflect current state
Good catch on this! This was happening when another menu opened (which in turn closes the original menu item).

This is fixed. To prevent this from happening again, and to make the code a bit more resilient, I refactored the CSS to style against the `aria-expanded` state of the `<button>` element, instead of the `hidden` CSS class of the parent `<section>` (which seemed weird to me TBH).

###  Issue 2 - Cursor displays with a pointer on non-clickable headers
This is actually a pre-existing issue (see https://designsystem.digital.gov/components/footer/#big-footer). However, this is a simple fix and I've taken care of it in this PR.

### Issue 3 - Inaccessible link items at transition width
I can't reproduce this one! I've tried both Chrome and FF. Can you still reproduce this? If so, can you give me more info on how to do so? 

Animated GIF 
footer-480
showing what I've done:

amyleadem (changes requested):
Thank you, @mherchel! I apologize for my delayed response. Everything is looking great. 

One request - can you update your branch with the latest updates from the `develop` branch? We merged some tweaks in [PR #4551](https://github.com/uswds/uswds/pull/4551) related to Issue 3 and I wanted to make sure they are included. 

Other than that, all looks good to me.

Apologies again for the delay. We appreciate your efforts!

mherchel:
Thanks! 

> can you update your branch with the latest updates from the develop branch? 

Just updated my fork. Tests are still passing on my local so 🤞

src/js/components/footer.js
```diff
@@ -3,37 +3,65 @@ const select = require("../utils/select");
...
+    const currentElementClasses = currentElement.getAttribute('class');
+
+    // Create the new element
+    const newElement = document.createElement(newElementType);
+    newElement.setAttribute('class', currentElementClasses);
+    newElement.classList.toggle(`${PREFIX}-footer__primary-link--button`, isMobile);
+    newElement.textContent = currentElement.textContent;
+
+    if (isMobile) {
+      newElement.setAttribute('aria-expanded', 'false');
```
mejiaj, line 47: We would need to include the `aria-controls` attribute on the button that points to the ID of each list. 

We did something similar on the File Input component where we generate a unique ID. Here's an example https://github.com/uswds/uswds/blob/7349ddfc07f3814be51f6e6143fd16b0530771f8/src/js/components/file-input.js#L109-L110

Probably best to create a new function for that and call it on `init()`.

---

We should also remove these `aria` attributes (expanded/controls) on large screens.

src/js/components/footer.js
```diff
@@ -3,37 +3,65 @@ const select = require("../utils/select");
...
 const toggleHidden = (isHidden) =>
-  select(COLLAPSIBLE).forEach((list) =>
-    list.classList.toggle(HIDDEN, isHidden)
+  select(BUTTON).forEach((button) =>
+    button.setAttribute('aria-expanded', !isHidden)
   );
 
-const resize = (event) => toggleHidden(event.matches);
+const resize = (event) => {
+  toggleHtmlTag(window.innerWidth < HIDE_MAX_WIDTH);
```
mejiaj, line 62: Would `event.matches` work for these function as well?

src/js/components/footer.js
```diff
@@ -3,37 +3,65 @@ const select = require("../utils/select");
...
+  const newElementType = isMobile ? 'button' : 'h4';
+
+  primaryLinks?.forEach(currentElement => {
+    const currentElementClasses = currentElement.getAttribute('class');
+
+    // Create the new element
+    const newElement = document.createElement(newElementType);
+    newElement.setAttribute('class', currentElementClasses);
+    newElement.classList.toggle(`${PREFIX}-footer__primary-link--button`, isMobile);
+    newElement.textContent = currentElement.textContent;
```
mejiaj, line 44: If the new element is button outside of a form, we should add `type="button"` from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type).

mejiaj (changes requested):
Thank you for your contribution, and even more for your patience! It's very much appreciated.

Overall it looks great, added some comments. In the past we've had a11y issues pop up for using `aria-expanded` without an `aria-controls` on the same element. Adding those IDs and then referencing them would help improve a11y.

Besides that this PR is in good shape.

mherchel:
Thanks! These changes are pretty straightforward. I'm kinda slammed this week, so it'll likely be next week when I get to them :)

src/js/components/footer.js
mherchel: ✅

src/js/components/footer.js
mherchel: Yes it does!

src/js/components/footer.js
mherchel: > We would need to include the aria-controls attribute on the button that points to the ID of each list.

This is done. I ended up using `Math.random()` instead of `Date` because the date wasn't generating different values. 

```
 const menuId = `${PREFIX}-footer-menu-list-${Math.floor(Math.random() * 100000)}`;
```

> We should also remove these aria attributes (expanded/controls) on large screens.

Good catch on that. This is done.

mherchel:
This should be good to go! Let me know if there are any more changes requested!

src/js/components/footer.js
```diff
@@ -1,39 +1,67 @@
...
-    list.classList.toggle(HIDDEN, isHidden)
-  );
+/**
+ * Swaps the <h4> element for a <button> element (and vice-versa) and sets id
+ * of menu list
+ *
+ * @param {Boolean} isMobile - If the footer is in mobile configuration
+ */
+function toggleHtmlTag(isMobile) {
+  const footer = document.querySelector(SCOPE);
```
mejiaj, line 34: This runs on all variants, let's limit it to the big footer variant only.

```suggestion
  const bigFooter = document.querySelector(SCOPE);

  if (!bigFooter) {
    return;
  }
```

mejiaj (changes requested):
@mherchel just a small issue, otherwise LGTM!

src/js/components/footer.js
mherchel: Good catch. This is done!

mherchel:
All changes are complete. Please re-review!

mejiaj (approved):
Looks good, thanks for making those changes!

mahoneycm (approved):
Looks great! 

Might be able to get rid of the `!important` tag by setting the `$chevron-icon-defaults` "color" to `buttonText` in the future

mherchel:
image

mherchel:
caf5608a-67ec-4f9f-acb5-db0052c33bed

PR #4484, opened 2022-01-28
Beta sass modules - fix module loop in function color()

amyleadem:
To prevent module looping, I rebuilt aspects of the function color() so that it points to assignments-theme-color instead of shortcodes-color-project (which is dependent on the color function). 

This is still unrefined and I don't fully understand the ramifications of the change on the system, but figured it could be a good conversation starter. 

Note: There is a chance the 'set-theme' and 'no-warn' flags were essential for the theme colors to be built correctly, and this scratches those right out. 

[Color function success criteria](https://github.com/uswds/uswds/pull/3258)


mejiaj:
Closing this as we decided to use internal function to set theme color.

PR #4487, opened 2022-01-29
Icon: Remove inline style in GitHub icon to fix CSP error

aduth:
## Description

This pull request seeks to resolve an error which can occur when using the USWDS Icon component SVG sprite in combination with a [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP), where the presence of an inline style tag within the GitHub icon can violate most common CSPs which do not include the `unsafe-inline` `style-src` directive.

image

## Additional information

The process for resolving this was to simply drop the SVG markup into the [SVGOMG online SVGO optimizer tool](https://jakearchibald.github.io/svgomg/). The produced output minified the markup and inlined the `fill-rule` as an HTML attribute.

An alternative solution could be to simply remove the style tag (and accompanying `class` attribute) and instead add the equivalent [`fill-rule="evenodd"` attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-rule) to the `<path>` element.

---


aduth:
>@aduth could you please include these changes in these other files too?

Sure, assuming it's the same icon, I copied the new version to replace those as well in 169e948.

mejiaj (approved):
Great, thanks a lot!

PR #4501, opened 2022-02-07
Fix core color contrast calculation:

### LINKED ISSUE #4146: Color Contrast Not Calculating Correctly
### Describe the bug
Using the color "dhs-green", #5e9732, color contrast requires the text be black, but USWDS is displaying white text for green buttons

1. Go to https://www.dhs.gov/xlibrary/dhsweb/_site/components.html
2. Scroll down to Buttons
3. Under "Secondary" see green buttons. Default button displays as "dhs-green" which should have black text, but shows white text

### Expected behavior
When using "dhs-green" as a background, USWDS should show black text

### Screenshots
Add screenshots to help provide context, if applicable.

### System setup
- **USWDS version:**
- **Device:**
- **Operating system:**
- **Browser and version:**

### Additional context
Add any other context about the problem.

amyleadem:
As outlined in issue #4146, the get-color-token-from-bg() function is returning a color token that does not meet accessibility standards. 

Problem: Colors that receive the color-grade of 45 fail to meet the WCAG magic number of 50 with the theme defaults (white, gray-90).

Solution: Altered the default fallback token from gray-90 to black.

Resolves #4146.

### DIFF
--- a/spec/unit/sass/tests.scss
+++ b/spec/unit/sass/tests.scss
@@ -408,12 +408,12 @@ $this-function: "get-color-token-from-bg()";
 @include describe("[function] #{$this-function}") {
   @include it("Returns the default from light bg.") {
     $test: get-color-token-from-bg("white");
-    $expect: $theme-text-color;
+    $expect: "black";
     @include assert-equal($test, $expect);
   }
   @include it("Returns the default reverse from dark bg.") {
     $test: get-color-token-from-bg("ink");
-    $expect: $theme-text-reverse-color;
+    $expect: "white";
     @include assert-equal($test, $expect);
   }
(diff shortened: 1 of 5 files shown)


amyleadem:
@mejiaj 

I updated the defaults in the function itself rather than updating the values in $project-defaults. Going this route might leave outstanding issues in these items: 
@function get-link-tokens-from-bg
@mixin set-link-from-bg
@mixin set-text-from-bg

Let me know if you think we should change the defaults in these as well or in $project-defaults.

amyleadem:
Closing this PR because it was edited inside the 2.x structure 
[New PR for 3.0](https://github.com/uswds/uswds/pull/4687) ➡️

PR #4468, opened 2022-01-19
USWDS - Add package exports for fonts and img src directories

pearl-truss:
## Description

After upgrading our project that uses USWDS to react-scripts v5.0.0 which uses Webpack 5. The following errors surfaced when trying to build our app

`
ModuleNotFoundError: Module not found: Error: Package path ./src/fonts/roboto-mono/roboto-mono-v5-latin-300.woff2 is not exported from package /Users/pearl/Projects/mymove/node_modules/uswds (see exports field in /Users/pearl/Projects/mymove/node_modules/uswds/package.json)
`

and

`
ModuleNotFoundError: Module not found: Error: Package path ./src/img/correct8.svg is not exported from package /Users/pearl/Projects/mymove/node_modules/uswds (see exports field in /Users/pearl/Projects/mymove/node_modules/uswds/package.json)
`

The changes in this PR resolves both issues

PR #4461, opened 2022-01-14
USCIS: Export other artifacts

tyduptyler13:
Webpack 5 enforces correct export rules and thus prevents importing things like uswds.css.

## Description

Upgrading to webpack 5 in our project caused our import of the css to break. The proposed change allows us to continue using this project without a fork/hack/or holding back a webpack upgrade.

## Additional information

I don't have the webpack change on hand, but if I run across it again i'll update this ticket.

- [x] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [x] Run `npm test` and make sure the tests for the files you have changed have passed.
- [x] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [x] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #4502, opened 2022-02-07
3.0 beta: export artifacts for use by Webpack 5

jkjustjoshing:
## Description

PRs #4468 and #4461 add exports to package.json so Webpack 5 can access font/img/etc assets. These changes are missing in the 3.0 betas

## Remaining issue

The structure of the `./src` directory changed from v2 to v3, which is causing the `./src/js/components` exports in the package.json file to not point at anything. This issue still needs to be resolved, but will require a little discussion in a ticket. EDIT ticket has been created #4507

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,14 @@
       "default": "./dist/js/uswds.min.js"
     },
     "./src/js/components": "./src/js/components/index.js",
-    "./src/js/components/*": "./src/js/components/*.js"
+    "./src/js/components/*": "./src/js/components/*.js",
+    "./src/fonts/*": "./src/fonts/*",
+    "./src/img/*": "./src/img/*",
+    "./js/*": "./dist/js/*",
+    "./css/*": "./dist/css/*",
+    "./scss/*": "./dist/scss/*",
+    "./img/*": "./dist/img/*",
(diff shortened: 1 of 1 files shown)


jkjustjoshing:
Converting this to draft - I realize that the Javascript file structure has changed, and I'll need to update that before submitting the PR

package.json
```diff
@@ -24,7 +24,14 @@
       "default": "./dist/js/uswds.min.js"
     },
     "./src/js/components": "./src/js/components/index.js",
-    "./src/js/components/*": "./src/js/components/*.js"
+    "./src/js/components/*": "./src/js/components/*.js",
```
mejiaj, line 27: Based on your comment [here](https://github.com/uswds/uswds/issues/4507#issuecomment-1040781754) does that mean you still want this missing reference to be included?

jkjustjoshing: > this missing reference

If you're referring specifically to the Javascript reference, no, that's less urgent at the moment. It would be needed for reducing the amount of unused Javascript included in our bundles (manual tree shaking), but not impacting our ability to get things built/shipped.

It's primarily the image and font assets that this is needed for.

mejiaj (approved):

Issue #4419, opened 2021-12-06
Copy change suggestion for the Banner component for improved accessibility using screen readers


AmandaD-USDS:
## Problem
The USDS team that is working to improve the [Child Tax Credit site](https://www.whitehouse.gov/child-tax-credit/) has been running some usability testing with people with disabilities. Some of these participants included those using a screen reader. We discovered that, when users are listening to a screen reader (these are often run at very fast speed), the copy on the USWDS banner component becomes confusing for them.

Because of this, we would like to suggest a slight change to the banner copy.

The reason for this suggested change is that, when the screen reader reads the copy, it comes out sounding like "Unofficial website of the United States government". This led one participant to comment,  "if this is the unofficial website, I wonder where the official websites are?". So we think this could be a positive change.

## Suggested Solution
We suggest a small change to the copy on the banner component to make it clearer for accessibility using a screen reader.

**Current copy:** <br>
An official website of the United States government.

**Suggested copy:**<br>
This is a United States government website.

## Alternatives
Any copy change that would not be misinterpreted by a screen reader to make the website sound "unofficial".

## Additional Context
This is the component we are referring to: https://designsystem.digital.gov/components/banner/

brunerae:
Decision at this time is to update the aria-label to remove "an" from the read-out so that it reads "official" instead of "an official".

PR #4512, opened 2022-02-10
Banner: Update banner copy and spec (#4419)

wpears:
## Description

 Updates the banner text to the more accessible text suggested in #4419

## Additional information
 - [The identifier component](https://designsystem.digital.gov/components/identifier/) is unchanged in this PR. The more flexible structure of that component makes the suggested text somewhat inappropriate, as the article preceding the `<Parent agency>` depends on the parent agency content, eg, `This is a United States Government website` vs `This is an Environmental Protection Agency website`.
 - One possible alternative is simply dropping the leading `An` from the original copy, ie, `Official website of the United States Government`. This would allow the same copy to exist for both `Banner` and `Identifier`.

### DIFF
--- a/spec/unit/banner/template.html
+++ b/spec/unit/banner/template.html
@@ -6,7 +6,7 @@
                     <img class="usa-banner__header-flag" src="../../dist/img/favicons/favicon-57.png" alt="U.S. flag">
                 </div>
                 <div class="grid-col-fill tablet:grid-col-auto">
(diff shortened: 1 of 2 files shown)


ethangardner:
This PR was superseded by https://github.com/uswds/uswds/pull/4925, which has been merged. The team appreciates the contribution but this will be closed since there's an alternate solution in place.

PR #4591, opened 2022-03-18
Add Public Sans Latin subset as font stack token

### LINKED ISSUE #4542: Default fonts should be subset consistently
### Describe the bug

USWDS currently ships with 4 default fonts: Merriweather, Public Sans, Roboto Mono, and Source Sans Pro. Of these, all except "Public Sans" are subset to include only Latin characters†. USWDS default fonts should be subset consistently, so that I am not surprised by differences in character availability when switching between fonts, and so that fonts are optimized if only certain subsets are intended to be supported by default.

†: Source Sans Pro is a little less clear on subset, and may be custom per https://github.com/uswds/uswds/pull/2934#issuecomment-472234616.

1. Download Public Sans as shipped through `uswds` npm package: https://unpkg.com/uswds@2.13.1/dist/fonts/public-sans/PublicSans-Regular.ttf
2. Go to https://opentype.js.org/glyph-inspector.html
3. In the file picker, choose the file downloaded in Step 1

**Expected:** I see a few hundred glyphs at most.
**Actual:** I see 600+ glyphs.

### System setup
- **USWDS version:** v2.13.1

### Additional context

- This was noticed as the result of performance difference when changing font to Public Sans on Login.gov, where previous fonts Source Sans Pro and Merriweather were on-average smaller than Public Sans (~20kb vs. ~32kb).
   - Slack context (TTS internal): https://gsa-tts.slack.com/archives/C0NGESUN5/p1645821094190849?thread_ts=1645133006.512749&cid=C0NGESUN5
- Per closure of #756, Latin may be only set intended to be supported by default?

aduth:
Closes #4542

## Description

This pull request adds a Public Sans Latin subset.

## Additional information

As mentioned in #4542, the purpose here is to offer a performant alternative for sites which don't need the full set of glyphs from the original Public Sans font. This is also aimed at reconciling some inconsistencies across built-in fonts, where most fonts already subset to Latin characters by default.

This is a conservative solution which aims to retain backward-compatibility and offer consumers the choice between the full font file vs. a common subset, without impacting existing usage. An alternative solution could be to change the existing font files to be subset-by-default.

This has a noticeable impact on performance, **reducing size of font files by ~35%**. Real-world measurements show **savings of 23.5kb** for a site using `400` (normal) and `700` (bold) font weights ([source](https://github.com/18F/identity-idp/pull/6094)).

These fonts were generated from the `.ttf` source files using using [glyphhanger](https://github.com/zachleat/glyphhanger):

```
glyphhanger --LATIN --formats=woff2,woff --subset=*.ttf
```

Open questions:

- [ ] This does not include `.ttf` file versions of the files. Should it, for consistency?
   - Related: Should USWDS bother defining `@font-face` for formats other than WOFF, considering [browser support](https://caniuse.com/woff)?
- [ ] Should files exist in a separate `fonts/public-sans-latin` folder, instead of together with non-subset font files?

---


aduth:
To improve broad support, it could be a good idea to ship different font subsets, and reference them in `@font-face` declarations using [`unicode-range`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range), so that the browser could optimize which file subsets should be loaded.

ryparker:
Thanks for the Public Sans subset proposal, and sorry it went without a review. I'm closing this during the backlog cleanup because we don't have capacity to review it now.

If you'd like to continue, comment here or reopen the PR. I'll help with the review and any rebase needed to pick it back up.

PR #4615, opened 2022-04-04
Pin dependencies to exact versions only

thisisdano:
Pin dependencies to exact versions only to give the project and team more control about which dependencies load on an `npm install`.


mejiaj (commented):
No breaking changes with this approach. 

The only concern would be an increase in size to `node_modules` if users have the same package, but different versions in their projects.

Alternatively, we could try:
1. Pinning `dependencies` only and keeping the carat for devDependencies
1. Rely on the `~` to only use patch versions

mejiaj (approved):

PR #4634, opened 2022-04-18
Update to Beta 5 and remove unused cover scripts

thisisdano:
We no longer use CodeClimate or measure code coverage as part of our CI. This PR removes the cover scripts and the `nyc` dependency. It also updates the current Beta version to `5`. Finally, it pins dependencies to specific versions, as we do in the current 2.x branch.


amyleadem:
@thisisdano Do we still need the [.nycrc](https://github.com/uswds/uswds/blob/library--develop/.nycrc) and reference to [`.nyc_output` in .gitignore](https://github.com/uswds/uswds/blob/library--develop/.gitignore#L12) after removing the `nyc` dependency?

amyleadem:
@thisisdano Another thought that might be out of scope - we have some devDependencies inside the [uswds-core/package.json](https://github.com/uswds/uswds/blob/dw-update-beta/packages/uswds-core/package.json#L28-L29). Do we want to pin those as well?

thisisdano:
No, I believe we can remove them. Will do.

thisisdano:
Let's leave the current dependencies in `uswds-core` for now, since we're not yet using them.

amyleadem:
Ok, I believe just one last note: I noticed some references to the `cover` task in gulpfile.js
https://github.com/uswds/uswds/blob/dw-update-beta/gulpfile.js#L11
https://github.com/uswds/uswds/blob/dw-update-beta/gulpfile.js#L39
https://github.com/uswds/uswds/blob/dw-update-beta/gulpfile.js#L46

And some references to `coverage` (if related):
https://github.com/uswds/uswds/blob/dw-update-beta/.gitignore#L11
https://github.com/uswds/uswds/blob/dw-update-beta/.npmignore#L7

amyleadem (approved):

amyleadem:
Looks good to me!

PR #4683, opened 2022-05-09
Normalize SVG icons with SVGO

### LINKED ISSUE #4306: Remove unnecessary cruft in SVGs
This is all bundled in this PR, but we really should have our SVGs be optimized and consistently formatted.

https://github.com/uswds/uswds/pull/4304

I don't know if `data-prefix="fas" data-icon="angle-double-left”` or `class="svg-inline--fa fa-angle-double-left fa-w-14"` are documented anywhere. Not sure why this info is doing in a SVG file. May not be useful, but might just cause some quirky behavior too.  

I don't know if it is still considered a best practice in SVG circles to use `<?xml version="1.0" encoding="UTF-8”?>` but it isn't actually necessary. 

Are there places where we actually want to use `fill-rule=“evenodd" in these simple SVGs? 

Might even be possible to re-use more of these SVGs. They are way more flexible than PNGs so maybe we could use less of them and just provide different color instructions via CSS.

thisisdano:
This uses the [SVGO](https://github.com/svg/svgo) defaults to normalize and compress all USWDS SVGs. Our config maintains both the `height`/`width` and `viewBox`. 

- Added `height="24"` and `width="24"` for some icons that were missing it
- Updated all icons with new config
- Install `svgo@2.8.0` as a dependency

Fixes #4306


mejiaj (approved):
Icons look good and I don't see visual problems with the icons, how the spritesheet is compiled, or using icons from sprite.

I did notice some icons had their ID's removed. This probably won't be a problem since the IDs seemed to be inconsistently used.

amyleadem:
This issue appears to have existed prior to this PR, and not sure if it can/should be resolved here, but wanted to flag that `navigate_far_before` and `navigate_far_after` do not appear to accept CSS control of `fill`, unlike the other icons:

image

PR #4708, opened 2022-05-23
USWDS - Header: Increase navigation link target size

### LINKED ISSUE #2836: 2.0 Feature Request - Make dropdown navigation behavior similar to mobile and side nav

mahoneycm:
## Preview

[Header →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/cm-navigation-link-target-3.0/?path=/story/components-header--default)

## Description

Fixes #2836

Originally #4427

Modeled some of the suggestions in #2836 and #3033 which give subnav menu items a larger targeting area.

The desired effect was accomplished by setting the submenu item anchor tag to `display: block;` and moving the padding from the parent tag to the anchor itself.

## Additional information

Current Nav menus

image
image

Suggested on #2836 
image

After this fix

Screen Shot 2022-05-23 at 12 07 59 PM
Screen Shot 2022-05-23 at 12 08 15 PM

### DIFF
--- a/packages/usa-nav/src/styles/_usa-nav.scss
+++ b/packages/usa-nav/src/styles/_usa-nav.scss
@@ -375,7 +375,6 @@ $expand-less-icon: map-merge(
     @include add-list-reset;
     background-color: color("primary-darker");
     width: units("card-lg");
-    padding: units(2);
     position: absolute;
     z-index: z-index(400);
   }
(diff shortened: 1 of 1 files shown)


mejiaj (approved):
Looks good, thanks for updating!

PR #4766, opened 2022-06-03
USWDS - Select: Include standard "multiple" stylings. Resolves #4423

### LINKED ISSUE #4423: The select element is not rendered properly when "multiple" is specified
### Describe the bug
According to the [HTML 5 standard](https://html.spec.whatwg.org/multipage/rendering.html#the-select-element-2), a select element whose `multiple` attribute is present is "expected to render as an `inline-block` box whose `height` is the height necessary to contain as many rows for items as given by the element's `display size`, or four rows if the [size] attribute is absent." This standard behavior has the benefit of making it obvious that multiple selections can be made and users have come to expect and depend on this visual clue. All of the major browsers implement this behavior by default. The USWDS component system suppresses this behavior, ignoring the `size` attribute as well as the standard's instructions for the `height` of the box in the absence of that attribute.

1. Install the latest USWDS package according to the [documentation's instructions](https://github.com/uswds/uswds#download-and-install)
2. Create a minimal HTML page using the example in those instructions
3. Add a `select` element with `class="usa-select"` to the page with the `multiple` and `size` attributes set
4. Bring up the page in any major browser on any desktop operating system

### Expected behavior
The `size` attribute of the element is honored (whether explicitly provided or implied by the standard's default).

### Screenshots
What the field looks like following the instructions above:
image

What the field looks like with `class="usa-select"` removed:
image

### System setup
- **USWDS version:** 2.12.2
- **Device:** any desktop or laptop computer
- **Operating system:** Linux, Windows, MacOS
- **Browser and version:** Any supported version of any browser supported by USWDS

### Additional context
Here is the repro HTML with `class="usa-select"` applied.
```html 
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
    <title>My Example Project</title>
    <script src="assets/uswds-2.12.2/js/uswds-init.min.js"></script>
    <link rel="stylesheet" href="assets/uswds-2.12.2/css/uswds.min.css" />
  </head>
  <body>
    <div class="grid-container">
      <label class="usa-label" for="repro" id="repro">Repro Select</label>
      <select class="usa-select" id="repro" name="repro" multiple size="6">
        <option value="value1">Option A in a multi-select field</option>        
        <option value="value2">Option B in a multi-select field</option>        
        <option value="value3">Option C in a multi-select field</option>        
        <option value="value4">Option D in a multi-select field</option>        
        <option value="value5">Option E in a multi-select field</option>        
        <option value="value6">Option F in a multi-select field</option>        
        <option value="value7">Option G in a multi-select field</option>        
        <option value="value8">Option H in a multi-select field</option>        
        <option value="value9">Option I in a multi-select field</option>        
      </select>
    </div>
    <script src="assets/uswds-2.12.2/js/uswds.min.js"></script>
  </body>
</html>
```
The suggestion in the documentation that checkboxes be considered instead of a multi-select box is not applicable or appropriate for `select` elements with dozens and dozens of options, as the user experience for such an approach would be unacceptably horrible. I have cut down this repro case to only a handful of options, in the spirit of keeping repro cases as small as possible, but it is not difficult to find real-world cases where a select field has too many options for the checkbox solution, and we have any number of such use cases in our systems at the National Cancer Institute. The HTML 5 standard provides a solution for this use case which has worked well for decades. Let's support what the standard calls for here.

mahoneycm:
## Preview

[Select →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/cm-select-multiple/?path=/story/components-form-inputs-select--select)

## Description

Resolves #4423 

> According to the [HTML 5 standard](https://html.spec.whatwg.org/multipage/rendering.html#the-select-element-2), a select element whose multiple attribute is present is "expected to render as an inline-block box whose height is the height necessary to contain as many rows for items as given by the element's display size, or four rows if the [size] attribute is absent." This standard behavior has the benefit of making it obvious that multiple selections can be made and users have come to expect and depend on this visual clue. All of the major browsers implement this behavior by default. The USWDS component system suppresses this behavior, ignoring the size attribute as well as the standard's instructions for the height of the box in the absence of that attribute.

This PR

- Adds expected styling for `select` when the multiple attribute is present
- Sets default styling as described by `HTML 5 standard`
- Adds storybook controls to enable the `multiple` attribute and set `size` attribute
- If `size = 0` it will act as if the `size` attribute is absent and do the standard default of four rows.

## Additional information

Standard select: 

image

Multiple Enabled:

image

New Storybook controls:

image


amyleadem (changes requested):
Looks like this is working well! 
A couple of thoughts:
- We should have a separate story in Storybook for when `multiple` is true
- The padding on the select box can cause text to get cut off if the `size` is less than the item count. It might be better to add the padding to the options when `multiple` is true.
image


- Additionally, we have this note in our [docs](https://designsystem.digital.gov/components/select/) for select:
    "When to consider something else: Multi-select. If you need to allow users to choose more than one option at once. Users often don’t understand how to choose multiple items from select elements. Use checkboxes instead." 
    - Are we now allowing multiple select as a recommended practice? If this is now an acceptable pattern, we should remove this line, provide use-case recommendations, and should add some notes about both `multiple` and `size` in our documentation. 
    
mahoneycm:
Thanks @amyleadem!

- Having a second story could be a good idea! I feel like it also depends on if we want this advertised as an option or not, because of the checklist note you mentioned
- I'm open to suggestions on how to handle the padding. I initially left it because when it is missing, it becomse a little less clear that you can scroll to see more options (i.e. if size is not set and it is default to 4 rows, but there are more than four options):  
image

_(of course the scroll bar is a good indicator)_

- I also asked Dan about the guidance re: checkboxes but the issue is that the multiple select is a native functionality that the HTML 5 standards require. Dan said the following: "So yes, I think that in most instances, checkboxes are preferable to multi-select.  Multi-select is bad UI, I’d say, like 99.99% of the time. That said, native multi-select should work with USWDS and should probably work as the user suggests."

mejiaj:
>The padding on the select box can cause text to get cut off if the size is less than the item count. It might be better to add the padding to the options when multiple is true.

I think the padding is coming from the default input styles. It seems expected to leave it in by default for this. 

I agree with Dan's comments. We're not recommending this as a new pattern, we're just making sure USWDS styles don't break the expected behavior. We also don't have research to suggest otherwise, so we can leave guidance as-is. Good consideration though!

I agree with @amyleadem, a separate story would be better.

packages/usa-select/src/usa-select--multiple.twig
```diff
@@ -0,0 +1,13 @@
...
+    <option value>- Select -</option>
+    <option value="value1">Option A</option>
+    <option value="value2">Option B</option>
+    <option value="value3">Option C</option>
+    <option value="value1">Option E</option>
+    <option value="value2">Option F</option>
+    <option value="value3">Option G</option>
+  </select>
+</form>
+
```
mejiaj, line 13: Can we move this to a directory `usa-select/test/test-patterns`?

Similar to what we did with this component test template
https://github.com/uswds/uswds/blob/1db0e68d7e5cf21ae52f2d8a29d5963c7aa1b681/packages/usa-combo-box/src/test/test-patterns/test-usa-combo-box.twig

packages/usa-select/src/usa-select.json
```diff
@@ -0,0 +1,4 @@
+{
+  "multiple": true,
+  "size - (defaults to 4 when undefined or 0)": 0
+}
```
mejiaj, line 4: Let's move this too or just add it to the new test template inline.

mejiaj (changes requested):
Thanks Charlie, I've added some feedback.

packages/usa-select/src/test/test-patterns/usa-select--multiple.twig
```diff
@@ -0,0 +1,13 @@
...
+    <option value>- Select -</option>
+    <option value="value1">Option A</option>
+    <option value="value2">Option B</option>
+    <option value="value3">Option C</option>
+    <option value="value1">Option E</option>
+    <option value="value2">Option F</option>
+    <option value="value3">Option G</option>
+  </select>
+</form>
+
```
mejiaj, line 13: Can you double-check the empty lines at end of file? Locally shows 2 empty lines, should be 1.

packages/usa-select/src/usa-select.stories.js
```diff
@@ -1,9 +1,15 @@
 import Component from "./usa-select.twig";
+import TestMultiple from "./test/test-patterns/usa-select--multiple.twig";
```
mejiaj, line 2: Let's just rename this to Test for now.

packages/usa-select/src/usa-select.twig
```diff
@@ -7,3 +7,4 @@
     <option value="value3">Option C</option>
   </select>
 </form>
+
```
mejiaj, line 10: This one also shows an extra line at end.

image

mejiaj (changes requested):
Tiny changes, otherwise LGTM!

mejiaj (approved):

amyleadem (approved):
Looks good to me!

PR #4872, opened 2022-07-20
Update social icons

thisisdano:
**Add LinkedIn icon.** We now have a LinkedIn icon included in our default icon sprite as `linkedin`.

**Use unmodified social media icons.** We now provide simpler social media icons, and removed any decoration not in the original icon. This means that the new icons are not in an enclosing circle unless the original icon is enclosed in a circle.

## Problem statement
Our social media icon set was missing the LinkedIn icon. Additionally, the social media icons we delivered were optimized for consistent presentation in our footer, inside a circular button. Since our footer is not the only place social media icons might be used, we should remove the circular outline around the images, since we do not want to imply that the circle is part of the icon itself.

## How to test
- Open the Icons component of our Storybook site and look for the `linkedin` icon
- In the Icons component of our Storybook site, check the following icons to assure that they do not have a circular button shape. Note that the Facebook and GitHub are circular and have not changed. 
  - Flickr
  - Instagram
  - Twitter
  - YouTube

### New Instagram icon in our Icons component
Screen Shot 2022-07-20 at 10 43 36 AM

### All social icons in the footer
Note that this a is just an example. This PR does not add social media icons to the footer.

Screen Shot 2022-07-20 at 10 34 21 AM


package.json
```diff
@@ -67,7 +67,7 @@
     "test:unit": "gulp unitTests",
     "version": "gulp release",
     "watch": "gulp watch",
-    "fix:icons": "npx svgo --folder ./packages/usa-icon/src/img/uswds-icons -q && npx svgo --folder ./packages/usa-icon/src/img/material-icons -q && npx svgo --folder ./packages/usa-icon/src/img/usa-icons-bg -q && npx svgo --folder ./packages/usa-icon/src/img/usa-icons -q",
+    "fix:icons": "npx svgo --folder ./packages/usa-icon/src/img/uswds-icons -q && npx svgo --folder ./packages/usa-icon/src/img/usa-icons-bg -q",
```
thisisdano, line 70: I also updated this script so it points to the proper directory

mejiaj (commented):
No issues with code, but the display on preview looks a jagged on OSX - Edge

image

Looks good in Firefox.

image

amyleadem (approved):
Looks good to me!

I checked the display of the new icons on the following browsers and all are presenting with a quality similar to previous versions:
- iOS safari
- Chrome
- Firefox
- Safari

PR #4884, opened 2022-07-27
Jb inclusive patterns input mask

jonathanbobel:
## Input Mask component

[Federalist Storybook - Input Mask] (https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/jb-inclusive-patterns-input-mask/?path=/story/components-form-inputs-text-input-mask--input-mask)
[Federalist Content Site - Input Mask] (https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.app.cloud.gov/preview/uswds/uswds-site/jb-inclusive-patterns-input-mask/components/input-mask/)
-->

## Description

Code for Input Masks - making it easier for users to fill out input content that is easily attached to a regex pattern
- Creates an array from all inputs with class="masked", placeholder and pattern attributes
- Code will look to pattern attribute to determine maxlength to limit input
- Creates mask overlay that when typing removes the characters and replaces them with the true input value
- Can be used in conjunction with validation, as it doesn't alter the actual contents of the input

## Additional information

Patterns include
* SSN
* US Telephone Number
* US Zip Code

This code can also be extended to additional Input Masks if the user is familiar with regex


jonadecker:
In the Federalist preview, the phone is [screen]reading as invalid data? This isn't happening on SSN or ZIP so I don't think it's an onFocus issue.

I'm also wondering whether this script allows you to choose not display pre-fill text? Our patterns advocate for not displaying pre-fill text because it usually disappears on entering an input field. That isn't true in this case, so I'm not sure it's a problem, it's just something we need to at least mention in the supporting patterns.

I do think we should also show hint text in the preview, even if the mask is displaying pre-fill values in the inputs, both as an affordance and to align with pattern writeups.

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,162 @@
...
+        input.setAttribute("data-placeholder", placeholder);
+        input.removeAttribute(`${PLACEHOLDER}`);
+      } else {
+        return;
+      }
+
+      // This section still needs some rework
+      text = `<span class="${SHELL}"><span aria-hidden="true" id="${input.id}Mask"><i></i>${placeholder}</span>${input.outerHTML}</span>`;
+
+      input.outerHTML = text; // eslint-disable-line no-param-reassign, no-unsanitized/property
```
mejiaj, line 54: We'll need to sanitize if it's taking user input. You can view an example https://github.com/uswds/uswds/blob/cda1459efd1c028fce62c960eafc2653993654f9/packages/usa-date-picker/src/index.js#L871-L877

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,162 @@
...
+      return [theIEl, placeholderVal];
+    },
+
+    // add event listeners
+    activateMasking(inputs) {
+      let i;
+      let l;
+
+      for (i = 0, l = inputs.length; i < l; i += 1) {
+        masking.maskedInputs[i].addEventListener(
```
mejiaj, line 73: Using the event delegation library in behaviors let's you add this at the end. Look at other example like date picker

https://github.com/uswds/uswds/blob/cda1459efd1c028fce62c960eafc2653993654f9/packages/usa-date-picker/src/index.js#L2113-L2120

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,162 @@
...
+        case 18: // option
+        case 16: // shift
+        case 37: // arrow keys
+        case 38:
+        case 39:
+        case 40:
+        case 9: // tab (let blur handle tab)
+          return;
+        // no default
+      }
```
mejiaj, line 100: Alternatively, there's a keymap functino you could use. See date picker example:

https://github.com/uswds/uswds/blob/cda1459efd1c028fce62c960eafc2653993654f9/packages/usa-date-picker/src/index.js#L2127-L2143

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,162 @@
...
+      let isInt;
+      let isLetter;
+
+      // strip special characters
+      const strippedValue = isCharsetPresent
+        ? value.replace(/\W/g, "")
+        : value.replace(/\D/g, "");
+
+      for (i = 0, j = 0; i < l; i += 1) {
+        // eslint-disable-next-line no-restricted-globals
```
mejiaj, line 127: What do `i` and `j` represent?

mejiaj (changes requested):
Overall the component functions well. Added some suggestions and questions below (mostly for code style and readability).

packages/usa-input-mask/src/styles/_usa-input-mask.scss
```diff
@@ -0,0 +1,48 @@
...
+    pointer-events: none;
+    color: color("gray-50");
+    font-family: family("sans");
+
+    i {
+      /* any of these 3 will work */
+      color: transparent;
+      opacity: 0;
+      visibility: hidden;
+    }
```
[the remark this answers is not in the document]
jonathanbobel: The <i> tag is rendered inside the mask and hidden as you type to convert the placeholder into the actual value of the input 

`const text = `<span class="${MASK}"><span class="${MASK_CONTENT}" aria-hidden="true" id="${input.id}Mask"><i></i>${placeholder}</span>${input.outerHTML}</span>`;` from index.js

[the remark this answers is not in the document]
jonathanbobel: https://github.com/uswds/uswds/blob/2652484b85dde97b5df330b1d422b033f676dba2/packages/usa-input-mask/src/index.js#L52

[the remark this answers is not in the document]
jonathanbobel: the <i> tag is being rendered, hidden and replaced by actual input content when the user enters information

jonathanbobel:
> 

Addressed the invalid data issue, and the hint area. The javascript here is looking for the placeholder (and classes applied) to know the length of the input, so it wouldn't work without the placeholder text overlaying. A user can always use the text-input component if they don't want this functionality, but we might have to update guidance to reflect.

jonathanbobel:
> In the Federalist preview, the phone is [screen]reading as invalid data. This isn't happening on SSN or ZIP so I don't think it's an onFocus issue.
> 
> I'm also wondering whether this script allows you to choose not to display pre-fill text? Our patterns advocate for not displaying pre-fill text because it usually disappears on entering an input field. That isn't true in this case, so I'm not sure it's a problem; it's just something we need to at least mention in the supporting patterns.
> 
> I do think we should also show hint text in the preview, even if the mask is displaying pre-fill values in the inputs, both as an affordance and to align with pattern writeups.

Addressed the hint and the invalid data screen reader issues

The javascript looks for the placeholder in order to render the mask and determine the input length, so if the user didn't want that functionality, they could use the existing text input component. We might have to update guidance to reflect this difference.

packages/usa-input-mask/src/index.js
jonathanbobel: I removed the key mapping here, it might have been overkill - I tested with VO screen reader + keyboard and I didn't notice a difference. We can add it back in if it's necessary.

packages/usa-input-mask/src/index.js
jonathanbobel: Followed Mitch's lead and used document.createElement(); for this section

packages/usa-input-mask/src/index.js
jonathanbobel: i represents the location in the mask of input + characters and spaces
j represents how many characters have been typed in a given input

There might be a better or cleaner way to go about this, could refactor at some point

packages/usa-input-mask/src/styles/_usa-input-mask.scss
```diff
@@ -0,0 +1,28 @@
...
+    color: color("gray-50");
+    font-family: family("sans");
+    border: 1px solid transparent;
+    i {
+      visibility: hidden;
+    }
+  }
+}
+
+usa-masked,
```
amyleadem, line 21: The class styles are currently not being applied to `.usa-masked`. Is this line needed?

packages/usa-input-mask/src/styles/_usa-input-mask.scss
```diff
@@ -0,0 +1,28 @@
...
+.usa-input-mask--content {
+  padding: units(1);
+  line-height: 1.4;
+  font-size: size('body', 'sm');
```
amyleadem, line 25: `.usa-input` typography is defined with the `block-input-general` placeholder, which uses form settings. Can we replace these typography declarations with the same placeholder?

```suggestion
@extend %block-input-general;
```

packages/usa-input-mask/src/styles/_usa-input-mask.scss
```diff
@@ -0,0 +1,28 @@
...
+  background-color: transparent;
+  text-transform: uppercase;
```
amyleadem, line 27: Do we need to force uppercase?

packages/usa-input/src/usa-input.twig
```diff
@@ -1,4 +1,4 @@
-<label class="usa-label" for="input-type-text">Text input label</label>
+<label class="usa-label test" for="input-type-text">Text input label</label>
```
amyleadem, line 1: What does `test` do?

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  const shell = document.createElement("span");
+  shell.classList.add(MASK);
+  shell.setAttribute("data-mask", placeholder);
+  console.log("shell ", shell); // eslint-disable-line no-console
+
+  const content = document.createElement("span");
+  content.classList.add(MASK_CONTENT);
+  content.classList.add("test");
```
amyleadem, line 34: What does `test` do?

packages/uswds-core/src/js/index.js
```diff
@@ -5,6 +5,7 @@ const comboBox = require("../../../usa-combo-box/src/index");
 const datePicker = require("../../../usa-date-picker/src/index");
 const dateRangePicker = require("../../../usa-date-range-picker/src/index");
 const fileInput = require("../../../usa-file-input/src/index");
+const inputMask = require("../../../usa-input-mask/src/index");
```
amyleadem, line 8: Can we put these in alphabetical order?

packages/uswds/_index.scss
```diff
@@ -47,6 +47,7 @@
 @forward "usa-tag/src/styles";
 @forward "usa-tooltip/src/styles";
 @forward "uswds-form-controls";
+@forward "usa-input-mask/src/styles";
```
amyleadem, line 50: I am thinking this should be included in uswds-form-controls instead. Can you move this `@forward` reference to packages/uswds-form-controls/_index.scss?

packages/usa-input-mask/src/usa-input-mask.twig
```diff
@@ -0,0 +1,42 @@
...
+    placeholder="XXX XX XXXX" 
+    pattern="^(?!(000|666|9))\d{3} (?!00)\d{2} (?!0000)\d{4}$" 
+    class="usa-input usa-masked" 
+    title="9-digit social security number" 
+    aria-describedby="ssnHint" 
+>
+
+<h3 class="site-preview-heading">US Telephone Number Input Mask</h3>
+<label class="usa-label" for="tel">US Telephone Number</label>
+<div class="usa-hint" id="pnHint">For example: (XXX) XXX-XXXX</div>
```
amyleadem, line 18: The screen reader readout isn't very helpful here. On VoiceOver, users hear "For example XXXXXXXXX". The placeholder in this component does show the visual example in the hint already - could we instead rewrite this hint to give clear instructions for what to do? I am thinking of something like "Provide your 10-digit telephone number, starting with the area code."

If we do not want to give that hint to visual users, could we place instructional copy in an aria-label on the `<input>` instead of using `aria-describedby`?

packages/usa-input-mask/src/styles/_usa-input-mask.scss
```diff
@@ -0,0 +1,28 @@
+@use "uswds-core" as *;
+
+.usa-input-mask {
+  position: relative;
+  line-height: 1;
```
amyleadem, line 5: Is this line-height needed?

packages/usa-input-mask/src/styles/_usa-input-mask.scss
```diff
@@ -0,0 +1,28 @@
...
+
+  &--content {
+    position: absolute;
+    left: 2px;
+    top: 0;
+    pointer-events: none;
+    color: color("gray-50");
+    font-family: family("sans");
```
amyleadem, line 13: Can this be defined with $theme-form-font-family? (Or declare the font-family using the placeholder below)

packages/usa-input-mask/src/usa-input-mask.twig
```diff
@@ -0,0 +1,42 @@
+<h3 class="site-preview-heading">Social Security Number Input Mask</h3>
+<label class="usa-label" for="ssn">Social Security Number</label>
+<div class="usa-hint flex" id="ssnHint">For example: XXX XX XXXX</div>
```
amyleadem, line 3: Is `.flex` needed here?

packages/usa-input/src/usa-input.twig
mitchmoccia: This has been removed and will be reflected in the next push I make tonight, thanks!

packages/usa-input-mask/src/index.js
mitchmoccia: Test removed

packages/uswds-core/src/js/index.js
mitchmoccia: Alphabetized... Turns out a couple more needed it (navigation was after search) P.S. Love the Alphabetical Sorter VSCode extension!

packages/usa-input/src/usa-input.twig
amyleadem: Thanks!

amyleadem:
@mitchmoccia @jonathanbobel Hello! Thanks for all your hard work on this. I added some notes in the `.scss` and `.twig` files. I'll come back first thing tomorrow morning with fresh eyes to dive in deeper into the JS. Please let me know if you have any questions!

mitchmoccia:
> @mitchmoccia @jonathanbobel Hello! Thanks for all your hard work on this. I added some notes in the `.scss` and `.twig` files. I'll come back first thing tomorrow morning with fresh eyes to dive in deeper into the JS. Please let me know if you have any questions!

Thanks Amy!

packages/usa-input-mask/src/styles/_usa-input-mask.scss
mitchmoccia: Done

packages/usa-input-mask/src/usa-input-mask.twig
mitchmoccia: Removed

packages/usa-input-mask/src/styles/_usa-input-mask.scss
mitchmoccia: changed to .usa-masked... They are both needed I believe. .usa-masked is applied to the input and .usa-input-mask--content is applied to the dynamically created shell span.

packages/usa-input-mask/src/styles/_usa-input-mask.scss
mitchmoccia: Done

packages/usa-input-mask/src/styles/_usa-input-mask.scss
mitchmoccia: Removed

packages/uswds/_index.scss
mitchmoccia: Done

packages/usa-input-mask/src/styles/_usa-input-mask.scss
mitchmoccia: Removed

packages/usa-input-mask/src/styles/_usa-input-mask.scss
```diff
@@ -0,0 +1,26 @@
...
+
+    i {
+      visibility: hidden;
+    }
+  }
+}
+
+.usa-masked,
+.usa-input-mask--content {
+  @extend %block-input-general;
```
amyleadem, line 23: More on alignment: I am wondering if we should switch to the mono font for this component. It would be a visual deviation from our other text input but would ensure that the characters of the input content lines up exactly with the mask content.

packages/usa-input-mask/src/styles/_usa-input-mask.scss
```diff
@@ -0,0 +1,26 @@
+@use "uswds-core" as *;
+
+.usa-input-mask {
+  position: relative;
+
+  &--content {
+    border: 1px solid transparent;
+    color: color("gray-50");
+    font-size: font-size($theme-form-font-family, $theme-body-font-size);
+    left: 2px;
```
amyleadem, line 10: Digging into the visual alignment of the mask content and input content and found that this left addition causes some horizontal mismatch. Can we remove this?

packages/usa-input-mask/src/test/input-mask-phone.template.html
```diff
@@ -0,0 +1,21 @@
+<div class="usa-input-masking">
+  <h3 class="site-preview-heading">US Telephone Number Input Mask</h3>
```
mejiaj, line 2: You can prevent these `.site-preview-heading`'s from showing up in the code examples by adding them to the site's component include.

[Alert example with heading](https://github.com/uswds/uswds-site/blob/main/_includes/code/components/alert.html#L1)

How it currently looks
image

mejiaj (changes requested):
Very nice! Component looks and works well. Also tested it in voiceover and didn't notice any issues. Added some suggestions below.

You also mentioned it can be used in conjunction with character count   —  does that work and is it recommended?

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+  console.log("shell ", shell); // eslint-disable-line no-console
+
+  const content = document.createElement("span");
+  content.classList.add(MASK_CONTENT);
+  content.setAttribute("aria-hidden", "true");
+  content.id = `${input.id}Mask`;
+  content.textContent = placeholder;
+
+  shell.appendChild(content);
+  input.parentNode.insertBefore(shell, input);
```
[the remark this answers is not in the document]
mitchmoccia: Added a form tag then set a CONTEXT variable to "form"... Then used closest(CONTEXT) like you suggested.

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+  shell.appendChild(input);
+};
+
+const setValueOfMask = (el) => {
+  const { value } = el;
+  const placeholder = el.getAttribute("data-placeholder");
```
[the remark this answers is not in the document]
mitchmoccia: Yes this value changes based on the pattern that is set (e.g. for phone ___-___-____). I tried implementing this and it didn't work initially, will revisit it and try again soon.

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+  const placeholderVal = `${placeholder.substr(value.length)}`;
+
+  const theIEl = document.createElement("i");
+  theIEl.textContent = value;
+  return [theIEl, placeholderVal];
+};
+
+const handleCurrentValue = (el) => {
+  const isCharsetPresent = el.getAttribute("data-charset");
```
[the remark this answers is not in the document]
mitchmoccia: data-charset is for alphanumeric values, like in a Canadian ZIP code (which we are currently not using, but it's there as an option). We could probably name this better? like data-alphanumeric what do you think? Also yes this needs to be added to the guidance! Thanks

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+  const placeholder = isCharsetPresent || el.getAttribute("data-placeholder");
+  const { value } = el;
+  const l = placeholder.length;
```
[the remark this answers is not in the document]
mitchmoccia: I changed l to len

packages/usa-input-mask/src/test/input-mask-phone.template.html
mitchmoccia: This has been reworked in uswds and uswds-site per the alert example you cited. In uswds, the component has examples that are now separated into multiple files, and in uswds-site we're using the .site-preview-heading to display the heading above each example.

mitchmoccia:
@mejiaj have a couple more comments to address but made some great progress... To answer your question about character count and input mask... I confirmed with Dan that we should not use character count with input mask, only free text fields. Thanks and more to come!

packages/usa-input-mask/src/index.js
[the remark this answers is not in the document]
mejiaj: @mitchmoccia I see. My question was originally meant if it changed based on input  —  which is where I was seeing it fired IIRC.

packages/usa-input-mask/src/index.js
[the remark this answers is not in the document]
mejiaj: Sorry if this is a silly question are the options strictly alpha, numeric, or both (alphanumeric)? The values would be a regex that developer sets?

I think the name could be flexible, but the most important thing is that it's documented. Perhaps `data-charset-type`.

packages/usa-input-mask/src/index.js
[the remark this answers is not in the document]
mejiaj: Thank you!

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+  let newValue = "";
+  let i;
+  let j;
+  let isInt;
+  let isLetter;
+
+  // strip special characters
+  const strippedValue = isCharsetPresent
+    ? value.replace(/\W/g, "")
+    : value.replace(/\D/g, "");
```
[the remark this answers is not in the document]
mitchmoccia: This has been abstracted into its own function

packages/usa-input-mask/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+
+  for (i = 0, j = 0; i < l; i += 1) {
+    isInt = !Number.isNaN(parseInt(strippedValue[j], 10));
+    isLetter = strippedValue[j] ? strippedValue[j].match(/[A-Z]/i) : false;
```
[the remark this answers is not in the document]
mitchmoccia: These have been abstracted into their own functions

packages/usa-input-mask/src/index.js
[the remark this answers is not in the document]
mitchmoccia: @mejiaj When the data-charset attribute is added, it's an alphanumeric placeholder that should match the regex in the pattern attribute. NOTE: If data-charset is not present, the input is strictly numeric.

The alphanumeric placeholder would take the form of X for letter and _ for number (this is customizable in the JS)

So for a Canadian postal code (*see description below), it would be:
"pattern": "\w\d\w \d\w\d",
"data-charset": "X_X _X_"

Let me know if that makes sense... I'm discussing with the team now if we should add a fourth example that shows some alphanumeric input masking, and if so what example would that be... Do you have any thoughts on that? Thanks!

* A Canadian postal code is a six-character uniformly structured, alphanumeric code in the form “ANA NAN” where “A” is an alphabetic character and “N” is a numeric character.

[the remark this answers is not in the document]
mitchmoccia: Small correction on the above... Just talked with Dan and the data-charset character placeholders should be # for number and A for letter like this `A#A #A#`. I'm going to be updating this in the code.

thisisdano:
@mitchmoccia I'm having trouble getting this to work for me. Did something go wrong?

packages/usa-input-mask/src/styles/_usa-input-mask.scss
thisisdano: I've done a lot of back-and-forth on this, and at the end of the day I don't think we're gaining as much from the font change as I'd hoped. In fact, the monospaced `space` character can make it look as though you have an extra space added in your entry. So I think the best option is to stick with the conventional font here.

packages/usa-input-mask/src/styles/_usa-input-mask.scss
thisisdano: This is a quirky one since removing it results in its own horizontal mismatch. I think we should leave it as-is for now, but but alignment on a short list of fast-follow improvements

amyleadem:
Just reviewed the interaction and behavior of the component. Looking good on my tests so far! I had a couple of notes about giving users more feedback and instructions. At this stage of the game, I expect these might need to be enhancements to add later, but I wanted to throw them out there in case there is time to consider them.

## Error handling
It would be helpful to give some kind of feedback for when a character is entered that doesn't fit the acceptable pattern. This feels especially important when using assistive technologies to navigate this component. Right now adding an item that doesn't fit just returns silence. 

## More detailed instructions
Additionally, more detail about what kind of entry is acceptable would be helpful. Adding something to indicate the number of characters expected along with explicit directions for the expected character pattern. So social security could have something as simple as "Expected input: Nine numbers. For example, 123 45 6789", but the alphanumeric example could have something like "Expected input: Six characters with alternating letters and numbers. For example, A1B 2C3." I expect this could be added either to `usa-hint` or as an `aria-label` on the input.

PR #4885, opened 2022-07-27
Optimize hero image

### LINKED ISSUE #4281: Reduce Hero Image Size
While [compressing some images in this PR](https://github.com/uswds/uswds/pull/4280) I came across /img/hero.gov which is (after a basic compression 624 KB. 

This is a pretty big image, but it made me wonder why a 2048 × 820px image is being used here. 

The Hero image is called up in the CSS (or more precisely the SCSS) `$theme-hero-image: "#{$theme-image-path}/hero.png";` but could we make this smaller? 

People use design systems like this as examples of best practices. Maybe this is just part of the old default theme, not sure. I couldn't find an example of a live version of the demo site to look at. Maybe it is somewhere in the documentation that I've missed. It should be run through Google Lighthouse as well as tools like https://www.websitecarbon.com/

thisisdano:
**Optimize hero image.** We replaced our default hero image (644 KB PNG) with an optimized image (147 KB JPG), saving 477 KB. We've also provided a next-generation image format version of the image (105 KB WEBP) as an example.

⚠️ This changes the value of `$theme-hero-image`. If your project uses this default hero image, you'll need to make sure to move the new asset (`hero.jpg`) to your project images directory.

Setting | Old default | New default
--- | --- | ---
`$theme-hero-image` | `"#{general.$theme-image-path}/hero.png"` | `"#{general.$theme-image-path}/hero.jpg"`

## Problem
Our default hero image isn't a good example of an optimized image. Even if teams don't use this generic image in their projects, we should deliver an optimized image. Additionally, we don't have a good way of distinguishing high-quality, high-resolution image originals from the optimized version (or versions) we ship. We need the uncompressed originals to make our optimized versions, but teams should not use these files in their projects. 

## Solution

- Replace the current hero PNG (644 KB) with an optimized JPG (147 KB).
- Update the value of `$theme-hero-image` to point to this new file (`"#{general.$theme-image-path}/hero.jpg"`).
- Include a WEBP version (105 KB) as an example.
- Include the original image prefixed with an underscore and ignore underscore-prefixed files when building our `dist/img` directory.

## How to test
1. View the component on Federalist ([Preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/dw-hero-image/iframe.html?args=&id=components-hero--hero&viewMode=story) →)
2. Use the inspector to see that it's using the new image (`hero.jpg`)
3. Pull the branch locally and build USWDS with `npx gulp`
4. Check the `dist/img` directory to find `hero.jpg` and `hero.webp`. You should NOT see `_hero.png` (the original)

## Related issues
- Fixes #4281


tasks/copy.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
   // Copy images to /dist directory
   copyImages() {
     dutil.logMessage("copyImages", "Copying images to /dist/img");
-    return src(["packages/**/src/img/**/*{png,jpg,gif,webp,svg,ico}"])
+    return src(["packages/**/src/img/**/[!_]*.{png,jpg,gif,webp,svg,ico}"])
```
mejiaj, line 32: Thoughts on moving processed images to a `dist` directory and keep unprocessed in `src`?

It would remove ambiguity in this glob and what the underscore means.

mejiaj (commented):
LGTM, just a question on moving processed images to `dist`.

tasks/copy.js
mejiaj: Dist directory issue has been moved to #4886. Resolving this thread.

mejiaj (approved):

kd-usgs:
hero.png is still huge. This has been a problem for years. It is a stock image that users will replace. Please consider removing it from USWDS.

PR #4719, opened 2022-05-26
USWDS - Validation: Adds complete/incomplete labels for screen readers. Resolves #4477

mahoneycm:
## Preview

[Validation →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/cm-validation-a11y/?path=/story/components-validation--validation)

## Description

Upon initialization, a new span is created for validation checklist items with the `usa-sr-only` class and a `data-checklist-label=hidden` data attribute. By default the span is set to "Incomplete"

When the validator updates the items to "checked", it will change the text content to `Complete` or `Incomplete` as necessary. 

Resolves #4477 

## Additional information

This PR also changes the validation event listener from `keyup change` to `input change` so that keys like shift wouldn't trigger the function

Before you hit Submit, make sure you’ve done whichever of these applies to you:

- [x] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [x] Run `npm test` and make sure the tests for the files you have changed have passed.
- [x] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [x] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #4914, opened 2022-08-08
USWDS - Bug: Fix invalid aria role in validator list items

### LINKED ISSUE #4905: USWDS - Bug: Fix invalid aria role in validator list items
### Describe the bug

Testing validator updates in uswds/uswds#4719 results in the following a11y errors:

```bash
Errors in http://localhost:4000/components/validation/:

 • Elements must only use allowed ARIA attributes
   (https://dequeuniversity.com/rules/axe/3.5/aria-allowed-attr?application=axeAPI)

   (#validate-code > li:nth-child(1))

   <li class="usa-checklist__item" data-validator="uppercase"
   aria-checked="false"> Use at least one u...</li>

 • Elements must only use allowed ARIA attributes
   (https://dequeuniversity.com/rules/axe/3.5/aria-allowed-attr?application=axeAPI)

   (#validate-code > li:nth-child(2))

   <li class="usa-checklist__item" data-validator="numerical"
   aria-checked="false"> Use at least one n...</li>
```

1. Use branch for uswds/uswds#4719
2. Build site
3. Create federalist preview with draft PR
5. View failed CircleCI tests
6. Click `details` link
7. See errors 
image

### Expected Behavior

A11Y tests should pass

### Related code

Work was done in USWDS PR uswds/uswds#4719.

### Screenshots

_No response_

### System setup

- USWDS v3.1.0

### Additional context

We should look into adding the necessary role 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked#associated_roles

Or change the markup so it the validation state is accurate without `aria-checked`

**After adding `aria-role="checkbox"`**

image

### Code of Conduct

amyleadem:
## Summary
**Updated the validation component so that users of assistive technologies can receive status updates about the completion of each validation requirement.** This was done by adding a script that generates appropriate ARIA attributes. 

### Details
**Added a script that generates an aria-label for each checklist item.** This allows assistive technologies users to navigate to each checklist item and receive the current status of each validation requirement. 

**Added a script that generates a new span to house the summary of the validation status updates.** This span has `aria-live` and `aria-atomic` attributes so that users of assistive technologies will be notified of status changes. 

**Removed the `aria-checked` attribute from list items in the validation component.** This attribute  is no longer needed now that a text element informs the user if the requirement is complete or incomplete.

**Added `data-validation-complete` and `data-validation-incomplete` attributes so that users can customize copy for `status complete` and `status incomplete` messages.**

## Related issue
Closes: #4905 

Related PRs: This PR incorporates the changes originally merged in https://github.com/uswds/uswds/pull/4719 that were later reverted in https://github.com/uswds/uswds/pull/4906. This PR includes the original update from #4719 and resolves the errors that caused the reversion.

## Preview link
Preview link: [Validation component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/al-validation-aria/?path=/story/components-validation--validation)

### Including the aria-checked attribute generates errors
Adding the `aria-checked` attribute to a `<li>` item results in the following a11y error:
```
Errors in http://localhost:4000/components/validation/:

 • Elements must only use allowed ARIA attributes
   (https://dequeuniversity.com/rules/axe/3.5/aria-allowed-attr?application=axeAPI)

   (#validate-code > li:nth-child(1))

   <li class="usa-checklist__item" data-validator="uppercase"
   aria-checked="false"> Use at least one u...</li>

 • Elements must only use allowed ARIA attributes
   (https://dequeuniversity.com/rules/axe/3.5/aria-allowed-attr?application=axeAPI)

   (#validate-code > li:nth-child(2))

   <li class="usa-checklist__item" data-validator="numerical"
   aria-checked="false"> Use at least one n...</li>
```
These errors reveal that items with the `aria-checked` attribute also need an appropriate role. Reference: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked#associated_roles

However, adding `role="checkbox"` to the `<li>` item also returns an error:
```
Detected the following accessibility violations!
       
       1. aria-allowed-role (ARIA role should be appropriate for the element)
       
          For more info, visit https://dequeuniversity.com/rules/axe/4.3/aria-allowed-role?application=axeAPI.
       
          Check these nodes:
       
          - html: <li class="usa-checklist__item" data-validator="uppercase" role="checkbox" aria-checked="true">Use at least one uppercase character</li>
            summary: Fix any of the following:
                       ARIA role checkbox is not allowed for given element
```

### VoiceOver produces a confusing readout
Additionally, VoiceOver generates undesirable and inconsistent readouts. Items can be read twice or only partially. 

### aria-checked
The `data-checklist-label` element now informs the keyboard user if the validation requirements are complete or incomplete. There is no longer a need to use an attribute like `aria-checked` to communicate `complete` or `incomplete` to the user. 

### Confusing readout
Changing the following items  generated a more consistent readout in VoiceOver:
- Generated an `aria-label` for each checklist item that includes the original text content plus the item's current status
- Generated a new sr-only span with `aria-live` and `aria-atomic` attributes that houses the summary of all checklist items plus their individual statuses. 

### Tasks completed in this PR:
1. Integrated changes created in `cm-validation-a11y` branch [0684b17](https://github.com/uswds/uswds/pull/4914/commits/0684b17e86b0bd8653edf8874ce94846b7d06e6b)
1. Removed the `aria-checked` attribute from the list item [f077275](https://github.com/uswds/uswds/pull/4914/commits/f0772753cfdce51c64355b745a979635070a162b)
1. Added `tabindex="0"` to make the items in the validation checklist items easier to navigate via tab [29c2b61](https://github.com/uswds/uswds/pull/4914/commits/29c2b618e75eb6e8ec4727323835e93e1f9077dd)
1. Created a new sr-only span element with `aria-live` and `aria-atomic` attributes that houses a status summary [07ad7de](https://github.com/uswds/uswds/pull/4914/commits/07ad7de7dfcd33d179958252706565b07c51284c)
1. Added optional `data-validation-complete` and `data-validation-incomplete` attributes [eeb2e18](https://github.com/uswds/uswds/pull/4914/commits/eeb2e188c324b59fcb0e116b899d114ef15515d8)

## Testing and review
Using the screen readers you have available,
1. Confirm that validation status is discoverable via keyboard navigation . 
1. Confirm that the validation status is read out in a way that makes sense.
1. Confirm that the validation component passes a11y tests. 
1. Add the following data attributes with custom values to `.usa-input` and confirm that the status message in `data-checklist-label` updates accordingly:
    - data-validation-complete
    - data-validation-incomplete
    
Tested on: 
- VoiceOver on Firefox, Safari, Edge for MacOs
- VoiceOver Gestures on iOS

---


amyleadem:
@mejiaj I wasn't sure how best to work with the previously merged PR, so I simply brought in the changes from that PR into this branch. Please let me know if there is a cleaner path.

mejiaj (changes requested):
Based on our testing, let's try using an aria label on the list items to improve how it gets announced via screenreader.

amyleadem:
@mejiaj I made some updates to improve the readout in VoiceOver. Surprisingly, an `<i>` element generated a more reliable readout than the `<span>` (at least in VO). Will you check this on your screen readers and let me know your thoughts?

mejiaj:
@amyleadem it does read better in voiceover! The only thing I've noticed is that the list item doesn't read out the updated status when I tab to it. Instead, it will just read the requirements **without** the current status.

The only way I can get the status to be read is to focus out and back in to the input element.

amyleadem:
@mejiaj Switched it up a bit and generated an `aria-label` for the checklist items instead of generating a new text element. Will you let me know what you think?

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,44 @@
...
+const validator = behavior(
+  {
+    "input change": {
+      [VALIDATE_INPUT]: change,
+    },
   },
-});
+  {
+    init(root) {
+      selectOrMatches(VALIDATE_INPUT, root).forEach((item) => {
```
mejiaj, line 21: Naming each validation input as `item` can be easily confused with checklist items. Let's rename to make it clearer that each item is a validation input.

```suggestion
      selectOrMatches(VALIDATE_INPUT, root).forEach((input) => {
```

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,44 @@
...
+        const validationParent = item.parentNode;
+        const checklistItems =
+          validationParent.querySelectorAll(CHECKLIST_ITEM);
+
+        item.setAttribute("aria-controls", "validate-code");
+        item.setAttribute("aria-live", "polite");
```
mejiaj, line 27: Shouldn't the `aria-live` attribute go to the region that's updated? In this case the unordered list with validation items (`.usa-alert--validation`).

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,44 @@
...
+        item.setAttribute("aria-atomic", true);
```
mejiaj, line 28: Looking at the MDN guidance for [aria-atomic](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-atomic) I can see this input changes. But wouldn't the best region be the unordered list again?

mejiaj, line 28: It might be better to move these scaffolding lines to its own function like `enhanceValidation`. As well as export them after init.

See example from accordion's JS

https://github.com/uswds/uswds/blob/8d2c7716890606e342877e916d1622eb8f1eaa0b/packages/usa-accordion/src/index.js#L84-L95

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,44 @@
...
+
+        let inputLabel = "";
```
mejiaj, line 30: This could be confused with the input's `label` element. Maybe we can rename for clarification. Something that tells us it's the validation item label or the input's status.

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,44 @@
...
+        checklistItems.forEach((listItem) => {
+          const itemStatus = `${listItem.textContent} status incomplete`;
+          listItem.setAttribute("tabindex", "0");
+          listItem.setAttribute("aria-label", itemStatus);
+          inputLabel += `${itemStatus}. `;
+        });
+
+        item.setAttribute("aria-label", inputLabel);
```
mejiaj, line 38: These could be it's own function too for setting up the unordered list.

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,44 @@
...
+        const checklistItems =
+          validationParent.querySelectorAll(CHECKLIST_ITEM);
+
+        item.setAttribute("aria-controls", "validate-code");
+        item.setAttribute("aria-live", "polite");
+        item.setAttribute("aria-atomic", true);
+
+        let inputLabel = "";
+        checklistItems.forEach((listItem) => {
+          const itemStatus = `${listItem.textContent} status incomplete`;
```
mejiaj, line 32: `itemStatus` label is fine, but we might want to consider allowing users to be able to set the label. 

For example, if I'm working on a multilingual site `status incomplete` would be nice to update via an optional data attribute.

packages/usa-validation/src/index.js
mejiaj, line 38: Wouldn't giving the input an aria-label cause conflicts with the value of the input element? Just trying to get a better understanding of user benefit of `aria-label` on the input itself.

mejiaj (commented):
Works well, on VoiceOver I was able to verify functionality has improved:
- [x] Elements read correctly
- [x] Validation requirements update and read properly

Component also still works when inserting dynamically.

I've added some suggestions for improving code and some questions about aria attributes on the input element.

packages/usa-validation/src/index.js
amyleadem: Thanks. Updated in [34c5be9](https://github.com/uswds/uswds/pull/4914/commits/34c5be9f6793fce4cc4571e1631a2901c35d2c2d)

packages/usa-validation/src/index.js
amyleadem: I agree that this shouldn't live on the input element. I have moved `aria-live` to the new `data-validation-status` span element.

packages/usa-validation/src/index.js
amyleadem: I agree that this shouldn't live on the input element. I have moved `aria-atomic` to the new `data-validation-status` span element.

amyleadem: I created an enhanceValidation() function that sets up all the aria pieces. I admit that I might not fully grasp the concept, so please reach out if this needs revision.

packages/usa-validation/src/index.js
amyleadem: I agree that this shouldn't live on the input element. I have moved `aria-live` to the new `data-validation-status` span element.

packages/usa-validation/src/index.js
amyleadem: Agreed. I have moved the `aria-label` off of the input element

amyleadem:
@mejiaj 
Thanks for the helpful review. I wanted to show you one more possible approach that has gotten the cleanest readout so far and would not interfere with with the `aria-label` on the input. This option generates a span that  is a peer to the input and checklist elements, has the `aria-live` and `aria-atomic` attributes, and gathers the status updates for all checklist items into a single element for readout. 

The current proposed solution also generates an `aria-label` for each checklist item that includes the item's current status, so users can still get check the status of each individual element. 

The only concern I have is that creating a fully new text element might be too clumsy and cause some unexpected problems. It would be great to hear your thoughts.

Additionally, I created an `enhanceValidation()` function, but I am not sure I fully understand how best to do this. Will you take a look and let me know if this is doing what you hoped?

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,57 @@
...
 
 function change() {
   validate(this);
 }
 
-const validator = behavior({
-  "keyup change": {
-    "input[data-validation-element]": change,
+const enhanceValidation = (target) => {
+  selectOrMatches(VALIDATE_INPUT, target).forEach((input) => {
```
mejiaj, line 14: The enhanceValidation() function is something we should be able to call on a single component.

Currently, if I created a dynamic validation component and ran `enhanceValidation()` I would enhance **all** validation components again. That's something we only want to run on init().

I would move the `selectOrMatches()` to `init()`.

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,57 @@
...
+  {
+    "input change": {
+      [VALIDATE_INPUT]: change,
+    },
   },
-});
+  {
+    init(root) {
+      enhanceValidation(root);
+    },
```
mejiaj, line 52: Moving the loop here will allow us to enhance single components. For example, creating a single dynamic component and appending to DOM.

It'd look something like this:

```suggestion
    init(root) {
      selectOrMatches(VALIDATE_INPUT, root).forEach((input) => enhanceValidation,
    }
```

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,57 @@
...
+      listItem.setAttribute("tabindex", "0");
+      listItem.setAttribute("aria-label", itemStatus);
+    });
+  });
+}
+
+const validator = behavior(
+  {
+    "input change": {
+      [VALIDATE_INPUT]: change,
```
mejiaj, line 46: The name could be improved to provide clarity. 
```suggestion
      [VALIDATE_INPUT]: handleChange,
```

Maybe you could even call it directly 

```suggestion
      [VALIDATE_INPUT]: validate,
```

packages/usa-validation/src/index.js
```diff
@@ -1,14 +1,57 @@
...
+    // Set up input attributes
+    input.setAttribute("aria-controls", "validate-code");
+
+    // Create container to hold aria readout
+    const statusSummaryContainer = document.createElement("span");
+    statusSummaryContainer.setAttribute("data-validation-status", "");
+    statusSummaryContainer.classList.add("usa-sr-only");
+    statusSummaryContainer.setAttribute("aria-live", "polite");
+    statusSummaryContainer.setAttribute("aria-atomic", true);
+    validationParent.append(statusSummaryContainer);
```
mejiaj, line 28: This could be moved to its own function.

```js
const createStatusElement = (validationContainer) => {
  const statusSummaryContainer = document.createElement("span");
  
  statusSummaryContainer.setAttribute("data-validation-status", "");
  statusSummaryContainer.setAttribute("aria-live", "polite");
  statusSummaryContainer.setAttribute("aria-atomic", true);

  statusSummaryContainer.classList.add("usa-sr-only");

  validationContainer.append(statusSummaryContainer);
}
```
```

Comment (mejiaj) [reply packages/usa-validation/src/index.js:28]: Wouldn't this status element have to be tied to the input as well? If so, we'd have to give it a unique ID and add it to the input's `aria-describedby`.

I didn't notice any major issues with it being read in voiceover. On rare occasion it would either cutoff the status announcement or not give it at all.

**Review by mejiaj — CHANGES_REQUESTED** [review]
@amyleadem it does read better. Added some thoughts below.

Overall, if we want the new status element we would have to link it to the input (which seems to be performing the same function as the list with `aria-live="polite"`).

I do see the potential of this improving the experience, especially after seeing GovUKs character count recommendations in #4714. If we want to go that route we'd probably need to:
- Link the input to this new status via a unique ID and `aria-describedby`
- Set a delay on the status so it can finish reading user input _and then_ give status

File: packages/usa-validation/src/index.js (hunk 7 above)
Comment (amyleadem) [reply packages/usa-validation/src/index.js]: Moved enhanceValidation to init in https://github.com/uswds/uswds/pull/4914/commits/a57aff1446383a37b762637069c4b63f9691c52c

**Comment by amyleadem** [comment]
@mejiaj 
Thanks a ton for the helpful review. I believe I have addressed all the comments above. Please let me know if you see anything else! 

Tasks completed:
- Moved enhanceValidation to init in https://github.com/uswds/uswds/commit/a57aff1446383a37b762637069c4b63f9691c52c
- Added aria-describedby to input in [a57aff1](https://github.com/uswds/uswds/pull/4914/commits/a57aff1446383a37b762637069c4b63f9691c52c#diff-a31baf8faa988bb0919458669b9013502ff4cc638eeaf017676c82b6a25d13aeR23)
- Split out the functions in https://github.com/uswds/uswds/commit/853fb7db8eb64803231807e0db68c71e7a132c31
- Added a delay for the screen reader readout in [9162036](https://github.com/uswds/uswds/pull/4914/commits/916203673d297cb78b67ba3874071af76937b450)
- Updated `handleChange()` to be more explicit in https://github.com/uswds/uswds/pull/4914/commits/aa38ecbcb2f829a4b6c03bc293407798d2a8433b

File: packages/uswds-core/src/js/utils/validate-input.js [hunk 11]
```diff
@@ -14,20 +14,52 @@ module.exports = function validate(el) {
...
+      let checkboxContent = `${validatorCheckbox.textContent} `;
+
+      if (el.hasAttribute("data-validation-complete")) {
+        statusComplete = el.getAttribute("data-validation-complete");
+      }
+
+      if (el.hasAttribute("data-validation-incomplete")) {
+        statusIncomplete = el.getAttribute("data-validation-incomplete");
+      }
+
```
mejiaj, line 46: We could simplify this by having these labels as a backup in case the dataset comes back null or undefined.

```suggestion
      const statusComplete = el.dataset.validationComplete || "status complete";
      const statusIncomplete = el.dataset.validationIncomplete || "status incomplete";
```

packages/uswds-core/src/js/utils/validate-input.js
```diff
@@ -14,20 +14,52 @@ module.exports = function validate(el) {
...
+      // move status updates to aria-label on checklist item
+      validatorCheckbox.setAttribute("aria-label", checkboxContent);
+
+      // Create a summary of status for all checklist items
+      statusSummary += `${checkboxContent}. `;
+
+      // Add summary to screen reader summary container, after a delay
+      ((debounce(() => {
+        statusSummaryContainer.textContent = statusSummary;
+      }, 1000)))();
```
mejiaj, line 62: Optional because we have the comment, but we can increase readability by extracting it to it's own `updateSRSummary()` and just calling it here.

mejiaj (changes requested):
Looks good, the experience has improved for users on screen readers with the delayed announcement.

Added some suggestions we could use to improve code readability.

packages/uswds-core/src/js/utils/validate-input.js
amyleadem: Clean! Updated in [aea87ce](https://github.com/uswds/uswds/pull/4914/commits/aea87ce06c60429928e9ff842b06531e7e9ca202)

packages/uswds-core/src/js/utils/validate-input.js
amyleadem: I like that it could make it more consistent with the new character count setup. Updated in [bfbfd52](https://github.com/uswds/uswds/pull/4914/commits/bfbfd52063589b239ac2bc3c215dd86c09c7bc5f)

(If I misunderstood your request, feel free to fix while I am out!)

mejiaj (approved):
Works well, tested for regressions and VoiceOver. Has a debounce function that was introduced in PR #4976.

thisisdano (approved):
Nice discussion and nice work!

PR #4918, opened 2022-08-15
[IDP] In-page navigation

mitchmoccia:
This is a follow-up PR for In-page navigation from this earlier PR:
https://github.com/uswds/uswds/pull/4860

https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/mm-in-page-nav4/?path=/story/components-in-page-navigation--default

This version addresses all the great comments left by @mejiaj. Looking forward to any and all feedback, thank you!


packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,85 @@
+@use "uswds-core" as *;
+
+.usa-in-page-nav-container {
+  align-items: flex-start;
+  display: flex;
+  margin: units(8) units(0);
```
amyleadem, line 6: @mitchmoccia Curious - why these margins are needed?

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,85 @@
...
+  main,
+  [class^="main"] {
+    max-width: 87.5rem;
+    width: 100%;
+  }
+
+  .usa-in-page-nav {
+    display: none;
+    margin-left: units(0.5);
+    margin-top: units(2.5);
```
amyleadem, line 18: @mitchmoccia Also curious why these margins are needed

amyleadem:
@mitchmoccia Starting to do some testing on appearance and behavior - it is working well so far!  I had a few questions about alignment and am flagging some questions in the code comments. I will keep looking poking at it this afternoon and will cycle back tomorrow for more review. If you have any questions in the meantime, please let me know!

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+const IN_PAGE_NAV_LINK_CLASS = `${IN_PAGE_NAV_CLASS}__link`;
+const IN_PAGE_NAV_ANCHOR_CLASS = "usa-anchor-tag";
+
+const IN_PAGE_NAV_LINK = `.${IN_PAGE_NAV_LINK_CLASS}`;
+const IN_PAGE_NAV_ANCHOR = `.${IN_PAGE_NAV_ANCHOR_CLASS}`;
+const IN_PAGE_NAV_TITLE = "On this page";
+const IN_PAGE_NAV = `.${PREFIX}-in-page-nav`;
+const CURRENT_CLASS = `${PREFIX}-current`;
+const MAIN_ELEMENT = "main";
+const SITE_SUBHEADING = "site-subheading";
```
amyleadem, line 19: Is the class name `.site-subheading` pulled from uswds-site? I don't see any references to it in USWDS so I am wondering if there is a more descriptive name

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+
+const createInPageNav = () => {
+  const sectionHeadings = getSectionHeadings();
+  const insertNode = document.querySelector(IN_PAGE_NAV);
+
+  const inPageNav = document.createElement("nav");
+  inPageNav.setAttribute("role", "navigation");
+  inPageNav.setAttribute("arial-label", "In-page navigation");
+
+  const inPageNavTitle = document.createElement("p");
```
amyleadem, line 107: Wondering if this should be a header tag instead. Might be helpful information for users navigating by heading with assistive technology.

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+  });
+};
+
+const createInPageNav = () => {
+  const sectionHeadings = getSectionHeadings();
+  const insertNode = document.querySelector(IN_PAGE_NAV);
+
+  const inPageNav = document.createElement("nav");
+  inPageNav.setAttribute("role", "navigation");
+  inPageNav.setAttribute("arial-label", "In-page navigation");
```
amyleadem, line 105: Found a typo with aria-label. 
Additionally, the readout on the aria-label reads "In page navigation, navigation". We should be able to safely remove the word "navigation" from this attribute. (Reference: [Mozilla](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#associated_wai-aria_roles_states_and_properties))

```suggestion
  inPageNav.setAttribute("aria-label", IN_PAGE_NAV_TITLE );
```

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
```diff
@@ -0,0 +1,31 @@
...
+          <h3>{{ item.section_heading }}</h3>
+        {% else %}
+          <h2>{{ item.section_heading }}</h2>
+        {% endif %}
+        <p>{{ item.content }}</p>
+      {% endfor %}
+    {% endif %}
+  </main>
+
+  <aside class="usa-in-page-nav"></aside>
```
amyleadem, line 23: @mitchmoccia  @mejiaj 
Wondering about the benefits of nesting the `<nav>` inside an `<aside>`. I believe it is valid (depending on how it is defined), but feels semantically unnecessary and appears to just add an extra layer for users to tab through.

Can we just make this element the `<nav>` itself or switch to a `<div>` (if it needs a wrapper)?

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -110,6 +110,15 @@ $theme-identifier-max-width: "desktop" !default;
...
 $theme-identifier-secondary-link-color: "base-light" !default;
 
+// In-Page Navigation
+$theme-in-page-nav-background-color: "white" !default;
+$theme-in-page-nav-background-padding: 2 !default;
+$theme-in-page-nav-background-radius: "lg" !default;
+$theme-in-page-nav-bar-color: "secondary-dark" !default;
+$theme-in-page-nav-bar-width: 2px !default;
+$theme-in-page-nav-font-family: "ui" !default;
+$theme-in-page-nav-top-position: 8 !default;
```
amyleadem, line 120: I like that this setting is customizable. 
I wonder if a smaller default might be better to start. 8 units just feels big on a short screen, so I am wondering if 4 might be better to start with?

mejiaj (changes requested):
Very nice rewrite @mitchmoccia! The code, but JS especially looks much more in-line with other component standards.

Added some comments and suggestions. Also, I'm sure you've got this in mind already, but I didn't notice any unit tests for this component.

packages/uswds-core/src/styles/settings/_settings-components.scss
mitchmoccia: Thanks Amy, yes I started with 4 but then translating that to the USWDS documentation site caused it to be too close to the fixed header. The 8 gives it a good feel, but I can change it in uswds and then customize it in my uswds-site working branch? What do you think?

Here's a screen shot that shows where it stops at when sticky is engaged and 8 (4rem) is used for top:

image

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -110,6 +110,15 @@ $theme-identifier-max-width: "desktop" !default;
 $theme-identifier-primary-link-color: default !default;
 $theme-identifier-secondary-link-color: "base-light" !default;
 
+// In-Page Navigation
+$theme-in-page-nav-background-color: "white" !default;
+$theme-in-page-nav-background-padding: 2 !default;
+$theme-in-page-nav-background-radius: "lg" !default;
+$theme-in-page-nav-bar-color: "secondary-dark" !default;
```
amyleadem, line 117: I like that this is a customizable setting. However, I am wondering if the default value for this should be in the `primary` family. My thinking is that our component defaults use primary or base colors, and to use secondary here feels like an intentional deviation. 
If this choice to use secondary is meant to communicate that users should *not* be using primary or base here, then we should make sure to note that in our guidance.

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -110,6 +110,15 @@ $theme-identifier-max-width: "desktop" !default;
 $theme-identifier-primary-link-color: default !default;
 $theme-identifier-secondary-link-color: "base-light" !default;
 
+// In-Page Navigation
```
amyleadem, line 113: I like the suite of customization settings for this component.

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -110,6 +110,15 @@ $theme-identifier-max-width: "desktop" !default;
...
+$theme-in-page-nav-background-color: "white" !default;
```
amyleadem, line 114: Thinking it might be helpful to define this with the theme body background color setting. It feels like it could be the most common use pattern.

```
@use "settings-color" as color;

...

$theme-in-page-nav-background-color: color.$theme-body-background-color !default;
```

mitchmoccia:
Amazing feedback @amyleadem and @mejiaj THANK YOU... I'll be addressing all of these ASAP

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
mejiaj: The [HTML5 standard mentions [html.spec.whatwg.org]](https://html.spec.whatwg.org/multipage/sections.html#the-aside-element:~:text=The%20element%20can%20be%20used%20for%20typographical%20effects%20like%20pull%20quotes%20or%20sidebars%2C%20for%20advertising%2C%20for%20groups%20of%20nav%20elements%2C%20and%20for%20other%20content%20that%20is%20considered%20separate%20from%20the%20main%20content%20of%20the%20page.):

>The element can be used for typographical effects like pull quotes or sidebars, for advertising, for groups of [nav](https://html.spec.whatwg.org/multipage/sections.html#the-nav-element) elements, and for other content that is considered separate from the main content of the page.

And they have an example below (third example down):

image

**TL;DR** I can see both sides of the argument. Semantically I _think_ it's okay. The final decision depends on the screen reader experience. Unless others understand the use case better, please chime in!

amyleadem: In the same spec, it does show this example for the nav element. https://html.spec.whatwg.org/multipage/sections.html#the-nav-element
> In the following example, there are two [nav](https://html.spec.whatwg.org/multipage/sections.html#the-nav-element) elements, one for primary navigation around the site, and one for secondary navigation around the page itself.

image

FWIW: I did poke around at other implementations and found that in its own implementation of a similar component (The "In this article" TOC on the right), Mozilla uses an `<aside>` with a nested `<section>` and no `<nav>`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside

Other popular design systems use `<nav>` with a nested list or just a `<div>` with a nested list. 

So my tl;dr is that it seems like there are a lot of possible correct paths here regarding if it is a nav or aside (or both), and that it might just depend on how we define "tangential" or "indirectly related" content. Users should still be able to find what they need with either. Change would only be needed here if we want to remove an extra layer when users navigate by element, but it shouldn't affect users who tab into it. 

[Reference:](https://html.spec.whatwg.org/multipage/sections.html#the-aside-element)
>The [aside](https://html.spec.whatwg.org/multipage/sections.html#the-aside-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) a section of a page that consists of content that is tangentially related to the content around the [aside](https://html.spec.whatwg.org/multipage/sections.html#the-aside-element) element, and which could be considered separate from that content. Such sections are often represented as sidebars in printed typography.

>The element can be used for typographical effects like pull quotes or sidebars, for advertising, for groups of [nav](https://html.spec.whatwg.org/multipage/sections.html#the-nav-element) elements, and for other content that is considered separate from the main content of the page.

packages/uswds-core/src/styles/settings/_settings-components.scss
amyleadem: It makes sense to me to use a shorter default value and then customize it for uswds-site, since the header nav on the USWDS site seems to be the exception rather than the rule.

packages/usa-in-page-navigation/src/index.js
mitchmoccia: Yes, site-subheading is from uswds-site _uswds-theme-custom-styles.scss line 652 
Should I change it to a style that comes with the in-page nav package?

packages/usa-in-page-navigation/src/index.js
mitchmoccia: Sure, h4 tag?

mitchmoccia: Oh the reason I made it a p tag was because the site-heading class was applied to a p tag in its use on uswds-site... Changing it to h4

packages/usa-in-page-navigation/src/index.js
mitchmoccia: Changing to IN_PAGE_NAV_TITLE so the reader says "On this page" or whatever the title is changed to.

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
mitchmoccia: Love the tag team analysis!! I did some quick testing in VoiceOver and it doesn't seem to make a difference... I did make the decision to use aside > nav based on the w3c example but will go with whatever makes most sense.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,85 @@
...
+
+.usa-in-page-nav-container {
+  align-items: flex-start;
+  display: flex;
+  margin: units(8) units(0);
+
+  // Set main landmark max width to ensure in-page nav will fit
+  main,
+  [class^="main"] {
+    max-width: 87.5rem;
```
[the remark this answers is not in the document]
mitchmoccia: Changing this to a setting $theme-in-page-nav-max-width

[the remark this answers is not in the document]
mitchmoccia: `$theme-in-page-nav-max-width: "desktop" !default;`

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,85 @@
...
+
+    nav {
+      background-color: color($theme-in-page-nav-background-color);
+      border-radius: radius($theme-in-page-nav-background-radius);
+      padding: units($theme-in-page-nav-background-padding);
+    }
+  }
+
+  .usa-in-page-nav-list {
+    background-color: initial;
```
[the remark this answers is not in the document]
mitchmoccia: Removed, I believe this is a remnant from the previous version.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,85 @@
...
+    border: none;
```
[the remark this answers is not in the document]
mitchmoccia: Removed, I believe this is a remnant from the previous version.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,85 @@
...
+    border-left: units(1px) solid color("base-lighter");
+    @include typeset($theme-in-page-nav-font-family, "sm", 2);
```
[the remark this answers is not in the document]
mitchmoccia: Done

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+const keymap = require("receptor/keymap");
+const behavior = require("../../uswds-core/src/js/utils/behavior");
+const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
+const { CLICK } = require("../../uswds-core/src/js/events");
+
+const IN_PAGE_NAV_CLASS = `${PREFIX}-in-page-nav`;
+const IN_PAGE_NAV_LIST_CLASS = `${IN_PAGE_NAV_CLASS}-list`;
+const IN_PAGE_NAV_ITEM_CLASS = `${IN_PAGE_NAV_CLASS}__item`;
+const IN_PAGE_NAV_LINK_CLASS = `${IN_PAGE_NAV_CLASS}__link`;
+const IN_PAGE_NAV_ANCHOR_CLASS = "usa-anchor-tag";
```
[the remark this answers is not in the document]
mitchmoccia: Fixed

thisisdano:
Having trouble getting this to work for me locally. Can you give me a hand, @mitchmoccia?

mitchmoccia:
> Having trouble getting this to work for me locally. Can you give me a hand, @mitchmoccia?

@thisisdano I just pushed some updates to mm-in-page-nav4 that were not present on remote... This includes your merge from idp-staging... Detailing issues and working with James here:
https://gsa-tts.slack.com/archives/C050HRGN7/p1667326494832779

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: I think I'll set it as `default` as in a component like Breadcrumb. That way it will pick up `$theme-body-background-color`

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: I think this should be `default` where the default is the text color as calculated by settings

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,85 @@
...
+    padding: units(0);
+
+    @include border-box-sizing;
+    @include nav-list("sidenav");
+
+    a {
+      color: color("base-darker");
```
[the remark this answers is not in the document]
thisisdano: I'm refactoring this be be better responsive to theme color settings

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
thisisdano: Since these determine the distance between the nav and both the top of the container and the text block, these should be settings to allow adaptation to individual projects

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+  const targetAnchor = document.getElementById(id);
+  const target = targetAnchor.parentElement;
+
+  if (target) {
+    target.setAttribute("tabindex", 0);
+    target.focus();
+    target.addEventListener(
+      "blur",
+      once(() => {
+        target.setAttribute("tabindex", -1);
```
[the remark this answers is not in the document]
thisisdano: @mitchmoccia Can you confirm that this is the intended behavior?

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+    listItem.appendChild(navLinks);
+  });
+
+  insertNode.appendChild(inPageNav);
+
+  const options = {
+    root: IO_ROOT,
+    rootMargin: IO_ROOT_MARGIN,
+    threshold: IO_THRESHOLD,
+  };
```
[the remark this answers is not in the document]
thisisdano: Done

packages/usa-in-page-navigation/src/index.js
thisisdano: I think this hed should get a BEM class related to the in-page-navigation component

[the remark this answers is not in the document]
thisisdano: Yes, let's follow this convention

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * Return a section id/anchor hash without the number sign
+ *
+ * @return {String} - Id value with the number sign removed
+ */
+const getSectionId = (value) => {
+  let id;
+
+  // Check if value is an event or element and get the cleaned up id
```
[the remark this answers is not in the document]
thisisdano: @mitchmoccia Can you take a look at this comment?

thisisdano:
I've made a number of updates to try an address some outstanding comments from James and Amy — and a couple observations of my own. 

### Use component BEM format throughout
I updated a few class names off the `usa-in-page-nav` base class:
- `usa-in-page-nav-list` → `usa-in-page-nav__list`
- `site-subheading` → `usa-in-page-nav__heading`
- `sub-item` → `usa-in-page-nav__item--sub-item`

### Normalized some of the JS constants
I tried to use more consistent naming conventions for the constants.

### Used more realistic Lorem Ipsum
I changed up some of the headings in the lorem ipsum, to appear more like read headings and give a better sense of the long and short entries we might expect to see in real life

### Simplified presentation for more resiliency across different theme colors
This is maybe the biggest one. I made a few updates to the presentation, with the goal of simplifying the color and style of the component and make it more able to respond automatically when users change their theme colors.

- **Used the default link colors throughout.** These use `set-link-from-bg` to get a "good" link color from the component background
- **Relied on more common link styling.** I removed the highlight background and stuck to default link colors for the nav links.
- **Used the calculated text color for the current color, and used a fatter default bar.** Together these help highlight the current item without requiring complex settings and calculations when the theme colors change
- **Removed sub-item inset and normalized the size of the entries.** This simplifies the presentation. We can use bold for entries and normal for sub-entries, and we avoid text shifting when it's bold-on-current. Removing the inset and normalizing the size allows a cleaner scan of the text while still communicating hierarchy.

### Added and changed some settings
I added and updated some settings.

A couple to help set the position of the nav, which might have to shift depending on the project layout:
- $theme-in-page-nav-margin-left: 4 !default;
- $theme-in-page-nav-margin-top: 2.5 !default;

Used "default" for some colors so they automatically adapt to upstream settings
- $theme-in-page-nav-background-color: "default" !default; (will take the value of $theme-background-color by default)
- $theme-in-page-nav-bar-color: "default" !default; (will take the value of $theme-text-color by default, and will automatically adapt if the background color changes)

Used a fatter bar so it's easier to see:
- $theme-in-page-nav-bar-width: "05" !default;


### Removed unnecessarily opinionated styles
I removed styles from the nav container so those who implement the in-page nav don't have unintended layout shifts.

Screen Shot 2022-11-02 at 4 50 08 PM
Screen Shot 2022-11-02 at 4 34 35 PM
Screen Shot 2022-11-02 at 4 35 20 PM
Screen Shot 2022-11-02 at 4 47 28 PM

thisisdano:
Note that we'll need to document these settings in the settings data so it appears on the settings page

mitchmoccia:
> Note that we'll need to document these settings in the settings data so it appears on the settings page

This is awesome @thisisdano thank you!

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
mitchmoccia: @mejiaj @thisisdano I borrowed this behavior from skipnav which sets the value of tabindex of the target (main) to -1... 

The purpose of it in in-page-nav is if a user reverse tabs, then don't let them go to the previous section's header (keeping them stuck in the content portion of the page), rather send them back to the in-page-nav... I think this intended behavior is sound, any thoughts on this?

packages/usa-in-page-navigation/src/test/template.html
```diff
@@ -0,0 +1,23 @@
...
+    <h2><a id="section_0" class="usa-anchor-tag"></a>Section 1</h2>
+    <p></p>
+    <h3><a id="section_1" class="usa-anchor-tag"></a>Section 1.1</h3>
+    <p>Section 1.1 content</p>
+    <h3><a id="section_2" class="usa-anchor-tag"></a>Section 1.2</h3>
+    <p>Section 1.2. content</p>
+  </main>
+
+  <aside class="usa-in-page-nav">
+    <nav role="navigation" arial-label="In-page navigation">
```
amyleadem, line 14: Little typo. Additionally, `<nav>` has an implicit role of "navigation", so I believe there is no need to make it explicit. (Reference: [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav))

```suggestion
    <nav aria-label="In-page navigation">
```

mitchmoccia: Just pushed the fix, thank you!

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,216 @@
...
+    behavior: "smooth",
+    top: el.offsetTop,
+    block: "start",
+  });
+};
+
+const createInPageNav = (inPageNavEl) => {
+  const sectionHeadings = getSectionHeadings();
+  const inPageNav = document.createElement("nav");
+  inPageNav.setAttribute("role", "navigation");
```
amyleadem, line 105: This line is probably safe to remove. 
`<nav>` has an implicit role of "navigation", so I believe there is no need to make it explicit. (Reference: [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav))

packages/usa-in-page-navigation/src/test/template.html
```diff
@@ -0,0 +1,23 @@
...
+    <p>Section 1.2. content</p>
+  </main>
+
+  <aside class="usa-in-page-nav">
+    <nav aria-label="In-page navigation">
+      <h4 class="site-subheading">On this page</h4>
+      <ul class="usa-in-page-nav-list">
+        <li class="usa-in-page-nav__item"><a href="#section_0" class="usa-in-page-nav__link usa-current">Section 1</a></li>
+        <li class="usa-in-page-nav__item sub-item"><a href="#section_1" class="usa-in-page-nav__link">Section 1.1</a></li>
+        <li class="usa-in-page-nav__item sub-item"><a href="#section_2" class="usa-in-page-nav__link">Section 1.2</a></li>
```
amyleadem, line 19: Looks like these items missed the BEM class update described [here](https://github.com/uswds/uswds/pull/4918#issuecomment-1301502261).

> * `usa-in-page-nav-list` → `usa-in-page-nav__list`
> * `site-subheading` → `usa-in-page-nav__heading`
> * `sub-item` → `usa-in-page-nav__item--sub-item`

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
```diff
@@ -0,0 +1,88 @@
...
+const behavior = require("../index");
+
+const HIDE_MAX_WIDTH = 639;
+const TEMPLATE = fs.readFileSync(path.join(__dirname, "/template.html"));
+const STYLES = fs.readFileSync(
+  `${__dirname}/../../../../dist/css/uswds.min.css`
+);
+const THE_NAV = ".usa-in-page-nav";
+const PRIMARY_CONTENT_SELECTOR =
+  ".usa-in-page-nav-container .usa-in-page-nav .usa-in-page-nav-list";
```
amyleadem, line 14: Updated per the BEM class update described [here](https://github.com/uswds/uswds/pull/4918#issuecomment-1301502261).

```suggestion
  ".usa-in-page-nav-container .usa-in-page-nav .usa-in-page-nav__list";
```

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -110,6 +110,18 @@ $theme-identifier-max-width: "desktop" !default;
 $theme-identifier-primary-link-color: default !default;
 $theme-identifier-secondary-link-color: "base-light" !default;
 
+// In-Page Navigation
+$theme-in-page-nav-background-color: "default" !default;
```
amyleadem, line 114: Flagging that values with color grades between 20-50 will return values with inaccessible color contrast.

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -110,6 +110,18 @@ $theme-identifier-max-width: "desktop" !default;
...
+$theme-in-page-nav-background-padding: 2 !default;
+$theme-in-page-nav-background-radius: "lg" !default;
+$theme-in-page-nav-bar-color: "default" !default;
```
amyleadem, line 117: Changing this value doesn't do anything &mdash;  I don't see that this variable is being used anywhere.

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -110,6 +110,18 @@ $theme-identifier-max-width: "desktop" !default;
...
+$theme-in-page-nav-bar-width: "05" !default;
```
amyleadem, line 118: Flagging that large values here will cause the bar to overlap list item text. 
`$theme-in-page-nav-bar-width: 3`:

image

amyleadem (dismissed):
@mitchmoccia @thisisdano 
Nice revisions. I have added some comments to the files - they are mostly simple items to fix or review. Let me know if you have questions!

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: Honestly, I wonder if we should just fix the bar width for our release version

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: ha whoops, lemme check!

packages/usa-in-page-navigation/src/test/template.html
thisisdano: Ah... forgot to update the tests!

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: hmmm. I'll look into what's happening with `set-link-from-bg()`...

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: On second though, I'm going to leave as-is and leave it up to the teams to use a working value when they customize...

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: At least we do get a good warning:

```
Warning: [In-page-navigation link] Neither the specified preferred link color token (`base-lighter`) nor the fallback link color token (`primary`) have AA contrast on a `blue-30` background. Using `primary`. Please check your source code and project settings.
```

But perhaps we should give folks a setting for link color...

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: OK, fixed.

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: Added a new setting for link color

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
mitchmoccia: @thisisdano @amyleadem just addressed this and pushed up the adjustment... When I removed role=navigation before, that was just on the unit test template... Here I have changed the dynamically added attribute from role navigation to aria-label In-page navigation.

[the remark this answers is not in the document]
thisisdano: I know we should have it on the `nav` but I'm not sure about the `aside` — how might we label it to be informative but not redundant with the nav?

mitchmoccia: Addressed this by changing role to your recommended aria-label... Commented on this below as well, thanks!

packages/uswds-core/src/styles/settings/_settings-components.scss
mitchmoccia: @thisisdano I just updated the settings on uswds-site to include the new setting you added $theme-in-page-nav-link-color:
https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/mm-inclusive-patterns-in-page-navigation/components/in-page-navigation/

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
amyleadem: My recommendation would be to define an aria-label with something like `IN_PAGE_NAV_TITLE` for the `aside`. 

Then the page landmarks (in VoiceOver, at least) would appear as:
On this page complementary
In-page navigation navigation 
main

_Note: In VoiceOver, it would be safe to remove the redundant word `navigation` from the `aria-label` on the `nav` so that it reads `aria-label="In-page"` instead. I am not sure how this presents in other AT though._

If scanning the page by element, the user would still get useful information, albeit maybe a little heavy-handed. 

It might be overly prescriptive but I like how it clarifies the landmarks menu.

mitchmoccia:
@thisisdano @mejiaj @amyleadem Just made a great update per Dan's suggestion... The in-page nav heading text and heading level are now customizable from the HTML via data attributes! The new attributes are data-title and data-heading-level both added to the aside element. 

I have also updated the unit tests and everything is looking good. Let me know how it looks on your end, thanks!

mitchmoccia:
I added two variables const 
IN_PAGE_NAV_TITLE = “On this page”;
IN_PAGE_NAV_HEADING_LEVEL = “h4"

and then check for the presence of the corresponding data attributes and if none, use the defaults...

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
mitchmoccia: Done! It's even better with the addition of data-attributes and the ability to customize that text in the HTML... It will look for the data attr. data-title and use that for the heading text AND the aria-label... If the data attr is not there, it will use the IN_PAGE_NAV_TITLE variable for both.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
+@use "uswds-core" as *;
+@use "sass:color";
+
+.usa-in-page-nav-container {
+  align-items: flex-start;
+  display: flex;
+
+  // Set main landmark max width to ensure in-page nav will fit
+  main,
+  [class^="main"] {
```
mejiaj, line 10: Do we need to support the `[class^="main"]` as well?

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+    max-width: units($theme-in-page-nav-main-content-max-width);
+    width: 100%;
+  }
+
+  .usa-in-page-nav__heading {
```
mejiaj, line 15: Nitpick, but I always like to order my selectors based on how they're ordered in markup. In this case `usa-in-page-nav ` would be before heading.

```
.usa-in-page-nav-container
  .usa-in-page-nav
    nav
      .usa-in-page-nav__heading
```

mejiaj: I've also noticed everything is nested inside of `usa-in-page-nav-container`. Was this intentional?

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+    margin-left: units($theme-in-page-nav-margin-left);
+    margin-top: units($theme-in-page-nav-margin-top);
+    max-width: units("card-lg");
+    order: 2;
+    position: sticky;
+    top: units($theme-in-page-nav-top);
+
+    @include at-media("tablet") {
+      display: block;
+      width: 100%;
```
mejiaj, line 39: This `width: 100%` could be a default style instead of inside media query.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    nav {
```
mejiaj, line 42: **Optional**, but I wish we used a class here. Even something like `.usa-in-page-nav__inner`. We've used a similar naming convention in banner, card, and header wrappers.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+      );    
+      border-radius: radius($theme-in-page-nav-background-radius);
+      padding: units($theme-in-page-nav-background-padding);
+    }
+  }
+
+  .usa-in-page-nav__list {
+    @include border-box-sizing;
+    @include nav-list("sidenav");
+    @include typeset($theme-in-page-nav-font-family, "sm", 2);
```
mejiaj, line 55: Do we need this `typeset()`? I see additional font styles set in `.usa-in-page-nav__item` on line 110.

Maybe we could move `font-size: font-size($theme-in-page-nav-font-family, "2xs");
` up and have nav items inherit.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+    border-left: units(1px) solid color.adjust(
+      color(
+        get-color-token-from-bg(
+          $bg-color: $theme-in-page-nav-background-color,
+          $context: "In-page navigation rail",
+        )
+      ), $alpha: -0.75
+    );
+    padding: units(0);
```
mejiaj, line 64: Only adding this because I've seen 0 values above without the wrapping `units()`.

```suggestion
    padding: 0;
```

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+            get-color-token-from-bg(
+              $bg-color: $theme-in-page-nav-background-color,
+              $preferred-text-token: $theme-in-page-nav-bar-color,
+              $context: "In-page navigation bar",
+            )
+          );
+          border-radius: 0;
+          bottom: units(-1px);
+          left: units(0);
+          top: units(0);
```
mejiaj, line 103: Again, for consistency.

```suggestion
          left: 0;
          top: 0;
```

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+          width: units($theme-in-page-nav-bar-width);
+          z-index: z(100);
+        }
+      }
+    }
+
+    .usa-in-page-nav__item {
+      @include u-border(0);
```
mejiaj, line 111: This produces two lines of css, you could simplify with `border: none;`.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+      font-size: font-size($theme-in-page-nav-font-family, "2xs");
+      font-weight: bold;
+      position: relative;
+
+      &.usa-in-page-nav__item--sub-item {
+        font-weight: normal;
+      }
+    }
+  }
+}
```
mejiaj, line 121: Needs empty line at end of file.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+      &.usa-current {
+        @include set-text-from-bg(
+          $bg-color: $theme-in-page-nav-background-color,
+          $context: "In-page-navigation current color"
+        );
+        font-weight: inherit;
+      }
+
```
mejiaj, line 83: There are four empty spaces on line 83.

image

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -0,0 +1,121 @@
...
+    @include at-media("tablet") {
+      display: block;
+      width: 100%;
+    }
+
+    nav {
+      @include set-text-and-bg(
+        $theme-in-page-nav-background-color,
+        $context: "In-page navigation background"
+      );
```
mejiaj, line 46: Empty spaces after semi-colon on 46.

image

mejiaj (commented):
Some great improvements here! Added some comments and questions.

I tested the for:
- [x] ANDI a11y plugin - no errors
- [x] AXE plugin - no errors
- [x] Checked display on site - minor with empty `<p></p>`
- [x] Checked output HTML of component - minor with empty `<p></p>`
- [x] Reviewed JS for issues and possible improvements
- [x] SASS formatting and possible improvements
- [x] Ran NPM test - It seems only `in-page` nav tests ran
  
image

Seeing the same in [CircleCI build](https://app.circleci.com/pipelines/github/uswds/uswds/4357/workflows/a1c6e38a-e36f-4a3c-b8e1-c48ac44a2215/jobs/8778).
image

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
thisisdano: removed

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
thisisdano: Changed to settings

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+    [CLICK]: {
+      [IN_PAGE_NAV_LINK](event) {
+        event.preventDefault();
+        if (this.disabled) return;
+        handleClickFromLink(this);
+      },
+    },
+    keydown: {
+      [IN_PAGE_NAV_LINK]: keymap({
+        Enter: handleEnterFromLink,
```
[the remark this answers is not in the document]
thisisdano: Let's consider this as a possible followup enhancement.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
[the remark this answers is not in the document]
thisisdano: Refactored

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: Changed to a `default` value that defaults to text color

packages/uswds-core/src/styles/settings/_settings-components.scss
thisisdano: And that's what I did!

packages/usa-in-page-navigation/src/test/template.html
```diff
@@ -0,0 +1,23 @@
+<div class="usa-in-page-nav-container">
+  <main id="main-content" class="main-content">
+    <h1>In page navigation test heading</h1>
+
+    <h2><a id="section_0" class="usa-anchor-tag"></a>Section 1</h2>
+    <p></p>
```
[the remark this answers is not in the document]
thisisdano: This should have placeholder content

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+ */
+const handleScrollToSection = (el) => {
+  window.scroll({
+    behavior: "smooth",
+    top: el.offsetTop,
+    block: "start",
+  });
+};
+
+const createInPageNav = (inPageNavEl) => {
```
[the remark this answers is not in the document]
thisisdano: @mitchmoccia Can you add the JSDoc documentation to this code?

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
mitchmoccia: @mejiaj also just pushed a fix for the extra spaces and formatting in the scss... Applied prettier to the file.

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
```diff
@@ -0,0 +1,31 @@
...
+    {% endif %}
+
+    {% if items %}
+      {% for item in items %}
+        {% if item.is_subheading %}
+          <h3>{{ item.section_heading }}</h3>
+        {% else %}
+          <h2>{{ item.section_heading }}</h2>
+        {% endif %}
+        <p>{{ item.content }}</p>
```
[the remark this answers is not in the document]
mitchmoccia: @mejiaj @thisisdano I believe this was happening due the p tags being added to the JSON and also being present in the twig file. I have removed the p tags wrapping the content in the twig file.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
mitchmoccia: The thinking here was that some sites might have a div with the class "main-content". I can remove if you think just main would cover it?

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
mitchmoccia: Yes there is a requirement detailed in the [Using the in-page navigation component](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/mm-inclusive-patterns-in-page-navigation/components/in-page-navigation/) notes that calls for the addition of a wrapper with the class usa-in-page-nav-container.

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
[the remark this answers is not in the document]
mitchmoccia: Confirmed this is fixed when removing the p wrappers in the twig file.

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
thisisdano: But does the CSS need this specificity? We may be able to remove some unnecessary specificity by flattening the code

thisisdano: I will try flattening and reordering

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
thisisdano: `usa-in-page-nav__nav` seems reasonable here

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
mitchmoccia: Done

packages/usa-in-page-navigation/src/index.js
```diff
@@ -0,0 +1,219 @@
...
+const handleClickFromLink = (el) => {
+  const id = getSectionId(el);
+  const allAnchors = getSectionAnchors();
+  const allAnchorsArr = Array.from(allAnchors);
+  const target = allAnchorsArr.find(
+    (anchor) => anchor.getAttribute("id") === id
+  );
+
+  handleScrollToSection(target);
+};
```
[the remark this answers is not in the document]
mitchmoccia: @thisisdano @mejiaj this has been resolved, thanks James!

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
```diff
@@ -82,12 +95,12 @@
 
     .usa-in-page-nav__item {
       @include u-border(0);
+      font-size: font-size($theme-in-page-nav-font-family, "2xs");
+      font-weight: bold;
       position: relative;
 
       &.usa-in-page-nav__item--sub-item {
-        a {
```
mitchmoccia: @thisisdano we lost the second level indent here. Is that preferred? I thought it looked weird with just the top level being bold. Here are screen shots comparing both. Before pushing this back, I wanted to confirm with you, should we leave the indent out? Thanks

image


image

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
mejiaj: @thisisdano @mitchmoccia sounds good, created issue based on this comment.

https://github.com/uswds/uswds/issues/5020

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
mejiaj: Created issue https://github.com/uswds/uswds/issues/5021.

thisisdano (approved):

packages/usa-in-page-navigation/src/styles/_usa-in-page-navigation.scss
thisisdano: Yes, we're leaving the indent out to simplify presentation since this component is restricted to only two levels

PR #4921, opened 2022-08-17
[Test] In page navigation with sidenav

amyleadem:
## Description
Test with both in-page nav and side nav in the same layout:
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/al-in-page-nav/?path=/story/components-in-page-navigation--test

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -166,12 +166,12 @@
       }
     },
     "node_modules/@babel/generator": {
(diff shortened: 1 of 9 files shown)


amyleadem:
Test no longer needed.

PR #4939, opened 2022-09-12
USWDS - JS: Remove domready for component JS init

### LINKED ISSUE #4897: USWDS - Feature: Use DOMContentLoaded instead of domready

Domready seems unmaintained and causes issues with JS initializing in storybookjs. We should use a modern technique for initializing JS when DOM is ready. It would help us:

1. Remove an unmaintained dependency
2. Use modern JS techniques
3. Improve how JS works in component library

Check for `DOMContentLoaded`

Alternative recommendation to use `defer` attribute with JS.

### Additional context

Related to
https://github.com/uswds/uswds/issues/4778

### Code of Conduct

mejiaj:
## Description

**Use native DOMContentLoaded insead of domready.** [Crossbrowser support [caniuse.com]](https://caniuse.com/domcontentloaded) is pretty good and removes an old IE dependency. Closes #4897.

## Additional information

Domready was a dependency that is no longer maintained. See last publish of 7 years ago on NPMJS - https://www.npmjs.com/package/domready.  It gave support to IE6-IE8 browsers to check if the DOM was ready to initialize javascript.

This PR replaces that legacy dependency with a native method to check if document is ready (`DOMContentLoaded`). 

#### Component initialization

~This also fixes the issue with JS not initializing as expected in Storybook. Previously, we had to refresh several times to get interactive components to work.~

The issue with Storybook initializing still exists and is reported see https://github.com/storybookjs/storybook/issues/15753.  The solution in this [comment](https://github.com/storybookjs/storybook/issues/15753#issuecomment-915142145) gave the best results (most components working), but there was an issue with the way modal is initialized that made it unusable.

#### New entrypoint

I've changed the entrypoint in Storybook's `preview.js` from the compiled `uswds.min.js` to `start.js`. This allows us to refresh and view changes without having to run `npx gulp compile` on change.

### Local development

- Open `package.json` and confirm `domReady` is removed from `dependencies` field
- Run a clean `npm install`
- Run `npm start`
- Confirm interactive components require a refresh
- Modify component JS and refresh  —  `npx gulp buildJS` should no longer be required for local development

### For users

- Run `npx gulp buildJS`
- Use compiled JS (`uswds.js` or `uswds.min.js`)
- Components should **still** initialize as expected in modern browsers (domReady included support for IE9+, Firefox 4+, Safari 3+)

---

### DIFF
--- a/packages/uswds-core/src/js/start.js
+++ b/packages/uswds-core/src/js/start.js
@@ -1,5 +1,3 @@
-const domready = require("domready");
-
 window.uswdsPresent = true; // GLOBAL variable to indicate that the uswds.js has loaded in the DOM.
 
 /**
@@ -15,7 +13,7 @@ const svg4everybody = require("./polyfills/svg4everybody");
 
 uswds.components = components;
 
-domready(() => {
+window.addEventListener("DOMContentLoaded", () => {
   const target = document.body;
   Object.keys(components).forEach((key) => {
     const behavior = components[key];
--- a/packages/uswds-core/src/js/start.js
+++ b/packages/uswds-core/src/js/start.js
@@ -20,7 +20,7 @@ const initComponents = () => {
     behavior.on(target);
   });
   svg4everybody();
-}
+};
 
 window.addEventListener("DOMContentLoaded", initComponents, { once: true });
 
--- a/packages/uswds-core/src/js/start.js
+++ b/packages/uswds-core/src/js/start.js
@@ -13,13 +13,16 @@ const svg4everybody = require("./polyfills/svg4everybody");
 
 uswds.components = components;
 
-window.addEventListener("DOMContentLoaded", () => {
+const initComponents = () => {
   const target = document.body;
   Object.keys(components).forEach((key) => {
     const behavior = components[key];
     behavior.on(target);
   });
   svg4everybody();
-});
+}
 
-module.exports = uswds;
+window.addEventListener("DOMContentLoaded", initComponents, { once: true });
+
+exports.default = uswds;
+exports.initComponents = initComponents;
--- a/packages/uswds-core/src/js/start.js
+++ b/packages/uswds-core/src/js/start.js
@@ -22,7 +22,11 @@ const initComponents = () => {
   svg4everybody();
 };
 
-window.addEventListener("DOMContentLoaded", initComponents, { once: true });
+if (document.readyState === "loading") {
+  document.addEventListener("DOMContentLoaded", initComponents, { once: true });
+} else {
+  initComponents();
+}
 
 exports.default = uswds;
 exports.initComponents = initComponents;
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,4 +1,4 @@
-import "../dist/js/uswds.min.js";
+import "../packages/uswds-core/src/js/start";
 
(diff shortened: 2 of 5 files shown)


mejiaj:
Unit tests passed, but a11y checks failed. Looking into it.

packages/uswds-core/src/js/start.js
```diff
@@ -15,13 +13,16 @@ const svg4everybody = require("./polyfills/svg4everybody");
...
   const target = document.body;
   Object.keys(components).forEach((key) => {
     const behavior = components[key];
     behavior.on(target);
   });
   svg4everybody();
-});
+};
+
+window.addEventListener("DOMContentLoaded", initComponents, { once: true });
```
mejiaj, line 25: Passing `once` is a good practice. Helps with Storybook because otherwise it would fire multiple times on load **or** switching between components.

packages/uswds-core/src/js/start.js
```diff
@@ -15,13 +13,16 @@ const svg4everybody = require("./polyfills/svg4everybody");
...
 
-module.exports = uswds;
+exports.default = uswds;
+exports.initComponents = initComponents;
```
mejiaj, line 28: uswds is the default export, but adding the `initComponents` function in case we want to re-trigger component initialization in the future.

For example: if we want to add a custom event for Storybook loading (different than DOM ready) and re-initialize _all_ components.

amyleadem (approved):
Everything seems to work as expected. Looks good to me!

thisisdano (commented):
This looks good and just needs a merge conflict resolved

packages/uswds-core/src/js/start.js
aduth, line 25: Not sure how much of a problem it would be, but `domready` also handled a scenario where the page is already loaded by the time the script is run, by checking [`document.readyState`](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState). Should that be accounted for in this replacement implementation?

It may only be an issue if the script was added after the initial page load, and I'm not sure how common that would be.

thisisdano: Thanks for the review, @aduth. It does seems like a rare situation, but it could be worth duplicating the functionality we're replacing.

mejiaj: @aduth I thought `DOMContentLoaded` and `document.readyState` were interchangeable, but I understand the concern.

> It may only be an issue if the script was added after the initial page load, and I'm not sure how common that would be.

Like adding the `start.js` or `uswds.js` script dynamically?

aduth: Yeah, the worry is that if `DOMContentLoaded` is only fired once at initial page load, if the script containing this code was added any time after that (for whatever reason), the component initialization would never be run because the event would never be fired again.

mejiaj: Updated based on this [MDN example](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete)

And created a very simple test in https://github.com/mejiaj/gulp-sandbox/tree/jm-test-uswds-replace-domready. Should work the same now based on your suggestion, thanks for taking a look!

thisisdano (approved):

PR #4955, opened 2022-09-20
USWDS - Tooltip: dynamic initialization

### LINKED ISSUE #4055: Allow for dynamic initialization of tooltip component

mejiaj:
## Description

**Tooltip initializes on first interaction.** Tooltip now checks if it has been initialized before toggling. If it hasn't initialized properly then it will call a setup attributes function. Closes #4055.

## Additional information

Component setup only occurred on document ready **or** when manually calling the `on/off` method on a specific element. This caused issues because:

1. Dynamically created components never initialized properly
2. Event listeners weren't attached properly

This PR does the following: 

1. Init **only** scaffolds component https://github.com/uswds/uswds/pull/4955/commits/e0a3cb48a1c86d790d23eee9315a92b4375e6947
2. Use behaviors to add event listeners that bubble https://github.com/uswds/uswds/pull/4955/commits/eb4c4d0c44b4a12eee088c4c30b29ea42b905147#diff-8d7e7a60d39759da3b847a457caf221ad9251a8e84ab2356905ac7e9ba36cf4cR372-R383
3. Scaffold component on first interaction (if it hasn't already  —  helps with dynamically added components) [92d15e3](https://github.com/uswds/uswds/pull/4955/commits/92d15e3f09ce59f78432c34eee103d8a4bd19f68)
4. Exports methods users can hook into (gives users more control over functionality) [585d2b3](https://github.com/uswds/uswds/pull/4955/commits/585d2b3af9e30056e5b7406fdc8046a19ead076f)
5. Updates JS entrypoint for Storybook  —  avoids building JS _every_ change, but still requires refresh

I've also removed the title attribute from tooltip trigger, instead of leaving it empty.

## How to test
- [ ] Confirm tooltip still functions as expected
- [ ] Create a tooltip in JS (added mine via devtools console) and insert into DOM ★
- [ ] Confirm tooltip works as expected

★ Dynamic tooltip example

```js
// Create trigger based on recommended markup from Tooltip page
const dynamicTooltip = document.createElement("button");

// Add attributes
dynamicTooltip.setAttribute("type", "button");
dynamicTooltip.setAttribute("data-position", "right");
dynamicTooltip.setAttribute("title", "A dynamic tooltip");
dynamicTooltip.classList.add("usa-button", "usa-tooltip");

// Let's give our tooltip trigger some text
dynamicTooltip.textContent = "My dynamic tooltip";

// Insert into DOM
document.body.appendChild(dynamicTooltip);
```

---

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,4 +1,6 @@
-import "../dist/js/uswds.min.js";
+// import "../dist/js/uswds.min.js";
+
+import "../packages/uswds-core/src/js/start";
 
 export const parameters = {
   actions: { argTypesRegex: "^on[A-Z].*" },
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -386,10 +386,10 @@ const tooltip = behavior(
         setUpAttributes(tooltipTrigger);
       });
     },
-    setUpAttributes,
+    setup: setUpAttributes,
     getTooltipElements,
-    showToolTip,
-    hideToolTip
+    show: showToolTip,
+    hide: hideToolTip,
   }
 );
 
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -33,8 +33,8 @@ const addListenerMulti = (element, eventNames, listener) => {
  * @returns {object} Elements for initialized tooltip; includes trigger, wrapper, and body
  */
 const getTooltipElements = (trigger) => {
-  const wrapper = trigger.closest(TOOLTIP);
-  const body = wrapper.querySelector(`.${TOOLTIP_BODY_CLASS}`)
+  const wrapper = trigger.parentNode;
+  const body = wrapper.querySelector(`.${TOOLTIP_BODY_CLASS}`);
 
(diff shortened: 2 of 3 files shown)


amyleadem (approved):
Very cool fix! I was able to dynamically add a tooltip that scaffolds itself and works as expected. Additionally, the tooltip works as expected with both mouse and keyboard in Safari, Firefox, Chrome and with touch in iOS and VoiceOver Gestures.

thisisdano (approved):
Such a good improvement and a model for other components!

jakerella:
❤️ THANK YOU!

PR #4963, opened 2022-09-27
USWDS - Header: Secondary nav a11y improvements

### LINKED ISSUE #4944: USWDS - Bug: secondary nav link separators are visible to screen readers
### Describe the bug

Secondary links line separator is being read by screen readers as follows: “Link About Vertical Bar, Link Subscribe Vertical Bar…"

1. Navigate to digital.gov homepage
2. Use JAWS or ChromeVox to read the navigation

### Expected Behavior

Screen reader should read “Link About, Link Subscribe …"

### Related code

We could style with `border-left` instead of `content`.

https://github.com/uswds/uswds/blob/8d2c7716890606e342877e916d1622eb8f1eaa0b/packages/usa-nav/src/styles/_usa-nav.scss#L345-L349

### Screenshots

Secondary navigation within the primary navigation.

Screen Shot 2022-09-14 at 4 34 26 PM

### System setup

This occurs on both PC and Mac with Chrome browser with USWDS 3.0

### Additional context

_No response_

### Code of Conduct

mejiaj:
## Description

**Secondary nav separator is no longer read on screenreaders.** The `|` pseudoelement is no longer read on screen readers. Closes #4944.

[Preview link](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-a11y-nav-separator/iframe.html?args=&id=components-header--extended&viewMode=story)

## Additional information

This solution converts the separator from a pseudoelement to a border. This simplifies styles and prevents the screenreaders from reading this element.

### Old
Header extended in Chrome.
image

### New

**Chrome**
image

**Firefox**
image

**Safari**
image

<details>
  <summary>Previous solution</summary>

⚠️ This solution **won't** work because Firefox fails to render the separator correctly. See [comment below](https://github.com/uswds/uswds/pull/4963#issuecomment-1413851140) for more information.

**Firefox**
image

---

  The separator uses an empty string for the pseudoelement's alt text.

```
content: "\007C" / "";
```

Based on [W3 Guidance](https://www.w3.org/TR/css-content-3/#alt)

image

- Use entity code for pipe character (More consistent rendering)
- Add empty alt text for pseudoelement to prevent character being read on screenreaders 

</details>

---


thisisdano:
Caniuse indicates this [this feature doesn't have great support yet](https://caniuse.com/mdn-css_properties_content_alt_text) — perhaps since it's still functionality that's only in [a draft spec](https://drafts.csswg.org/css-content-3/#alt)?

That said, it [this issue in the JAWS repo](https://github.com/FreedomScientific/VFO-standards-support/issues/367) makes it look like JAWS _does_ support it?

mejiaj:
@thisisdano we were able to test in JAWS thanks to @amycole501. She confirmed this fixes the issue.

I'm happy to change it to border if we want to avoid introducing features that are still in draft.

mejiaj:
Firefox doesn't render correctly, so this solution won't work.

From @nick-mon1:
>### Question
>Firefox says this is an invalid value and the override fails. If I remove / "" then it works.
>Any suggestions for Firefox support?

https://github.com/GSA/digitalgov.gov/pull/5763

packages/usa-nav/src/styles/_usa-nav.scss
```diff
@@ -342,10 +342,10 @@ $expand-less-icon: map-merge(
...
       padding-left: units(0.5);
 
-      & + .usa-nav__secondary-item::before {
-        color: color("base-lighter");
-        content: "|";
-        padding-right: units(0.5);
+      & + .usa-nav__secondary-item {
+        border-left: 1px solid color("base-lighter");
+        margin-left: px-to-rem(5px);
+        padding-left: units(1);
```
mejiaj, line 348: # Alternative

Alternate solution keeps the pseudo element, but creates the shape. I liked that it's similar to the current, but it's more CSS.

```scss
      & + .usa-nav__secondary-item::before {
        background-color: color("base-lighter");
        display: inline-block;
        width: units("1px");
        height: units(2);
        content: "";
        margin-right: units(0.5);
        vertical-align: middle;
      }
```

mahoneycm: I noticed that this solution causes some slight content shift brought on by the `height` setting. Here's a video of me clicking back and forth between the development build and local


https://user-images.githubusercontent.com/25211650/216679764-2c1de705-f81b-4187-bd30-41877693a7a2.mov

mahoneycm (approved):
I think either solution would work! I like trying to keep the original border height but the benefit of less css and not having the content shift slightly might be the better move forward.

packages/usa-nav/src/styles/_usa-nav.scss
```diff
@@ -342,10 +342,10 @@ $expand-less-icon: map-merge(
       display: inline;
       padding-left: units(0.5);
 
-      & + .usa-nav__secondary-item::before {
-        color: color("base-lighter");
-        content: "|";
-        padding-right: units(0.5);
+      & + .usa-nav__secondary-item {
+        border-left: 1px solid color("base-lighter");
+        margin-left: px-to-rem(5px);
```
amyleadem, line 347: @mejiaj 
I am curious about why the value is `5px` here instead of matching the `padding-left` value of `units(1)`. I did a little digging, and when I set the `margin-left` to `units(1)`, I saw that the horizontal alignment between the elements was a bit off. Is this the reason for the `5px`? If so, the misalignment seemed to resolve itself when I replaced `float:right` with `display:flex` on `.usa-nav__secondary-links`. Doing so also had the added benefit of making the border height a little smaller. What do you think? Am I making too many assumptions here?

amyleadem (changes requested):
I like the border solution. I think it is a nice, simple way to approach it. I had one question about some CSS. Let me know what you think.

amycole501 (commented):
Using JAWS, I didn't hear any indication of the pipe being read aloud. I tried in Chrome, Firefox and Edge. It does, though, say "Blank" instead. Not ideal but it is better than describing the pipe as a vertical line.

mejiaj:
# Additional updates

- Removed an unnecessary `float: right` [[ef09fcb](https://github.com/uswds/uswds/pull/4963/commits/ef09fcb5e79574b58d7e74095a7072836016eec3)]
- Use flexbox and column gap for more consistent spacing (no magic numbers like `5px`) [73dacc4](https://github.com/uswds/uswds/pull/4963/commits/73dacc491ab001997cfa5238fa1a3e73ba114a55)

amyleadem (approved):
Approved!

thisisdano (approved):
What a smart solution!

PR #5008, opened 2022-10-26
usa-header--extended usa-logo text width: Fixed to work with theme setting

RSD-accessboard:
## Description

USWDS documentation on header says there is a theme setting to adjust the logo text width. It does not work with header--extended. This fixes that. 
      max-width: $theme-header-logo-text-width;

[USWDS settings](https://designsystem.digital.gov/documentation/settings/) 

Verify where else this variable is used and if the default value needs to be updated. Documentation says default value is 33%, header-extended uses default value 50%.


mejiaj (approved):
Looks good, thanks for making this change!

Tested extended header (desktop width) in:
- [x] Firefox
- [x] Chromium
- [x] Safari 

**Safari**
image

**Chromium**
image

**Firefox**
image

PR #5031, opened 2022-11-09
Query the DOM for .usa-modal-wrapper before building or destroying mo…

rachidatecs:
## Description

Query the DOM for .usa-modal-wrapper before building or destroying modals. This is useful for JS frameworks (React, Vue, Angular) where modal components are dynamically created, to avoid extra inits and bad teardown calls

### DIFF
--- a/packages/usa-modal/src/index.js
+++ b/packages/usa-modal/src/index.js
@@ -303,10 +303,12 @@ modal = {
   focusTrap: null,
   toggleModal,
   on(root) {
-    this.init(root);
+    if (!document.querySelector(".usa-modal-wrapper"))
+      this.init(root);
   },
   off(root) {
-    this.teardown(root);
+    if (document.querySelector(".usa-modal-wrapper"))
+      this.teardown(root);
   }
 };
 


mahoneycm:
Hello @rachidatecs ,

We were unable to replicate the bad inits and teardowns issue (#5062) that this should resolve, preventing us from doing further testing.

The direction for this solution doesn't follow existing USWDS component patterns. If we were to move forward with a similar solution, let's look at how we can include this check within the Modal logic, instead of within `on / off` or initialization functionality

mahoneycm (commented):
Hello @rachidatecs ,

We were unable to replicate the bad inits and teardowns issue (#5062) that this should resolve, preventing us from doing further testing.

The direction for this solution doesn't follow existing USWDS component patterns. If we were to move forward with a similar solution, let's look at how we can include this check within the Modal logic, instead of within `on / off` or initialization functionality

mahoneycm:
This work is being continued in #5315. We'll continue to track the work there. Closing this ticket in the meantime! Thank you for your contribution here @rachidatecs!

Please reach out if you have any further questions!

Issue #5030, opened 2022-11-09
USWDS - Feature: Add customizable data- attribute for in-page nav to specify which heading tags should be used


mitchmoccia:
### Is your feature request related to a problem? Please describe.

Currently, the in-page nav component searches a page for h2 and h3 tags and dynamically builds the nav links based on the content of those tags. It also makes h2-based, created links bold while making h3-based links normal weight.

### Describe the solution you'd like

We could offer a way to easily customize the way in-page nav searches for page headings to dynamically build the navigation block. What if a developer only wanted to use h2 tags? e.g.:

```
data-headings="h2"

```

If no data- attribute is present, the default one in the JS would be used.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Feature+Request%22) for duplicate feature requests.

PR #5034, opened 2022-11-10
In-page nav - Add customizable data- attribute for which heading[s] gets queried

mitchmoccia:
This implements the feature detailed here:
https://github.com/uswds/uswds/issues/5030

This is an important one that will allow for the customization of what headings should be queried when the in-page nav is being dynamically built.

A developer will be able to easily add the data- attribute `data-headings` to the aside element in the HTML to define which headings get queried.

Some examples include:
`data-headings="h2"`
`data-headings="h2 h3"`
`data-headings="h2 h3 h4"`
`data-headings="h1 h2"`

**The headings should be listed with a space in between.**

NOTE: This is most likely needed ASAP for the USWDS documentation site since including all h3s in the query is problematic on some of the pages... It might be better to use `data-headings="h2"` so only h2s are built in the in-page nav. Let's discuss.

### DIFF
--- a/packages/usa-in-page-navigation/src/index.js
+++ b/packages/usa-in-page-navigation/src/index.js
@@ -8,6 +8,7 @@ const { CLICK } = require("../../uswds-core/src/js/events");
 const CURRENT_CLASS = `${PREFIX}-current`;
 const IN_PAGE_NAV_TITLE = "On this page";
 const IN_PAGE_NAV_HEADING_LEVEL = "h4";
+const IN_PAGE_NAV_OFFSET_TOP = 0;
 const IN_PAGE_NAV_CLASS = `${PREFIX}-in-page-nav`;
 const IN_PAGE_NAV_ANCHOR_CLASS = `${PREFIX}-anchor`;
 const IN_PAGE_NAV_NAV_CLASS = `${IN_PAGE_NAV_CLASS}__nav`;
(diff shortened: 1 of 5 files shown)


packages/usa-in-page-navigation/src/index.js
```diff
@@ -51,12 +53,20 @@ const setActive = (el) => {
...
  *
+ * @param {String} headings A string of one or more headings to query for
  * @return {HTMLElement[]} - An array of DOM nodes
  */
-const getSectionHeadings = () => {
-  const sectionHeadings = document.querySelectorAll(
-    `${MAIN_ELEMENT} h2, ${MAIN_ELEMENT} h3`
-  );
+const getSectionHeadings = (headings) => {
+  const headingList = headings.indexOf(" ") ? headings.split(" ") : headings;
```
mejiaj, line 60: Is this `headings.indexOf(" ")` to check for multiple values?

mejiaj: Should we validate any of the values passed into `data-headings`? Or at least return an error if a user tries to pass an invalid header _ex: j1 j2 instead of h1 h2_ ?

For example, validate against a list of expected values? Could be a `const VALID_HEADINGS = ["h1", "h2", "h3", "h4", "h5", "h6"]`

mejiaj (commented):
Looks good overall, just a question on the new `data-headings` setting.

packages/usa-in-page-navigation/src/index.js
mitchmoccia: Done! Outstanding suggestion, made it so, and it throws an error like the screen shot below if they use something other than a valid heading tag:

image

mitchmoccia: > Is this `headings.indexOf(" ")` to check for multiple values?

Yes, I know they could try to use some other delimiter like a comma, but was thinking we could detail this in the guidance... And with the valid headings check implemented, we should be good? What do you think?

thisisdano:
Holding on this one, pending further discussion

packages/usa-in-page-navigation/src/index.js
mejiaj: That might not be necessary, but that's a good question. I feel like it's good enough to throw an error once the expected heading element is invalid.

FWIW; I had thought something similar like passing a value `[h1, h2, h3]` to the data-attribute for clarity. Unfortunately, it doesn't make things any easier for anyone.

amyleadem:
Moving this work into https://github.com/uswds/uswds/pull/5444. Thanks for getting this started, @mitchmoccia!

PR #4576, opened 2022-03-11
Beta: Repair typography vertical margins

amyleadem:
### Description:
Repair typography vertical margin settings errors found in https://github.com/uswds/uswds/issues/4574

### Solution notes:
Discovered that pulling in the typeset-p/typeset-heading (or usa-prose-p/usa-prose-heading) styles as a mixin instead of a placeholder compiles as `*+.usa-prose>h2` instead of `.usa-prose>*+h2`, which results in incorrect visual display. 

Closes https://github.com/uswds/uswds/issues/4574

PR #5043, opened 2022-11-15
[Test] Prose placeholder fix

### LINKED ISSUE #4962: USWDS - Bug: Unable to find %usa-prose-p selector when importing uswds-typography in one file and uswds-core in another
### Describe the bug

I’m trying to configure USWDS inside an `@layer` rule using the `load-css` function from `sass:meta`. When I try to `@use "uswds-core"` in my own Sass modules so that I can access the tokens for my styles, I get a `target selector not found` error for `%usa-prose-p`.

1. Create a `_uswds-theme.scss` module [per the documentation](https://designsystem.digital.gov/documentation/settings/)
2. Create a `_uswds.scss` that forwards the theme file and `uswds-typography`
3. Create another Sass file with `@use "uswds-core" as uswds` to define a rule that uses a USWDS token
4. Create a Sass file that imports both `_uswds.scss` and your third Sass file using `load-css`
5. Build Sass file number 4

### Expected Behavior

The output should contain the styles from `uswds-typography` as well as whatever styles you added to your Sass file in step 3.

### Related code

I have a demo of the bug here: https://github.com/esheehan-gsl/uswds-layer-test

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

amyleadem:
## Summary
Fixed compilation issues related to `usa-prose-` placeholders. 

## Breaking change
:warning: This is potentially a breaking change. This PR removes `usa-prose-` mixins and placeholders. Users should instead use `typeset-` mixins in their place. 

## Preview link
Preview link: [Prose component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-prose-fix/?path=/story/components-prose--default)

## Problem statement
From https://github.com/uswds/uswds/issues/4962: 
> I’m trying to configure USWDS inside an @layer rule using the load-css function from sass:meta. When I try to @use "uswds-core" in my own Sass modules so that I can access the tokens for my styles, I get a target selector not found error for %usa-prose-p.

## Solution
Sass seems to get confused when `* + &` is brought in from another file via a mixin (more details in https://github.com/uswds/uswds/pull/4576). Additionally, Sass can throw errors in limited circumstances when we rely on `@extend` and placeholders.

The key conclusion is that `* + &` does not currently seem to be a viable addition to mixins or placeholders. By moving all `* + &` out of mixins and placeholders and directly into the style definitions, we can remove the risk of Sass confusion.

Note: The downside of this solution is that it requires some code repetition, but that might be slightly mitigated by creating variables/settings for the heading and paragraph top margins. 

## Tasks completed
1. Moved * + & references out of mixins/placeholders and directly into the style definitions.
1. Removed redundant and unused `usa-prose-` mixins and replaced them with `typeset-` mixins. 

## Testing and review
1. Confirm that Sass compiles without error
1. Confirm that 

This will need more testing, but I experimented with moving `* + &` references out of mixins/placeholders and directly into the style definitions. 

This was done because:
1. Sass seems to get confused when `* + &` is brought in from another file via a mixin (more details in https://github.com/uswds/uswds/pull/4576).
3. Sass can throw errors in limited circumstances when we rely on `@extend` and placeholders.

This PR represents only one possible path for moving forward. The downside is that it requires some code repetition, but that might be slightly mitigated by creating variables/settings for the heading and paragraph top margins. There are likely other CSS/Sass solutions that can generate the same result. 

Additionally, this PR in its current state removes unused mixins and placeholders, but we might want to retain those in case users might be calling these items in their projects.

The key conclusion is that `* + &` does not currently seem to be a viable addition to mixins or placeholders.


packages/uswds-core/src/styles/mixins/typography/usa-prose.scss
```diff
@@ -5,58 +5,3 @@
 @use "./headings" as *;
 @use "./typeset" as *;
 
-@mixin usa-prose-link {
```
amyleadem, line 8: The usa-prose mixins are all redundant with the typeset mixins

packages/usa-content/src/styles/_usa-content.scss
```diff
@@ -7,7 +7,3 @@
     max-width: measure($theme-text-measure);
   }
 }
-
-%usa-heading {
-  @include typeset-heading;
-}
```
amyleadem, line 13: Removed unused items

packages/usa-paragraph/src/styles/_usa-paragraph.scss
```diff
@@ -1,9 +1,9 @@
...
-%usa-paragraph {
-  @include typeset-p;
-}
-
 .usa-paragraph {
   @include typeset-p;
+
+  * + & {
+    margin-top: 1em; // TODO: add as var to settings?
+  }
```
amyleadem, line 8: `* + &` gets confused when added to mixins, so moved these items directly inside the style rules.

packages/usa-prose/src/styles/_usa-prose.scss
```diff
@@ -9,4 +9,19 @@
...
+    h4,
+    h5,
+    h6 {
+      margin-top: 1.5em;
+    }
+
+    p {
+      margin-top: 1em;
+    }
+  }
```
amyleadem, line 26: `* + &` gets confused when added to mixins, so moved these items directly inside the style rules.

packages/uswds-core/src/styles/mixins/typography/typeset.scss
```diff
@@ -39,10 +39,6 @@ Sets:
   @include u-margin-y(0);
   clear: both;
 
-  * + & {
-    margin-top: 1.5em; // TODO: add as var to settings?
-  }
-
```
amyleadem, line 45: `* + &` gets confused when added to mixins, so moved these items directly inside the style rules.

packages/uswds-core/src/styles/mixins/typography/typeset.scss
```diff
@@ -55,10 +51,6 @@ Sets:
   margin-top: 0;
   max-width: measure($theme-text-measure);
 
-  * + & {
-    margin-top: 1em; // TODO: add as var to settings?
-  }
-
```
amyleadem, line 61: `* + &` gets confused when added to mixins, so moved these items directly inside the style rules.

packages/uswds-core/src/styles/mixins/typography/usa-content-styles.scss
```diff
@@ -3,17 +3,16 @@
 @use "./usa-prose" as *;
 @use "./usa-list" as *;
 @use "./usa-table-styles" as *;
-@use "../../placeholders/typography" as *;
```
amyleadem, line 6: Removed this file and the placeholders inside

packages/uswds-core/src/styles/mixins/typography/usa-content-styles.scss
```diff
@@ -3,17 +3,16 @@
...
 
 @mixin usa-paragraph-style {
   p {
-    @extend %usa-prose-p;
+    @include typeset-p;
```
amyleadem, line 9: The usa-prose mixins are all redundant with the typeset mixins, so I replaced them here.

packages/uswds-core/src/styles/placeholders/_index.scss
```diff
@@ -1,3 +1,2 @@
 @forward "forms";
 @forward "list";
-@forward "typography";
```
amyleadem, line 3: Removed this file and the placeholders inside

packages/usa-prose/src/styles/_usa-prose.scss
mahoneycm: I came to a similar conclusion! I was able to get the same desired effect by placing these rules within the `& > {}` selector above it.

I `npm link`ed this to the [test repo](https://github.com/esheehan-gsl/uswds-layer-test) for #4962 and it build correctly 👍🏼

mahoneycm:
I used `npm link` to link this PR to a site branch and built the site. The styles output appropriately and matched the live site. I also tried linking it on the [test repo](https://github.com/esheehan-gsl/uswds-layer-test) for #4962 and it built correctly!

@mejiaj  I was going to offer to put up a PR with this change but I don't think the link will transfer. It would get overwritten if `npm install` was ran at least.

mahoneycm (commented):
So all the styles appear to be outputting as intended. I did notice that _only in storybook_ the source for the prose styles says `button-unstyled.scss`

image


But clicking through to the linked stylesheet just looks like the regular button-unstyled page with no mention of prose.

When I link it to site and inspect the output style is linked it `_usa-prose` as expected. I changed the value to ensure that the styles were updating as I expected and it all worked fine.

image

@amyleadem  What do you make of this?

amyleadem:
Closing because it is no longer needed. The official fix will be handled in https://github.com/uswds/uswds/pull/5158

PR #5049, opened 2022-11-21
USWDS: Modal with nested forms. Resolves #4953

### LINKED ISSUE #4953: USWDS - Bug: Date Picker component doe not function inside modal
### Describe the bug

When a Date Picker component is inside a modal, nothing happens upon clicking the calendar icon.

1. Create a modal with a form containing a Date Picker component
2. Click Date Picker calendar icon
3. Nothing happens:-!

### Expected Behavior

All form components should function inside a modal even if the component is post-rendered. Clicking the Date Picker calendar icon inside a modal should still cause the calendar to pop-up and allow a date to be selected.

### Related code

https://codepen.io/hobbesz/pen/ZEoLjYw

### Screenshots

_No response_

### System setup

- USWDS: 3.1.0

### Additional context

_No response_

### Code of Conduct

mahoneycm:
[Preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-modal-nested-components/iframe.html?args=&id=components-modal--nested-forms&viewMode=story)

## Description

Resolves #4953 

In its current iteration, the modal component watches all clicks while the modal is open and looks to see if it contains the `CLOSER_ATTRIBUTE` to trigger the modal to close. If it does not contain this attribute, it stops propagation which causes any other click events to fail.

As pointed out in #4953, this would cause some form elements to not function properly if they are nested in a modal. Consider login/registration pop-ups or scheduling assistants.

Based on the blame for that line of code, it appears the original modal creator tried to remove the stop propagation before but needed to maintain it because of a mobile safari bug. I'm not positive of what the bug was at this time but I believe [this issue](https://stackoverflow.com/questions/21029660/stoppropagation-on-ipad-iphone-in-safari) on stack overflow may be related. Here the user has a modal that closes when the input is tapped.

In this PR I've removed the `stopPropagation` function and created a modal with two form elements within it. After testing on various browsers and versions of mobile safari, it appears to work as expected.

Browsers Tested:
- [x] Various iPhones with different Mobile Safari versions (Latest is v16.0, oldest tested was v10.3)
- [x] Desktop browsers(Chrome, Safari, Edge)
- [x] Various Android devices and browsers (Chrome, Firefox, Samsung Browser) 

## To Test

- Visit Nested Forms modal page
- Open Modal
- Test drop down components
- Click out of modal view or use closer buttons to close

Currently pressing the escape key closes the entire modal, not just the focused element. We can discuss if we'd like to change this moving forward

## Screenshots

Screen Shot 2022-11-21 at 12 09 41 PM
Screen Shot 2022-11-21 at 12 09 57 PM


amyleadem (commented):
Nice work! The writeup is excellent and the fix feels clean. Thanks for building out a new  story for demo and testing. This is all super helpful! 

Everything seems to be working as expected in my browser testing. What do you think about adding a unit test for this functionality?

Also, I do think that there is an opportunity to improve  the `esc` functionality here. I think it would be best if pressing `esc` while the form elements are open should close only the form element, not the modal. This could be out of scope for this issue though and might require further discussion. 
 
Tested in:
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] iOS Safari
- [x] Edge for mac
- [x] VoiceOver
- [x] VoiceOver Gestures

mejiaj (approved):
Looks good! 

Works on Safari iPhone 6 with iOS 13.4 (browserstack). I also tried testing this on an iPhone 5, but it didn't support JS template literals (`).

mahoneycm:
@amyleadem @mejiaj Created a unit test for nested forms using a combo box! Passing as expected currently and breaks with `stopPagination()`

packages/usa-modal/src/test/modal-nested-form.spec.js
```diff
@@ -0,0 +1,161 @@
...
+      it("makes all other page content invisible to screen readers", () => {
+        const activeContent = document.querySelectorAll(
+          "body > :not([aria-hidden])"
+        );
+
+        assert.strictEqual(activeContent.length, 1);
+        assert.strictEqual(activeContent[0], modalWrapper);
+      });
+
+      it("should show the list by clicking the toggle button", () => {
```
amyleadem, line 123: Wondering if this could be more clear. Maybe something like "Displays combobox list when clicked"?

packages/usa-modal/src/test/modal-nested-form.spec.js
```diff
@@ -0,0 +1,161 @@
+const assert = require("assert");
```
amyleadem, line 1: I am curious why a new `spec.js` file and `template` were created for nested forms. The files seem very similar. Would it be possible to just add these updates and the ones from `modal-nested-form.template.html` to `modal.spec.js` and `test/template.html` instead of creating new files?

amyleadem (changes requested):
Thanks for adding the unit test! The component and test both appear to work as expected. I had a couple of comments about the [file structure](https://github.com/uswds/uswds/pull/5049/files#r1063803315) and the [test description](https://github.com/uswds/uswds/pull/5049/files#r1063801517), but other than that I think it is looking good.

packages/usa-modal/src/test/modal-nested-form.spec.js
mahoneycm: Oh yeah I copied this from the combo box test, I'll update this to be more clear 😁

packages/usa-modal/src/test/modal-nested-form.spec.js
mahoneycm: I believe I had a discussion with @mejiaj about the test and we mentioned creating a new test with the nested form. I also referenced combo box which has a different template and test for each of its variants.

Happy to make this the sole test if we prefer to do it all in one sweep!

amyleadem: @mejiaj do you have a preference on this?

amyleadem (approved):
Nice work! I approve, but am happy to defer if @mejiaj weighs in with a preference for multiple test files as described in [this thread](https://github.com/uswds/uswds/pull/5049#discussion_r1063803315).

packages/usa-modal/src/test/modal-nested-form.spec.js
mejiaj: @amyleadem @mahoneycm initially I suggested this as it's own test to keep the template and code simple. It looks like the test template and code are pretty much identical. In that case it makes sense to combine into one.

I'm okay if we simplify the test code to bare minimum or combine them into a single `modal.spec.js`.

mahoneycm:
@thisisdano Ready for final review!

mejiaj:
@mahoneycm going to move this back to review since I haven't had a chance to look over the requested changes.

packages/usa-modal/src/usa-modal.stories.js
```diff
@@ -15,3 +15,6 @@ Large.args = LargeContent;
 
 export const ForcedAction = Template.bind({});
 ForcedAction.args = ForcedActionContent;
+
+export const NestedForms = Template.bind({});
+NestedForms.args = NestedFormsContent;
```
mejiaj, line 20: Can we mark this as a test so there's no confusion between this and modal variants?

I'd delete the JSON file for it and give this the same treatment as our other test stories.

packages/usa-modal/src/usa-modal.twig
```diff
@@ -17,6 +17,34 @@
         {% if description.text -%}
         <div class="usa-prose">
           <p id="{{ description.id }}">{{ description.text }}</p>
+          {% if nestedForms -%}
```
mejiaj, line 20: Let's delete and create a simpler twig test template.

packages/usa-modal/src/test/modal.spec.js
```diff
@@ -25,22 +28,30 @@ tests.forEach(({name, selector: containerSelector}) => {
     let openButton1;
     let openButton2;
     let overlay;
+    let comboList;
+    let CBToggle;
```
mejiaj, line 32: Let's use camel casing and use a more descriptive name.

mejiaj (changes requested):
Functionality and test is good! Added some comments so this work is more in line with other component tests.

packages/usa-modal/src/usa-modal.stories.js
mahoneycm: Resolved in https://github.com/uswds/uswds/pull/5049/commits/64e549d73e429c31e1391559272b917e1ddc8b25

packages/usa-modal/src/usa-modal.twig
mahoneycm: Resolved in [a93f022](https://github.com/uswds/uswds/pull/5049/commits/a93f022bfcb154c7665861b98338d2e482f5b13b)

packages/usa-modal/src/test/modal.spec.js
mahoneycm: Resolved in https://github.com/uswds/uswds/pull/5049/commits/577d1b94baf2d62a48859bce2fd3e3424df0c8cd

mejiaj (changes requested):
@mahoneycm no issues with functionality, everything looks good. But ran into an issue on `npm run build:html`, which is what we use for the HTML code examples on USWDS-Site. Could you please take a look?

The compile error:

```bash
> @uswds/uswds@3.3.0 build:html
> webpack build --config ./webpack.twig.config.js && npm run prettier:html

assets by status 709 KiB [cached] 193 assets
./html-templates/main.js 1 bytes [built] [code generated]

ERROR in Conflict: Multiple assets emit different content to the same filename usa-modal.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-modal~forced-action.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-modal~large.html

webpack 5.58.1 compiled with 3 errors in 1898 ms
```

mejiaj (approved):
I've made a small change to the test pattern structure in fec99fafb34988b4d7030a5a573175dc588caa0e. This doesn't affect the functionality ― just uses the standard test template pattern.

thisisdano (approved):

PR #5004, opened 2022-10-24
USWDS - Input & Prefix/Suffix: Add disabled styles. Resolves #4913

mahoneycm:
## Description

[Preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-prefix-suffix-disable/?path=/story/components-form-inputs-input-prefix-or-suffix--prefix)

Resolves #4913 

### Input disabled styling

Currently, disabled form elements other than `button` rely on default browser styles to give visual queues that form elements are disabled. Chrome and Edge very lightly dim the field, but is sometimes hard to notice. Other browsers such as Safari and Firefox do not add any visual indication that these fields are inaccessible to the user. 

@amyleadem summed it up very nicely in her comment on #4322 which addresses this issue for `select / dropdown` components

> @thisisdano @mejiaj I have included a screenshot that showcases disabled visual cues in Chrome, Firefox, Safari, and Edge for OS. This screenshot displays the following:
> 
> * Standard .usa-select element
> * Disabled .usa-select element
> * Disabled select element without .usa-select class
> * Standard .usa-input element
> * Disabled .usa-input element
> * Disabled input element without .usa-input class
> 
> Image
> 
> I also have a link to a [live demo](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-test-disabled-select/iframe.html?args=&id=components-form-inputs-select--select&viewMode=story) in PR #4969 in case it is hard to glean the style differences from the screenshot.
> 
> Things to note:
> 
> * Each browser has its own set of default styles for disabled form elements.
> * USWDS overrides some default disabled styles in a way that makes it less clear visually that the element is disabled.
> 
> I lean towards wanting to take a look at developing `disabled` styles for USWDS form elements that provide clearer visual cues that the element is disabled. However, we would need to be careful of poor contrast ratio and look into accessibility implications.

 I agree with Amy in leaning towards developing our own `disabled` styles to have a consistent and easy-to-understand style across all USWDS form elements and browsers.

This PR adds `disabled` styles for `input` elements but can be extended to other form elements such as text fields, drop downs, and combo boxes if we chose to take this route.

### Prefix / Suffix

[...]

PR #5063, opened 2022-12-06
Form Inputs: Normalize disabled forms styles

### LINKED ISSUE #5051: USWDS - Disabled styles: Apply consistent styles to disabled elements

Related to #5004 & #4913

Disabled styles are not consistent across browsers and can cause visual confusion for users

Add `disabled` modifier to other form elements to ensure consistent styling across disabled forms.

_No response_

### Additional context

_Screenshots go from Active inputs, default styles when disabled, new styles with USWDS disabled (Chrome)_
image

Other browsers lack disabled styles so there is no indication that the element is disabled other than it is not focusable. Example from safari:
Screen Shot 2022-11-22 at 3 30 00 PM

### Code of Conduct

mahoneycm:
[Preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-disabled-forms-styles/)

Resolves #5051
Includes work from: 
#5004 
#5128 
#5146 

## Description

Creates and normalizes disabled styles across USWDS forms.

## Breaking change

Existing `usa-combo-box`, `usa-date-picker`, `usa-file-input` components with `aria-disabled="true"` may be affected now that the javascript for each component looks for that attribute during initialization

While we advise against using the usa-button--disabled class, this would break markdown for devs who have implemented it

## Related PRs

[USWDS Site change log PR →](https://github.com/uswds/uswds-site/pull/2089)

## The Problem

Currently, Chrome and Edge support disabled styles for forms. Both implement a pretty minimal opacity shift to disabled elements to help users understand an element is disabled. The shift is pretty minor:

_Enabled vs disabled in Chrome_
Screen Shot 2022-12-08 at 12 46 27 PM

Safari & Firefox do not offer any visual queues an element is disabled at this time. This can cause 

## The solution

Create USWDS styles for disabled forms to add visual context for users.

Text Input Disabled 
image

## Affected components

All components within the `Forms Inputs` category on Storybook as well as Button

- [x] Button
- [x] Character Count
- [x] Checkbox _(Disabled Styles existed previously)_ 
- [ ] Checklist _(Didn't seem necessary. Not a form input)_ 
- [x] Combo Box
- [x] Date Picker
- [x] Date Range Picker
- [x] File Input
- [x] Text Input Mask
- [x] Input Prefix or Suffix _(Work completed in #5004)_ 
- [x] Text Input _(Work completed in #5004)_
- [x] Memorable Date
- [x] Radio _(Disabled Styles existed previously)_  
- [x] Range
- [x] Select
- [x] Time Picker _(Storybook seems to be having trouble rendering this one)_   

## Updates made
- All form controls and buttons have matching `disabled` and `aria-disabled` states
    - Every disabled element and its overlaid text meets WCAG AA contrast:
        - Disabled backgrounds are defined with **`$theme-color-disabled` ("gray-20")**
        - Text on default backgrounds are defined with the new variable: **`$theme-color-disabled-text` (”gray-50”)**
        - Text on reverse backgrounds are defined with the new variable: **`$theme-color-disabled-text-reverse` ("gray-40")**
        - Text on disabled backgrounds are defined with the new variable: **`$theme-color-text-on-disabled` ("gray-70")**
    - Disabled styles now receive `cursor: not-allowed`
    - Disabled styles do not show hover states
    - Removed all disabled classes (This only applies to user-added disabled styles. Dynamically-added styles are still allowed)
    - Dynamically-added disabled styles are triggered for both `disabled` and `aria-disabled`
- All disabled states are consistently demonstrated in Storybook
    - A new `Disabled Forms` story has been added that shows all disabled form elements on one page
    - All `disabled`/`aria-disabled` controls in storybook are consistent
    - For buttons, `aria-disabled` is now included in the “show all states” control, which is displayed by default
- Input prefix/suffix: removed need for a disabled class
    - The CSS was re-written to absolutely position the icons instead of relying on disabled parent class
    - Removed `index.js` and related test files because they were no longer needed (the JS moved focus to the disabled group, which no longer exists
- Removed test pattern twig for `usa-select—multiple`; can rely on standard twig instead

## New `disabled` utility mixin
The new `u-disabled` mixin overrides `background-color` and cursor styles of elements to display as disabled. Includes overrides for `hover`, `active`, and `focus` states.

Users can pass in `$bg-color` & `$text-color` parameters as necessary to add custom colors if needed.

An example of this can be found for `usa-button--outline` variants to ensure color compliance

```scss
@include u-disabled("transparent", $theme-color-disabled-text);
```

## New Disabled Color Tokens
| Token Name | Color |
| ------------- | ------------- |
| `$theme-color-disabled-text`  | "gray-50"  |
| `$theme-color-disabled-text-reverse` | "gray-40"  |
| `$theme-color-text-on-disabled` | "gray-70" |

## Preview
[Disabled Forms →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-disabled-forms-styles/?path=/story/patterns-forms--disabled-form-elements)

## To Test

- Visit each of the form components
- View disabled variant or turn on `disabled` and/or `aria-disabled`
- You may need to refreseh the page in order for the components javascript to initialize and the view to update
- Confirm:
   - `disabled` elements have a grey background and a `not-allowed` cursor on hover
   - `aria-disabled` styles mimic `disabled styles` _(Components will still be accessible if not `disabled`)_ 
   - Both style work together and add relative attribute to html code

---

### Notes & Other Considerations

After discussion, we have decided to allow disabled classes that style elements that:

- Do not receive the disabled attribute
- Are dynamically placed
   This will allow us to correctly style elements as disabled without putting the burden on the user.

Further findings from my disabled landscape analysis can be found at #5117 

⚠️ We should consider adding new disabled tokens to [theme color tokens](https://designsystem.digital.gov/design-tokens/color/theme-tokens/) and create a page for the disabled utility mixin in [Utilities](https://designsystem.digital.gov/utilities/)

Moving forward, we should also create guidance on our website about accessibility and disabled form elements to assist users in understanding when it is and is not appropriate to use these new styles and mixin


packages/usa-checkbox/src/usa-checkbox.stories.js
```diff
@@ -5,6 +5,11 @@ import TestComponent from "./test/test-patterns/test-usa-checkbox.twig";
 export default {
   title: "Components/Form Inputs/Checkbox",
   argTypes: {
+    disabled: {
+      control: {type: 'boolean'},
+      defaultValue: false,
+      name: "disabled"
+    },
```
amyleadem, line 12: What do you think about using a radio controls in these stories like we do for Date picker? It makes it easier to isolate the test.

```suggestion
    disabled_state: {
      name: "Disabled state",
      control: { type: "radio" },
      options: ["none", "disabled", "aria-disabled"]
    },
```

packages/usa-character-count/src/usa-character-count.stories.js
```diff
@@ -1,4 +1,5 @@
 import Component from "./usa-character-count.twig";
+import disabledOption from "./index.json";
```
amyleadem, line 2: What do you think about changing the two booleans into a single radio control like we do in the Date range picker stories? It makes it easier to isolate the testing so that you are only checking one attribute at a time. 

I also have a slight lean towards making separate stories for Disabled and Aria disabled for all our components like we have for Date Range Picker. This might just be a personal preference though.

packages/usa-combo-box/src/styles/_usa-combo-box.scss
```diff
@@ -56,13 +56,25 @@ $close: map-merge(
   margin-bottom: 0;
   max-width: none;
   padding-right: calc(2.5em + 3px);
+
+  &:disabled,
+  &[aria-disabled="true"] {
+    background-color: color("disabled");
+    cursor: not-allowed;
+  }
```
amyleadem, line 64: I see these are rules repeated a lot. Could we create a mixin for this?

packages/usa-character-count/src/usa-character-count.stories.js
mahoneycm: The reason I originally decided to have them be separate booleans instead of the radio buttons, is because `aria-disabled` doesn't apply the disabled _state_ to the components. You are still able to set text in inputs/text fields, dropdowns can be opened, checkboxes can be checked/unchecked.

My thought was this would allow users to test them separately _and together_ and that having them in conjunction would allow them to be selected by the screen reader while being actively disabled and not interactive.

>  ...the aria-disabled="true" only semantically exposes these elements as being disabled. Web developers must manually ensure such elements have their functionality suppressed when they are exposed in the disabled state.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled

It does appear that [some users think that you do not need both at this time](https://stackoverflow.com/questions/38059140/when-to-use-the-disabled-attribute-vs-the-aria-disabled-attribute-for-html-eleme#:~:text=You%20will%20be%20happy%20to%20know%20that%20both%20NVDA%20and%20JAWS%20skip%20the%20field%20(or%20if%20explicitly%20focused%2C%20announce%20that%20is%20disabled).), as `disabled` seems to accept keyboard focus for screen readers (at least in modern browsers / our storybook example).

I'm definitely open to suggestions on what we think is the best path moving forward. You might be much more familiar with aria attributes than I am so I'm happy to change this back to the radio button slike date range picker if you think that makes more sense!

packages/usa-combo-box/src/styles/_usa-combo-box.scss
mahoneycm: Oh! I didn't even consider that! I like that idea!

packages/usa-checkbox/src/usa-checkbox.twig
```diff
@@ -1,15 +1,15 @@
 <fieldset class="usa-fieldset">
   <legend class="usa-legend">Select any historical figure</legend>
   <div class="usa-checkbox">
-    <input class="usa-checkbox__input" id="check-historical-truth" type="checkbox" name="historical-figures" value="sojourner-truth" checked {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
+    <input class="usa-checkbox__input" id="check-historical-truth" type="checkbox" name="historical-figures" value="sojourner-truth" checked {% if disabled %} disabled {% endif %} {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
```
mejiaj, line 4:
```suggestion
    <input class="usa-checkbox__input" id="check-historical-truth" type="checkbox" name="historical-figures" value="sojourner-truth" checked {% if disabled %} disabled {% endif %} {% if aria_disabled %}aria-disabled="true"{% endif %}>
```

packages/usa-checkbox/src/usa-checkbox.twig
```diff
@@ -1,15 +1,15 @@
...
     <label class="usa-checkbox__label" for="check-historical-truth">Sojourner Truth</label>
   </div>
   <div class="usa-checkbox">
-    <input class="usa-checkbox__input" id="check-historical-douglass" type="checkbox" name="historical-figures" value="frederick-douglass" {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
+    <input class="usa-checkbox__input" id="check-historical-douglass" type="checkbox" name="historical-figures" value="frederick-douglass" {% if disabled %} disabled {% endif %} {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
```
mejiaj, line 8:
```suggestion
    <input class="usa-checkbox__input" id="check-historical-douglass" type="checkbox" name="historical-figures" value="frederick-douglass" {% if disabled %} disabled {% endif %} {% if aria_disabled %}aria-disabled="true"{% endif %}>
```

packages/usa-checkbox/src/usa-checkbox.twig
```diff
@@ -1,15 +1,15 @@
...
     <label class="usa-checkbox__label" for="check-historical-douglass">Frederick Douglass</label>
   </div>
   <div class="usa-checkbox">
-    <input class="usa-checkbox__input" id="check-historical-washington" type="checkbox" name="historical-figures" value="booker-t-washington" {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
+    <input class="usa-checkbox__input" id="check-historical-washington" type="checkbox" name="historical-figures" value="booker-t-washington" {% if disabled %} disabled {% endif %} {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
```
mejiaj, line 12:
```suggestion
    <input class="usa-checkbox__input" id="check-historical-washington" type="checkbox" name="historical-figures" value="booker-t-washington" {% if disabled %} disabled {% endif %} {% if aria_disabled %}aria-disabled="true"{% endif %}>
```

packages/usa-combo-box/src/styles/_usa-combo-box.scss
```diff
@@ -56,13 +56,23 @@ $close: map-merge(
...
 button.usa-combo-box__toggle-list,
 button.usa-combo-box__clear-input {
   &:focus {
     outline-offset: -4px;
   }
+
+  &:disabled,
+  &[aria-disabled="true"] {
+    @include u-disabled;
+  }
```
mejiaj, line 75: I see this is set on the toggle list. Is there an example of how this would look?

packages/usa-combo-box/src/styles/_usa-combo-box.scss
```diff
@@ -56,13 +56,23 @@ $close: map-merge(
   margin-bottom: 0;
   max-width: none;
   padding-right: calc(2.5em + 3px);
+
+  &:disabled,
+  &[aria-disabled="true"] {
+    @include u-disabled;
+  }
```
mejiaj, line 63: Wouldn't adding disabled styles to `%block-input-styles` make it easier to style or does that cause issues with other components?

packages/usa-select/src/content/usa-select~multiple.json
```diff
@@ -0,0 +1,5 @@
+{
+  "disabled": false,
+  "aria_disabled": false,
+  "size": 0
+}
```
mejiaj, line 5: You can set basic defaults and override a value or add a new prop easily.

Checkout banner settings for an example:
https://github.com/uswds/uswds/blob/8b78b05795cfa5d4998bfd19ae1977b1108edbe6/packages/usa-banner/src/usa-banner.stories.js#L21-L25

packages/usa-select/src/styles/_usa-select.scss
```diff
@@ -30,6 +30,12 @@
     background-image: none;
   }
 
+  &:disabled,
+  &[aria-disabled="true"] {
+    @include u-disabled;
+    opacity: 1;
```
mejiaj, line 36: I was confused when I originally saw the opacity being set. But then I checked in Chrome devtools.

image


Can we add a note that this undoes a default browser style?

mejiaj (commented):
Looks good so far, added some suggestions and questions.

packages/usa-character-count/src/usa-character-count.stories.js
```diff
@@ -7,3 +8,4 @@ export default {
 const Template = (args) => Component(args);
 
 export const CharacterCount = Template.bind({});
+CharacterCount.args = disabledOption;
```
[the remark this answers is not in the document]
mejiaj: Done in b1ca9f0601641496b338f02e974b2fbe75123453.

packages/usa-checkbox/src/usa-checkbox--tile.twig
```diff
@@ -2,7 +2,7 @@
   <fieldset class="usa-fieldset">
     <legend class="usa-legend">Select any historical figure</legend>
     <div class="usa-checkbox">
-      <input class="usa-checkbox__input usa-checkbox__input--tile" id="check-historical-truth-2" type="checkbox" name="historical-figures-2" value="sojourner-truth" checked {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
+      <input class="usa-checkbox__input usa-checkbox__input--tile" id="check-historical-truth-2" type="checkbox" name="historical-figures-2" value="sojourner-truth" checked {% if disabled %} disabled {% endif %} {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
```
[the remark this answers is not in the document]
mejiaj: Simplified logic, simplified storybook controls, updated formatting, and removed previous aria-disabled test in 9f13406e9f61d7efe1495c65877527df287be629.

packages/usa-checkbox/src/usa-checkbox.twig
mejiaj: Fixed in a41689ab584c9ee5d22e7935f8113c839033fec4.

packages/usa-combo-box/src/styles/_usa-combo-box.scss
mejiaj: Inspected classes, they're the arrow dropdown and clear icons once component is initialized and option has been selected.

image

packages/usa-select/src/styles/_usa-select.scss
mejiaj: Added comment in 6758d68ceb5a64b900155cde8ab9cbd4c29f910a.

packages/usa-combo-box/src/styles/_usa-combo-box.scss
mejiaj: Has a negative effect on `usa-range` component because it also uses `%block-input-styles`.

https://github.com/uswds/uswds/blob/c662f50928025ed1dc8c40e04b258e7f879cfb17/packages/usa-range/src/styles/_usa-range.scss#L53

packages/usa-select/src/content/usa-select~multiple.json
mejiaj: Simplified story in c662f50928025ed1dc8c40e04b258e7f879cfb17. Now the `disabled/aria_disabled` settings are defaults and story uses a single template.

mejiaj:
@amyleadem ready for re-review. I've made the following updates:

- Simplified stories (removed JSON files exclusively for `disabled/aria-disabled`)
- Simplified logic in templates (ex: a41689ab584c9ee5d22e7935f8113c839033fec4)
- Removed unnecessary controls in combo box (71ee0662c6bced30435be98ed64fbcffb7ddb8cc)
- Removed duplicate selector in usa-range (4770bca0944cd927ec65132bfa1438d74c9c9dff)
- Run prettier formatting on stories and SCSS files

# How to test
- [ ] Build HTML templates without errors
- [ ] Form control stories should have `disabled/aria-disabled` controls
- [ ] Form controls should have `disabled/aria-disabled` styles

packages/usa-input-prefix-suffix/src/styles/_usa-input-prefix-suffix.scss
```diff
@@ -13,6 +13,11 @@
     @include focus-outline();
   }
 
+  &--disabled {
+    background-color: color("disabled");
+    cursor: not-allowed;
```
amyleadem, line 18: Should this use the mixin? 
```suggestion
    @include u-disabled;
```

packages/usa-range/src/usa-range.json
```diff
@@ -0,0 +1,4 @@
+{
+  "disabled": false,
+  "aria_disabled": false
+}
```
amyleadem, line 4: @mejiaj Did you want to remove the json files that had only disabled info?

packages/uswds-core/src/js/index.js
```diff
@@ -42,4 +43,6 @@ module.exports = {
   timePicker,
   tooltip,
   validator,
+
+  memorableDate,
```
amyleadem, line 47: Can this be added in alphabetical order?

amyleadem:
@mejiaj I found a few items that were missing controls or had broken styles: 

- The combo box does not have aria-disabled styles once initialized
- File input does not have an aria-disabled control
- Text input does not have disabled or aria-disabled controls
- Radio does not have disabled control
- The time picker doesn’t have disabled or aria-disabled styles once initialized

There were also a few items about the Storybook experience that I wanted to flag:
- There is an inconsistency in how users view the disabled states. Some use controls while others put the disabled styles in the twig. Some have disabled stories and others don’t. I think there are a lot of right ways to do it — It would just be good if the experience was consistent.
- I still think that we should change the disabled/aria-disabled boolean controls to radio buttons like we do for Date Range Picker. I don’t know when users would use both attributes at the same time since they serve different purposes. It would be nice if it was easier to test just one at a time.
- What do you think about making stories for Disabled and aria-disabled for each component? (Like we do with date picker) This one might just be personal preference but I like that it makes disabled styles more noticeable for testing. 

And finally,  I noticed some visual inconsistencies in the various form components when they were placed together on the page. I threw together a [quick PR](https://github.com/uswds/uswds/pull/5080) to demonstrate. I added notes in the screenshot below:

image

amyleadem (changes requested):
@mejiaj I found a few items that were missing controls or had broken styles: 

- The combo box does not have aria-disabled styles once initialized
- File input does not have an aria-disabled control
- Text input does not have disabled or aria-disabled controls
- Radio does not have disabled control
- The time picker doesn’t have disabled or aria-disabled styles once initialized

There were also a few items about the Storybook experience that I wanted to flag:
- There is an inconsistency in how users view the disabled states. Some use controls while others put the disabled styles in the twig. Some have disabled stories and others don’t. I think there are a lot of right ways to do it — It would just be good if the experience was consistent.
- I still think that we should change the disabled/aria-disabled boolean controls to radio buttons like we do for Date Range Picker. I don’t know when users would use both attributes at the same time since they serve different purposes. It would be nice if it was easier to test just one at a time.
- What do you think about making stories for Disabled and aria-disabled for each component? (Like we do with date picker) This one might just be personal preference but I like that it makes disabled styles more noticeable for testing. 

And finally,  I noticed some visual inconsistencies in the various form components when they were placed together on the page. I threw together a [quick PR](https://github.com/uswds/uswds/pull/5080) to demonstrate. I added notes in the screenshot below:

image

amyleadem:
Thinking about it a bit more: this might be out of scope -- or not anything we want to pursue at all --  but I am wondering if there are ways to give non-color cues that the element is disabled. 

I struggle because the medium-gray background with a dark border does not read as disabled to me and it could be invisible/not meaningful to users with who need higher contrast. 

One option I see is to do something like a dotted border (though maybe those are hard to see as well):
image

Or maybe we could even be more explicit and add an icon and/or text to indicate that the field is disabled. Here is just a rough idea: 
image

Disabled form patterns are tricky, so just throwing it out there to see if this is a conversation we even want to have.

packages/usa-input-prefix-suffix/src/styles/_usa-input-prefix-suffix.scss
mahoneycm: Resolved in [12c3713](https://github.com/uswds/uswds/pull/5063/commits/12c371351d9459d2c23509ee84a10068ba74e675)

packages/uswds-core/src/js/index.js
mahoneycm: Resolved in [6e9498e](https://github.com/uswds/uswds/pull/5063/commits/6e9498ee385ba6f68fd043dc15169fdfc5045d0d)

packages/usa-range/src/usa-range.json
mahoneycm: This example is resolved in [50f287c](https://github.com/uswds/uswds/pull/5063/commits/50f287c95f7a67eaa014099df3d2e27682721456)

Happy to move all disabled options into booleans! We'll discuss if disabled variants should also have their own stories as well

mahoneycm:
@mejiaj @amyleadem 

OK SO I addressed _almost_ all of the topics here:

> @mejiaj I found a few items that were missing controls or had broken styles:
> 
> * ~The combo box does not have aria-disabled styles once initialized~ _// Resolved_
> * ~File input does not have an aria-disabled control~ _// Resolved_
> * Text input does not have disabled or aria-disabled controls _// Lets discuss how all states should be displayed for this component_
> * ~Radio does not have disabled control~ _// Resolved_
> * ~The time picker doesn’t have disabled or aria-disabled styles once initialized~ _// Resolved_
> 
> There were also a few items about the Storybook experience that I wanted to flag:
> 
> * There is an inconsistency in how users view the disabled states. Some use controls while others put the disabled styles in the twig. Some have disabled stories and others don’t. I think there are a lot of right ways to do it — It would just be good if the experience was consistent.
> * ~I still think that we should change the disabled/aria-disabled boolean controls to radio buttons like we do for Date Range Picker. I don’t know when users would use both attributes at the same time since they serve different purposes. It would be nice if it was easier to test just one at a time.~
> _// I switched all disabled controls from boolean switches to radio buttons_
> * What do you think about making stories for Disabled and aria-disabled for each component? (Like we do with date picker) This one might just be personal preference but I like that it makes disabled styles more noticeable for testing.
> 
> And finally, I noticed some visual inconsistencies in the various form components when they were placed together on the page. I threw together a [quick PR](https://github.com/uswds/uswds/pull/5080) to demonstrate. I added notes in the screenshot below:
> 
> image

For the above screenshot I fixed the following:

- Adjusted disabled Radio + Checkbox text opacity adjustment from  `-0.7` to `-0.4`. An opacity of `0.6` seems to be the magic number for color contrast compliance
- Adjusted calendar Icon opacity for `date-picker`
- Adjusted disabled file input text opacity from `0.5` to `0.7` for color contrast compliance
- Resolved disabled initialization error

## Things to discuss moving forward

- I like the idea of adding dashed or dotted borders to disabled forms. I think it visually adds a nice touch that makes it feel more disabled. Can even be done to checkboxes if we'd like
  - On that note, I didn't notice a difference in the border style for `file-input` in it's regular and disabled states. Inspecting the styles gave me the same results
- Do we want disabled component stories to highlight disabled variants?
- Do we want to change the variant displays for components like `Text Input` where all of its versions are displayed on one story?

mahoneycm:
## DISABLED STORIES ARE NOW NORMALIZED ACROSS THE BOARD

I've also cleaned up the **Text Input** and **Prefix/Suffix** story so there's an emphasis on controls instead of all of the state variants on one story

### Radio/Checkbox Tile collor

I realized that the radio/checkbox tiles are currently set to `color-input-
disabled` so it is harder to distinguish between enabled and disabled items
Screen Shot 2022-12-29 at 4 42 16 PM

How do we feel about switching the default tile color to black so the disabled color is more noticeable?

Screen Shot 2022-12-29 at 4 48 57 PM

Alternatively, a dashed line does the trick but also seems a bit extra
Screen Shot 2022-12-29 at 4 52 45 PM

@amyleadem @mejiaj

packages/usa-input-prefix-suffix/src/usa-input-prefix-suffix.stories.js
```diff
@@ -3,6 +3,14 @@ import suffix from "./usa-input-suffix.twig";
 
 export default {
   title: "Components/Form Inputs/Input Prefix or Suffix",
+  argTypes: {
+    state: {
+      name: "State",
+      control: "radio",
+      options: ["default", "error", "disabled", "aria_disabled"],
+      defaultValue: "default"
```
amyleadem, line 11: This control is called `state` while others are called `disabled_state` and this includes an error state option. Is there something specific about the input prefix/suffix error state that it needs to shown in this component? Wondering if it is safe to remove and just rely on the text input component for checking error/success states.

Also, I noticed there are no disabled/aria-disabled storiesfor this component.

```suggestion
    disabled_state: {
      name: "Disabled state",
      control: { type: "radio" },
      options: ["none", "disabled", "aria_disabled"],
      defaultValue: "none"
    },
```

packages/usa-radio/src/usa-radio--tile.twig
```diff
@@ -2,15 +2,21 @@
   <fieldset class="usa-fieldset">
     <legend class="usa-legend">Select one historical figure</legend>
     <div class="usa-radio">
-      <input class="usa-radio__input usa-radio__input--tile" id="historical-truth-2" type="radio" name="historical-figures-2" value="sojourner-truth" checked {% if aria_disabled == true %}aria-disabled="true"{% endif %}>
+      <input class="usa-radio__input usa-radio__input--tile" id="historical-truth-2" type="radio" name="historical-figures-2" value="sojourner-truth" checked
```
mejiaj, line 5: There are a lot of attributes in these radios, maybe we can increase readability by separating them out into their own line.

```suggestion
<input 
  class="usa-radio__input usa-radio__input--tile" 
  id="historical-truth-2" 
  type="radio" 
  name="historical-figures-2" 
  value="sojourner-truth" 
  checked 
  {% if disabled_state == "disabled" %}disabled{% endif %} 
  {% if disabled_state == "aria_disabled" %}aria-disabled="true"{% endif %}
>
```

packages/usa-range/src/styles/_usa-range.scss
```diff
@@ -32,12 +32,24 @@
...
+  @include u-disabled;
+}
+
+@mixin range-track--disabled {
+  @include range-track;
+  @include u-disabled;
+}
+
+@mixin range-ms-fill--disabled {
+  @include range-ms-fill;
```
mejiaj, line 46: Is everything from this mixin needed? It looks like some of the styles set with `range-ms-fill` end up being overwritten by `u-disabled`.

If we're also using it below in `-ms-fill-upper` why not consider something like breaking out the required styles from `range-track to something like `range-track-base` and importing that in the places where you **only** need those base styles?

packages/usa-select/src/styles/_usa-select.scss
```diff
@@ -30,6 +30,17 @@
...
   }
 
+  &:disabled,
+  &[aria-disabled="true"] {
+    @include u-disabled;
+    opacity: 1; // Undo chromium browser style
+
+    option {
+      color: color(ink); // Undo browser style
+      opacity: 0.7;
```
mejiaj, line 40: Again, just a note on consistency of value. We should try to stick with either `0.6` or `0.7`.

packages/usa-time-picker/src/index.js
```diff
@@ -62,7 +62,7 @@ const transformTimePicker = (el) => {
 
   const selectEl = document.createElement("select");
 
-  ["id", "name", "required", "aria-label", "aria-labelledby"].forEach(
+  ["id", "name", "required", "aria-label", "aria-labelledby", "disabled", "aria-disabled"].forEach(
```
mejiaj, line 65: What are these attributes for? There are enough now where we might want to extract it into a `const`.

mejiaj (changes requested):
@mahoneycm visually the styles look great and thank you for all the work you've put into this. 

I've added some comments for readability and consistency. I also had a question on the new JS for memorable date.

⚠️ We'll also need to check `npm run build:html` to make sure that the output still looks correctly and builds without errors.

packages/usa-memorable-date/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+    childEl.setAttribute('disabled', '')
+  })
+}
+
+const memorableDate = behavior(
+  {},
+  {
+    init(root) {
+      selectOrMatches(MEMORABLE_DATE_CLASS, root).forEach((MemDateEl) => {
+        if (MemDateEl.classList.contains(MD_DISABLED)) {
```
amyleadem, line 35: Thinking out loud - should adding aria-disabled be automated as well?

packages/usa-input-mask/src/usa-input-mask.stories.js
```diff
@@ -4,10 +4,20 @@ import {
   PhoneContent,
   ZipContent,
   AlphanumericContent,
+  DisabledContent,
+  AriaDisabledContent
```
amyleadem, line 8: Curious why these are controlled via JSON rather than via the argType like the other components.

amyleadem (changes requested):
@mahoneycm Looking good! The Storybook experience is easy to navigate and feels more consistent.

I discovered a few things during review: 
- File input - Seeing a standard cursor on disabled/aria-disabled inputs
- Text input and Input prefix/suffix - Do not have disabled and aria-disabled stories
- Disabled buttons - Thinking it should have styles that match disabled inputs since they will appear in/near forms (Namely, add cursor: not-allowed)
- Small tweak, but it might be clearer to make the `disabled_state` control option read `aria-disabled` instead of `aria_disabled`
- (Optional) What do you think about adding a test story that shows all the form elements together so that we can easily check how any visual change interacts with the other elements? (Possible template in https://github.com/uswds/uswds/pull/5080)
- (Optional) In general, I still believe it would be helpful if we gave more visual cues for the disabled state.  Some opportunities:
    - Memorable date and Select - can we remove/replace the “select” text from disabled input? (This is optional, but feels like the "select" with a down caret is suggestive of an action being available.)
    - Radio and checkbox - Subtle difference between disabled and activated. Can we add some other visual cue? Maybe a disabled symbol (like in the cursor: not-allowed) in place of the empty circle?
    - I am interested in hearing more opinions about changing border styles, adding disabled icons, etc

packages/usa-input-prefix-suffix/src/usa-input-prefix-suffix.stories.js
mahoneycm: Resolved in [687156b](https://github.com/uswds/uswds/pull/5063/commits/687156b0f3418558f3d26662cba65c522a01ebe2)

packages/usa-character-count/src/usa-character-count.twig
```diff
@@ -8,7 +8,10 @@
         id="with-hint-input"
         maxlength="25"
         name="with-hint-input"
-        aria-describedby="with-hint-input-info with-hint-input-hint">
+        aria-describedby="with-hint-input-info with-hint-input-hint"
+        {% if disabled_state == 'disabled' %} disabled {%- endif %}
```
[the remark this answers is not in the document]
mahoneycm: Since `aria-disabled` is an option of `disabled_state`, this will enable `disabled` and `ariaDisabled` at the same time. The default `disabled_state` is `none`, so `disabled_state` will always come back as true because there's always a value.

I could switch the default value to `false`, but if we're just checking for `disabled_state` we can't switch between `disabled` and `aria_disabled`

I could do 
```twig
{% if disabled_state == 'disabled' %} disabled
{% elseif disabled_state == 'aria_disabled' %} aria-disabled="true" {%- endif %}
```
If we prefer it to be an if/else

packages/usa-combo-box/src/index.js
```diff
@@ -124,6 +124,20 @@ const disable = (el) => {
...
+ *
+ * @param {HTMLInputElement} el An element within the combo box component
+ */
+const ariaDisable = (el) => {
+  const { inputEl, toggleListBtnEl, clearInputBtnEl } = getComboBoxContext(el);
+
+  clearInputBtnEl.hidden = true;
+  clearInputBtnEl.setAttribute("aria-disabled", true);
+  toggleListBtnEl.setAttribute("aria-disabled", true);
+  inputEl.setAttribute("aria-disabled", true);
```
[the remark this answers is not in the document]
mahoneycm: I referenced the pre-existing combo-box `disabled` function for this one. I believe that during the `enhanceComboBox` function, it adds `usa-sr-only` class to the select element, and then rebuilds the select as an input with the clear and toggle list buttons. In order to maintain functionality, you need to remove the disabled attribute from select, and add it to the newly generated input and buttons

https://github.com/uswds/uswds/blob/ac850d8bf854a4af17f8e0360e86197837e5915b/packages/usa-combo-box/src/index.js#L276-L286

packages/usa-date-picker/src/index.js
```diff
@@ -710,9 +710,23 @@ const disable = (el) => {
...
+ * Check for aria-disabled on initialization
+ *
+ * @param {HTMLElement} el An element within the date picker component
+ */
+const ariaDisable = (el) => {
+  const { externalInputEl, toggleBtnEl } = getDatePickerContext(el);
+
+  toggleBtnEl.setAttribute("aria-disabled", true);
+  toggleBtnEl.style.opacity = "0.6";
+  externalInputEl.setAttribute("aria-disabled", true);
```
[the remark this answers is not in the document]
mahoneycm: This falls into the same boat as the combo box functionality I commented on above

packages/usa-file-input/src/index.js
```diff
@@ -121,6 +132,7 @@ const buildFileInput = (fileInputEl) => {
   const box = document.createElement("div");
   const instructions = document.createElement("div");
   const disabled = fileInputEl.hasAttribute("disabled");
+  const ariaDisabled = fileInputEl.hasAttribute("aria-disabled");
```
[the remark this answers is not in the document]
mahoneycm: I based this check on the way were already checking for `disabled` and tried to match the use.

https://github.com/uswds/uswds/blob/ac850d8bf854a4af17f8e0360e86197837e5915b/packages/usa-file-input/src/index.js#L134-L135

Which runs the `if()` as a part of the `buildFileInput` function

https://github.com/uswds/uswds/blob/ac850d8bf854a4af17f8e0360e86197837e5915b/packages/usa-file-input/src/index.js#L157-L165

Would you like me to change them both to `isDisabled` & `isAriaDisabled` for better readability?

packages/usa-file-input/src/styles/_usa-file-input.scss
```diff
@@ -188,14 +188,14 @@
   pointer-events: none;
 
   .usa-file-input__instructions {
-    opacity: 0.5;
+    opacity: 0.7;
```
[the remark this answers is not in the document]
mahoneycm: On second inspection it actually looks like `.7` isn't passing for regular-sized text. I'll have to bump this to `.8`

60% opacity works for most elements but because the background on the file input gets changed to a light gray, we need this text to be darker.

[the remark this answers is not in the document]
mahoneycm: What do we think of `.8` and removing the link color from the disabled version?

image

packages/usa-input-mask/src/content/usa-input-mask-alphanumeric.json
```diff
@@ -6,5 +6,6 @@
   "inputmode": "text",
   "placeholder": "___ ___",
   "pattern": "\\w\\d\\w \\d\\w\\d",
-  "datacharset": "A#A #A#"
+  "datacharset": "A#A #A#",
+  "disabled_state": "none"
```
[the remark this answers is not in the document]
mahoneycm: I might be misunderstanding but the state controls are still available in Storybook just like the other component!

image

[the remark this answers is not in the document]
mahoneycm: Resolved in [fd7c14a](https://github.com/uswds/uswds/pull/5063/commits/fd7c14af34351f23a2009fdbda15941811db9190)

packages/usa-date-picker/src/usa-date-picker.twig
```diff
@@ -14,7 +14,8 @@
       type="text"
       aria-labelledby="appointment-date-label"
       aria-describedby="appointment-date-hint"
-      {% if disabled_state == "disabled" %} disabled{% endif %}
-      {% if disabled_state == "aria-disabled" %} aria-disabled="true"{% endif %}>
+      {% if disabled_state == 'disabled' %} disabled {%- endif %}
+      {% if disabled_state == 'aria_disabled' %} aria-disabled="true" {%- endif %}
+      >
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [fd7c14a](https://github.com/uswds/uswds/pull/5063/commits/fd7c14af34351f23a2009fdbda15941811db9190)

packages/usa-input-prefix-suffix/src/usa-input-prefix-suffix.stories.js
```diff
@@ -3,6 +3,14 @@ import suffix from "./usa-input-suffix.twig";
 
 export default {
   title: "Components/Form Inputs/Input Prefix or Suffix",
+  argTypes: {
+    state: {
+      name: "State",
+      control: "radio",
+      options: ["default", "error", "disabled", "aria_disabled"],
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [687156b](https://github.com/uswds/uswds/pull/5063/commits/687156b0f3418558f3d26662cba65c522a01ebe2)

packages/usa-memorable-date/src/index.js
```diff
@@ -0,0 +1,43 @@
+const behavior = require("../../uswds-core/src/js/utils/behavior");
```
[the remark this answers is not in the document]
mahoneycm: Oh! Thank you for catching this! I created this dynamic disabled stylings before I went on holiday leave and then implemented the storybook controls without utilizing it. 

I just need to match it with an `aria-disabled` version and I'll change the template

packages/usa-radio/src/usa-radio--tile.twig
mahoneycm: Resolved in [2afdb02](https://github.com/uswds/uswds/pull/5063/commits/2afdb022226a20eac994818285486a1ad4b7bfef)

packages/usa-range/src/styles/_usa-range.scss
mahoneycm: It looks like I was able to remove these disabled styles and simply add the `u-disabled` mixin once `usa-range` is disabled/aria-disabled

Resolved in [a753027](https://github.com/uswds/uswds/pull/5063/commits/a753027d552ef32bc708778b380c4c3079ed6d17)

packages/usa-select/src/styles/_usa-select.scss
mahoneycm: I originally put this as 0.7 because it matched chrome disabled styles for all browsers. It appears that `option` is not targetting the select text like I thought it was. This was displaying the text as `ink` with no opacity shift.

For the time being, I've changed this to `.8` as this seems to be the magic opacity dark grey text on the disabled background color and matches my suggestion for file input up above.

It looks like: 
- `opacity: 0.6` is the golden number for disabled elements with a white background
- `opacity: 0.8` for elements with text on a disabled background

### Screenshots

Before:
image


After:
image

Passes on:
- [x] : Chrome
- [x] : Firefox
- [x] : Safari

### Disadvantage
This does adjust the opacity of the background as well, so potentially causing inconsistency between disabled colors on inputs.

packages/usa-input-mask/src/usa-input-mask.stories.js
mahoneycm: This component's variants & arguments were mostly set by JSON files, so at the time I had done this to match the other variants. It wasn't until I started editing components that didn't have JSON files that I started to set the controls via argTypes instead. I can change this one to match the others now that that's been the standard for the disabled arguments!

mahoneycm: Quick question actually.

If I do that, beneath each of the other variant exports I'll need to define the `disabled_state` default value. Since this component has so many variants, I feel like the code will be cleaner if we leave it defined in JSON files. How do you feel about that?

mahoneycm:
Update from #5128 regarding disabled classes and dynamically created components

> After discussion, we have decided to allow disabled classes that style elements that:
>
> 1. Do not receive the disabled attribute
> 2. Are dynamically placed
> This will allow us to correctly style elements as disabled without putting the burden on the user.

mejiaj (changes requested):
@mahoneycm , everything is looking good for the most part. Added some comments on some quirks I found when testing functionality. The only major one is button group segmented styles; could you please take a look?

Also, don't forget to run prettier on styles/js. Thanks!

amyleadem:
Note: I have made updates to this code in PR https://github.com/uswds/uswds/pull/5167. It is set to merge into this branch.

amyleadem (approved):
Looks good to me! I had one small comment about removing the state control for the text input state showcase story, but that is it from me! Thanks for all your hard work on this.

I completed the following checks:
- [x] Confirm that both `aria-disabled` and `disabled` have consistent styling
- [x] Confirm that components do not rely on a user-added disabled class (dynamically-added disabled classes are ok)
- [x] Confirm that Storybook experience is consistent for `aria-disabled` and `disabled` controls
- [x] `npm run test` , `npm run start` run without error
- [x] Checked visual styles in Chrome, Safari, Edge for Mac, Firefox

packages/usa-input/src/usa-input.stories.js
```diff
@@ -1,9 +1,42 @@
...
+  state: "disabled",
+};
+
+export const AriaDisabled = Template.bind({});
+AriaDisabled.args = {
+  state: "aria-disabled",
+};
+
+export const StateShowcase = ShowcaseTemplate.bind({});
+StateShowcase.args = {};
```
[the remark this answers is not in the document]
mahoneycm: Great catch! Thanks for showing me how to do this :)

amyleadem (commented):
Wanted to note that I checked the disabled form elements in (an emulated) forced colors mode and found that they were inconsistently applied.  Should we make sure those are consistent across the board as well?

Sample of some disabled components in forced-colors mode:
image

mahoneycm:
> Wanted to note that I checked the disabled form elements in (an emulated) forced colors mode and found that they were inconsistently applied. Should we make sure those are consistent across the board as well?
> 
> Sample of some disabled components in forced-colors mode: image

I think this is a great idea! I can play with including it in the disabled mixin. Do you think this should be a part of this PR or should this be an enhancement with a new ticket?

mahoneycm:
@amyleadem After playing around with it for a bit, I think we'll need to comb through and correct some visual bugs. Simply setting the outline helps a good amount of them but I'm noticing other issues like missing icons or weird visual quirks. Not a one size fits all answer, unfortunately! I'll create a ticket for this and start checking out each component.

mejiaj:
@mahoneycm @amyleadem I've added some small changes:

1. Moved button setting from JSON to Story [79e0a5de91c6fa852542a083dc174379a54ec0c9]. This was a story specific setting, not a component setting so I moved it.
2. Removed check from input prefix's error state [82d6958abc7599999d99a99c144e441fe188992b]. It's not possible to toggle disabled/aria-disabled on the error state, so I've removed to clean up markup.
3. Removed a11y error for duplicate ID [fff820dac5afc2fd8394040337cbd76d3b4c146a]. Odd that our tests didn't catch this. I caught it with AXE devtools.

packages/uswds-core/src/styles/mixins/utilities/_disabled.scss
```diff
@@ -0,0 +1,23 @@
...
+  &:hover,
+  &.usa-button--hover,
+  &:active,
+  &.usa-button--active,
+  &:focus,
+  &.usa-focus {
+    background-color: color($bg-color);
+    color: color($text-color);
+  }
+}
```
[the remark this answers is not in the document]
mahoneycm: It looks like we can just remove them. I hovered, clicked, and added the classes to buttons with `disabled` or `aria-disabled` and I didn't notice any changes.

@amyleadem I believe these came from your branch, what do you think?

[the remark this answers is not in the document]
amyleadem: I don't see a need to keep them. I support whatever approach @mahoneycm's tests support. Good catch!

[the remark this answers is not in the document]
mahoneycm: Looks like I spoke too soon here. The classes partnered with an _accent class_ seem to override the disabled styles without `usa-button--hover/active` in the mixin

Before removing:
image

Screenshot 2023-03-29 at 5 00 17 PM

After:
Screenshot 2023-03-29 at 5 03 32 PM

Screenshot 2023-03-29 at 5 01 12 PM

I moved the declaration to `button-disabled.scss` and they're displaying correctly now!

Resolved in [e7c7f10](https://github.com/uswds/uswds/pull/5063/commits/e7c7f108afe7dc90b9764122c356dd922ab2d1cf)

mejiaj (approved):
Double-checked the button and button group styles. No issues found!

amyleadem (approved):
All changes since my last approval look good!

packages/usa-file-input/src/index.js
```diff
@@ -147,6 +158,11 @@ const buildFileInput = (fileInputEl) => {
     disable(fileInputEl);
   }
 
+  // Aria-disabled styling
+  if (ariaDisabled) {
+    ariaDisable(fileInputEl);
```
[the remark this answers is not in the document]
thisisdano: @mahoneycm We should probably fix this too, if it looks like a simple fix, even though I'd guess this is more of an edge case, having an error state on a disabled element

thisisdano (changes requested):
@mejiaj Had a couple observations that are worth addressing, especially the Combo box issue

thisisdano (approved):
My comments above came from an incomplete and incorrect understanding of `aria-disabled` — which is, by design, responsible only for conveying disabled state to screen reader. See [this MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled).

Since this PR is about normalizing styles and not about JS, I think we can ignore my comments above and proceed.

PR #5068, opened 2022-12-13
In-Page Navigation: Update page URL when scrolling to section

### LINKED ISSUE #5056: USWDS - Feature: In-page navigation: Update URL fragment when navigating

As a user, I'd like to be able to either share or bookmark a specific part of a page that I access using the in-page navigation component.

When using the in-page navigation to navigate to a part of the page, ideally the hash fragment portion of the browser URL would be updated to reflect the ID of the element in focus.

Why?

- To match the native browser behavior of [an anchor element which links to an element on the same page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#linking_to_an_element_on_the_same_page)
- To make it easier to share the link to a specific part of the page

N/A

### Additional context

Steps to reproduce:

1. Go to https://designsystem.digital.gov/components/accordion/
2. Click "Component code" link in the "On this page" in-page navigation

Expected: Browser URL is https://designsystem.digital.gov/components/accordion/#accordion-code
Actual: Browser URL is https://designsystem.digital.gov/components/accordion/

### Code of Conduct

aduth:
Closes #5056

## Description

Enhances the In-Page Navigation component to update the current URL when scrolling to a section. 

## Additional information

See #5056 for more details about rationale.

### DIFF
--- a/packages/usa-in-page-navigation/src/index.js
+++ b/packages/usa-in-page-navigation/src/index.js
@@ -99,7 +99,9 @@ const handleScrollToSection = (el) => {
     block: "start",
   });
 
-  window.location.hash = el.id;
+  if (window.location.hash.slice(1) !== el.id) {
+    window.history.pushState(null, "", `#${el.id}`);
+  }
 };
 
 /**
--- a/packages/usa-in-page-navigation/src/index.js
+++ b/packages/usa-in-page-navigation/src/index.js
@@ -98,6 +98,8 @@ const handleScrollToSection = (el) => {
     top: el.offsetTop - inPageNavScrollOffset,
     block: "start",
   });
+
+  window.location.hash = el.id;
 };
 
 /**
@@ -146,13 +148,14 @@ const createInPageNav = (inPageNavEl) => {
     const anchorTag = document.createElement("a");
     const textContentOfLink = el.textContent;
     const tag = el.tagName.toLowerCase();
+    const linkHref = `#section_${i}`;
 
     listItem.classList.add(IN_PAGE_NAV_ITEM_CLASS);
     if (tag === "h3") {
       listItem.classList.add(SUB_ITEM_CLASS);
     }
 
-    navLinks.setAttribute("href", `#section_${i}`);
+    navLinks.setAttribute("href", linkHref);
     navLinks.setAttribute("class", IN_PAGE_NAV_LINK_CLASS);
     navLinks.textContent = textContentOfLink;
 
@@ -162,6 +165,10 @@ const createInPageNav = (inPageNavEl) => {
 
     inPageNavList.appendChild(listItem);
     listItem.appendChild(navLinks);
+
+    if (window.location.hash === linkHref) {
+      handleScrollToSection(anchorTag);
+    }
   });
 
   inPageNavEl.appendChild(inPageNav);
--- a/packages/usa-in-page-navigation/src/index.js
+++ b/packages/usa-in-page-navigation/src/index.js
@@ -103,6 +103,21 @@ const handleScrollToSection = (el) => {
 };
 
 /**
+ * Scrolls the page to the section corresponding to the current hash fragment, if one exists.
+ */
+const scrollToCurrentSection = () => {
+  const hashFragment = window.location.hash.slice(1);
+  if (hashFragment) {
+    const anchorTag = document.querySelector(
+      `.${IN_PAGE_NAV_ANCHOR_CLASS}#${hashFragment}`
+    );
+    if (anchorTag) {
+      handleScrollToSection(anchorTag);
+    }
+  }
+};
+
+/**
  * Create the in-page navigation component
  *
  * @param {HTMLElement} inPageNavEl The in-page nav element
@@ -148,14 +163,13 @@ const createInPageNav = (inPageNavEl) => {
     const anchorTag = document.createElement("a");
     const textContentOfLink = el.textContent;
     const tag = el.tagName.toLowerCase();
-    const linkHref = `#section_${i}`;
 
     listItem.classList.add(IN_PAGE_NAV_ITEM_CLASS);
     if (tag === "h3") {
       listItem.classList.add(SUB_ITEM_CLASS);
     }
 
-    navLinks.setAttribute("href", linkHref);
+    navLinks.setAttribute("href", `#section_${i}`);
     navLinks.setAttribute("class", IN_PAGE_NAV_LINK_CLASS);
     navLinks.textContent = textContentOfLink;
 
@@ -165,10 +179,6 @@ const createInPageNav = (inPageNavEl) => {
 
     inPageNavList.appendChild(listItem);
     listItem.appendChild(navLinks);
-
-    if (window.location.hash === linkHref) {
-      handleScrollToSection(anchorTag);
-    }
   });
 
   inPageNavEl.appendChild(inPageNav);
@@ -235,6 +245,7 @@ const inPageNavigation = behavior(
     init(root) {
       selectOrMatches(`.${IN_PAGE_NAV_CLASS}`, root).forEach((inPageNavEl) => {
         createInPageNav(inPageNavEl);
+        scrollToCurrentSection();
       });
     },
   }
--- a/packages/usa-in-page-navigation/src/index.js
+++ b/packages/usa-in-page-navigation/src/index.js
@@ -258,7 +258,7 @@ const handleEnterFromLink = (event) => {
   } else {
     // throw an error?
   }
-  handleScrollToSection(target);
+  handleScrollToSection(targetAnchor);
 };
 
 const inPageNavigation = behavior(
--- a/packages/usa-in-page-navigation/src/index.js
+++ b/packages/usa-in-page-navigation/src/index.js
@@ -13,6 +13,7 @@ const IN_PAGE_NAV_SCROLL_OFFSET = 0;
 const IN_PAGE_NAV_ROOT_MARGIN = "0px 0px 0px 0px";
 const IN_PAGE_NAV_THRESHOLD = "1";
 const IN_PAGE_NAV_CLASS = `${PREFIX}-in-page-nav`;
+const IN_PAGE_NAV_ANCHOR_CLASS = `${PREFIX}-anchor`;
 const IN_PAGE_NAV_NAV_CLASS = `${IN_PAGE_NAV_CLASS}__nav`;
 const IN_PAGE_NAV_LIST_CLASS = `${IN_PAGE_NAV_CLASS}__list`;
 const IN_PAGE_NAV_ITEM_CLASS = `${IN_PAGE_NAV_CLASS}__item`;
@@ -53,6 +54,18 @@ const getSectionHeadings = () => {
 };
 
 /**
+ * Return a node list of section anchor tags
+ *
+ * @return {HTMLElement[]} - An array of DOM nodes
+ */
+const getSectionAnchors = () => {
+  const sectionAnchors = document.querySelectorAll(
+    `.${IN_PAGE_NAV_ANCHOR_CLASS}`
+  );
+  return sectionAnchors;
+};
+
+/**
  * Generates a unique ID for the given heading element.
  *
  * @param {HTMLHeadingElement} heading
@@ -60,10 +73,6 @@ const getSectionHeadings = () => {
  * @return {string} - Unique ID
  */
 const getHeadingId = (heading) => {
-  if (heading.id) {
-    return heading.id;
-  }
-
   const baseId = heading.textContent
     .toLowerCase()
     .replace(/[^a-z\d]/g, "-")
@@ -181,12 +190,10 @@ const createInPageNav = (inPageNavEl) => {
   inPageNavList.classList.add(IN_PAGE_NAV_LIST_CLASS);
   inPageNav.appendChild(inPageNavList);
 
-  const observeSections = new window.IntersectionObserver(setActive, options);
   sectionHeadings.forEach((el) => {
-    observeSections.observe(el);
-
     const listItem = document.createElement("li");
     const navLinks = document.createElement("a");
+    const anchorTag = document.createElement("a");
     const textContentOfLink = el.textContent;
     const tag = el.tagName.toLowerCase();
 
@@ -195,15 +202,25 @@ const createInPageNav = (inPageNavEl) => {
       listItem.classList.add(SUB_ITEM_CLASS);
     }
 
-    // eslint-disable-next-line no-param-reassign
-    el.id = getHeadingId(el);
+    const headingId = getHeadingId(el);
 
-    navLinks.setAttribute("href", `#${el.id}`);
+    navLinks.setAttribute("href", `#${headingId}`);
     navLinks.setAttribute("class", IN_PAGE_NAV_LINK_CLASS);
     navLinks.textContent = textContentOfLink;
 
+    anchorTag.setAttribute("id", headingId);
+    anchorTag.setAttribute("class", IN_PAGE_NAV_ANCHOR_CLASS);
+    el.insertAdjacentElement("afterbegin", anchorTag);
+
     inPageNavList.appendChild(listItem);
     listItem.appendChild(navLinks);
+  });
+
+  const anchorTags = getSectionAnchors();
+  const observeSections = new window.IntersectionObserver(setActive, options);
+
+  anchorTags.forEach((tag) => {
+    observeSections.observe(tag);
   });
 
   inPageNavEl.appendChild(inPageNav);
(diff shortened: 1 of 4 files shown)


packages/usa-in-page-navigation/src/index.js
```diff
@@ -162,6 +165,10 @@ const createInPageNav = (inPageNavEl) => {
 
     inPageNavList.appendChild(listItem);
     listItem.appendChild(navLinks);
+
+    if (window.location.hash === linkHref) {
+      handleScrollToSection(anchorTag);
+    }
```
mejiaj, line 171: Thoughts on moving this into it's own function and call it after the component has been created?

Something similar to `handleClickFromLink()`. It might help improve clarity.

mejiaj (commented):
A nice improvement. Small comment on separating the location check to improve readability. 

Otherwise LGTM, thanks. 😄

packages/usa-in-page-navigation/src/index.js
aduth: @mejiaj Are you thinking it would be something we would call in the behavior's `init`, like `scrollToCurrentSection`, where the function would try to find and scroll to the heading associated with the current URL fragment?

mejiaj: @aduth yes, that seemed like the best place at the time of original review. Maybe I misunderstood the intent.

aduth: I think that could make sense to separate. I pushed up some changes in 786d92e8d. Is that what had in mind?

mejiaj (commented):
@aduth sorry, I missed your response. Yes, that's what I was thinking. Separating the function to keep the code more readable.

The only quirk I see is that smooth scrolling isn't working. If I click on a branch locally it just jumps immediately to the section:
[aduth-inpage_nav-2023-01-26 at 09.27.57.webm](https://user-images.githubusercontent.com/3385219/214881339-9e4ea5a9-5f41-4176-8175-1117411d4ba7.webm)


Here's how it looks in [develop](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=components-in-page-navigation--default&viewMode=story)

[develop-inpage_nav-2023-01-26 at 09.44.03.webm](https://user-images.githubusercontent.com/3385219/214881194-a6f9e2d0-a4f6-4d49-918f-9e00f5ece7d8.webm)

aduth:
Good catch @mejiaj . I think that's likely a result of how the `handleScrollToSection` function is now assigning the URL hash, effectively bypassing the smooth scrolling.

My week is starting rather busy unfortunately, but I'll plan to take a closer look once time allows.

Some potential avenues for fixes I might want to explore:

- Invert hash assignment to happen by default link click behavior, rather than `window.location.hash` assignment
- Consider [`scroll-behavior` CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior), which may interoperate better with `window.location.hash` assignment

mejiaj:
@aduth understandable and your help is much appreciated! 

Those ideas sound good. `scroll-behavior` CSS looks to have much better support now, interesting.

amyleadem:
@aduth  @mejiaj 
I like this added functionality! 
This might be out of scope for this PR, but I see an opportunity to make the url more readable by changing the generated id (and resulting hash) from `section_X` to the header title string (possibly truncated, with dashes in place of spaces). Curious to hear what you think.

aduth:
@amyleadem Yes, I agree, I wish the generated hashes were more accurate based on the title of the section. This behavior is inherited from [code that already exists](https://github.com/uswds/uswds/blob/v3.3.0/packages/usa-in-page-navigation/src/index.js#L155-L159) and ideally I'd hoped not to have to alter that. However, I think if my pull request is the first time those generated hashes are made user-facing in that it'd be expected a user might copy the URL from their address bar, it could be good a good idea to consider that in-scope of this work.

aduth:
@mejiaj @amyleadem Apologies for the delay in revisiting this; it's been a hectic few weeks for me 😅 

I pushed up two additional commits which should hopefully address the feedback:

* [d82568b](https://github.com/uswds/uswds/pull/5068/commits/d82568b69d3f7904a29833544a9b7a7d14420cbb) ought to address the issue where smooth scroll was not happening as expected
* [66e0232](https://github.com/uswds/uswds/pull/5068/commits/66e0232f398ef2da7aaf252838f8f57137edfb56) introduces a revised way for generating IDs based on heading content (rather than a generic "section_#"). I've handled a few edge cases already in the logic, but it's probably worth taking a detailed look at how this might handle more extreme cases.

aduth:
> Actually, I might need to make a couple more revisions, since [66e0232](https://github.com/uswds/uswds/commit/66e0232f398ef2da7aaf252838f8f57137edfb56) appears to have broken the behavior of the intersection observer (updating the hash as the user scrolls). There weren't existing tests for this, so I might look to add those as well to prevent future issues.

Nevermind, I misunderstood what the intersection observer was for. I see that it's meant to be for setting the current section as "active", which appears to be working correctly still in my testing.

amyleadem:
Thanks so much for your work on this @aduth! I am excited about the updated hash. I will continue digging in for a review, but wanted to flag a couple quirks I found while doing initial functionality testing:

### Current status is not always triggered when scrolling up
If you are low on the page (for example, in the "vitae ligula" section) and click on an item in the in-page nav that is above where you are (say, the "Consectetuer adipiscing elit" section), the selected item does not receive the current status and does not show the vertical scroll bar. For the current status to activate, you need to scroll up a tiny bit. 

This issue does not exist on the [current implementation of the component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=components-in-page-navigation--default&viewMode=story) found in `develop`.

image

In this PR, the full headline must be visible before triggering the current state. In the current build, only the bottom part of the headline must be visible to trigger the current state:

image

amyleadem (changes requested):
Thanks so much for your work on this @aduth! I am excited about this upgrade! I will continue digging in for a deeper review, but wanted to flag a couple quirks I found during functionality testing:

### Current status is not always triggered when scrolling up
 If you are in a low section on the page (for example, in the "vitae ligula" section) and click on an item in the in-page nav that is above where you are (say, the "Consectetuer adipiscing elit" section), the selected item does not receive the current status and does not show the vertical scroll bar. For the current status to activate, you need to scroll up a tiny bit. This issue does not exist on the [current implementation of the component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=components-in-page-navigation--default&viewMode=story) found in `develop`.

This issue appears in Chrome, Safari, and Edge for Mac. It does NOT happen in Firefox. 

image

One thing I noticed is that in this PR, the full headline must be visible before triggering the current state. In the current implementation, only the bottom part of the headline must be visible to trigger the current state. I am guessing this is contributing to the issue:

image

### Links now turn purple after click
In-page nav links are now registering as visited when clicked.  It makes sense to me to style these items so that the `:visited` links look the same as the unvisited links, but I can see how there might be differing opinions on that. 

image

I will return to review this more in-depth tomorrow. Let me know if you have any questions in the meantime.

Thanks again!

aduth:
@amyleadem Those are both great points, let me see if I can fix 'em up!

aduth:
Ok, pushed a couple more revisions!

- e0d704e should fix the issue with "active" links not being highlighted as expected, by restoring the original behavior of using a newly-created anchor link rather than setting the ID on the heading directly. While I kinda hoped to be able to avoid the new element, this simplifies the diff, avoids the issues with the "active" states, and may avoid some compatibility issues if the consuming project is doing its own things with the heading attributes.
- 74f658c avoids the visited coloring of the navigation items. The code for this is a bit more complex than I might have anticipated 😄 , but there didn't seem to be a way with the current `set-link-from-bg` mixin to force a specific visited color.

amyleadem (commented):
Everything seems to be working well, @aduth! 
I'll hold off on approving until @mejiaj takes a deeper look at the JS, but it is looking good to me. Thank you!

- [x] Run `npm start`, `npm test` , `npm run lint`, `npm run prettier:js` without error
- [x] Confirm that `usa-current` is appropriately applied when moving to a lower section (Both during scrolling and after selecting an in-page nav item)
- [x] Confirm that `usa-current` is appropriately applied when moving to a higher section  (Both during scrolling and after selecting an in-page nav item)
- [x] Confirm smooth scroll when a new in-page nav item is selected
- [x] Confirm that in-page navigation visually matches the existing component in `develop`
- [x] Confirm that a url hash is generated from the header text
    - Note: Hash is generated when an item is selected via mouse or keyboard; it is not generated during scroll. This probably is expected behavior but wanted to flag it just in case.
- [x] Confirm that the generated hash handles special characters
    - Tested the following string in the header: "-#!A-b@--1---2._~!@#$%^&*()_+`=[{]}|;:'<>?,./-*"; results came out as expected:
        - Generated url hash is: #a-b-1-2
        - Generated nav item text is: -#!A-b@--1---2._~!@#$%^&*()_+`=[{]}|;:'<>?,./-*
- [x] Confirm that a unique hash is generated when two headers have the same text
    - Added a second "Vitae ligula" header in tests; results came out as expected:
        - Generated url hash is: "#vitae-ligula-2"
- [x] When loading a url that includes a generated section hash (ex: localhost:6006/iframe.html?args=&id=components-in-page-navigation--default&viewMode=story#nullam-sit-amet-enim), confirm that `usa-current` is applied to the appropriate nav item and the appropriate section is shown in the viewport
- [x] Test functionality in Safari, Firefox, Chrome, Edge

packages/usa-in-page-navigation/src/index.js
```diff
@@ -65,6 +65,39 @@ const getSectionAnchors = () => {
...
+  do {
+    id = baseId;
+
+    // To avoid conflicts with existing IDs on the page, loop and append an
+    // incremented suffix until a unique ID is found.
+    suffix += 1;
+    if (suffix > 1) {
+      id += `-${suffix}`;
+    }
+  } while (document.getElementById(id));
```
mejiaj, line 96: I really like this idea and could see it as its own utility in the future.

Only concern is it looks like we're looping over the headers twice? Since we're calling `getHeadingId(el)` inside of the `sectionHeadings.forEach((el)`.

packages/usa-in-page-navigation/src/index.js
```diff
@@ -98,6 +131,23 @@ const handleScrollToSection = (el) => {
     top: el.offsetTop - inPageNavScrollOffset,
     block: "start",
   });
+
+  if (window.location.hash.slice(1) !== el.id) {
+    window.history.pushState(null, "", `#${el.id}`);
+  }
```
mejiaj, line 137: Nice feature! I was able to go forwards/back on previously clicked in-page nav items.

mejiaj (commented):
@aduth a few questions, but overall this looks pretty good.

I tested:
- [x] Smooth scrolling in Chrome/Firefox
- [x] Regex patterns match and work as expected
  - Tested following `section_heading` values in `usa-in-page-navigation.json`:
    - …All previous headers
    - `--Aliquam erat volutpat: velit vitae ligula volutpat--`
    - `Duplicate Regex tester`
    - `Duplicate Regex tester`
    - `$$$Regex tester----hello&world! Lorem$Ipsum!`
- [x] Compared functionality and visuals with previous component
- [x] Mobile behavior (⚠️ Seems to scroll further than previous)
- [x]  Push state forward/backward works as expected (⚠️ Focus state disappears)

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
```diff
@@ -48,28 +49,68 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
+        HTMLElement.prototype,
+        "offsetTop"
+      );
+      Object.defineProperty(HTMLElement.prototype, "offsetTop", {
+        get() {
+          const heading = this.closest("h2,h3");
+
+          let index = 0;
+          let sibling = heading;
+          while (true) {
```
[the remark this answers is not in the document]
aduth: This tries to simulate a vertical offset in order to test that `offsetTop` used for scrolling is behaving as expected, since JSDOM doesn't emulate that. The logic essentially considers computes the index of the current heading and multiplies that by a constant number. I'm open to other ideas for how to ensure the scrolling occurs as expected.

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
```diff
@@ -84,5 +125,47 @@ tests.forEach(({ name, selector: containerSelector }) => {
       resizeTo(1024);
       assertHidden(theList, false);
     });
+
+    it("assigns id to section headings", () => {
+      const ok = [
```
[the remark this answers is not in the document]
aduth: Yes, correct, this tests against the template.

packages/usa-in-page-navigation/src/index.js
aduth: While this is technically a loop, I'd expect it would only run once in the vast majority of cases, since it would only ever run more than once if there was already an ID somewhere in the page which was the same as our computed ID using the heading's text.

packages/usa-in-page-navigation/src/index.js
aduth: Yeah this was the trick to solve [the issue you identified earlier](https://github.com/uswds/uswds/pull/5068#pullrequestreview-1271308864), since assigning the hash fragment this way appears to work the same as `window.location.hash = el.id` with the exception of scrolling, which is what we wanted to avoid.

aduth:
Thanks @mejiaj for the review! I pushed a few clarifying code comments in b051d44 based on some of the questions you raised.

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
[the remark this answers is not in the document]
mejiaj: I see, thanks. I couldn't find any alternative solutions to `offsetTop` in their [issues queue](https://github.com/jsdom/jsdom/issues?q=offsetTop+sort:updated-desc).

I guess what threw me off initially was the `while (true)` inside of `Object.defineProperty`.

mejiaj (approved):
Thanks for all your work on this. I appreciate the code comments.

amyleadem:
@aduth 
cc: @mejiaj 

Thanks for all your work on this. I did another review and everything is working as expected, except I found two items related to keyboard navigation:
- Selecting an item with `enter` does not generate a hash (but it successfully generates a hash on click)
- Tab navigation and Focus states do not work in Safari (this appears to be an existing bug that also happens on [develop](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=components-in-page-navigation--default&viewMode=story))

Outside of that, it is looking good to me!

aduth:
Great catch @amyleadem ! I'll take a look at the issue with changing hash on pressing <kbd>Enter</kbd>.

For the second item, I suspect this might be due to a Safari default setting which doesn't navigate links using <kbd>Tab</kbd> by default, but you can change this in Preferences > Advanced. Can you confirm that it works as expected after clicking this checkbox?

Screen Shot 2023-03-07 at 12 03 06 PM

amyleadem:
@aduth Thanks for the quick response! Updating that setting in Safari does fix the problem. I am able to tab into the component with this setting turned on. 

Since it was a pre-existing issue, it should be fine to ignore the Safari issue in this PR. I have opened issue https://github.com/uswds/uswds/issues/5169 so that we can evaluate if this meets our browser compatibility standards. @mejiaj, let me know if you have any objections to this.

amyleadem:
Apologies, I accidentally closed this due to some clumsy clicking. It was not intentional!

aduth:
The changes in 3163fd1 should hopefully address the issues with the <kbd>Enter</kbd> keypress navigation, if you're able to take another look @amyleadem

amyleadem (approved):
This looks good to me! Thanks for your effort here!

thisisdano (approved):
This is a nice new feature with a thoughtful implementation! Thanks, and thanks for the thorough reviews!

thisisdano:
Oh bother. This came in after I'd already created the release branch, then I guess I never updated the release branch. We'll get a 3.4.1 out asap to fix this!

thisisdano:
@aduth This should now be live in [USWDS 3.4.1](https://github.com/uswds/uswds/releases/tag/v3.4.1). Sorry for the mix-up!

aduth:
No worries, thanks for the quick turnaround on the patch release!

PR #5090, opened 2022-12-30
[alpha] USWDS - Core: use ESM

mejiaj:
## Summary

**USWDS Core JS now uses ES module syntax.** Scripts in uswds-core (including unit test) use ESM syntax with `.mjs` file extension. Files were previously using CommonJS syntax. Part one of migration.

## Breaking change

:warning: This is potentially a breaking change. Core file extensions have changed from `.js → .mjs` to prevent Node errors with using ESM.

## Related issue

This is the first of a few PRs for converting JS. Related to #4515 and #4987.
 
## Preview link

Preview link: n/a

## Major changes

This PR does the following:
- Adds MJS extension support to eslint and tsconfig [dc423d4eff54f7306c6aba98988a4d2cf68d4d13, ae88c3ec88dcb590da121c3c18633b2674909aed]
- Converted util scripts to ESM [8c6f646e11ba4dca8334a22edfab21b59152017d]
- Converted polyfills to ESM [1f342dc238964eaa938bc08b7c6f299b3537cf16]
- Converted core scripts to use ESM [05290fba63cfbcbe08720290755e68f8d8f8beb1]
- Updated gulp tasks [8e23c7d626839650717cbccfa22582f091aeb312]
  - Replaced `gulp-spawn-mocha` dependency because it was tied to Mocha v6, which doesn't support ESM
  - Removed duplicate typecheck task
  - Update mocha to v10
  - Updated entrypoint references to JS 
    ⚠️ Note the compile task will fail because component scripts still use CJS

## Testing and review

1. Confirm that `test.js` is only running tests on core files
2. Run `npm test`
3. All unit tests (both SASS and JS) should pass
4. Confirm that all uswds-core scripts (except `uswds-init.js`) use `import` instead of `require()`
5. Confirm that all uswds-core scripts use MJS extension (except `uswds-init.js`)

⚠️ If you encounter an error on `npm install` — that's expected. The code is compiled via a NPM post install hook. Because only core has been migrated you might see an error like the one below:

```bash
[10:06:07] Error: Can't walk dependency graph: Cannot find module './config' from '/Users/aleadem/Dev/github/uswds/packages/uswds-core/src/js/start.mjs'
    required by /Users/aleadem/Dev/github/uswds/packages/uswds-core/src/js/start.mjs
    at /Users/aleadem/Dev/github/uswds/node_modules/resolve/lib/async.js:167:35
    at load (/Users/aleadem/Dev/github/uswds/node_modules/resolve/lib/async.js:186:43)
[...]

PR #5091, opened 2023-01-03
[alpha] USWDS - Components: use ESM syntax

### LINKED ISSUE #4515: Use ECMAScript Modules instead of CommonJS
ECMAScript Modules (aka "ES6 Modules" or "ESM") is a module system that natively works in most browsers. Newer versions of Node.js also natively support it. The USWDS javascript code should switch from the current module system, CommonJS, to using ECMAScript Modules.

**What makes ESM better?**

 * Tools like Rollup and Webpack can tree-shake ESM dependencies, but can't tree-shake CommonJS dependencies. If we want to do the following:
    ```js
    import { accordion } from 'uswds/src/js/components'
    import domready from 'domready'
    domready(() => {
        accordion.on()
    })
    ```
    Webpack or Rollup knows that we only need the `accordion` component's javascript, and it can delete all other component javascript.
    Caveat: tree shaking can be finicky, and other changes may need to be made to get it fully working. Those changes could be fast-follows to the initial work of getting ESM working.
 * ESM can be transpiled down to CommonJS, and then run through browserify like we do today. But in a few years when browser support for ESM is sufficient, the transpilation can simply be turned off and we can switch to directly taking advantage of native browser module support.

**Additional context**

In some ways this is the javascript sister issue to #2854.

### LINKED ISSUE #4987: Change all JS imports from 'required' to 'import'

mejiaj:
## Summary

Part two of migration to ESM. First part is migrating uswds-core https://github.com/uswds/uswds/pull/5090. This PR converts component JS and unit tests to ESM, JS bundling (with gulp-esbuild), and updates package exports.

This PR **only** migrates component code. Migrating the entire codebase has bigger implications. We would need to:

1. Add `"type": "module"` to `package.json` [low LOE]
2. Handle gulp scripts
     a. Keep as-is, but use CJS extension [very low LOE]
     b. Convert to ESM _(minor issues with importing JSON files)_ [low LOE]
3. Update storybook to _at least_ `7.0.0-alpha.0` [high LOE]

### For users

ESM is the standard module syntax which is being used by most, if not all JS frameworks. It's the recommended standard for build tools like esbuild and webpack. 

### For core developers

For core developers, we'll be able to:
1. Use this standard module syntax in development
2. VS Code has auto-completion for imports
4. Use ESM-only packages (so far only `del` — used in `clean.js` and `svg-sprite.js`). This isn't a major benefit and we should move away or limit use of gulp.

## Breaking change

> **Warning**
> This would be a breaking change for CommonJS users. Although none of the projects on the [Implementations page](https://designsystem.digital.gov/documentation/implementations/) use CommonJS. Bundlers like RollupJS _can_ handle importing ESM as CJS.

Users who import `uswds.js` or `uswds.min.js` are **not** affected.

## Related issue
Closes #4515, closes #4987. 

## Preview link

Preview link: [Library preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/alpha-esm--components/
)

## Major changes

- Updated node requirements for ESM (node engine 12 & typescript) [1cc70d7]
- Updated mocha requirements for ESM [abcdca3]
- Converted component scripts and unit tests to ESM [6972a90 & 8c7073a]
- Removed `object-assign` and `classList` polyfills [ce77e83]
- Replaced browserify for bundling with gulp-esbuild [cde4022]
- Updated package.json entrypoints [a860f1f]

## Testing and review

1. Test via `npm test` should pass (specifically typecheck, lintSass, and unit tests)
2. JS build task `npx gulp buildJS` should run without errors
3. Bundled JS `uswds.js` should still initialize and make interactive components work
5. You should be able to import USWDS [bundler like webpack required]:
    ```js
    // Either this  
    import USWDS from "@uswds/uswds/src/js";
    const { accordion } = USWDS;
  
    // Or this
    import accordion from "@uswds/uswds/js/usa-accordion";
      
    // In react
    useEffect(() => {
      // initialize accordion
      accordion.on(ref);
  
      return () => {
        accordion.off();
      }
    });
    ```

## Dependency updates

| Dependency name    | Previous version | New version |
| ------------------ | :--------------: | :---------: |
| classlist-polyfill |      1.0.3       |      -      |
| object-assign      |      4.1.1       |      -      |
| @babel/core        |      7.16.7      |      -      |
| @babel/preset-env  |      7.16.7      |      -      |
| @types/node        |     16.11.19     |  18.11.18   |
| babelify           |      10.0.0      |      -      |
| browserify         |      17.0.0      |      -      |
| gulp-esbuild       |        -         |   0.11.0    |
| gulp-mocha         |        -         |    8.0.0    |
| gulp-spawn-mocha   |      6.0.0       |      -      |
| mocha              |      6.2.3       |   10.2.0    |
| typescript         |      4.4.4       |    4.9.4    |


.eslintrc.yml
```diff
@@ -16,9 +16,7 @@ rules:
...
   no-unsanitized/property: error
 overrides:
-  - { files: ['*.spec.js'],
-      rules: {
-        no-unsanitized/method: off,
-        no-unsanitized/property: off
-        }
-      }
+  - {
+      files: [".js", ".mjs"],
```
mejiaj, line 20: Adds support for `.mjs` files.

gulpfile.js
```diff
@@ -9,7 +9,7 @@ const { noCleanup, noTest } = require("./tasks/flags");
 const { buildSprite, buildSpriteStandalone } = require("./tasks/svg-sprite");
 const { compileJS, typeCheck } = require("./tasks/javascript");
 const { unitTests, sassTests } = require("./tasks/test");
-const { lintSass, typecheck } = require("./tasks/lint");
+const { lintSass } = require("./tasks/lint");
```
mejiaj, line 12: Duplicate typecheck task. It's also included in `javascript.js`.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "3.3.0",
   "description": "Open source UI components and visual style guide for U.S. government websites",
   "engines": {
-    "node": ">= 4"
+    "node": ">= 12"
```
mejiaj, line 6: Minimum supported engine for ESM.

packages/_usa-password/src/index.mjs
```diff
@@ -0,0 +1,17 @@
+import behavior from "../../uswds-core/src/js/utils/behavior.mjs";
+import toggleFormInput from "../../uswds-core/src/js/utils/toggle-form-input.mjs";
+import { CLICK } from "../../uswds-core/src/js/events.mjs";
+import { prefix as PREFIX } from "../../uswds-core/src/js/config.mjs";
```
mejiaj, line 4: File extension for relative files is a requirement for mocha. We could turn this off, but I wanted to see how it affected package exports.

For now, it seems like there is **no** effect.

packages/usa-accordion/src/test/accordion.spec.mjs
```diff
@@ -1,6 +1,11 @@
...
-const fs = require("fs");
-const Accordion = require("../index");
+import assert from "assert";
+import fs from "fs";
+import { fileURLToPath } from "url";
+import path from "path";
+import Accordion from "../index.mjs";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
mejiaj, line 8: This ESM method of referencing `__dirname` requires 4 lines (2 imports and the `const`), instead of the CJS method of simply using `__dirname`.

packages/usa-combo-box/src/index.mjs
```diff
@@ -1,9 +1,9 @@
-const keymap = require("receptor/keymap");
-const selectOrMatches = require("../../uswds-core/src/js/utils/select-or-matches");
-const behavior = require("../../uswds-core/src/js/utils/behavior");
-const Sanitizer = require("../../uswds-core/src/js/utils/sanitizer");
-const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
-const { CLICK } = require("../../uswds-core/src/js/events");
+import keymap from "receptor/keymap/index.js";
```
mejiaj, line 1: **TODO: Investigate further**

The component worked with importing keymap without extension. Updated to this path because mocha couldn't find the import.

packages/usa-footer/src/index.mjs
```diff
@@ -88,10 +88,12 @@ module.exports = behavior(
       this.mediaQueryList = window.matchMedia(
         `(max-width: ${HIDE_MAX_WIDTH - 0.1}px)`
       );
+      // ! Deprecated feature
       this.mediaQueryList.addListener(resize);
     },
 
     teardown() {
+      // ! Deprecated feature
       this.mediaQueryList.removeListener(resize);
```
mejiaj, line 97: **TODO: Deprecated feature**

The `mediaQueryList.addListener` and `removeListener` are deprecated features per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener). Probably why the unit test `footer.spec.mjs` imports `mq-polyfill` because of this. Removing the polyfill causes the test to fail.

packages/uswds-core/src/js/utils/test/select.spec.mjs
```diff
@@ -1,11 +1,19 @@
...
+import path from "path";
+import assert from "assert";
+import select from "../select.mjs";
 
-const TEMPLATE = fs.readFileSync(`${__dirname  }/select.template.html`);
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
amyleadem, line 11: These are declared twice. Is that intentional?

mejiaj: @amyleadem for `__filename` and `__dirname`? If so, I added a comment above https://github.com/uswds/uswds/pull/5091#discussion_r1061589161.

Basically the original `__dirname` only works in CJS. To get the same functionality in ESM we need a few more lines of code.

ethangardner:
I rescoped the work yesterday to go package-by-package to help control the risk as we migrate to ESM format for the internals of the project. That work will live in https://github.com/uswds/uswds/issues/6845 as the epic. I'll close this PR and linked issues as a result. I also added a new issue https://github.com/uswds/uswds/issues/6844 to address the deprecation that was pointed out here: https://github.com/uswds/uswds/pull/5091#discussion_r1061604520.

Issue #5054, opened 2022-11-28
USWDS - Input Mask refactor


jonadecker:

mitchmoccia:
The input mask rewrite is in a good place and a PR has been created against the idp-staging branch:

https://github.com/uswds/uswds/pull/5104

Component page:
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/mm-input-mask/?path=/story/components-form-inputs-text-input-mask--ssn

Updated guidance page:
https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/mm-input-mask-guidance/components/input-mask/

Highlights include:
- Countless improvements implemented based on feedback from version 1 and more
  - Primarily instant feedback and error handling has been implemented (similar to character count)
- Feedback from all our syncs integrate
- JSDocs added to each function
- Unit tests worked out for all five examples
- Guidance created on how to implement the new version

mejiaj:
@amyleadem do you feel this is possible for 3.7.0 (end of this sprint) or should we bump back?

brunerae:
This now duplicates #5442

PR #5104, opened 2023-01-20
USWDS input mask version 2

mitchmoccia:
This is a pretty large rewrite of the input mask component with a ton of improvements. I tried to keep the requirements to implement as close to version 1 as possible, however, there are some additional requirements that a developer must implement if they upgrade to this version.

The most important update required is the addition of a `.usa-input-mask` wrapper and the class `.usa-input-mask__field` added to the input element. This allowed for some great feedback if a user types an incorrect character. This also follows the character count architecture FYI.

Here is updated input mask implementation guidance featuring previews with working masks:
https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/mm-input-mask-guidance/components/input-mask/

Here is the component page:
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/mm-input-mask/?path=/story/components-form-inputs-text-input-mask--ssn

UPDATE: Here is the uswds-site PR for the guidance udpates:
https://github.com/uswds/uswds-site/pull/1995

Looking forward to feedback!


packages/usa-input-mask/src/content/usa-input-mask-date-short.json
```diff
@@ -0,0 +1,10 @@
+{
+  "hint": "For example, 01/01/1970",
+  "id": "us-date",
+  "inputmode": "numeric",
+  "label": "US Date",
+  "mask": "99/99/9999",
+  "pattern": "\\d{2}/\\d{2}/\\d{4}",
```
amyleadem, line 7: I wonder if this accepted pattern could be more constrained so that users can't input an invalid date like 33/33/3333? Not sure if that is out of scope...

packages/usa-input-mask/src/content/usa-input-mask-date-short.json
```diff
@@ -0,0 +1,10 @@
+{
+  "hint": "For example, 01/01/1970",
+  "id": "us-date",
+  "inputmode": "numeric",
+  "label": "US Date",
```
amyleadem, line 5: Should all labels with `US` instead say `U.S.`? It seems to be how we use it elsewhere on the site. Not sure if there is a universal stance on this.

packages/usa-input-mask/src/content/usa-input-mask-date-short.json
```diff
@@ -0,0 +1,10 @@
+{
+  "hint": "For example, 01/01/1970",
```
amyleadem, line 2: It is unclear if this example is telling the user to input MM-DD-YYYY or DD-MM-YYYY. Could the example be something that more clearly describes the format? It might not matter for our examples, but I want to make sure we are guiding users to write helpful hints.

packages/usa-input-mask/src/content/usa-input-mask-alphanumeric.json
```diff
@@ -1,10 +1,11 @@
 {
-  "label": "Alphanumeric",
-  "hint": "For example, A1B 2C3",
+  "forceupper": "true",
```
amyleadem, line 2: Small change, but I found the `forceupper` text field control to be a bit clunky in Storybook. If we remove the quotes, the control will appear as a boolean toggle instead of a text field, which seems easier to control.

```suggestion
  "forceupper": true,
```

image
image

packages/usa-input-mask/src/content/usa-input-mask-date-short.json
mitchmoccia: Hi Amy! Thanks for commenting on this... We could and maybe should do this! I wanted to explain a little on why this would be "separate" from the inner workings of the input mask JS and why, even with the pattern set to force an actual valid date, a user would still be able to type in 33/33/3333.

The pattern attribute only affects CSS (which is currently not implemented or represented in the examples) and is not needed for the input mask to work. (what's really used is the mask attribute 99/99/9999 in this case, which only allows numbers).

In other words, if we added a more complex regex pattern check, it would NOT affect or limit the mask in anyway, it would just trigger the `:invalid` pseudo class as "true" (meaning someone could implement a red border or background on the field as instant feedback). It would still allow for a date like 33/33/3333 but would have a red border.

To truly check for a valid date (like disallowing 13 as a month), I believe that would have to be done in the JS and may be out of scope for this version.

This makes me wonder if we should implement the `:invalid` CSS and recommend its use like the following for all mask examples:

```
input:invalid {
  border: red solid 3px;
}
```

Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern

I'll bring it up to Dan at our sync today, thanks again!

amyleadem: Thanks for the explanation! It all makes sense. It makes me wonder if the input mask is an appropriate solution for Date, or if it might be better to point users to the Date picker or memorable date components for that kind of entry. Would be curious to hear your thoughts!

packages/usa-input-mask/src/index.js
```diff
@@ -1,124 +1,1091 @@
...
-    input.setAttribute("maxlength", placeholder.length);
-    input.setAttribute("data-placeholder", placeholder);
-    input.removeAttribute(`${PLACEHOLDER}`);
+
+const MESSAGE_INVALID_CLASS = `${MASKED_CLASS}__status--invalid`;
+const STATUS_MESSAGE_CLASS = `${MASKED_CLASS}__status`;
+const STATUS_MESSAGE_SR_ONLY_CLASS = `${MASKED_CLASS}__sr-status`;
+const STATUS_MESSAGE = `.${STATUS_MESSAGE_CLASS}`;
+const STATUS_MESSAGE_SR_ONLY = `.${STATUS_MESSAGE_SR_ONLY_CLASS}`;
+const DEFAULT_STATUS_LABEL = `character is required here`;
```
amyleadem, line 19: Should this be a data attribute that the user can set in case they want to use another language?

mitchmoccia: I followed the character count paradigm for this which has hardcoded English messages. I think a data attribute would be good though! https://github.com/uswds/uswds/blob/develop/packages/usa-character-count/src/index.js

packages/usa-input-mask/src/content/usa-input-mask-date-short.json
mitchmoccia: The sense we got from Dan on this is that we shouldn't try to build in more intelligence e.g. to validate a date. A few reasons include:
- keeping it simple (e.g. only validate if it's a letter or number)
- having to potentially account for a variety of other date formats (e.g. if a user wanted to put dd/mm/yyyy or mm/yy etc.)
- would we need to build in intelligence for other masks (e.g. area codes, SSNs etc. and limit it to only valid inputs)

As such, we are actually going to remove the `pattern` attribute all together, as well as, the new pattern guidance as it may add some confusion to implementing a mask. Updates coming soon!

packages/usa-input-mask/src/content/usa-input-mask-date-short.json
mitchmoccia: Changed this to 01/31/1970, also did this in unit test to keep it consistent.

mejiaj (changes requested):
Submitting initial review for functionality. Overall I like the new features and I'm curious as to the reasons for the rewrite. The JS especially has gone through a lot of changes.

As a developer what changes would I need to make (if any) to upgrade to this version? Do we have that available as guidance?

Some standout items from notes:
1. I like that you can't paste invalid characters into SSN.
2. Entering <kbd>Space</kbd> multiple times has weird results (example in comments).
3. Status messages disappear when switching between variants (maybe storybook bug).
4. Default status message doesn't return once input has returned to "blank" state (Met input requirements, but then deleted them to go back to initial state)
5. Microsoft edge on windows shows HTML form message `Please match the requested format` when filled out (probably something with the pattern?)
6. Trying to delete chunks of text via <kbd>ctrl+backspace</kbd> on windows makes the input mask disappear. You can reproduce in OSX too by holding the same key combination.
7. **NVDA** quirks on reading "U.S."
8. **NVDA** reads SSN numbers as hundreds/thousands (when reading hint).
9. **NVDA** alphanumeric doesn't read anything on page load or tab/focus.

I tested general functionality here on OSX, Windows, and NVDA.

mitchmoccia:
Hey @mejiaj, thanks as always for the great review and comments... I wanted to answer your primary question and then take some time to go through the rest. Here are some of the details on why a rewrite was initiated:
- error/invalid character handling
  - the new mask gives instant feedback for invalid characters, similar to how the character counter behaves
  - btw I saw the issues you found with spaces which I will address first
- weirdness with the current mask https://designsystem.digital.gov/components/input-mask/
  - when you type an invalid character, it shows the character for a split second and then dissappears (screen readers call it out as well, this was perhaps the biggest issue)
  - with the current mask, there is some weirdness in a field with format characters (like telephone)... move to the right of a format character and try to use the arrow keys
  - you can't highlight all and delete on some masks (like telephone)
  - you can't use a leading format character
    - so with the current mask, you can't have (___) ___-____ we were forced to go with ___-___-____

There's a bunch more, mostly covering edge cases but there's a lot of nuance addressed (e.g. not being able to arrow to or click to the left of a leading format character). 

Going to work on addressing all your concerns this week. Thanks again James!

mejiaj:
> Hey @mejiaj, thanks as always for the great review and comments... I wanted to answer your primary question and then take some time to go through the rest. Here are some of the details on why a rewrite was initiated:
> 
> * error/invalid character handling
>   
>   * the new mask gives instant feedback for invalid characters, similar to how the character counter behaves
>   * btw I saw the issues you found with spaces which I will address first
> * weirdness with the current mask https://designsystem.digital.gov/components/input-mask/
>   
>   * when you type an invalid character, it shows the character for a split second and then dissappears (screen readers call it out as well, this was perhaps the biggest issue)
>   * with the current mask, there is some weirdness in a field with format characters (like telephone)... move to the right of a format character and try to use the arrow keys
>   * you can't highlight all and delete on some masks (like telephone)
>   * you can't use a leading format character
>     
>     * so with the current mask, you can't have (**_) _**-**__ we were forced to go with _**-**_-**__
> 
> There's a bunch more, mostly covering edge cases but there's a lot of nuance addressed (e.g. not being able to arrow to or click to the left of a leading format character).
> 
> Going to work on addressing all your concerns this week. Thanks again James!

Happy to help and I appreciate your detailed response. I wasn't aware of these quirks! These are nice features and the component is complex, so reach out on Slack if you need some quick eyes on something. 😄

amyleadem:
@mitchmoccia
I am excited about the addition of a status field and the smoother visual experience here! Thanks for ironing out these kinks. I have only been able to test behavior in Chrome and VoiceOver so far, but I wanted to share some initial thoughts. Some of these are just notes and might not actually be fixable, but wanted to share all the same!

- I think we should reconsider using Date as an example here. The component can not actually check for the specific numbers that would be required, and we have solutions for inputting a date in Date Picker and Memorable Date that have more user friendly formats.
- I noticed that the component shows an error status when the user inputs the special characters that are expected in the pattern (and show in the placeholder). For example typing a `/` where it should appear in the date example creates an error status even though it is technically the correct character.
- Typing a bunch of spaces into the mask creates strange behavior (For me it writes a period and then removes a character from the placeholder with each space added.)
- I want to advocate for more descriptive examples in `usa-hint`. I like the idea of something very clear, like: `Provide your 10-digit phone number, starting with the area code. For example: (123) 456-7890`
- I noticed that if I try to paste in a number that doesn’t follow the pattern (For example, pasting `145-62-0495` into the alphanumeric example), it doesn’t do anything and gives no feedback.
- I am able to add some common keyboard characters (&, ^, #, etc) to the input when numeric characters are accepted

packages/usa-input-mask/src/content/usa-input-mask-date-short.json
```diff
@@ -0,0 +1,10 @@
...
+  "id": "us-date",
```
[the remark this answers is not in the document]
mitchmoccia: Changed this to `date-short`... Was trying to keep from using just "date" as the ID. `date-short` is what we're calling it in the code preview section.

packages/usa-input-mask/src/usa-input-mask.twig
```diff
@@ -1,15 +1,21 @@
...
+          inputmode="{{ inputmode }}"
+          mask="{{ mask }}"
+          name="{{ id }}"
+          pattern="{{ pattern }}"
+          placeholder="{{ placeholder }}"
+          type="{{ type }}"
+      >
+    </div>
+    <span id="{{ id }}-hint-info" class="usa-input-mask__message"></span>
+  </div>
```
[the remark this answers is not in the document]
mitchmoccia: Ah got you... Yeah I sort of went off of character count on the new version of input mask, that's why form-group is nested.

packages/usa-input-mask/src/index.js
```diff
@@ -1,124 +1,1091 @@
...
+    throw new Error(`${INPUT} is missing outer ${MASKED}`);
+  }
+
+  const messageEl = maskedEl.querySelector(MESSAGE);
+
+  if (!messageEl) {
+    throw new Error(`${MASKED} is missing inner ${MESSAGE}`);
+  }
+
+  const el = inputEl;
```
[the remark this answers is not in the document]
mitchmoccia: Simplified this by destructuring `value` straight from `inputEl`.

packages/usa-input-mask/src/index.js
```diff
@@ -1,124 +1,1091 @@
...
+ * @param {HTMLElement} inputEl - The input element associated with the status message to be hidden.
+ */
+const hideMessage = (inputEl) => {
+  const { maskedEl } = getMaskedElements(inputEl, true);
+  const statusMessage = maskedEl.querySelector(STATUS_MESSAGE);
+  const srStatusMessage = maskedEl.querySelector(STATUS_MESSAGE_SR_ONLY);
 
-const isLetter = (value) => (value ? value.match(/[A-Z]/i) : false);
+  statusMessage.textContent = "";
+  srUpdateStatus(srStatusMessage, "");
```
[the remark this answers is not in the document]
mitchmoccia: Check the character count error functionality... `usa-input-mask__message` holds that initial message, while `usa-input-mask__status` is what gets populated with the status message.

`MESSAGE` is used in the getMaskedElements function to confirm that the `usa-input-mask__message` element exists, otherwise it throws an error (just like character count).

`const messageEl = maskedEl.querySelector(MESSAGE);`

I'm working on getting the initial message to show again after valid or invalid inputs are entered.

packages/usa-input-mask/src/index.js
```diff
@@ -1,124 +1,1091 @@
...
+const handleKeyUp = (inputEl, event) => {
+  const keyCode = event.which;
+  const el = inputEl;
+
+  if (keyCode === undefined && el.value.length > 0) {
+    const pastedText = getPastedText(inputEl, el.value);
+    el.value = "";
+    pasteTextToInput(inputEl, pastedText, 0);
+
+    if (pastedText === "") {
```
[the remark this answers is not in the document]
mitchmoccia: I disallowed space characters in the same way the current input mask version does. This seems to have resolved this issue.

packages/usa-input-mask/src/index.js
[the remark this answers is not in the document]
mitchmoccia: I addressed your initial comment in this section. The default message is now visible in all cases that an invalid input is not detected.

https://user-images.githubusercontent.com/19338309/215361701-c405bdf9-65e2-416b-8219-6b37ddaa4b3a.mov

packages/usa-input-mask/src/index.js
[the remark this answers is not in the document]
mitchmoccia: There is a weird addition to this story... On Mac, there is a setting to allow a double space to produce a period lol... Check this out https://osxdaily.com/2019/03/27/disable-period-typing-shortcut-mac/
I have disallowed spaces using the following code so the period issue is not happening now, and the mask is not getting fouled up:

```
 if (keyCode === KEYS.space) {
    el.value = el.value.replace(/\s+/g, "");
    return false;
  }
```

mejiaj:
@thisisdano hastily merged `develop`, but reverted back to [5f3a992](https://github.com/uswds/uswds/pull/5104/commits/5f3a992d0c84ce6d50ededb1bacda12f2de59859).

amyleadem:
Closing this PR in favor of https://github.com/uswds/uswds/issues/5054

PR #5114, opened 2023-01-27
USWDS - File Input: Move imageId for larger scope. Resolves #5088

### LINKED ISSUE #5088: USWDS - Bug: Preview Image ID logic issue
### Describe the bug

If a file takes a long time to load in the file input, the loading spinner is not replaced with a preview image
image
This error appears in the console:
`Uncaught TypeError: Cannot read properties of null (reading 'setAttribute')
    at w.onloadend (index.js:292:22)`

1. Go to [USWDS File input example page](https://designsystem.digital.gov/components/file-input/)
2. Click 'choose from folder' link in file input component
3. Choose a file that will take at least 1 second to load into input - my case is a .xls file greater than 1 MB from a network shared drive
4. File will load, but spinner is not replaced with preview image and error is displayed in console.

### Expected Behavior

The spinner should be replaced with correct preview image and no errors are shown in console.

### Related code

In index.js for the file input, both the `reader.onloadstart` and `reader.onloadend` functions call `createUniqueID(makeSafeForID(fileName))`. This function uses `Date.now()` as part of creating the unique ID for the preview image element. 

If enough time passes between `onloadstart` and `onloadend`, the preview image IDs created don't match. `onloadstart` creates the preview with the spinner, but when `onloadend` queries the DOM to remove the spinner and add the correct preview image, it can't find the preview image element because it is searching with a different ID that the element was created with, thus the spinner isn't removed and an error is thrown because `onloadend` couldn't find the preview image element.
Here is an example using the same code `Math.floor(Date.now() / 1000)`
id created on page load - 
image
id created 750 ms later using setTimeout - 
image

### Screenshots

_No response_

### System setup

- USWDS Version 3.1.0
- Device: HP Laptop
- Operating System: Windows 10 Enterprise
- Browser: Edge 108.0.1462.54

### Additional context

_No response_

### Code of Conduct

mahoneycm:
## Summary
This change declares `imageId` outside of the `onloadstart()` and `onloadend()` functions to ensure the same target is selected and replaced by the preview image if the attachment takes too long to attach.

## Related issue
Closes #5088 

## Preview link
Preview link: [File Input →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-file-input-imageID/iframe.html?args=&id=components-form-inputs-file-input--default&viewMode=story)

`createUniqueID()` uses the current time to create the `ImageId` variable. When a large file takes time to upload, it can cause `onloadstart` and `onloadend` to generate different image IDs. 

## Solution
By declaring the variable before the functions,  we can use `createUniqueID()` only once and pass the variable through both functions.

## Testing and review
1. Upload larger files to file input  _// (I had to use a 600mb dmg file to replicate the error)_
2. Confirm that the spinner is replaced with preview image
3. Make multiple attempts and try file input variants to confirm it still works as expected


mahoneycm:
@mejiaj @amyleadem Bumping for review if either of you have time today or next week!

amyleadem:
Hi @mahoneycm, I did some preliminary testing and found that the loader now either freezes for a bit before showing the preview image OR doesn’t appear at all. Are you seeing the same thing?
So far I have tested in Chrome, Edge, and Safari and they all have the same issue for me. Here are some screen grabs:

1. **Frozen loader before preview**
75 MB file:
    
    [https://user-images.githubusercontent.com/93996430/217358593-ef1ee2c0-5a85-42fc-a8f6-dc2217844d56.mov](https://user-images.githubusercontent.com/93996430/217358593-ef1ee2c0-5a85-42fc-a8f6-dc2217844d56.mov)
    
    317 MB file:
    
    [https://user-images.githubusercontent.com/93996430/217361799-30f576d3-4364-4e84-a278-528ecaa0af84.mov](https://user-images.githubusercontent.com/93996430/217361799-30f576d3-4364-4e84-a278-528ecaa0af84.mov)
    
2. **No loader before preview:**
75 MB file:
    
    [https://user-images.githubusercontent.com/93996430/217359585-fa861151-3569-494c-901a-1df709df7808.mov](https://user-images.githubusercontent.com/93996430/217359585-fa861151-3569-494c-901a-1df709df7808.mov)

mejiaj:
@amyleadem @mahoneycm I wasn't able to reproduce with a 651MB ZIP (using slow 3G network throttling). Attached video below of current component on `develop`.

[file-input-develop-2023-02-08 at 10.19.45.webm](https://user-images.githubusercontent.com/3385219/217589169-45365c0f-c27b-49e7-b9fa-081f60073a62.webm)

mahoneycm:
> Hi @mahoneycm, I did some preliminary testing and found that the loader now either freezes for a bit before showing the preview image OR doesn’t appear at all. Are you seeing the same thing? So far I have tested in Chrome, Edge, and Safari and they all have the same issue for me. Here are some screen grabs:
> 
> 1. **Frozen loader before preview**
>    75 MB file:
>    
>      
>        
>        
>    
>        Screen.Recording.2023-02-07.at.1.32.07.PM.mov
>        
>      
>    
>      
>    
>      
>    
>    
>    317 MB file:
>    
>      
>        
>        
>    
>        Screen.Recording.2023-02-07.at.1.44.13.PM.mov
> 2. **No loader before preview:**
>    75 MB file:
>    
>      
>        
>        
>    
>        Screen.Recording.2023-02-07.at.1.35.15.PM.mov

@amyleadem Yes I'm seeing the same thing, but only when the icon swap happens **successfully**. If the attachment takes too long I still see the infinite spinning loader. 

Here's a clip where I'm attempting to attach the same file on the current federalist build and on this PR branch. 


https://user-images.githubusercontent.com/25211650/217589902-6c9b0d99-a64a-461c-9095-ca61e7007b5d.mov


There are times where the network connection is fast enough and the current build does successfully attach the file which results in the spinner pause and file image replacement though. I think if the file upload takes a few second or on the turnover of a new minute it causes the ID's to mismatch and doesn't replace the spinner correctly.

We can take a look at what is causing that pause as well either in this ticket or in a new issue

mahoneycm:
@mejiaj When I first tried to confirm this bug I had trouble recreating it as well. The network throttling didn't seem to affect it for some reason but I kept trying different files and I began seeing the bug. I believe it has to do with the time you start the attachment vs when the attachment ends.

It does seem to be rare though so potentially an edge case.

mejiaj:
@amyleadem with the examples above can you interact at all with the browser? Like right click or select text.

I was able to reproduce the loader icon freezing, but that's because the entire browser was frozen for a second.

amyleadem:
@mejiaj I am not able to interact with the browser when the loader icon is frozen. The browser itself does seem to be frozen during that time.

mahoneycm:
@mejiaj @amyleadem since the freeze seems to be related to the browser freezing, should we go ahead and push this along?

mejiaj (approved):
The ID is unique and doesn't change once the file is done loading.

thisisdano (approved):

thisisdano:
I'll draw up a changelog PR for this change

PR #5123, opened 2023-01-30
USWDS - PR Template: Update breaking changes section

### LINKED ISSUE #5089: USWDS - PR template: Update breaking change section
Consider making the breaking change section of the PR more actionable. This way it's easier to indicate the status of the change and forces the user to select one of the three options.

https://github.com/uswds/uswds/blob/e75f3f347170ebfe472bba9719935a4f5ba38b69/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L35-L44

```diff
## Breaking change 
- _Indicate if this update is a breaking change._ 
+ <!-- For clarity, indicate if this is a breaking change. Include a statement here even if there is no breaking change. -->

- <!-- 
- Use clear language to: 
- 1. Indicate if this update is a breaking change. 
- 2. Explain what actions are required for the user to remediate the break. 
-   For clarity, include a statement here even if there is no breaking change: 
-   "This is not a breaking change." 
-   ":warning: This is potentially a breaking change." 
-   ":warning: This is a breaking change." 

+ This is not a breaking change.
+ :warning: This is potentially a breaking change.
+ :warning: This is a breaking change.
```

amyleadem:
# Summary
Update the breaking changes section in the PR template to be clearer and easier to use.

## Breaking change

## Related issue
Closes #5089 

## Preview link
Preview link: N/A

## Testing and review
Confirm that the instructions for identifying a breaking change are clear and easy to use.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,7 +24,8 @@ USWDS - [Package]: [Brief statement describing what this pull request solves]
 eg: "USWDS - Button: Increase font size"
  -->
 
-## Summary
+# Summary
+
 _Provide a one or two sentence summary of the update that can be used in the changelog._
 <!--
 A successful summary is written in the past tense and includes:
@@ -33,18 +34,15 @@ See [USWDS release notes](https://github.com/uswds/uswds/releases) for examples.
 -->
 
 ## Breaking change
-_Indicate if this update is a breaking change._
-<!--
-Use clear language to:
-1. Indicate if this update is a breaking change.
-2. Explain what actions are required for the user to remediate the break.
-   For clarity, include a statement here even if there is no breaking change:
-   "This is not a breaking change."
-   ":warning: This is potentially a breaking change."
-   ":warning: This is a breaking change."
--->
+
+_Indicate if this update is a breaking change with one of the following statements:_
+This is not a breaking change.
+:warning: This is potentially a breaking change.
+:warning: This is a breaking change.
+<!-- If applicable, explain what actions are required for the user to remediate the break. -->
 
 ## Related issue
+
 Closes #_[issue_no]_
 <!--
 Every pull request should resolve an open issue.
@@ -53,10 +51,12 @@ https://github.com/uswds/uswds/issues/new/choose.
 -->
 
 ## Preview link
+
 Preview link:
 <!-- If available, provide a link to a demo of the solution in action. -->
 
 ## Problem statement
+
 _Summarize the problem this PR solves in a clear and concise statement._
 <!--
 A successful problem statement conveys:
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -33,18 +34,15 @@ See [USWDS release notes](https://github.com/uswds/uswds/releases) for examples.
...
-Use clear language to:
-1. Indicate if this update is a breaking change.
-2. Explain what actions are required for the user to remediate the break.
-   For clarity, include a statement here even if there is no breaking change:
-   "This is not a breaking change."
-   ":warning: This is potentially a breaking change."
-   ":warning: This is a breaking change."
--->
+
+_Indicate if this update is a breaking change with one of the following statements:_
```
amyleadem, line 38: @mejiaj I took a stab at including some visible short, clear instructions so that it can more closely match the format in the other sections. Let me know what you think!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -33,18 +34,15 @@ See [USWDS release notes](https://github.com/uswds/uswds/releases) for examples.
...
+This is not a breaking change.
+:warning: This is potentially a breaking change.
+:warning: This is a breaking change.
```
mahoneycm, line 41: Just an idea, it might be nice to have these as a bulleted list since we're listing out options

> _Indicate if this update is a breaking change with one of the following statements:_
> 
> - This is not a breaking change.
> - :warning: This is potentially a breaking change.
> - :warning: This is a breaking change.

I think the indention also helps with readability

.github/PULL_REQUEST_TEMPLATE.md
mejiaj: Glad you like the suggestions, this looks good. We might want to consider making the `one` bold to make it standout a little more.

```suggestion
_Indicate if this update is a breaking change with **one** of the following statements:_
```

.github/PULL_REQUEST_TEMPLATE.md
mejiaj: A list does help initially, but when updating we'll just end up removing this formatting. I'm inclined to try this out _without_ the list formatting.

That way we can just keep the statement that applies.

mejiaj (commented):
@amyleadem minor comment on adding emphasis to `one` of the options. LGTM otherwise!

.github/PULL_REQUEST_TEMPLATE.md
amyleadem: I agree with @mejiaj on this one. In other circumstances I would advocate for list formatting, but since this content exists only for users to quickly delete two of the three options, I am inclined to say that less formatting (and quicker deleting) is better. We can always adjust if we see that users are getting confused.

amyleadem:
@mejiaj I like the added emphasis. Good suggestion!

Updated the screenshot:
image

mejiaj (approved):
Nice work, looks great!

mahoneycm (approved):
Thanks Amy!

thisisdano (approved):

Issue #5050, opened 2022-11-22
USWDS - Enhancement: Allow customization of in-page nav target


danny-englander:
### Describe the bug

The new In-page navigation component picks up what I would classify as unwanted headings from elsewhere on the page. If you review the screen shot below, you will see that there are headings that do not add any value to the in page nav and are also confusing.

<img width="1529" alt="2022-11-22_09-12-33" src="https://user-images.githubusercontent.com/609623/203337126-71de9eb8-b1d2-43be-add8-213fc793a3ac.png">

### Steps to reproduce the bug

- Create a page that has a breadcrumb above the content area. For example

```html
<nav role="navigation" aria-labelledby="system-breadcrumb" class="usa-breadcrumb">
    <h2 id="system-breadcrumb" class="visually-hidden"><a id="section_0" class="usa-anchor"></a>Breadcrumb</h2>
    <ol class="usa-breadcrumb__list">
          <li class="usa-breadcrumb__list-item">
                  <a class="usa-breadcrumb__link" href="/"><span>Home</span></a>
              </li>
          <li class="usa-breadcrumb__list-item">
                  <a class="usa-breadcrumb__link" href="/uswds-components"><span>USWDS 3 Component Library</span></a>
              </li>
        </ol>
  </nav>
```

- The below on the page, create a content area and in-page nav. 
- Note that that the hidden h2 is picked up by in-page nav and it is also visually hidden as well. 

### Expected Behavior

 I would recommend the following improvements / fixes

- Create a data attribute to specifically name the content container that the in-page nav picks up headings for. For example,  `data-content-container="my-content-region"` -- only use headings from this region and no others elsewhere on the page
- ignore visually hidden heading tags

### System setup

- Using UWDS 3.3.0

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

mejiaj:
@mitchmoccia would you be able to confirm whether this is a bug or not?

mitchmoccia:
@danny-englander @mejiaj Hi Danny, I like your suggestion. Currently the in-page nav uses the `main` element as the region that headings will be queried and used for the nav. 

Could you please confirm:
- do you have a `main` element?
- is the nav code you posted inside that main element?

[...]

PR #5124, opened 2023-02-02
In-page nav main content section requirement changed from the main element to a class

mitchmoccia:
This update makes in-page nav way more flexible by changing the main content portion requirement from the `main` element to any element with the class `.usa-in-page-nav__main`.

https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/mm-in-page-nav-parent-update/iframe.html?args=&id=components-in-page-navigation--default&viewMode=story

This was suggested in the following issue:
~Closes https://github.com/uswds/uswds/issues/5050~ striking this because it no longer closes this issue (Amy)

Also, the HTML structural requirement has been updated to bring it more inline with other components. Currently, the parent wrapper is `.usa-in-page-nav__container`. The new structure changes that to `.usa-in-page-nav` and overall looks like this now:

image

Notice the main content section is a `div` with the class `.usa-in-page-nav__main` instead of forcing the `main` element.

Finally, unit test assertions have been vastly improved in this PR. 

Looking forward to questions and comments, thanks!!


packages/usa-in-page-navigation/src/index.js
```diff
@@ -14,13 +14,14 @@ const IN_PAGE_NAV_ROOT_MARGIN = "0px 0px 0px 0px";
...
 const IN_PAGE_NAV_ANCHOR_CLASS = `${PREFIX}-anchor`;
+const IN_PAGE_NAV_CONTAINER_CLASS = `${IN_PAGE_NAV_CLASS}__container`;
+const IN_PAGE_NAV_MAIN_CLASS = `${IN_PAGE_NAV_CLASS}__main`;
 const IN_PAGE_NAV_NAV_CLASS = `${IN_PAGE_NAV_CLASS}__nav`;
 const IN_PAGE_NAV_LIST_CLASS = `${IN_PAGE_NAV_CLASS}__list`;
 const IN_PAGE_NAV_ITEM_CLASS = `${IN_PAGE_NAV_CLASS}__item`;
 const IN_PAGE_NAV_LINK_CLASS = `${IN_PAGE_NAV_CLASS}__link`;
 const IN_PAGE_NAV_TITLE_CLASS = `${IN_PAGE_NAV_CLASS}__heading`;
 const SUB_ITEM_CLASS = `${IN_PAGE_NAV_ITEM_CLASS}--sub-item`;
-const MAIN_ELEMENT = "main";
```
aduth, line 23: I'm not a maintainer, just a fellow contributor, but: Since this would be a breaking change as proposed, would another option be to support an optional attribute on the In-Page Navigation element to override the selector for the main element, so you could do something like ... ?

```html
<aside
  class="usa-in-page-nav__container"
  data-main-selector=".usa-in-page-nav__main"
></aside>
```

Or at least, could backward-compatibility be maintained by continuing to support the `main` element, if present?

amyleadem (commented):
Note: The effort to create a custom selector for the main content region is now being worked in https://github.com/uswds/uswds/pull/5387

Leaving this PR open because we need to discuss if we want to include the other changes (including the HTML and test updates) as shown in this PR. 

cc: @mejiaj @mahoneycm

amyleadem:
Closing this PR in favor of https://github.com/uswds/uswds/pull/5387

PR #5128, opened 2023-02-02
USWDS - Remove disabled classes. Resolves #5116

### LINKED ISSUE #5116: USWDS - Disabled styles: Phase out disabled classes (Interim)
Wherever possible, we should remove any classes that were created specifically for disabled styles. Instead, styles should be applied to both the `disabled` and `aria-disabled` attributes.

mahoneycm:
## Summary
This PR takes a look at existing disabled classes and attempts to replace them by just using the `disabled` and `aria-disabled` attributes in their place.

In it's current state, USWDS contains two `--disabled` classes and one placeholder.

### `usa-button--disabled` class

There was an existing `usa-button--disabled` class that acquired all of the same style definitions as button classes that had the `disabled` or `aria-disabled` attributes. The classes themselves were not used on our storybook or USWDS Site. 

There is a note on site that states to **not** use the class unless for testing or debugging. It might be best to simply remove the class as well as the note on site.

> Use the browser-native disabled attribute for any disabled button. Don’t use usa-button--disabled, which is intended only for debugging and prototyping. [^1]

[^1]: https://designsystem.digital.gov/components/button/#:~:text=Use%20the%20browser%2Dnative%20disabled%20attribute%20for%20any%20disabled%20button.%20Don%E2%80%99t%20use%20usa%2Dbutton%2D%2Ddisabled%2C%20which%20is%20intended%20only%20for%20debugging%20and%20prototyping.

### `usa-file-input--disabled` class

File input uses a class to apply disabled styling to the child elements of the target `<div>`. Generally, this is acceptable behavior and helps to assign styles since the `<div>` element does not receive the `disabled` attribute.

After discussion, we have decided to allow disabled classes that style elements that:
1. Do not receive the disabled attribute
2. Are dynamically placed

This will allow us to correctly style elements as disabled without putting the burden on the user.

### `usa-date-picker--disabled` placeholder
Date picker uses a disabled placeholder but applies these settings to elements with the `disabled` or `aria-disabled` attributes, so we can safely ignore.

## Breaking change
This is **_potentially_** a breaking change

While we advise against using the `usa-button--disabled` class, this would break markdown for devs who have implemented it

## Related issue
Closes #5116

## Preview links
[Button →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-remove-disabled-classes/?path=/story/components-button--default)
[File Input (disabled) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-remove-disabled-classes/iframe.html?args=&id=components-form-inputs-file-input--disabled&viewMode=story)

## Problem statement
Rather than relying on specific state classes to target disabled components, we should instead target these elements using the `disabled` or `aria-disabled` attributes.

## Solution
1. Remove unused and redundant disabled button classes
2. Ensure disabled classes are dynamically set using the related form element as to not put the need on the user.

## Testing and review
1. Review the current state of disabled buttons
2. Confirm no visual discrepancies
3. Tests File Input to make sure dynamic creation works as expected


packages/usa-file-input/src/index.js
```diff
@@ -12,7 +12,6 @@ const BOX_CLASS = `${PREFIX}-file-input__box`;
 const INSTRUCTIONS_CLASS = `${PREFIX}-file-input__instructions`;
 const PREVIEW_CLASS = `${PREFIX}-file-input__preview`;
 const PREVIEW_HEADING_CLASS = `${PREFIX}-file-input__preview-heading`;
-const DISABLED_CLASS = `${PREFIX}-file-input--disabled`;
```
amyleadem, line 15: Right now, adding `aria-disabled="true"` to the input element does not cause the component to look disabled. Since this class is no longer available, we should make sure that `aria-disabled` triggers the same styling as `disabled`.

amyleadem (changes requested):
Looking good so far. I added some comments for a couple items I saw. 

I also wanted to flag that this work might conflict with the input prefix/suffix disabled fix: https://github.com/uswds/uswds/pull/5004/files. We should remove the `--disabled` classes from that fix as well. 

Let me know if you have questions.

Tests performed:
- [ ] Confirm that all disabled-specific classes have been removed (wherever possible)
- [ ] Confirm that both `aria-disabled` and `disabled` trigger disabled styles
    - [x] Buttons
    - [ ] File input

packages/usa-button/src/styles/_usa-button.scss
```diff
@@ -197,10 +197,6 @@ $button-stroke: inset 0 0 0 units($theme-button-stroke-width);
   padding: units(2) units(3);
 }
 
-.usa-button--disabled {
-  @include button-disabled;
-}
-
 .usa-button--outline-disabled,
 .usa-button--outline-inverse-disabled,
```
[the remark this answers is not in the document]
mahoneycm: Oh thanks for catching this! Removed in https://github.com/uswds/uswds/pull/5128/commits/da03960fe11fec71a59ef572a23a5ac6b1d565bc

packages/usa-file-input/src/index.js
mahoneycm: I've brought over some of the javascript and storybook settings from #5063 so that the component will dynamically add the `aria-disabled` attribute to the parent class in the same way it currently does with `disabled`

Now when the input has the attribute, the `usa-file-input` div will receive it as well and style it appropriately.

If you manually add `aria-disabled="true"` to the input after init it will not inherit the same styles, but this is to be expected because of the way the component is built and styled through the new divs and is consistent with how the component was styled previously

mahoneycm:
@amyleadem @mejiaj Restored `file-input`'s dynamically set disabled glass based on our discussion yesterday and updated the PR description to reflect our decision to allow dynamically set disabled classes on divs that do not receive the `disabled` attributes moving forward.

amyleadem (approved):
Looks good to me! Only thing I would consider is flagging this as a potentially breaking change in the PR description. We do advise against using the disabled button classes in our guidance, but it might be worth flagging just in case. 

- [x] Confirm that all instances of disabled classes have been accounted for in this PR
- [x] Confirm that all remaining instances of disabled classes are dynamically added (no additional burden on user)
- [x] Confirm that none of the button variants rely on a disabled class
- [x] Confirm that both `aria-disabled` and `disabled` show disabled button styles
- [x] `npm run test` completes without error

mahoneycm:
@mejiaj Bumping for review!

mejiaj (approved):

mejiaj:
@mahoneycm I've reviewed and approved. Please request re-review (if needed) via the  the arrow on the reviewers panel.

mahoneycm:
Work implemented in #5063

mejiaj:
Removing this issue from milestone to avoid confusion. Work was done in #5063.

PR #5146, opened 2023-02-13
USWDS: File Input: Remove aria disabled file input. Resolves #5144

### LINKED ISSUE #5144: USWDS - File input: Remove aria-disabled from from parent wrapper
Currently, the file input component adds `aria-disabled` to its parent wrapper, `usa-file-input`. In JAWS, this prevents the label from being read, which violates best practice. We should remove the `aria-disabled` attribute from `usa-file-input` element and confirm that the removal does not cause harm.

mahoneycm:
## Summary
Remove `aria-disabled` attribute from `usa-file-input` element to allow screen readers to read out label tag.

## Breaking change
This is **not** a breaking change

## Related issue
Closes #5144 

## Preview link
[File Input (Disabled) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-remove-aria-disabled-file-input/iframe.html?args=&id=components-form-inputs-file-input--disabled&viewMode=story)

## Problem statement
Currently, when File Input is initialized, the `aria-disabled` attribute is added to the parent element. This causes the label to no longer be read by screen readers, which goes against best practices for screen readers.

## Solution
Remove JS which sets this attribute. Now the File Input is still read as disabled, but the label is also ready by screenreaders

## Testing and review
1. Visit the File Input Disabled story.
2. Activate screen reader
3. Check that the label is read
4. Check that File Input component is read as disabled.


amyleadem (commented):
This looks good to me!

I confirmed the following items:
- [x] The `aria-disabled` attribute is no longer added to the `usa-file-input` parent element
- [x] There was no negative impact on the VoiceOver readout. It still reads the label and stated that the input was dimmed. (This is unchanged from the experience in VO for `develop`)

@amycole501 I am going to add you to the review so you can confirm that the JAWS readout follows best practices. I will hold off on approval until I hear your findings.

mejiaj (approved):

mejiaj:
@amyleadem can you follow up with @amycole501 on this?

amycole501:
Tested in JAWS; label was read accurately. Button action reads "No file selected, button unavailable."
Tested in NVDA; label was read accurately. Button action reads "Clickable button unavailable, no file selected."

amyleadem (approved):
Approved!

amycole501 (approved):
Sounded fine in both NVDA and JAWS

mahoneycm:
Work implemented in #5063

PR #5166, opened 2023-03-01
USWDS - Links: External link labels

### LINKED ISSUE #5164: External links are not indicated on screen readers
## Description

External links have no screen reader identifier to match the “external link” icon. 

This issue was created in response to this [related Slack thread](https://gsa-tts.slack.com/archives/C3F14AHSQ/p1649690481675279).

1. Open any page on https://designsystem.digital.gov/
2. Activate screen reader
3. Navigate to any external link (indicated by the external link icon)
image
5. Check if the readout gives indication that the link is pointing to an external url

### Voiceover:
image

### NVDA:

NVDA also failed to mention it was an external link

Screen Reader: NVDA
System: Windows 10 Pro
Browser: Chrome

mejiaj:
# Summary

**External links labeled for screen readers.** Adds a label for screen readers that clearly marks external links **and** whether they open in a new tab or current.

## Breaking change

## Related issue

Closes #5164.

## Preview link

~External link test →~ _Removed in [917b3e9](https://github.com/uswds/uswds/pull/5166/commits/917b3e914567dbedd19f678e38ff3b926cdd8087)_
 
[Links [story mode]](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-external-link-sr/iframe.html?id=components-link--link&viewMode=story)

## Problem statement

1. Users of assistive technologies should be able to differentiate internal vs external link.
2. Users of assistive technologies should be able to differentiate new tab vs current. 
3. Currently there's only a visual indicator for external links.
4. Fixing this will bring parity to the external link component for all users.

## Solution

This solution:

- Adds a label via the `::before` pseudo element. _Pseudo element was chosen to allow users to easily update and add this feature into their sites._
- Two new settings to allow developers to customize labels

⚠️ **Possible limitation:** Section508.gov warns against placing meaningful text in pseudo elements. We should be careful when testing this feature and implementing more hidden labels in the future.

>Development Considerations
>
>Avoid using inserted CSS content (pseudo elements ::before and ::after) to add meaningful content.

Source: [https://www.section508.gov/content/guide-accessible-web-design-development](https://www.section508.gov/content/guide-accessible-web-design-development/#:~:text=Avoid%20using%20inserted%20CSS%20content%20(pseudo%20elements%20%3A%3Abefore%20and%20%3A%3Aafter)%20to%20add%20meaningful%20content.)

### Alternatives

JavaScript could be used to inject hidden span into each external link.

## Major changes

Two new settings for Link component:

| Variable | Value | Type | Description |
|:--------|:--------|--------|:--------|
| ~`$theme-external-link-sr-label`~ | ~`"External link"`~ | ~string~ |  Removed in 94d4274. ~Base label indicating external link.~ |
| `$theme-external-link-sr-label-tab-new` | `"External, opens in a new tab."` | string | Screen reader-only text for links that open in new tab. |
| `$theme-external-link-sr-label-tab-same` | `"External, opens in current tab." ` | string |  Screen reader-only text for links that open in current tab. |

### End result

The following text gets announced _before_ an external link:

```
// External link in current tab
"External, opens in current tab."
```

```
// External link in new tab
"External, opens in a new tab."
```

## Testing and review

1. Screen readers (JAWS/VoiceOver/NVDA) **should** read hidden label
2. Screen reader indicates whether it opens in new tab (`target="_blank")` or current (fallback).

<details><summary>Previous test (click to expand)</summary>

1. ~Screen readers (JAWS/VoiceOver/NVDA) **should** read the hidden label identifying external links.~
2. ~Need to find the most clear and obvious label:~
    1. ~❌ `[External link]`~
    2. ~❌ `External link:` [⚠️ Punctuation might negatively impact RTL languages]~
    3. ~✅ `External link`. This was found to provide the best experience across screen readers (JAWS, VoiceOver, and Narrator).~

~I was able to get VoiceOver to read, Option 1 (`[External]`) seemed like the best experience to me. Need to hear how these sound on other screen readers.~

</details> 

## Related PRs

Updated site guidance uswds/uswds-site#2056
  

---


mejiaj:
@amycole501 this is ready for your review. Here's the [preview link](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-external-link-sr/?path=/story/components-link--test).

amycole501:
I can confirm that using JAWS the words "external link" were read on all three link examples. From a usability standpoint, the third link was best.  The first link example announces the brackets first, the second one mentions a colon but the third one only says "external link" along with the link wording and that's easiest to understand.

In Narrator all three links announced the same as the third link. So if given the choice, I'd go with how you coded the third link.

amyleadem (approved):
I like this. Feels like a straightforward solution and it works well in VoiceOver.

One thing that might be worth considering is adding a comma to the end of the default content (`$theme-external-link-sr-label: "External link,"`). It seems to give a tiny pause in between the pseudo text and the link text, which I found helpful. 

Curious to hear @amycole501's thoughts.

Tested on VoiceOver in Chrome, Safari, Firefox, and Edge. 

Also, the preview link is broken in the description. I am not sure if you intended to create a test story, but I went ahead and tested just on the main link component story.

amycole501 (commented):
I do hear "external link". The only note I'd add is if the external link opens the visitor to a new window. That does also need to be announced. If it opens the external link in the same window nothing additional needs to be announced.

amycole501 (changes requested):
Two things to consider:
1. If someone is implementing the external link feature and the link opens in a new window, the person who adds the link needs to add some sort of way for a screen reader to announce it opens in a new window. Some guidance is here https://www.w3.org/TR/WCAG20-TECHS/G201.html
2.  Consider also if someone is implementing this link feature using a CMS-driven environment. Not all WYSIWYG editors allow the user to add link help text or descriptive hints that say "link opens in a new window". Not sure of the solution but wanted to raise it as a potential scenario.

packages/usa-link/src/usa-link.twig
```diff
@@ -2,7 +2,13 @@
 
 <p>This is <a class="usa-link usa-color-text-visited" href="javascript:void(0);">a visited link</a>.</p>
 
-<p>This is a link that goes to an <a class="usa-link usa-link--external" href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.</p>
+<p>
+  This is a link that opens in <em>current</em> tab and goes to an <a class="usa-link usa-link--external" href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.
+</p>
+
+<p>
+  This is a link that opens in a <em>new</em> tab and goes to an <a class="usa-link usa-link--external" rel="noreferrer noopener" target="_blank"  href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.
```
mejiaj, line 10: 💡 I've explicitly added `rel="noreferrer noopener"` as a best practice; even though most modern browsers implicitly add it.

image
[MDN Browser compatibility →](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility)

amyleadem: Since this is a best practice, would it be beneficial to include guidance on `noreferrer noopener` in the documentation?

amycole501 (commented):
Listened in JAWS and used ANDI to verify the external link text was announced. Sounded good, clear, and was easy to understand.

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -123,6 +123,11 @@ $theme-in-page-nav-margin-left: 4 !default;
 $theme-in-page-nav-margin-top: 2.5 !default;
 $theme-in-page-nav-top: 4 !default;
 
+// Link
+$theme-external-link-sr-label: "External link" !default;
```
amyleadem, line 127: No action is required on this, but wondering if we can reduce some audio redundancy by removing the word “link” here so that it is just “External”. In Voiceover, that would mean the readout is “link, _external_, opens in a new tab [link text]”. I am not sure if other screen readers announce “link” in the same order though, so this might just reflect a VoiceOver bias.

packages/usa-link/src/usa-link.twig
```diff
@@ -2,7 +2,13 @@
 
 <p>This is <a class="usa-link usa-color-text-visited" href="javascript:void(0);">a visited link</a>.</p>
 
-<p>This is a link that goes to an <a class="usa-link usa-link--external" href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.</p>
+<p>
+  This is a link that opens in <em>current</em> tab and goes to an <a class="usa-link usa-link--external" href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.
```
amyleadem, line 6: No action needed, just noting that opening `source.gif` is still a surprise every time

amyleadem (approved):
I like this addition. The technical solution feels clean and the readout is helpful. I like that it doesn’t require any extra user action when setting up a link and that the readout copy is customizable.

Confirming that the VoiceOver readout is “link, external link, opens in current tab [link text]” or “link, external link, opens in new tab [link text]”, depending on the link target. 

Added a few comments but no action is needed.

- [x] Confirm that customizing the new theme settings updates the text as expected
- [x] Confirm that VoiceOver identifies external links
- [x] Confirm that VoiceOver identifies if the link opens in a new window

packages/uswds-core/src/styles/settings/_settings-components.scss
mejiaj: @amycole501 did you notice this issue on NVDA/JAWS? If so, I can shorten it to just "External".

packages/usa-link/src/usa-link.twig
mejiaj: I've kept `noreferrer` since only one is needed in [e348c59], but it might not be needed at all since modern browsers add it implicitly.

packages/usa-link/src/usa-link.twig
amycole501: Now that @amyleadem  mentions it yes, it does say "link" twice (which is tricky to discern since some of the links say "link" in them. 

Additionally, I hear JAWS trying to say the word "tab" which runs into the phrase "external website" both in ANDI and in JAWS. I'm attaching how it looks in ANDI so you can get an idea of how it sounds. Like a run-on sentence.

external link text

mejiaj: @amycole501 @amyleadem I've added the following changes:

- Simplified the CSS
- Removed the base variable `$theme-external-link-sr-label`. 
- Added punctuation to the label ex: `"External, opens in a new tab."` to prevent screen readers from jumbling the words together.

amycole501 (commented):
This sounds much better; the words aren't jumbled together and it's clear when listening in JAWS that the links open either in the same/current tab or a new tab.

packages/usa-link/src/usa-link.twig
```diff
@@ -2,9 +2,15 @@
 
 <p>This is <a class="usa-link usa-color-text-visited" href="javascript:void(0);">a visited link</a>.</p>
 
-<p>This is a link that goes to an <a class="usa-link usa-link--external" href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.</p>
+<p>
+  This is a link that opens in <em>current</em> tab and goes to an <a class="usa-link usa-link--external" rel="noreferrer" href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.
```
amyleadem, line 6:
```suggestion
  This is a link that opens in the <em>current</em> tab and goes to an <a class="usa-link usa-link--external" rel="noreferrer" href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.
```

amyleadem (approved):
Good improvements! 

FYI - I made a small commit ([a00a4ed](https://github.com/uswds/uswds/pull/5166/commits/a00a4edeecc8a31dceaaf81b194f4955cbb36433)) to add a missing "the" to a sentence in the twig file. I also updated the PR description to reflect the new  default values and descriptions for `$theme-external-link-sr-label-tab-new` and `$theme-external-link-sr-label-tab-same`.

amycole501 (commented):
looks good

amycole501 (approved):

thisisdano:
I'm going to draw up a changelog PR for this one

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -123,6 +123,10 @@ $theme-in-page-nav-margin-left: 4 !default;
 $theme-in-page-nav-margin-top: 2.5 !default;
 $theme-in-page-nav-top: 4 !default;
 
+// Link
+$theme-external-link-sr-label-tab-new: "External, opens in a new tab." !default;
+$theme-external-link-sr-label-tab-same: "External, opens in current tab." !default;
```
thisisdano, line 128:
```suggestion
$theme-external-link-sr-label-tab-same: "External." !default;
```

Since links open in the current tab by default, is it overkill to announce the default behavior?

mejiaj: Good point, added your suggestion ty!

aduth:
I'm curious about the choice to announce the label _before_ the link text. Most guidance I've seen (including [WCAG G201](https://www.w3.org/TR/WCAG20-TECHS/G201.html)) announces the label after the link text, which makes sense to me because it's a supplementary detail, where the primary detail is the link text itself. And I'd worry it might be slightly disorienting to announce that something will open in a new tab before even knowing what it is that'll be opened in a new tab.

In terms of backwards-compatibility, I found my way here after realizing that our external links were announcing "opens in new tab" twice to screen readers, since we had already implemented this type of label before USWDS, so now we have a redundant label. This was somewhat unexpected. Related to my next point, we may just end up disabling the setting by setting it to an empty string, though this seems to work only by accident rather than being fully supported.

Lastly, this seems to assume a site is English-only. Our site supports multiple languages, and it's not obvious how we'd be expected to use these labels. In our case, we may just "disable" the setting by configuring it to an empty string and continuing to use [our custom implementation](https://github.com/18F/identity-idp/blob/e8da68ffa3403b27da5d86ff0551287144a8b7b9/app/helpers/link_helper.rb#L12). The only alternative I could imagine is creating language-specific stylesheets, each with the Sass variable assigned to the label to be used in that language.

mejiaj:
@aduth good points. I've created #5942 to capture your suggestions.

It makes sense to place the label after and allow devs to opt out.

Another potential alternative is to use the [attr() - MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/attr) to allow users to pass their own strings via CSS. The only downside is it'd have to be manually added per link. We hope to address this type of feature the web component implementation.

Initially, this was added to allow basic support of announcing links with further enhancements based on feedback. @jaclinec have we tested or are planning to test links?

jaclinec:
@mejiaj we tested links in 2021 (see [research findings](https://github.com/uswds/uswds/wiki/2021-07-29-External-Link-Indicator-Research-Findings)), but the research did not include participants who use screen readers so it may not be useful to this conversation. We don't have immediate plans to test links for accessibility but it is in the backlog.

PR #5187, opened 2023-03-15
Alert: Remove invalid and redundant alert height style

aduth:
# Summary

Removes an unnecessary assignment of `height` styling in the `add-alert-icon` mixin.

- This produces invalid styles because `$theme-alert-icon-size` is a unitless number (e.g. `2`) and should be wrapped with the `unit` function
- This is redundant anyways because the `height` is already assigned by the inclusion of `add-color-icon` mixin above using the `height` value from `$this-icon-object`

## Testing and review

Observe the issue by inspecting the `::before` pseudo-element of `.usa-alert__body` in the [Alert component preview page](https://designsystem.digital.gov/components/alert/).

Screen Shot 2023-03-15 at 9 33 24 AM

The changes proposed in this pull request will cause the second (invalid) highlighted `height` style to be removed.

### DIFF
--- a/packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
+++ b/packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
@@ -116,7 +116,6 @@
     @include add-color-icon($this-icon-object, $bgcolor);
     content: "";
     display: block;
-    height: $theme-alert-icon-size;
     // padding - optical spacing value
     left: units($theme-site-margins-mobile-width) - units($theme-alert-bar-width);
     position: absolute;


mahoneycm (approved):
Confirming this effectively removes the redundant height definition from the alert icon!

The icon height is still defined through `$theme-alert-icon-size` via `$this-icon-object` directly above in the CSS. Removing the height definition from `$theme-icon-object` also does not use the height defined within the `&::before` selector so we can safely remove.

mejiaj (approved):

aduth:
This might not actually be so harmless in all cases, since some CSS post-processors may try to correct the style by adding a missing unit.

For example, [Lightning CSS](https://lightningcss.dev/) will change the `4` to `4px`, which will have an impact on the rendered style of the alert.

```bash
$ node -p "require('lightningcss').transform({ code: Buffer.from('body { height: 4; }') }).code.toString()"
body {
  height: 4px;
}
```

thisisdano (approved):

mahoneycm:
Created change log PR for site: https://github.com/uswds/uswds-site/pull/2094

PR #5197, opened 2023-03-23
USWDS - Pagination: Remove presentation role

### LINKED ISSUE #5022: USWDS - Bug: Axe a11y error in pagination
### Describe the bug

We use axe-core for accessibility testing on components. Versions above `5.4.0` introduce new rules (some based on WCAG 2.2). One new rule in particular is now reporting an error on the pagination component.

Originally reported in this [PR comment [USWDS - Deps: Update dependencies [minor and patch]]](https://github.com/uswds/uswds/pull/5019#issuecomment-1307796280)

1. Update `axe-core` dependency to `>=5.4.0`
2. Run `npm run test:a11y`
3. Confirm a11y error

### Expected Behavior

There should be no errors.

### Related code

https://github.com/uswds/uswds/blob/69cf6929585b5c95ed9700a4502830f84e497e95/packages/usa-pagination/src/usa-pagination.twig#L5

### Screenshots

_No response_

### System setup

USWDS 3.2.0

### Additional context

We need to either remove presentation role or add `listitem` role to sibling links.

### Code of Conduct

mahoneycm:
# Summary

This PR removes the role attribute from our pagination component. It also updates `axe-core` and `axe-storybook-testing` dev dependencies.

## Breaking change

## Related issue

Closes #5022 

## Preview link
[Pagination →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-pagination-remove-role/iframe.html?args=&id=components-pagination--default&viewMode=story)

Preview link:

## Problem statement

WCAG 2.1 & 2.0 guidelines require `ul` & `ol` to contain only `li` **content** elements. Elements containing `role=presentation` are not allowed under this guideline.

## Solution
Remove role from pagination component.

## Testing and review

1. Use [pagination component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-pagination-remove-role/iframe.html?args=&id=components-pagination--default&viewMode=story) with screen reader
2. There should be no change in readouts from its [develop counterpart](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?id=components-pagination--default&viewMode=story)
3. Run `npm test:a11y` and ensure no failing tests
4. Use axe-core browser plugin on component page to ensure no failing tests

## Dependency updates

| axe-core |     [4.3.4]      |   [^4.6.3]   |
| [@chanzuckerberg/axe-storybook-testing]     |        [5.0.1]        |   [^6.3.0]   |


package.json
```diff
@@ -111,7 +111,7 @@
     "@18f/identity-stylelint-config": "1.0.0",
     "@babel/core": "7.16.7",
     "@babel/preset-env": "7.16.7",
-    "@chanzuckerberg/axe-storybook-testing": "5.0.1",
+    "@chanzuckerberg/axe-storybook-testing": "^6.3.0",
```
amyleadem, line 114: Should these be pinned like the others?

```suggestion
    "@chanzuckerberg/axe-storybook-testing": "6.3.0",
```

package.json
```diff
@@ -122,7 +122,7 @@
     "@types/node": "16.11.19",
     "ansi-colors": "4.1.1",
     "autoprefixer": "10.4.1",
-    "axe-core": "4.3.4",
+    "axe-core": "^4.6.3",
```
amyleadem, line 125:
```suggestion
    "axe-core": "4.6.3",
```

amyleadem (commented):
This looks good! I had one question about pinning the dependencies. Let me know if you had questions. 

Tests completed:
- [x] Confirm the error in `develop`
- [x] Confirm dependency updates are most current available
- [x] Run `npm install`, `npm start`, `npm run test:a11y` without error
- [x] Confirm no errors in axe devTools

amyleadem (approved):
This looks good to me! 
Small change to the accessibility tree, but I didn't notice a change in the readout for VoiceOver.

amycole501:
Listening in NVDA I heard on this link:
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-pagination-remove-role/iframe.html?args=&id=components-pagination--default&viewMode=story

"list with 9 items"

while on this link 
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?id=components-pagination--default&viewMode=story

I heard
"List with 7 items"

I think NVDA ignores the ellipses as links which is fine as longs they aren't clickable for anyone.

They sounded the same to me in JAWS, however. They both said "7 items"

From a user perspective it's a little confusing to be on the last page in this case, 24, and still see the "next" option when there's no next page. In JAWS and NVDA it also says "page 24 last page" then "Next page". Weird but mostly confusing.

amycole501 (commented):
I put some comments into the ticket; overall I didn't hear any "roles" announced but just noticed a very minor inconsistency in how the links were announced.

mahoneycm:
@amycole501 The example link is only _partially_ a live demo. Since there are no pages to paginate through, it does not update the pagination when clicked

If you edit the controls to make the "current" page the last page, the next button is removed to avoid confusion :)

image

As for the ellipses, they should never be clickable, only used to separate the first and last page links from the current pages if they are many pages away.

amycole501:
That helps, thanks. I think the way it's coded now is fine, it's usable and clear enough to be able to navigate and it doesn't have the "role" included which is good. I will say that JAWS does read "elipses" out loud but they aren't clickable. If we want to change that wording that's read to something clearer like "ellipses indicating non-visible pages" that might be clearer.

mejiaj:
IIRC part of the reason we added `role="presentation"` was to convey there are pages in between being represented by the dots.

So I like @amycole501's suggestion of adding an aria label that tells the user non-visible pages exist and are being represented by ellipsis. This not only gets rid of the error, but improves the experience.

mejiaj (changes requested):
IIRC part of the reason we added `role="presentation"` was to convey there are pages in between being represented by the dots.

So I like @amycole501's suggestion of adding an aria label that tells the user non-visible pages exist and are being represented by ellipsis. This not only gets rid of the error, but improves the experience.

mahoneycm:
@amycole501 @mejiaj 

I added `aria-label="ellipses indicating non-visible pages" to the ellipses `<li>` tag. With VoiceOver it reads the label and then "ellipses" for the span itself

> Ellipses indicating non-visible pages... _(press right arrow)_ ...ellipses

Alternatively, I could hide the ellipses from the screen reader and add a second `sr-only` span for the screen reader

```twig
  <li class="usa-pagination__item usa-pagination__overflow" >
    <span aria-hidden="true">
      …
    </span>
    <span class="usa-sr-only">ellipses indicating non-visible pages</span>
  </li>
  ```
This causes the screen reader to not read both the ellipses and the label.

Let me know which direction is preferred or if there is a better option that I'm unaware of!

**Review by mejiaj — APPROVED** [review]

File: packages/usa-pagination/src/usa-pagination.twig [hunk 3]
```diff
@@ -2,7 +2,7 @@
 {% from "@components/usa-pagination/src/includes/_pagination-arrow.twig" import paginationArrow %}
 
 {% set overflow %}
-  <li class="usa-pagination__item usa-pagination__overflow" role="presentation">
+  <li class="usa-pagination__item usa-pagination__overflow" aria-label="ellipses indicating non-visible pages">
```
thisisdano, line 5:
```suggestion
  <li class="usa-pagination__item usa-pagination__overflow" aria-label="ellipsis indicating non-visible pages">
```

thisisdano:
I'll draw up a changelog PR for this change

thisisdano (approved):

timveld:
While redesigning our own pagination I came across this discussion.
I absolutely agree that you had to remove the role="presentation", this violated the aria specification.
However, I think the aria-label on the list items containing the ellipses is superfluous.

Technically, Aria permits the placement of an aria-label attribute on a li element.
However, all screen readers I tested (JAWS, NVDA, iPhone) ignore this attribute and just read "...".  

Those screen readers *do* read an aria-label if placed on an ul element, but apparently screen reader designers agree that aria-labels on list items should be ignored and I agree with them. Naming list items just doesn't make sense semantically, and allowing authors to do that could result in terribly complicated and senseless designs which would greatly confuse users. Therefore I'd make the point that the aria specification should prohibit aria-label on list items, but this is not the right venue for that discussion. Point is that the text "ellipses indicating non-visible pages" does not describe the list item.Instead, it describes the list item's content so it should replace or somehow be linked to that content.

The aria-label is, at least from a practical point of view, superfluous as screen readers intentionally ignore it. Superfluous aria should be removed in all cases, "the first rule of aria is not to use it".

 I see two alternatives:
* Implement the second suggestion @mahoneycm made in his apr 6 comment i.e. make "..." aria-hidden and instead place the sr-only text "ellipses indicating non-visible pages". While this could help some blind users to better understand the component, it comes at the cost of reading a substantial amount of extra text which many screen reader users won't appreciate. Also, it amounts to providing a text alternative for punctuation.
* Just remove the aria-label and assume that blind users will understand that in this context "page 1, page 2, …, page 9" means pages 3 - 8 are not currently visible. I think that's a reasonable assumption, so for now that's the design I'd go for.  

But I realize that opinions may vary. @amycole501 which were your reasons for suggesting the additional text of "ellipses indicating non-visible pages"?

mahoneycm:
Oh I do see what @timveld is saying though. It is more descriptive for the "ellipses indicating non-visible pages" to be on the `...` element instead of it's parent. WIth VO, it currently reads:

> "ellipses indicating non-visible pages, group, 3 of 9", 
> "ellipses", 
> "end of ellipses indicating non-visible pages"

I would agree that either of the suggested methods would likely be a better user experience than this. This feels unnecessarily verbose.

amycole501:
I want to make sure I'm testing the same thing as when I listen to the live and test versions of pagination I don't hear anything verbose in neither JAWS nor NVDA. I'm using Chrome; I can test in other browsers, however. I just want to make sure we're comparing apples to apples. But in practice I agree that putting the ARIA on the ellipses instead of the parent makes the most sense.

timveld:
I'll split my answer up into two parts:

#### User panel
@amycole501 Excellent suggestion to ask your user panel! Though we don't have a formal user panel I'll also discuss this with some of our regular testers. I suggest asking them two questions:
Given a pagination component contained in a landmark labelled "pagination" and ellipses indicating non-visible pages:
1. How would you prefer the ellipses, visually represented as "...", to be read?
a) "..." (pronunciation will depend on your screen reader settings and speech synthesizer)
B) "Ellipses"
C) "...  indicating non-visible pages"
d) "ellipses indicating non-visible pages"
2. How would you prefer the links to the pages, visually represented as the page numbers ("1", "2" and so on) to read?
A) "1"
B) "page 1"
My opinion as an expert and screen reader user:
1. A, because this use of "..." is a commonly understood textual convention, matches the visual display and respects my preferences as reflected in my screen reader's speech settings and choice of speech synthesizer
2. A, because the necessary context is provided by the "pagination" label on the landmark and it saves me unnecessary repetitions of the word "page" (which, on our Dutch website, would be "pagina" so it wastes even more time)

#### Technical

I never suggested "putting the aria on the ellipses", that would also be ignored by screen readers. We can't and don't need to use aria for providing the additional hint, sr-only text is the only way to provide it.

Interesting that VoiceOver on Mac doesn't ignore the aria-label. But the way it reads the aria-label illustrates my point: semantically this text doesn't belong in an aria-label. And it also illustrates the point that there is no such thing as "harmless superfluous aria". There are so many different browser and AT combinations that there will Always be one which reads things in a way we didn't expect. Aria is a double-edged sword. Only ever apply it if:
* You are sure there is no alternative

* You understand why you need to add it and how exactly it will support users

* You have made sure you followed the specification
* You have it tested by actual screen reader users
Short summary: "the first rule of aria is not to use it".

PR #5200, opened 2023-03-27
USWDS - In-Page Navigation: Fix error when clicking link starting with number

### LINKED ISSUE #5199: USWDS - Bug: In-page navigation errors when heading starts with number
### Describe the bug

When an In-Page Navigation component includes links which start with a number, clicking on those links does not navigate to the relevant section, and instead logs a JavaScript error to the console.

Example:

```
Uncaught DOMException: Failed to execute 'querySelector' on 'Document': '#1-check-your-current-uswds-code-and-settings-versions-2' is not a valid selector.
```

This is likely related to #5068, which added the dynamically-created links based on heading content.

The reason for the error is likely due to CSS syntax restrictions, which prevent an selector ID from beginning with a number:

>In CSS, identifiers (including element names, classes, and IDs in [selectors](https://www.w3.org/TR/CSS2/selector.html)) [...] cannot start with a digit, two hyphens, or a hyphen followed by a digit.

https://www.w3.org/TR/CSS2/syndata.html#value-def-identifier

1. Go to https://designsystem.digital.gov/documentation/migration/
2. Click a link in the In-Page Navigation

### Expected Behavior

Page scrolls to the relevant section

### Related code

_No response_

### Screenshots

_No response_

### System setup

- USWDS version: 3.4.1
- Operating system: macOS 12.6.3
- Browser: Chrome Version 111.0.5563.110 (Official Build) (arm64)

### Additional context

_No response_

### Code of Conduct

aduth:
# Summary

**Fix an error when clicking a link starting with a number.** Certain links would not scroll as expected when clicked.

## Related issue

Fixes #5199

## Problem statement

See #5199

## Solution

This implements the first of the proposed options outlined at https://github.com/uswds/uswds/issues/5199#issuecomment-1485125100 .

This has a few advantages, in that it limits the scope of the changes and maintains backwards compatibility. The only downside that I can consider is that these IDs are still technically not valid for use as CSS selectors, in case someone would need to reference it as such. As demonstrated in the included tests, a workaround is to reference the ID as using an attribute selector (e.g. `[id="1-installation"]`).

## Testing and review

1. Run Mocha tests
2. Perhaps this could be incorporated into demo content so it could be observed in the Storybook preview?

### DIFF
--- a/packages/usa-in-page-navigation/src/index.js
+++ b/packages/usa-in-page-navigation/src/index.js
@@ -232,7 +232,7 @@ const createInPageNav = (inPageNavEl) => {
  * @param {HTMLElement} el An element within the in-page nav component
  */
 const handleClickFromLink = (el) => {
-  const elementToScrollTo = document.querySelector(el.hash);
+  const elementToScrollTo = document.getElementById(el.hash.slice(1));
   handleScrollToSection(elementToScrollTo);
 };
 
--- a/packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
+++ b/packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
@@ -132,6 +132,7 @@ tests.forEach(({ name, selector: containerSelector }) => {
         "h2 ~ h3 > .usa-anchor#section-1-1",
         "h2 ~ h3 ~ h3 > .usa-anchor#section-1-2-2",
         "h2 ~ h3 ~ h3 ~ h3 > .usa-anchor#section-1-3",
+        "h2 ~ h3 ~ h3 ~ h3 ~ h3 > .usa-anchor[id='1-4-section-1-4']",
       ].every((selector) => document.querySelector(selector));
 
       assert(ok);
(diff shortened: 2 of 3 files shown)


mahoneycm (approved):
This works like a charm! I think it would be useful to add it to the twig template for Storybook.

## Tested:
1. Added numbers to the start of headings in the storybook preview
2. Clicked through in-page navigation to confirm it scrolled to the correct heading
3. Confirmed no errors in the console
4. Logged `document.getElementById(el.hash.slice(1))` and confirmed the correct element was being selected.
5. Ran gulp test for in-page navigation

mejiaj (approved):
Thanks for the PR!  I think it would be useful to add it to the twig template for Storybook.

## Tested:
- Component works as expected in StorybookJS
- No issues in unit test

amyleadem (approved):
Good improvement. Thank you! 
Works as expected in Safari, Edge, Chrome, Firefox. 

I agree that it would be great to see an example of this reflected in the Storybook twig template.

thisisdano (approved):
Nice clean solution

PR #5216, opened 2023-04-06
USWDS - Settings: Update sass map pattern

### LINKED ISSUE #5215: USWDS - Feature: Update sass maps to allow single setting updates

Currently, if you choose to update a USWDS setting within a sass map, you must define each setting within the map even if you are keeping the default values for the other.

Related: #5011 

We can use `map-merge` with a default values map to only replace updated settings keys without needing to redefine default values. 

_No response_

### Additional context

_No response_

### Code of Conduct

mahoneycm:
# Summary

Updates sass map pattern to allow users to only define overrides while keeping defaults for the other values

## Breaking change

This is _potentially_ a breaking change.

## Related issue

Closes #5215

## Related pull requests

Update sass map instructions on utility pages: https://github.com/uswds/uswds-site/pull/2077

## Problem statement

Currently, if you choose to update a USWDS setting within a sass map, you must define each setting within the map even if you are keeping the default values for the other.

## Solution

We can use map-merge with a default values map to only replace updated settings keys without needing to redefine default values.

This solution
- Doesn't break existing sass map settings users have previously set
- _BUT_ if users left an element out of the map, (effectively disabling it) this PR would re-enable the default setting.

This PR updates `$theme-[map-name]`s to `$theme-[map-name]-complete` wherever the setting is called.

This allows users still use existing setting names, such as `$theme-utility-breakpoints`, to update which breakpoints are active, while using the `-complete` map to restore defaults to undefined elements.

> Note: I considered naming the `complete` maps as `-defaults` but I believe this makes less contextual sense wherever the maps are being called.

## Updated maps

- [x] `$theme-namespace`
- [x] `$theme-utility-breakpoints`
- [x] `$[utility-module]-settings` 

## Ignored maps

These maps were either, empty, set to `false` by default, or only had example values. In all three settings, we would want the users override to be the only value, which is currently how they are set up.

- `$theme-fontface-tokens`
- `$theme-font-[family]-src`
- `$[utility-module]-manual-calues`

## Testing and review

1. Checkout this [test uswds-sandbox repo](https://github.com/uswds/uswds-sandbox/tree/cm-breakpoint-map)
3. The test repo should already have this branch installed as it's USWDS package
4. Run `npm run start` on the test repo
5. View each test page and review instructions
7. Visit `_uswds-theme.scss` and expirament enabling/disabling different settings
8. Revisit index page in browser and view the changes
9. Confirm that
   - [ ] : Default breakpoints are working as expected
   - [ ] : Updating one breakpoint, doesn't deactivate the others
   - [ ] : Updated breakpoints work as expected
10. Look over all maps listed on the [settings](https://designsystem.digital.gov/documentation/settings/#general-settings-2) page
11. Inspect their default values in the USWDS settings and verify there aren't any additional updates to make

These tests are simple and mostly just enable different background colors as a quick visual feedback. Feel free to experiment with other utility settings to verify the new pattern is working!

### Default maps for reference:
 
**Breakpoints**

```scss
@use "uswds-core" with (
  $theme-utility-breakpoints: (
      "card": false,
      "card-lg": false,
      "mobile": false,
      "mobile-lg": true,
      "tablet": true,
      "tablet-lg": false,
      "desktop": true,
      "desktop-lg": false,
      "widescreen": false,
    )
);
```

**Namespace**
```scss
$theme-namespace: (
  "grid": (
    namespace: "grid-",
    output: true
  ),
  "utility": (
    namespace: "u-",
    output: false,
  ),
),
```

### DIFF
--- a/packages/uswds-core/src/styles/settings/_settings-utilities.scss
+++ b/packages/uswds-core/src/styles/settings/_settings-utilities.scss
@@ -15,6 +15,8 @@ https://designsystem.digital.gov/utilities
 ----------------------------------------
 */
 
+@use "sass:map";
+
 $utilities-use-important: false !default;
 $output-these-utilities: default !default;
 
@@ -29,7 +31,7 @@ used by utilities or layout grid
 */
 
 $theme-utility-breakpoints: () !default;
-$theme-utility-breakpoints-complete: map-merge((
+$theme-utility-breakpoints-complete: map.merge((
   // 160px:
   "card": false,
   // 240px:
--- a/packages/uswds-core/src/styles/settings/_settings-utilities.scss
+++ b/packages/uswds-core/src/styles/settings/_settings-utilities.scss
@@ -72,518 +72,575 @@ Settings
 ----------------------------------------
 */
 
-$add-aspect-settings: (
+$add-aspect-settings: () !default;
+$add-aspect-settings-complete: map.merge((
   output: true,
   responsive: false,
   active: false,
   focus: false,
   hover: false,
   visited: false,
-) !default;
+), $add-aspect-settings) !default;
 
-$add-list-reset-settings: (
+$add-list-reset-settings: () !default;
+$add-list-reset-settings-complete: map.merge((
   output: true,
   responsive: false,
   active: false,
   focus: false,
   hover: false,
   visited: false,
-) !default;
+), $add-list-reset-settings) !default;
 
-$align-items-settings: (
+$align-items-settings: () !default;
+$align-items-settings-complete: map.merge((
   output: true,
   responsive: false,
   active: false,
   focus: false,
   hover: false,
   visited: false,
-) !default;
+), $align-items-settings) !default;
 
-$align-self-settings: (
+$align-self-settings: () !default;
+$align-self-settings-complete: map.merge((
   output: true,
   responsive: false,
   active: false,
   focus: false,
   hover: false,
   visited: false,
-) !default;
+), $align-self-settings) !default;
 
-$background-color-settings: (
-  output: true,
-  responsive: false,
-  active: false,
-  focus: false,
-  hover: true,
-  visited: false,
-) !default;
-
-$border-settings: (
-  output: true,
-  responsive: true,
-  active: false,
-  focus: false,
-  hover: true,
-  visited: false,
-) !default;
-
-$border-color-settings: (
-  output: true,
-  responsive: true,
-  active: false,
-  focus: false,
-  hover: true,
-  visited: false,
-) !default;
-
-$border-radius-settings: (
-  output: true,
-  responsive: true,
-  active: false,
-  focus: false,
-  hover: false,
-  visited: false,
-) !default;
-
-$border-style-settings: (
-  output: true,
-  responsive: false,
-  active: false,
-  focus: false,
-  hover: false,
-  visited: false,
-) !default;
-
-$border-width-settings: (
-  output: true,
-  responsive: false,
-  active: false,
-  focus: false,
-  hover: false,
-  visited: false,
-) !default;
-
-$bottom-settings: (
-  output: true,
-  responsive: false,
-  active: false,
-  focus: false,
-  hover: false,
-  visited: false,
-) !default;
-
-$box-shadow-settings: (
+$background-color-settings: () !default;
+$background-color-settings-complete: map.merge((
   output: true,
   responsive: false,
   active: false,
   focus: false,
   hover: true,
   visited: false,
-) !default;
+), $background-color-settings) !default;
 
-$circle-settings: (
+$border-settings: () !default;
+$border-settings-complete: map.merge((
+  output: true,
+  responsive: true,
+  active: false,
+  focus: false,
+  hover: true,
+  visited: false,
+), $border-settings) !default;
+
+$border-color-settings: () !default;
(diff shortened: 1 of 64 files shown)


mejiaj (changes requested):
A nice enhancement! I tested in `uswds-sandbox` and can confirm:

- [x] Settings for `mobile` and `card` are specified in `src/_styles/_uswds-theme.scss`
- [x] Settings are added to the defaults, giving me 5 total breakpoints (from three)
- [x] Removing the overrides for `mobile` and `card` will fallback to default (`mobile-lg, tablet, desktop`)
- [x] ⚠️ Passing an invalid breakpoint `jumbo: true` results in no errors. _Current behavior on `develop`._

BTW, instead of manually linking via `npm link` you can install a git branch directly. You just need to run `npm i "uswds/uswds#cm-theme-setting-maps"` in your test branch.

---

Just a few requests:

1. The description marks this as potentially breaking, could you add examples of where this would be a breaking change?
2. Can you create a PR for necessary site updates?
3. Can you link that PR to this one in the [related pull requests section](https://github.com/uswds/uswds/blob/develop/.github/PULL_REQUEST_TEMPLATE.md#related-pull-requests)

mahoneycm:
@mejiaj @amyleadem If we like this implementation, should I go ahead and begin to look updating other maps to match this pattern?

The reason I ask, is because of the note I want to update on Site about updating _all_ values of a map.

> When modifying any settings stored as a map, you need to set <em>all</em> elements of the map, not just the element you wish to change.

I feel like the best route is to make updating map values uniform across the board so we only need to offer guidance on map settings once.

---

@mejiaj to answer your question of where this would be a breaking change:

Before this update, if a user updated breakpoint maps and excluded a breakpoint, it would be disabled. For example:

```scss
// User defined breakpoints
$theme-utility-breakpoints: (
  mobile: true,
  desktop: true,
)
```

`mobile` & `desktop` become the only enabled breakpoints. `mobile-lg` and `tablet` which are enabled by default, become disabled.

**After** this update, the default breakpoints will become enabled again despite not being defined, so the same settings as above would result in:

```scss

// User defined breakpoints
$theme-utility-breakpoints: (
  mobile: true,
  desktop: true,
)

// Results in the following breakpoints being enabled
mobile: true,
mobile-lg: true,
tablet: true,
desktop: true
```

I wanted to flag this because I thought it could potentially cause visual regressions for users who aren't expecting these values to become enabled.

packages/uswds-core/src/styles/settings/_settings-utilities.scss
```diff
@@ -28,7 +28,8 @@ used by utilities or layout grid
 ----------------------------------------
```
amyleadem, line 28: I am wondering if it would be helpful if we added a note here that users should edit `$theme-utility-breakpoints` in the settings configuration and not `$theme-utility-breakpoints-complete`.

amyleadem (commented):
This enhancement is great @mahoneycm! To me it makes configuring USWDS sass maps much more intuitive and the solution feels elegant. Thank you for taking this on!

I have completed the following tests:

Confirmed that editing a single value (Example: `”card”: true`):
1. [x] Adds the `card` breakpoint to the compiled CSS
2. [x] Does not remove existing breakpoints (ex: `mobile-lg` ) from the compiled CSS

---

This is up for debate, but I do think this functionality should be consistent across all USWDS settings maps before we release this enhancement. However, I haven’t fully investigated what would need to be included in that, so it might make sense to work on it in more than one PR and potentially release incrementally.

Additionally, we should prepare the following documentation updates before merging:

- Add a summary of the break in the PR description so users can easily find what they need to check/fix
- Consider adding a comment in `_settings-utilities.scss` that lets users know that they should update `$theme-utility-breakpoints` in their settings config and not `$theme-utility-breakpoints-complete`.
- Check all references to `$theme-utility-breakpoints` in uswds-site
- Update or remove the note `When modifying any settings stored as a map, you need to set *all*  elements of the map, not just the element you wish to change.` in uswds-site
- Add changelog items for the Settings page and any affected utility pages. The changelog should indicate what might possibly break.

mahoneycm:
@amyleadem @mejiaj Ready for re-review! Restructured the PR to reflect all the sass maps affected and not just breakpoints!

amyleadem (approved):
This is great, @mahoneycm! I am excited about this update. I also really appreciate the test pages you set up. They were very helpful. Good work!

- [x] Confirmed that customizing the value of a single map key affects *only* the designated key and does *not* affect other items in the map. Tested the following settings:
    - [x] **$theme-utility-breakpoints**
    - [x] **$theme-namespace**
    - [x] **$background-color-settings**
    - [x] **$border-color-settings**
- [x] Confirmed all Sass maps inside `packages/uswds-core/src/styles/settings/` have been updated
    - [x] Confirmed all settings maps have been updated with `map.merge` or `map.deep-merge` and each have a `-complete`  sibling
    - [x] Confirmed all original utility settings and defaults are still there and that their defaults have not changed

mejiaj (approved):
A nice improvement. I've tested:

- [x] References are accurate and up-to-date.
- [x] Selected breakpoints can be overridden without issues.

mahoneycm:
@thisisdano ready for final review!

thisisdano (approved):

thisisdano:
A great improvement!

PR #5228, opened 2023-04-10
USWDS - Tooltip: Add top as default position

### LINKED ISSUE #5218: USWDS - Bug: Tooltip documentation claims the tooltip positioned "top" when not specified; Incorrectly positions bottom left
### Describe the bug

The [tooltip documentation](https://designsystem.digital.gov/components/tooltip/#using-the-tooltip-component-2) says: 

> By default, tooltips appear on the top of their related element.

and

> * data-position="top": On top, horizontally centered. If the data-position attribute is omitted, the tooltip will appear on top by default.

However, when `data-position` is not specified at all, the tooltip is incorrectly positioned bottom left of trigger.

1. Create a button with `class="usa-tooltip"` and `title` but no `data-position` attribute.

### Expected Behavior

Tooltip should appear as if `data-position="top"` is set.

### Related code

```
<div class="padding-8">
  <div class="margin-4">
    <button type="button" class="usa-button usa-tooltip" title="Tooltip">Show tooltip</button>
  </div>
</div>
```

[Codepen demo](https://codepen.io/danbrady/pen/QWZwjaN)

### Screenshots

image

### System setup

- 3.4.1
- Desktop
- MacOS 11.7.4
- Chrome v111.0.5563.146 

### Additional context

Would be happy to tackle this if confirmed!

### Code of Conduct

danbrady:
# Summary

**Added default positioning.** Updated the default position to "top" if `data-position` attribute is not specified.

## Breaking change

## Related issue

Closes #5218

## Problem statement

1. Tooltip should appear on "top" when `data-position` omitted
2. Tooltip appears incorrectly positioned (at bottom left, with triangle pointing down)
3. If `data-position` is not specified, tooltip appears on broken

## Solution

1. Detect if `data-position` on trigger is set
2. Use set position or "top" if not

## Testing and review

1. Used this markup to test: `<button type="button" class="usa-button usa-tooltip" title="Show me correctly">Show on top</button>`
2. I didn't add this variation into the examples, but let me know if that is desired.

### DIFF
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -315,10 +315,14 @@ const setUpAttributes = (tooltipTrigger) => {
   const tooltipContent = tooltipTrigger.getAttribute("title");
   const wrapper = document.createElement("span");
   const tooltipBody = document.createElement("span");
-  const position = tooltipTrigger.getAttribute("data-position")
-    ? tooltipTrigger.getAttribute("data-position")
-    : "top";
   const additionalClasses = tooltipTrigger.getAttribute("data-classes");
+  let position = tooltipTrigger.getAttribute("data-position");
+
+  // Apply default position if not set as attribute
+  if (!position) {
+    position = "top";
+    tooltipTrigger.setAttribute("data-position", position);
+  }
 
   // Set up tooltip attributes
   tooltipTrigger.setAttribute("aria-describedby", tooltipID);


mejiaj (approved):
Thanks for your contribution @danbrady! I tested the following markup:

```html
<button type="button" class="usa-button usa-tooltip"  title="No position set on this tooltip">No position set</button>
```

And was able to get the expected behavior.

image

packages/usa-tooltip/src/index.js
```diff
@@ -369,7 +369,10 @@ const tooltip = behavior(
       [TOOLTIP_TRIGGER](e) {
         const { trigger, body } = getTooltipElements(e.target);
 
-        showToolTip(body, trigger, trigger.dataset.position);
+        // Default to "top" if `data-position` not specified
+        const position = trigger.dataset.position || "top";
```
mejiaj, line 373: This was supposed to be handled in  `:318`. We check for position in `setUpAttributes` and use the default. We should revisit this.

---


Allternatively, we could pass it as a default parameter in `showTooltip`:

```js
const showToolTip = (tooltipBody, tooltipTrigger, position="top") => {
```

danbrady: You're right, that's a better solution. Updated the PR. Thanks @mejiaj!

mahoneycm (approved):
Love having this as a fallback! Working great on my local 👍

amyleadem (commented):
Thanks for this submission @danbrady!  I confirmed that the tooltip positioning now works as expected when no `data-position` attribute is added in the markup. 

I just had one question:
@mejiaj - I noticed you flagged line 318 ([now 320](https://github.com/uswds/uswds/blob/68cd5186e97e2bf7d6d618540cfea27c82c43cdc/packages/usa-tooltip/src/index.js#L320)). Is the top assignment there still needed?

mejiaj:
> I just had one question: @mejiaj - I noticed you flagged line 318 ([now 320](https://github.com/uswds/uswds/blob/68cd5186e97e2bf7d6d618540cfea27c82c43cdc/packages/usa-tooltip/src/index.js#L320)). Is the top assignment there still needed?

Yes, but it could be improved by setting the data-position to top via `setAttribute()` or `dataset.position = "top"`.  Currently, it won't do anything if we default to string `"top"`.

danbrady:
FWIW, it appears that removing lines [318-320](https://github.com/uswds/uswds/blob/68cd5186e97e2bf7d6d618540cfea27c82c43cdc/packages/usa-tooltip/src/index.js#L320-L322) as well as 'position' from the returning object on [355](https://github.com/uswds/uswds/blob/68cd5186e97e2bf7d6d618540cfea27c82c43cdc/packages/usa-tooltip/src/index.js#L355) will have no impact to the functionality to tooltip, even before this PR. The positioning happens when `showTooltip` is called and it passes the value from `trigger.dataset.position` on [374](https://github.com/uswds/uswds/blob/68cd5186e97e2bf7d6d618540cfea27c82c43cdc/packages/usa-tooltip/src/index.js#L374).

I can update the PR to remove these line and keep the position param with a default value of 'top'. Or, alternatively, update the `setUpAttributes` method to do the position detection or  fallback to 'top' (via setAttribute) and remove the default position param value. Is either one of those preferred?

mejiaj:
@danbrady removing `318-320` might not have any effect on the functionality, but setting the attribute would be helpful in making the experience consistent and possibly easier to debug.

danbrady:
@mejiaj Ok, updated! Let me know if that's more inline. I kept all the same logic in place for consistency, but safely added default "top" positioning.

amyleadem (approved):
I like this solution. It works well!

I tested this tooltip html with no `data-position` attribute:
```
<button type="button" class="usa-button usa-tooltip" title="No position set on this tooltip">No data-position set</button>
```

And I saw that the tooltip is successfully assigning top position:
image

mejiaj (approved):
Nice fix, thank you!

**Tested**
```
<div class="margin-4">
    <button type="button" class="usa-button usa-tooltip" title="Forgot to position">No position set</button>
</div>
```

**Result**
image

thisisdano (approved):

thisisdano:
Thank you!

PR #5233, opened 2023-04-13
USWDS - Validation: Allow textareas to use validation (Enhancement)

### LINKED ISSUE #4702: Validation Support for textarea input types
When text areas are used in a form, it would be nice to be able to use the data validation engine with those text areas. 

I could write javascript to perform the work, but USWDS has these great data validation properties that would be awesome to use on text areas.

danbrady:
# Summary

**Textarea components can use validation** Now you can use validation on textareas, just like you do on inputs.

## Breaking change

## Related issue

Closes #4702

## Problem statement

Currently, validation can only be used on `input` elements. This solves most validation needs. However, `textarea` is a form element in which validation might be desired (e.g. not empty, minimal number of characters or words.)

## Solution

1. Current selector for validation is `input[data-validation-element]`
2. Updated validation logic to include `textarea[data-validation-element]`

## Testing and review

1. Added `textarea` to tests

## Dependency updates
None.

### DIFF
--- a/packages/usa-validation/src/index.js
+++ b/packages/usa-validation/src/index.js
@@ -3,7 +3,8 @@ const validate = require("../../uswds-core/src/js/utils/validate-input");
 const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
 const selectOrMatches = require("../../uswds-core/src/js/utils/select-or-matches");
 
-const VALIDATE_INPUT = "input[data-validation-element]";
+const VALIDATE_INPUT =
+  "input[data-validation-element],textarea[data-validation-element]";
 const CHECKLIST_ITEM = `.${PREFIX}-checklist__item`;
 
 // Trigger validation on input change
--- a/packages/usa-validation/src/test/template-textarea.html
+++ b/packages/usa-validation/src/test/template-textarea.html
@@ -1,0 +1,13 @@
+<div>
+  <ul class="usa-checklist" id="textarea_validation_list">
+    <li class="usa-checklist__item" data-validator="empty">Cannot by empty</li>
+  </ul>
+  <textarea
+    class="usa-textarea"
+    id="comment"
+    name="comment"
+    aria-describedby="textarea_validation_list"
+    data-validate-empty="\S"
+    data-validation-element="textarea_validation_list"
+  ></textarea>
+</div>
--- a/packages/usa-validation/src/test/validator-textarea.spec.js
+++ b/packages/usa-validation/src/test/validator-textarea.spec.js
@@ -1,0 +1,81 @@
+const fs = require("fs");
+const path = require("path");
+const assert = require("assert");
+const validator = require("../index");
+
+const TEMPLATE = fs.readFileSync(
+  path.join(__dirname, "/template-textarea.html")
+);
+
+const TEXTAREA_SELECTOR = "[data-validation-element]";
+const VALIDATORS = "[data-validator]";
+const VALIDATOR_SUMMARY = "[data-validation-status]";
+const CHECKED_CLASS = "usa-checklist__item--checked";
+
+const textareaSelector = () => document.querySelector(TEXTAREA_SELECTOR);
+
+const keyup = (el) => {
+  el.dispatchEvent(new Event("change", { bubbles: true }));
+};
+
+const tests = [
+  { name: "document.body", selector: () => document.body },
+  { name: "textarea", selector: textareaSelector },
+];
+
+tests.forEach(({ name, selector: containerSelector }) => {
+  describe(`validator component initialized at ${name}`, () => {
+    const { body } = document;
+
+    let validated;
+    let validators;
+
+    beforeEach(() => {
+      body.innerHTML = TEMPLATE;
+
+      validator.on(containerSelector());
+
+      validated = textareaSelector();
+      validators = Array.from(document.querySelectorAll(VALIDATORS));
+    });
+
+    afterEach(() => {
+      validator.off(containerSelector());
+      body.textContent = "";
+    });
+
+    describe("initialization", () => {
+      it("creates a hidden span element for accessibility status summary readout", () => {
+        if (validators) {
+          assert.notStrictEqual(
+            document.querySelector(VALIDATOR_SUMMARY),
+            null
+          );
+        }
+      });
+    });
+
+    describe("validation state", () => {
+      it(`adds .${CHECKED_CLASS} for all successful validations`, () => {
+        validated.value = "This is one great comment!";
+        keyup(validated);
+        validators.forEach((checkbox) => {
+          assert.strictEqual(checkbox.classList.contains(CHECKED_CLASS), true);
+        });
+      });
+
+      it(`removes .${CHECKED_CLASS} for failed validations`, () => {
+        validated.value = "";
+        keyup(validated);
+        validators.forEach((checkbox) => {
+          const checked = checkbox.classList.contains(CHECKED_CLASS);
+          if (checkbox.getAttribute("data-validator") === "empty") {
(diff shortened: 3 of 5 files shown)


amyleadem:
Thanks for this submission! I have tagged this for review and we will take a look at our next available opportunity. If you have any questions in the meantime, please let us know!

danbrady:
Good question, @mejiaj. Is it fair to think character count is validation for a **single, specific requirement**? (I.e. cannot be more than N characters.) From the [docs](https://designsystem.digital.gov/components/character-count/):

> Character count helps users know how much text they can enter when there is a limit on the number of characters.

(Admittedly, this is probably the most useful validation for a `textarea`.)

However, allowing `textarea` to use the validation package allows it to provide visual feedback for multiple validation requirements that a `textarea` could have. For example: must have a _minimal_ number of characters/words.

The use cases are likely rare, but are not non-existent.

packages/usa-validation/src/usa-validation.twig
```diff
@@ -1,17 +1,33 @@
...
+				<ul class="usa-checklist" id="validate-comment">
+					<li class="usa-checklist__item" data-validator="empty">Cannot be empty</li>
+				</ul>
+			</div>
+		</div>
+
+		<label class="usa-label" for="comment">Comment</label>
+		<textarea class="usa-textarea" id="comment" name="comment" data-validate-empty="\S" data-validation-element="validate-comment"></textarea>
+		<input class="usa-button" type="submit" value="Submit comment">
+	</fieldset>
```
mejiaj, line 32: Can we separate these tests, one for input and another for text area?

packages/usa-validation/src/test/validator.spec.js
mejiaj, line 1: Should we add a test that mentions initialization on textarea?

mejiaj (changes requested):
Looks good as far as extending the functionality to textarea. Just some comments below.

packages/usa-validation/src/test/validator.spec.js
danbrady: Agreed. I'll separate `textarea` into separate story, test, etc. and add this to it instead of mixing it up with `input`. I'll update the PR...

danbrady:
Ok, @mejiaj, I split textarea out. A few observations when looking at the tests on the original `input` validation:

https://github.com/uswds/uswds/blob/18e648028a624f387d7999fd6f2c287c56f021ad/packages/usa-validation/src/test/validator.spec.js#L10

I don't see an element with that `data-checklist-label` selector in any template?

https://github.com/uswds/uswds/blob/18e648028a624f387d7999fd6f2c287c56f021ad/packages/usa-validation/src/test/validator.spec.js#L47-L53

`queryElement` isn't a valid method (perhaps `querySelector` was intended?), however execution never makes it to that point to throw an error because the `validators` object is always empty...

https://github.com/uswds/uswds/blob/18e648028a624f387d7999fd6f2c287c56f021ad/packages/usa-validation/src/test/validator.spec.js#L13

https://github.com/uswds/uswds/blob/18e648028a624f387d7999fd6f2c287c56f021ad/packages/usa-validation/src/test/validator.spec.js#L37-L39

(It's looking for the validators _within_ the `input` and not the validation checklist.)

Amazingly, all the test pass! You can even change all the asserts to anything and it still passes. 😄 

I addressed all these in the `textarea` tests. Please let me know if I'm on the right track with them? And if I am, should I update the tests for `input` as part of this PR? Thanks!

mejiaj:
Thanks for making these changes and apologies for the late reply.
 
### Data checklist label

> I don't see an element with that `data-checklist-label` selector in any template?

Looks like a regression. This was introduced in #4914. Looks like in rework it was removed in commit [c284450](https://github.com/uswds/uswds/pull/4914/commits/c284450b7f8aa0f4cffc872fb8f809c48f8f39b7), but the test wasn't updated. Thanks for catching.

### Unit tests

#### Validation list items

> `queryElement` isn't a valid method (perhaps `querySelector` was intended?), however execution never makes it to that point to throw an error because the `validators` object is always empty...

Yes, that looks incorrect.

#### Incorrect validators

> https://github.com/uswds/uswds/blob/18e648028a624f387d7999fd6f2c287c56f021ad/packages/usa-validation/src/test/validator.spec.js#L13
> 
> https://github.com/uswds/uswds/blob/18e648028a624f387d7999fd6f2c287c56f021ad/packages/usa-validation/src/test/validator.spec.js#L37-L39
> 
> (It's looking for the validators _within_ the `input` and not the validation checklist.)

This should fix this. Validators are set in list items and should be checked there.

It's also a good callout for improving our testing, thank you.

mejiaj:
## Todo

- [ ] Remove unused data attributes
- [ ] Remove unused references in unit tests

mejiaj (approved):

packages/usa-validation/src/index.js
```diff
@@ -3,7 +3,8 @@ const validate = require("../../uswds-core/src/js/utils/validate-input");
 const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
 const selectOrMatches = require("../../uswds-core/src/js/utils/select-or-matches");
 
-const VALIDATE_INPUT = "input[data-validation-element]";
+const VALIDATE_INPUT =
+  "input[data-validation-element],textarea[data-validation-element]";
```
mahoneycm, line 7: ## Validate input selector

Curious if there is a slightly cleaner way to define this selector. Probably not a blocker because this works well but may be room for a future enhancement in the future 🤔

packages/usa-validation/src/test/template-textarea.html
mahoneycm, line 1: ## Validation Alert: Defined width (future enhancement)

Currently, the validation alert matches the size of the input it's tied to. With text area, users can resize the text box using the corner of the input and it causes the alert to grow with it.

I'm curious if it'd be a good idea to set a max width on the alert or to limit the text area to only extend vertically. Flagging for consideration!

[Kapture 2024-02-05 at 17.21.47.webm](https://github.com/uswds/uswds/assets/25211650/bab1e004-60f7-4f6c-a2e8-2f3acff75548)

mahoneycm (commented):
Component, stories, and tests look good! Left two comments for potential future enhancements for the team.

CC: @amyleadem @mejiaj

packages/usa-validation/src/test/template-textarea.html
amyleadem: I don't mind this behavior, actually, but that might just be personal preference. Happy to discuss this as a possible future change if desired. I don't think it needs to be a blocker for this PR.

amyleadem (approved):
Looks good to me!

Here is what I tested:
- [x] Confirm that the validation component works with `textarea` elements
- [x] Confirm that a new story for the `textarea` variant exists and works as expected
- [x] Confirm unit tests work as expected

thisisdano (approved):
LGTM

aduth:
It looks like the changes here may have introduced an accessibility build failure which is causing other pull requests to fail.

```
  1) Components/Validation
       Textarea Validation:
     Detected the following accessibility violations!

     1. label (Form elements must have labels)

        For more info, visit https://dequeuniversity.com/rules/axe/4.8/label?application=axeAPI.

        Check these nodes:

        - html: <textarea class="usa-textarea" id="comment" name="comment" data-validate-empty="\S" data-validation-element="validate-comment"></textarea>
          summary: Fix any of the following:
                     Form element does not have an implicit (wrapped) <label>
                     Form element does not have an explicit <label>
                     aria-label attribute does not exist or is empty
                     aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
                     Element has no title attribute
                     Element has no placeholder attribute
                     Element's default semantics were not overridden with role="none" or role="presentation"
```

PR #5000, opened 2022-10-19
USWDS: Banner - Remove grid-layout dependency. Resolves #4868

mahoneycm:
## Component 
[Preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-usa-banner-package/iframe.html?args=&id=components-banner--default&viewMode=story)

## Description

This PR removes the grid-layout dependency to cut down on the component's package size. It uses flex to maintain the same look as before.

Resolves #4868 

### Isolate Banner package
- In `uswds/_index.scss` remove all packages besides `usa-banner`
- Remove `/src/styles` so it only reads `@forward "usa-banner"`

### Check output css file size
- Run `npx gulp buildSass` to compile CSS
- Check `dist/css/uswds.css` filesize _(I use a vscode plugin but you can cd the directory to it and run `open .`_
- With this change it should be around 17kb. Before it was about 47kb.

If you do the above steps on another branch, you should see the larger filesize.

Before you hit Submit, make sure you’ve done whichever of these applies to you:

- [x] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [x] Run `npm test` and make sure the tests for the files you have changed have passed.
- [ ] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [x] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #5234, opened 2023-04-13
USWDS - Footer: Remove additional space. Fixes #5214 (Bug)

### LINKED ISSUE #5214: USWDS - Bug: Agency contact center in the footer component is not right-aligned in the grid
### Describe the bug

The grid gap classes on that part of the footer cause a space on the right that misaligns what seems to be right-justified content and causes it to misalign from the grid. This misalignment becomes apparent if you have other right-justified content close-by:

Screen Shot 2023-04-05 at 10 09 33 AM

An easy fix that worked for me is to remove `mobile-lg:grid-gap-2` on .usa-footer__logo and set the distance between the agency logo and title through CSS margin instead:

Screen Shot 2023-04-05 at 10 09 33 AM

I also removed the grip-gap on the parent grid-row in that usa-footer__secondary-section, but mileage could vary by use-case (in other word, some more work could be needed to fully align content in that section of the footer)

 

1. Examine the alignment of the Agency section of the footer with the grid, specifically on the right side.

### Expected Behavior

I expect the 'Agency Contact' section that is comprisd of the social nav, Agency Contact Center header and agency contact info to be right justified and align with the grid on the right.

### Related code

_No response_

### Screenshots

Screen Shot 2023-04-05 at 10 16 43 AM
Screen Shot 2023-04-05 at 10 09 33 AM

### System setup

- 3.4.1
- OS
- Chrome

### Additional context

_No response_

### Code of Conduct

danbrady:
# Summary

**Fix alignment in Footer** Removes additional space that prevents the correct right alignment of elements.

## Breaking change

## Related issue

Resolves #5214

## Preview link

Current state:
image

Proposed fix:
image

## Problem statement

Additional padding on right side of footer within the secondary section could cause misalignment with other sections.

## Solution

Modified HTML structure to use a layout that prevents additional padding/space. (No new styles/classes created.)

## Testing and review

1. Update was validated in browser across all breakpoint views
2. Additional validation of correct layout across all breakpoints/platforms would be good! Or please let me know if there is automated visual regression testing that I should be using :)

### DIFF
--- a/packages/usa-footer/src/styles/_usa-footer.scss
+++ b/packages/usa-footer/src/styles/_usa-footer.scss
@@ -208,6 +208,10 @@ $-chevron-expand-more: map-merge(
 
   > .grid-container {
     @include grid-container($theme-footer-max-width);
+
+    > .grid-row {
+      justify-content: space-between;
+    }
   }
 
   a {
--- a/packages/usa-footer/src/usa-footer.twig
+++ b/packages/usa-footer/src/usa-footer.twig
@@ -10,17 +10,15 @@
   <div class="usa-footer__secondary-section">
     <div class="grid-container">
       {% block secondary %}
-        <div class="grid-row flex-align-center">
-          <div class="mobile-lg:grid-col-6">
-            <div class="usa-footer__logo grid-row mobile-lg:grid-gap-2">
-              <div class="mobile-lg:grid-col-auto">
-                <img class="usa-footer__logo-img" src="{{ img_path }}/logo-img.png" alt="">
-              </div>
-              <div class="mobile-lg:grid-col-auto">
-                <p class="usa-footer__logo-heading">
-                  {{ footer.secondary_section.agency_name |e }}
-                </p>
-              </div>
+        <div class="grid-row grid-gap">
+          <div class="usa-footer__logo grid-row mobile-lg:grid-col-6 mobile-lg:grid-gap-2">
+            <div class="mobile-lg:grid-col-auto">
+              <img class="usa-footer__logo-img" src="{{ img_path }}/logo-img.png" alt="">
+            </div>
+            <div class="mobile-lg:grid-col-auto">
+              <p class="usa-footer__logo-heading">
+                {{ footer.secondary_section.agency_name |e }}
+              </p>
             </div>
           </div>
           <div class="usa-footer__contact-links mobile-lg:grid-col-6">
--- a/packages/usa-footer/src/usa-footer.twig
+++ b/packages/usa-footer/src/usa-footer.twig
@@ -10,15 +10,17 @@
   <div class="usa-footer__secondary-section">
     <div class="grid-container">
       {% block secondary %}
-        <div class="grid-row grid-gap">
-          <div class="usa-footer__logo grid-row mobile-lg:grid-col-6 mobile-lg:grid-gap-2">
-            <div class="mobile-lg:grid-col-auto">
-              <img class="usa-footer__logo-img" src="{{ img_path }}/logo-img.png" alt="">
-            </div>
-            <div class="mobile-lg:grid-col-auto">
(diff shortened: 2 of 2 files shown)


mejiaj (commented):
Looks like the issue is coming from the nested grid row's. Specifically `.usa-footer__logo.grid-row.mobile-lg:grid-col-6 mobile-lg:grid-gap-2`.

Maybe we should consider simplifying the markup by removing the grid rows for footer logo, social logos, and addresses. I know we kept the grid classes in Banner rework #5000, but there's a lot more here so worth removing.

mahoneycm (commented):
@mejiaj Going off of your comment, it looks like there are margins being set by `.grid-row.grid-gap>*` that is causing that space on the right.

Should we consider this a breaking change because we're editing markdown?

mejiaj:
> @mejiaj Going off of your comment, it looks like there are margins being set by `.grid-row.grid-gap>*` that is causing that space on the right.
> 
> Should we consider this a breaking change because we're editing markdown?

Yes, it'd be a breaking change because of markup. I've added the `Needs: discussion` tag to the original issue, so we can discuss before making the change.

danbrady:
Happy to tackle simplifying the HTML/CSS upon your discussion. Let me know if we'd like to remove the grid utilities on just the sections @mejiaj mentioned, or the whole component. (Or any other requirements that emerge.)

mejiaj:
@danbrady after discussing with the team let's move forward with removing grid classes from component ― starting with the problematic areas.

danbrady:
@mejiaj To confirm, the goal for the next step is to remove all grid classes from problem areas (footer logo, social logos, and addresses) and get equivalent layout using flex utilities, but leave the grid classes (and `usa-layout-grid` package) in for the rest of the footer(s)?

mejiaj:
@danbrady sorry I missed this message. 

Yes, we're keeping grid classes (to prevent breaking change), but removing the grid layout dependency in SASS.

danbrady:
No problem, @mejiaj! That's the direction I took. I have a WIP PR (#5289) for removing the grid dependency but need some guidance, when you have the chance. I'll tag you for review. Thanks!

danbrady:
This updates the markup to remove the errant space defined in the issue. I've updated the markup, but only to use existing classes so I don't believe it's a breaking change? Thanks!

mahoneycm:
Hey there @danbrady!

Any changes to component markup are considered breaking changes because they require users to manually update the markup in their component code.

It'd probably be a better idea to return to a scss based solution for this one.

mejiaj (changes requested):
@danbrady thanks for creating this. 

As @mahoneycm's mentioned; any markup changes would be considered breaking. Would you be able to follow his suggestion and use a SCSS-based approach?

danbrady:
Hi @mahoneycm / @mejiaj!  Sure, I'll take a look and see if we can fix it with SCSS alone and not change any markup. We'll probably have to override grid/flex classes for this case.

So I understand for next time, can you explain to me why this instance would be a breaking change? I understand that if a user updates markup and it refers to updated CSS that hasn't been published it would actually break a component. However, while this change does change the markup, it only changes markup that refers to existing/unchanged styles. If it doesn't reference any new or changed classes, the change should operate ok whether the user updates the markup or not, no? 

(BTW, I'm not disagreeing. Just want to learn :)

mejiaj:
@danbrady historically these types of changes bring difficulties to users, especially those on small teams trying to stay up-to-date with USWDS.

We're working on improving how we communicate breaking changes & how we can make upgrading easier. In the meantime, we're trying to avoid changing markup.

mejiaj (approved):

mahoneycm (approved):
Short and effective! Love to see it. Thanks @danbrady ! Looks great

thisisdano (approved):
Nice scoped fix

PR #5241, opened 2023-04-17
USWDS - USWDS: Forward dependencies by entrypoint reference

aduth:
# Summary

This updates how the `uswds` package internally references its forwarded dependencies.

**Why?**

- If the idea is to treat packages as a formal construct referenced by their top-level handle, this convention should extend to how internal packages reference each other
- It's generally more readable
- It's consistent with other forwarded references in the same file (e.g. `@forward "uswds-typography";`)
- It improves compatibility with how downstream projects may try to override USWDS behaviors ([example context](https://github.com/18F/identity-style-guide/pull/336#pullrequestreview-1388712031))
- It avoids potential conflicts if a package's entrypoint defines dependencies which would not otherwise by loaded through direct references to the deep path
- There's no reason not to?

To the last point, I can't tell if there was a reason for having forwarded this way, aside from avoiding duplicate references to shared dependencies. But, best I can tell, Sass is pretty good about deduplicating dependencies that have already been loaded.

An attempt at a CHANGELOG:

>**Improve consistency and portability of default package forwarding.** Restructure internal references to packages to maintain formal entrypoint consistency.

## Problem statement

As alluded earlier, in the Login.gov Design System, we're trying an approach for overriding USWDS behaviors by adding "intercept" files for USWDS packages we customize, which load the original file as well as any customizations.

The folder structure ends up looking like...

```
packages/
├ usa-accordion/_index.scss # Original USWDS package
└ _usa-accordion.scss       # "Intercept" file
```

Where an "intercept" file contains something like...

```scss
@forward './usa-accordion/index';

.usa-accordion {
   # My custom styles
}
```

So that a consumer of our design system can still use packages in the same way as USWDS, while receiving any overrides that we define:

```scss
@forward 'usa-accordion';
```

The problem is that this only works for overriding a package if it's referenced by its entrypoint, and because `uswds/_index.scss` bypasses the entrypoint, the intercepts are not effective.

By changing `uswds/_index.scss` to reference the entrypoints, it effectively works the same as it does today, while supporting the override behavior described above.

## Testing and review

There is not expected to be any meaningful impact to the styles generated after these changes.


aduth:
Reading through documentation, I stumbled upon this section, which may explain why it had been implemented this way in the first place. Though I'd be curious to see what the practical performance difference would end up being.

https://designsystem.digital.gov/documentation/migration/#using-package-source

aduth:
> Reading through documentation, I stumbled upon this section, which may explain why it had been implemented this way in the first place. Though I'd be curious to see what the practical performance difference would end up being.
> 
> > You can improve performance by manually forwarding all your project component’s dependencies, then pointing your Sass entry point directly at each package source, bypassing each package’s dependency import.
> 
> https://designsystem.digital.gov/documentation/migration/#using-package-source

I started doing some benchmarking here and immediately noticed that this would increase the default build time dramatically (~3.7s to ~7.1s with the default Dart implementation), so I'm going to close this out for now. I think it's worth better understanding why Sass is doing so much work to when loading the same dependencies multiple times, but in the meantime it's probably best to keep it as-is.

PR #5208, opened 2023-03-28
USWDS - Modal: Implement behaviors and add conditional initialization

mahoneycm:
# Summary

**Implements behaviors utility and queries DOM before initialization**. Now `modal` follows the pattern set by other dynamically created components by using the behaviors utility. Additionally, the component will query the DOM before building the component to prevent errors caused by multiple initializations being called. This is beneficial for component-level initialization.

## Breaking change

This is not a breaking change.

## Related issue

Closes #5062 

## Preview link
[Modal (federalist preview)](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-modal-behavior/iframe.html?args=&id=components-modal--default&viewMode=story)

[Test repo (Follow instructions below)](https://github.com/rachidatecs/uswds-modals)

## Problem statement

When using modern JS frameworks such as Angular or React, calling on the component to initialize multiple times caused the modal's functionality to break. The error occurred because the initialization process selects all `usa-modal` divs and rebuilds them using their ID. The newly generated div removes the ID and moves it to the wrapper. 

Once initialization is called again, it selects the new `usa-modal` div _without_ the ID, which causes build errors due to the ID being `null`

## Solution

1. During initialization, check for previously declared `modalId`before running `setUpModal` function

## Testing and review

1. Checkout this PR
2. Confirm modal js follows standard dynamic component patterns regarding behaviors util
3. Run `npm link`
4. Clone this [test repo](https://github.com/rachidatecs/uswds-modals)
5. Run `npm install` `npm run start`
6. Confirm error on the mulitple modals page
7. Return to code and run `npm link @uswds/uswds`
8. Inspect multiple modal page and confirm modals are working

## Notes

Output HTML before change:
image

After:
image

Before opening this PR, make sure you’ve done whichever of these applies to you:
- [x] Confirm that this code follows the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [x] Run `git pull origin [base branch]` to pull in the most recent updates from your base and check for merge conflicts. (Often, the base branch is `develop`).
- [x] Run `npm test` and confirm that all tests pass.

PR #5242, opened 2023-04-17
USWDS - Banner: Add accordion dependency

### LINKED ISSUE #5179: USWDS - Bug: Banner has issues with accordion initialization
### Describe the bug

Banner fails to initialize properly because:
1. Component doesn't have its own init method.
2. There's a requirement to USA accordion JS that's not called.

1. Create a banner component in react
3. Try to initialize based on [react hooks example](https://designsystem.digital.gov/documentation/developers/#:~:text=React%20hooks%20example%3A)
  
  ```js
// UsaBanner.jsx

import { useEffect } from "react";
import banner from "@uswds/uswds/js/usa-banner/";

function UsaBanner() {

  useEffect(() => {
    const ref = document.body;
    // window.uswds
    // initialize
    banner.on(ref);

    return () => {
      banner.off();
    };
  });

…component markup
   ```

### Expected Behavior

Component should initialize in react without requiring uswds-init.

### Screenshots

_No response_

### System setup

- USWDS version: 3.4.0

### Additional context

<details><summary>Outdated notes</summary>
<p></p>

**~Affected components~**
- ~Banner~
- ~Header~
- ~Modal~

~These components currently depend on `usa-js-loading` class from `uswds-init.js`. Users of dynamic frameworks will have unexpected results when trying to initialize.~

~For example, in the case of banner: the banner will appear expanded and won't toggle properly.~

</details> 

### Code of Conduct

mejiaj:
# Summary

**Banner initializes without external dependencies.** USA banner module now initializes on it's own without the user having to separately initialize accordion.

> **Note**
>Because accordion classes are used within banner we have to initialize accordion from within banner. This means that a user could initialize accordion and expect their banners to work and vice versa.
>
>We don't see this issue when projects import `uswds.js` because it includes accordion already. Now you should be able to import banner and only banner and have it work as expected.

## Breaking change

## Related issue

Closes #5179. 

## Related pull requests

1. Need to create changelog entry in banner in `uswds-site`.
2. Also need to update [interactive components spreadsheet 🔒](https://docs.google.com/spreadsheets/d/1fAGnDf1qPdMin2je2RNFBCpMPGbLXs6RZxcKH-KzaxM/edit#gid=0)

## Preview link

Preview link: [Banner init test](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-banner-init-test/iframe.html?args=&id=components-banner--test&viewMode=story)

## Problem statement

1. Banner should initialize **without** user having to import and init accordion.
2. Banner should work independently, meaning one **shouldn't** trigger another
3. Banner **shouldn't** initialize or have any effect on USA Accordion

## Solution

1. Banner JS now imports accordion [4550a41].
2. Created a setting (`hasInit - boolean`) in accordion to check if `accordion.init()` has run.
3. Call `accordion.init()` in banner if it hasn't. Allows us to import Banner modularly without user having to import and initialize accordion.
4. Removed accordion import dependency in unit test [41144b0].
5. ~Banner **only** initializes accordion on individual banners.~

### Alternatives

1. Remove `usa-accordion` dependency and use `toggle.js` utility to recreate functionality.
2. Update banner guidance to note accordion dependency.

1. ~**Run `accordion.on()` on banner init.** This will also init accordions on the page, which will cause issues for users.~
3. ~**Import toggle and selectors from accordion.** We'd be recreating accordion functionality. More code reuse, but no real performance benefit.~

## Major changes

Banner script now imports accordion. Meaning users can follow [Guidance](https://designsystem.digital.gov/documentation/developers/#js-customization-2:~:text=import%20accordion%20from%20%22%40uswds/uswds/js/usa%2Daccordion%22%3B) to import banner via 

```js
import banner from "@uswds/uswds/js/usa-banner"
```

Initialize it with `banner.on()` and have component work properly. There are some new changes behind the scenes to check if accordion has initialized.

## Testing and review

<details><summary>Traditional site</summary>
<p></p>

1. Check out testing branch `jm-banner-init-test`
3. Start StorybookJS `npm start`
6. Visit [Banner init test page](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-banner-init-test/iframe.html?args=&id=components-banner--test&viewMode=story)
7. Accordion & banner should work without any issues
8. Accordion should only initialize once
9. Locally, visit `uswds-core/src/js/start.js`
10. Init **Banner** only, everything should work
11. Init **Accordion** only, everything should work
12. Init **all** components, components should work without conflicts
13. Run `npm run test:unit`; banner & accordion should only contain their respective scripts & run without errors
</details> 

<details><summary>Frameworks</summary>
<p></p>

1. Visit [uswds/uswds-sandbox at jm-test-uswds-5179](https://github.com/uswds/uswds-sandbox/tree/jm-test-uswds-5179)
2. Run `npm install`
3. Run `npm start`
4. Accordion and banner should work without conflicts.
5. Import **only** banner; component should initialize accordion.
6. Import both banner and accordion. Both components should work without conflicts.
</details> 

### Additional information
- _Additional instructions in [start.js 21-27](https://github.com/uswds/uswds/blob/e499670aaaaf8af76df3c428f2e10b5e893c8c81/packages/uswds-core/src/js/start.js#L21-L27)_
- Created branch in uswds-sandbox [`jm-test-uswds-5179`] for additional testing

<details><summary>Previous testing instructions</summary>

Keeping for historical record.

1. Visit preview link for [Banner init test](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-banner-init-test/iframe.html?args=&id=components-banner--test&viewMode=story)
14. Each banner should work
15. One banner should have no effect on another
16. Accordion component **should not** work. If so, it would mean that banner is unexpectedly initializing accordions. In projects this would cause unintended side effects.
17. `npm run test:unit` should run without errors.

</details> 

## Compiled package size

| uswds.js | 798KB | 801KB   |
| uswds.min.js     | 87KB  |   87KB  |

### DIFF
--- a/packages/usa-accordion/src/index.js
+++ b/packages/usa-accordion/src/index.js
@@ -82,10 +82,18 @@ const accordion = behavior(
   },
   {
     init(root) {
-      select(BUTTON, root).forEach((button) => {
-        const expanded = button.getAttribute(EXPANDED) === "true";
-        toggleButton(button, expanded);
-      });
+      // Check if Banner has previously initialized accordion.
+      if (!this.hasInit) {
+        select(BUTTON, root).forEach((button) => {
+          const expanded = button.getAttribute(EXPANDED) === "true";
+          toggleButton(button, expanded);
+        });
+
+        this.hasInit = true;
+      }
+    },
+    teardown() {
+      this.hasInit = false;
     },
     ACCORDION,
     BUTTON,
@@ -93,6 +101,7 @@ const accordion = behavior(
     hide: hideButton,
     toggle: toggleButton,
     getButtons: getAccordionButtons,
+    hasInit: this.hasInit || false
   }
 );
 
--- a/packages/usa-banner/src/index.js
+++ b/packages/usa-banner/src/index.js
@@ -5,7 +5,6 @@ const accordion = require("../../usa-accordion/src/index");
 
 const HEADER = `.${PREFIX}-banner__header`;
 const EXPANDED_CLASS = `${PREFIX}-banner__header--expanded`;
-const BUTTON = `.${PREFIX}-banner__button[aria-controls]`;
 
 /**
  * Toggle expanded banner class.
@@ -14,7 +13,7 @@ const BUTTON = `.${PREFIX}-banner__button[aria-controls]`;
  */
 const toggleBanner = function toggleEl(event) {
   event.preventDefault();
-  this.closest(HEADER).classList.toggle(EXPANDED_CLASS);
+  event.target.closest(HEADER).classList.toggle(EXPANDED_CLASS);
 };
 
 const banner = behavior(
@@ -25,13 +24,18 @@ const banner = behavior(
   },
   {
     init(root) {
-      const bannerButtons = root.querySelectorAll(BUTTON);
-
-      bannerButtons.forEach((button) => {
-        const parentBanner = button.closest(accordion.ACCORDION);
-
-        accordion.on(parentBanner);
-      });
+      // Initialize accordion if it hasn't already.
+      // Required for modular import of Banner.
+      if (!accordion.hasInit) {
+        accordion.on(root);
+        accordion.hasInit = true;
+      }
+    },
+    teardown(root) {
+      if (accordion.hasInit) {
+        accordion.off(root);
+        accordion.hasInit = false;
+      }
     },
   }
 );
--- a/packages/usa-banner/src/index.js
+++ b/packages/usa-banner/src/index.js
@@ -1,17 +1,30 @@
 const behavior = require("../../uswds-core/src/js/utils/behavior");
 const { CLICK } = require("../../uswds-core/src/js/events");
 const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
+// const select = require("../../uswds-core/src/js/utils/select");
+const accordion = require("../../usa-accordion/src/index");
 
 const HEADER = `.${PREFIX}-banner__header`;
 const EXPANDED_CLASS = `${PREFIX}-banner__header--expanded`;
+// const EXPANDED = "aria-expanded";
+// const BUTTON = `.${PREFIX}-banner__button[aria-controls]`;
 
 const toggleBanner = function toggleEl(event) {
   event.preventDefault();
   this.closest(HEADER).classList.toggle(EXPANDED_CLASS);
 };
 
-module.exports = behavior({
-  [CLICK]: {
-    [`${HEADER} [aria-controls]`]: toggleBanner,
+const banner = behavior(
+  {
+    [CLICK]: {
+      [`${HEADER} [aria-controls]`]: toggleBanner,
+    },
   },
-});
+  {
+    init(root) {
+      accordion.on(root);
+    }
+  }
+);
+
+module.exports = banner;
--- a/packages/usa-banner/src/index.js
+++ b/packages/usa-banner/src/index.js
@@ -1,14 +1,17 @@
 const behavior = require("../../uswds-core/src/js/utils/behavior");
 const { CLICK } = require("../../uswds-core/src/js/events");
 const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
-// const select = require("../../uswds-core/src/js/utils/select");
 const accordion = require("../../usa-accordion/src/index");
 
 const HEADER = `.${PREFIX}-banner__header`;
 const EXPANDED_CLASS = `${PREFIX}-banner__header--expanded`;
-// const EXPANDED = "aria-expanded";
-// const BUTTON = `.${PREFIX}-banner__button[aria-controls]`;
+const BUTTON = `.${PREFIX}-banner__button[aria-controls]`;
 
+/**
+ * Toggle expanded banner class.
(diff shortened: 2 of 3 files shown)


mahoneycm (approved):
Working as described and doesn't initiate the accordion! Thanks for making the test page 👍

packages/usa-banner/src/index.js
```diff
@@ -1,17 +1,39 @@
...
   },
-});
+  {
+    init(root) {
+      const bannerButtons = root.querySelectorAll(BUTTON);
+
+      bannerButtons.forEach((button) => {
+        const parentBanner = button.closest(accordion.ACCORDION);
+
+        accordion.on(parentBanner);
```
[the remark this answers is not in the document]
mejiaj: Hey @aduth, the initial solution just called `accordion.on()` on init and it was initializing accordions as well. That's been changed.

Right now it only initializes accordions in banners.

mejiaj:
Forgot to share, the test template was created with other pre-existing components and data. 

If you'd like to do the same in the future you can see how I did that here:
https://github.com/uswds/uswds/blob/36d31a1b6b846aa682b7ee938b9d4ca0cfaf8636/packages/usa-banner/src/test/test-patterns/test-usa-banner-init.twig

amyleadem (approved):
Looks good to me. I successfully performed the test described in the PR description.
I really appreciate the well thought-out test page and guidance in this PR. It helped me better understand the problem you were solving. Thanks for the extra work there, @mejiaj!

packages/usa-banner/src/index.js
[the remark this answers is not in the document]
mejiaj: Sorry, misunderstood. Yes, it gets initialized again when accordion is imported. 

I've moved this PR to draft to rethink this approach.

packages/usa-banner/src/index.js
```diff
@@ -1,17 +1,43 @@
...
 
+/**
+ * Toggle expanded banner class.
+ *
+ * @param {*} event - The clicked banner button.
+ */
 const toggleBanner = function toggleEl(event) {
   event.preventDefault();
-  this.closest(HEADER).classList.toggle(EXPANDED_CLASS);
+  event.target.closest(HEADER).classList.toggle(EXPANDED_CLASS);
```
mejiaj, line 16: Using `event.target` instead of `this` to avoid potential scoping issues.

mejiaj:
Re-requesting review. Unfortunately couldn't separate banner from accordion (see PR notes). But now they can be imported individually without conflicts.

mahoneycm (approved):
Everything is working great! Tested by:
- Switching order of component initialization,
- Initializing only banner/accordion
- Running `test`

Everything worked as expected!

Should we consider making `hasInit` a pattern for dynamic components? In #5208 I run a conditional check to `usa-modal` for a class that is created during initialization, but if we wanted to be consistent I could use `hasInit` instead!

packages/usa-accordion/src/index.js
```diff
@@ -82,17 +82,26 @@ const accordion = behavior(
   },
   {
     init(root) {
-      select(BUTTON, root).forEach((button) => {
-        const expanded = button.getAttribute(EXPANDED) === "true";
-        toggleButton(button, expanded);
-      });
+      // Check if Banner has previously initialized accordion.
+      if (!this.hasInit) {
```
aduth, line 86: Is this something which could make sense to implement as common in the `behavior` function initialization for all components?

mejiaj: I'm conflicted on this and would prefer to limit this feature for now. Mainly to test and be able to have the ability to quickly change if needed.

That said, I'm also looking forward to the teams feedback. If it's something we should move forward with on all components, that's totally possible too.

mahoneycm:
@mejiaj  Apologies! Linked the wrong PR. It's #5208

mejiaj:
@mahoneycm this could help resolve that problem, but I'd like to understand first why the modal component is initializing multiple times. For example, with two modals it initializing four times instead of the expected two.

I hesitate to introduce this as a fix for what might be a framework specific issue in implementation.

amyleadem (approved):
I was able to successfully perform the tests outlined in the PR description.  Everything is working as expected.

packages/usa-accordion/src/index.js
thisisdano: I'm comfortable with leaving this scoped to these specific components for now. But I agree it could make sense to move into `behavior` as we identify the need.

thisisdano (approved):

amyleadem:
According to [Slack](https://gsa-tts.slack.com/archives/C050HRGN7/p1685986536170779?thread_ts=1685740500.304699&cid=C050HRGN7), this PR was reverted because of some JS issues when bundled. 

> Seems to be our bundler browserify is changing `this` to `void(0)` for some reason.

@mejiaj Does it make sense to re-open this effort?

PR #5248, opened 2023-04-20
[Example] Comparing al-input-mask to mm-input-mask

amyleadem:
⚠️ This PR should never be published. 
It exists to track the changes from `mm-input-mask`


amyleadem:
Closing because the comparison is not accurate. Suspecting that the confused commit history in the base branch is causing this.

PR #5263, opened 2023-04-28
USWDS - Tooltip: Fix handling of children icons

aduth:
# Summary

**Fixes issues when using icons inside a tooltip button.** Resolved errors resulting from moving the cursor over the icon, as well as improve compatibility with icon styling.

## Preview link

## Problem statement

This seeks to address several issues:

- Errors result when moving the cursor over content inside a tooltip trigger, such as an icon. You can demonstrate this by mousing over icons in the project's own [Tooltip Test story](http://localhost:6006/?path=/story/components-tooltip--test) ([see screenshot](https://user-images.githubusercontent.com/1779930/235170171-44bed292-9e46-4565-8980-047f50f98f71.png))
   - `Uncaught TypeError: Cannot read properties of null (reading 'classList')`
- The tooltip's SVG element selector styles (`.usa-tooltip__trigger > svg`) can introduce undesirable side-effects when applying tooltip behaviors to a button which includes both an icon and text (e.g. like in #4493, [see screenshot of effects](https://user-images.githubusercontent.com/1779930/235170756-dba27841-7f68-4f36-ac25-d5cc2b6f04a9.png))
- Using `mouseover` events with `behavior`/`receptor`'s event delegation implementation is likely to have a negative performance impact, since `mouseover` events are likely being handled thousands of times on a page for user interactions unrelated to tooltips
   - Example: Try entering `document.addEventListener('mouseover', () => console.log('mouseover'));` into your developer tools console on this page and mousing over the page
   - Similar to what's described in "Additional information" of #4256

## Solution

The proposed solution leverages event bubbling of targeted event handlers to address all of the above issues:

- When event handler is bound to the trigger element, `currentTarget` can be used as a guaranteed reference to the tooltip trigger, addressing the errors mentioned above
- Similarly, event bubbling removes the need for styles which remove `pointer-events` from an icon, which are assumed to have been needed since otherwise the icon would potentially block events meant for the trigger
- Binding the events to the trigger elements avoids unnecessary handling of noisy `mouseover` / `mouseout` events for elements on the page unrelated to the tooltip

## Testing and review

1. Confirm in the [Tooltip Test preview](http://localhost:6006/?path=/story/components-tooltip--test) that:
   1. No errors occur when mousing over any of the tooltips, including over icons in the footer links
   2. The tooltips appear as expected

### DIFF
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -307,6 +307,26 @@ const hideToolTip = (tooltipBody) => {
 };
 
 /**
+ * Handles events to activate tooltip.
+ * @param {MouseEvent|KeyboardEvent} event Mouseover or focusin event initiating activation.
+ */
+const onTooltipActivate = (event) => {
+  const { trigger, body } = getTooltipElements(event.currentTarget);
+
+  showToolTip(body, trigger, trigger.dataset.position);
+};
+
+/**
+ * Handles events to deactivate tooltip.
+ * @param {MouseEvent|KeyboardEvent} event Mouseout or focusout event initiating deactivation.
+ */
+const onTooltipDeactivate = (event) => {
+  const { body } = getTooltipElements(event.currentTarget);
+
+  hideToolTip(body);
+};
+
+/**
  * Setup the tooltip component
  * @param {HTMLElement} tooltipTrigger The element that creates the tooltip
  */
@@ -357,37 +377,25 @@ const setUpAttributes = (tooltipTrigger) => {
   return { tooltipBody, position, tooltipContent, wrapper };
 };
 
+/**
+ * Set up tooltip trigger event handlers.
+ * @param {HTMLElement} tooltipTrigger The element that creates the tooltip
+ */
+const setUpEvents = (tooltipTrigger) => {
+  tooltipTrigger.addEventListener("mouseover", onTooltipActivate);
+  tooltipTrigger.addEventListener("focusin", onTooltipActivate);
+  tooltipTrigger.addEventListener("mouseout", onTooltipDeactivate);
+  tooltipTrigger.addEventListener("focusout", onTooltipDeactivate);
+};
+
 // Setup our function to run on various events
 const tooltip = behavior(
-  {
-    "mouseover focusin": {
-      [TOOLTIP](e) {
-        const trigger = e.target;
-        const elementType = trigger.nodeName;
-
-        // Initialize tooltip if it hasn't already
-        if (elementType === "BUTTON" && trigger.hasAttribute("title")) {
-          setUpAttributes(trigger);
-        }
-      },
-      [TOOLTIP_TRIGGER](e) {
-        const { trigger, body } = getTooltipElements(e.target);
-
-        showToolTip(body, trigger, trigger.dataset.position);
-      },
-    },
-    "mouseout focusout": {
-      [TOOLTIP_TRIGGER](e) {
-        const { body } = getTooltipElements(e.target);
-
-        hideToolTip(body);
-      },
-    },
-  },
+  {},
   {
     init(root) {
       selectOrMatches(TOOLTIP, root).forEach((tooltipTrigger) => {
         setUpAttributes(tooltipTrigger);
+        setUpEvents(tooltipTrigger);
       });
     },
     setup: setUpAttributes,
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -5,7 +5,6 @@ const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
 const isElementInViewport = require("../../uswds-core/src/js/utils/is-in-viewport");
 
 const TOOLTIP = `.${PREFIX}-tooltip`;
-const TOOLTIP_TRIGGER = `.${PREFIX}-tooltip__trigger`;
 const TOOLTIP_TRIGGER_CLASS = `${PREFIX}-tooltip__trigger`;
 const TOOLTIP_CLASS = `${PREFIX}-tooltip`;
 const TOOLTIP_BODY_CLASS = `${PREFIX}-tooltip__body`;
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -329,7 +329,7 @@ const onTooltipDeactivate = (event) => {
  * Setup the tooltip component
  * @param {HTMLElement} tooltipTrigger The element that creates the tooltip
  */
-const setUpAttributes = (tooltipTrigger) => {
+const setup = (tooltipTrigger) => {
   const tooltipID = `tooltip-${Math.floor(Math.random() * 900000) + 100000}`;
   const tooltipContent = tooltipTrigger.getAttribute("title");
   const wrapper = document.createElement("span");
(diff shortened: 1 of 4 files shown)


packages/usa-tooltip/src/index.js
```diff
@@ -353,37 +372,25 @@ const setUpAttributes = (tooltipTrigger) => {
...
-  },
+  {},
   {
     init(root) {
       selectOrMatches(TOOLTIP, root).forEach((tooltipTrigger) => {
         setUpAttributes(tooltipTrigger);
+        setUpEvents(tooltipTrigger);
       });
     },
     setup: setUpAttributes,
```
mahoneycm, line 396: Since we're exporting `setUpAttributes`, I think it would make sense to export `setUpEvents` as well.

mahoneycm (changes requested):
This works like a charm! The updated JS is more readable in my opinion and all three bugs in the problem statement are resolved. Great work!

Just one small suggested change to export the `setUpEvents` method.

Thanks @aduth !

- [x]  Tooltip displays as expected
- [x]  Errors are no longer occurring when button icons are hovered
- [x]  Unit tests work appropriately
- [x]  JS is improved by this pr

packages/usa-tooltip/src/index.js
aduth: > Since we're exporting `setUpAttributes`, I think it would make sense to export `setUpEvents` as well.

Are you thinking `setup` would call both `setUpAttributes` and `setUpEvents`, or a separate property for just `setUpEvents` ? It's not clear to me how the `setup` public API is meant to be used.

I'm thinking maybe we could do something like this? What do you think?

```ts
function setup(tooltipTrigger) {
  setUpAttributes(tooltipTrigger);
  setUpEvents(tooltipTrigger);
}

// ...

const tooltip = behavior(
  {
    init(root) {
      selectOrMatches(TOOLTIP, root).forEach(setup);
    },
    setup,
    // ...
  },
);
```

aduth: I couldn't recall the original motivation for creating a separate "setUpEvents" function aside from the fact that these event handlers aren't "attributes" as such. But based on your comment, seeing how `setup` should fully set up the component, I decided to just move the event binding into the existing function and rename it from `setUpAttributes` to just `setup`. I think that should simplify things overall and ensure that "setup" does all of the setup pieces.

packages/usa-tooltip/src/index.js
```diff
@@ -373,31 +373,22 @@ const setUpAttributes = (tooltipTrigger) => {
...
   {
     init(root) {
-      selectOrMatches(TOOLTIP, root).forEach((tooltipTrigger) => {
-        setUpAttributes(tooltipTrigger);
-        setUpEvents(tooltipTrigger);
-      });
+      selectOrMatches(TOOLTIP, root).forEach(setup);
     },
-    setup: setUpAttributes,
+    setup,
```
mahoneycm, line 391: ## Setup pattern

It appears that tooltip is the only component that follows the pattern of renaming and exporting a `setup` pattern. 

Initially, I was just referring to exporting both `setUpAttributes` and `setUpEvents` as methods.

```diff
const tooltip = behavior(
  {},
  {
    init(root) {
      selectOrMatches(TOOLTIP, root).forEach((tooltipTrigger) => {
        setUpAttributes(tooltipTrigger);
        setUpEvents(tooltipTrigger);
      });
    },
+   setUpAttributes,
+   setUpEvents
    getTooltipElements,
    show: showToolTip,
    hide: hideToolTip,
  }
);
```

We could instead follow the pattern set by Modal:

It uses a `setUpModal` function to handle all setup tasks

https://github.com/uswds/uswds/blob/3175c0258a63e6e464ea8284bb90f452f46463a1/packages/usa-modal/src/index.js#L286-L304

Then it calls that as part of its `init`.

We could follow this pattern by maintaining each function and calling them in one `setupTooltip` function.

```js
const setUpAttributes = (tooltipTrigger) => {
   ...
}

const setUpEvents = (tooltipTrigger) => {
  ...
}

const setupTooltip = (tooltipTrigger) => {
  setUpAttributes(tooltipTrigger);
  setUpEvents(tooltipTrigger);
}
```

What do you think @amyleadem @mejiaj ? Would it be a regression to remove the `setup` method? If so, we could export it as `setup: setUpTooltip`

mejiaj: @mahoneycm this is a public method, so it would be a breaking change if users rely on it and we remove it.

Like you mentioned, there are alternatives like:
1. Keeping the method, but calling a different function.
2. Providing an alternative method and guidance for migrating.

I do agree this might not follow a conventional pattern, but that seems out-of-scope for the work outlined in the problem statement. If you agree can you create a separate issue to address this?

mejiaj (changes requested):
Thanks for the PR. Tooltip **shouldn't** break if there's an icon inside.

Some questions/requests:
1. Support icons in tooltips without removing behaviors?
2. Are you seeing a performance impact specifically with tooltip and its use of `mouseover`?
3. Can you add an example of an icon in tooltip in the test template?

aduth:
> * Support icons in tooltips without removing behaviors?

Can you clarify what you mean by this?

As part of adding a new Storybook section in 2e91d66, I noticed that at least the `pointer-events: none` could be good to keep to avoid some flickering effect with mousing over icons in the button trigger.

> * Are you seeing a performance impact specifically with tooltip and its use of `mouseover`?

I haven't actually measured the impact, and the changes to event binding are based on an understanding of how the event delegation is applied, and wanting to avoid unnecessary work when mousing over unrelated elements on the page.

I could back out these changes if there's a desire to keep this pull request simpler or to avoid explicit event binding in favor of the more conventional `behavior`-based event delegation.

> * Can you add an example of an icon in tooltip in the test template?

I wasn't sure if you're referring to an icon within the tooltip itself, or an icon in a tooltip trigger button. The latter is what this pull request is trying to address. The [existing test template](http://localhost:6006/?path=/story/components-tooltip--test) footer example includes part of the issue here when a button contains only an icon. In 2e91d66 I also pushed an example where there's both a button and some text. This is improved slightly in this pull request, and would hopefully be fully improved with changes in #5398.

Before|After
---|---
Screenshot 2024-02-07 at 10 27 54 AM|Screenshot 2024-02-07 at 10 29 58 AM

sperro:
I also see the error Uncaught TypeError: Cannot read properties of null (reading 'classList') when using my keyboard to tab to the tooltip even if it just a regular button with no icon.  Will this address that issue with keyboard tabbing as well?

sperro:
@aduth I noticed the issue prior to 3.8.1 release. I just tested in 3.8.1 and it actually seems worse. Previously the error was only seen while tabbing to set focus via keyboard but I also saw it now with mouseover. Its the same error you posted in your Problem statement so I think its related. 

> Errors result when moving the cursor over content inside a tooltip trigger, such as an icon. You can demonstrate this by mousing over icons in the project's own [Tooltip Test story](http://localhost:6006/?path=/story/components-tooltip--test) ([see screenshot](https://user-images.githubusercontent.com/1779930/235170171-44bed292-9e46-4565-8980-047f50f98f71.png))
Uncaught TypeError: Cannot read properties of null (reading 'classList')

aduth:
Another workaround, and one I'm considering for my use-case: If you need children elements but they're not meant to be interacted with (e.g. an icon), you can apply `pointer-events: none` CSS styling to the element, similar to what USWDS already does for specific element types:

https://github.com/uswds/uswds/blob/369cf3fdf9abf721a357e864a0f6adbf667ba862/packages/usa-tooltip/src/styles/_usa-tooltip.scss#L45-L48

sperro:
I was able to resolve the issues I was seeing by re-initializing the tooltip component after my page loaded. I have an angular application. I added import for  
`import * as tooltip from '@uswds/uswds/packages/usa-tooltip/src/index';
`
And in my lifecycle function triggered after my component is initialized I called 
`tooltip.on();`

ryparker:
@coderabbitai full review

PR #5264, opened 2023-04-28
USWDS - Tooltip: Fix negative margin offset calculations

aduth:
# Summary

**Improves tooltip positioning.** Fixes issues which can occur in some cases with tooltip button positioning.

## Preview link

## Problem statement

The Tooltip component's logic for calculating offsets incorrectly assumes that offsets will always be positive numbers, resulting in invalid style assignments when the offsets are negative.

This causes the tooltips to not appear flush with the trigger.

Screenshot:

image

Example:

```js
// Ok! 😀
const topMargin = 10;
const leftMargin = 10;
e.style.margin = `-${topMargin}px 0 0 -${leftMargin / 2}px`;
// "-10px 0 0 -5px"

// Not Ok! 😭
const topMargin = -10;
const leftMargin = -10;
e.style.margin = `-${topMargin}px 0 0 -${leftMargin / 2}px`;
// "--10px 0 0 --5px"
```

## Solution

The solution is to move the negation to the interpolated value, not the string template.

```js
// Ok! 😀
const topMargin = 10;
const leftMargin = 10;
e.style.margin = `${-topMargin}px 0 0 ${leftMargin / -2}px`;
// "-10px 0 0 -5px"

// Still Ok! 😌
const topMargin = -10;
const leftMargin = -10;
e.style.margin = `${-topMargin}px 0 0 ${leftMargin / -2}px`;
// "10px 0 0 5px"
```

## Testing and review

Check that the newly-added examples to [the Tooltip Test story](http://localhost:6006/?path=/story/components-tooltip--test) all show the tooltip flush with the button.

### DIFF
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -142,6 +142,12 @@ const showToolTip = (tooltipBody, tooltipTrigger, position) => {
   const positionBottom = (e) => {
     resetPositionStyles(e);
 
+    const bottomMargin = calculateMarginOffset(
+      "bottom",
+      e.offsetHeight,
+      tooltipTrigger
+    );
+
     const leftMargin = calculateMarginOffset(
       "left",
       e.offsetWidth,
@@ -149,8 +155,10 @@ const showToolTip = (tooltipBody, tooltipTrigger, position) => {
     );
 
     setPositionClass("bottom");
-    e.style.left = `50%`;
-    e.style.margin = `${TRIANGLE_SIZE}px 0 0 ${-leftMargin / 2}px`;
+    e.style.left = `50%`; // center the element
+    e.style.bottom = `-${TRIANGLE_SIZE}px`; // consider the pseudo element
+    // apply our margins based on the offset
+    e.style.margin = `0 0 ${-bottomMargin}px ${-leftMargin / 2}px`;
   };
 
   /**
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -132,7 +132,7 @@ const showToolTip = (tooltipBody, tooltipTrigger, position) => {
     e.style.left = `50%`; // center the element
     e.style.top = `-${TRIANGLE_SIZE}px`; // consider the pseudo element
     // apply our margins based on the offset
-    e.style.margin = `-${topMargin}px 0 0 -${leftMargin / 2}px`;
+    e.style.margin = `${-topMargin}px 0 0 ${-leftMargin / 2}px`;
   };
 
   /**
@@ -150,7 +150,7 @@ const showToolTip = (tooltipBody, tooltipTrigger, position) => {
 
     setPositionClass("bottom");
     e.style.left = `50%`;
-    e.style.margin = `${TRIANGLE_SIZE}px 0 0 -${leftMargin / 2}px`;
+    e.style.margin = `${TRIANGLE_SIZE}px 0 0 ${-leftMargin / 2}px`;
   };
 
   /**
@@ -171,7 +171,7 @@ const showToolTip = (tooltipBody, tooltipTrigger, position) => {
     e.style.left = `${
       tooltipTrigger.offsetLeft + tooltipTrigger.offsetWidth + TRIANGLE_SIZE
     }px`;
-    e.style.margin = `-${topMargin / 2}px 0 0 0`;
+    e.style.margin = `${topMargin / -2}px 0 0 0`;
   };
 
   /**
@@ -199,7 +199,7 @@ const showToolTip = (tooltipBody, tooltipTrigger, position) => {
     setPositionClass("left");
     e.style.top = `50%`;
     e.style.left = `-${TRIANGLE_SIZE}px`;
-    e.style.margin = `-${topMargin / 2}px 0 0 ${
+    e.style.margin = `${-topMargin / 2}px 0 0 ${
       tooltipTrigger.offsetLeft > e.offsetWidth ? leftMargin : -leftMargin
     }px`; // adjust the margin
   };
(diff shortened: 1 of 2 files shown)


packages/usa-tooltip/src/test/test-patterns/test-usa-tooltip-utilities.twig
```diff
@@ -116,3 +116,15 @@
...
+<p>In this scenario:</p>
+<ul>
+  <li>The tooltip should be positioned accounting for any offsets that might apply from the trigger</li>
+</ul>
+<button type="button" class="{{ tooltip.classes }} margin-top-1" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Show on top</button>
+<button type="button" class="{{ tooltip.classes }} margin-top-2" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Show on top</button>
+<button type="button" class="{{ tooltip.classes }} margin-top-3" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Show on top</button>
+<button type="button" class="{{ tooltip.classes }} margin-top-4" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Show on top</button>
+<button type="button" class="{{ tooltip.classes }} margin-top-5" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Show on top</button>
+<button type="button" class="{{ tooltip.classes }} margin-top-6" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Show on top</button>
```
mejiaj, line 130: For simpler testing/review you could use the utility name as the text. _It wasn't immediately clear to me until I looked at the utils what was supposed to happen._

```suggestion
<button type="button" class="{{ tooltip.classes }} margin-top-1" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Margin top 1</button>
<button type="button" class="{{ tooltip.classes }} margin-top-2" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Margin top 2</button>
<button type="button" class="{{ tooltip.classes }} margin-top-3" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Margin top 3</button>
<button type="button" class="{{ tooltip.classes }} margin-top-4" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Margin top 4</button>
<button type="button" class="{{ tooltip.classes }} margin-top-5" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Margin top 5</button>
<button type="button" class="{{ tooltip.classes }} margin-top-6" data-position="top" data-classes="{{ tooltip.util_classes }}" title="Top">Margin top 6</button>
```

mejiaj (approved):
Looks good, added suggestion for simplifying testing.

image

mahoneycm (commented):
This is looking good so far! Top tooltip placement has definitely improved as a result of this work!

## Left and Bottom offsets

`top` and `right` tooltips seem to be working as expected but I noticed `left` and `bottom` tooltips are still showing offset bugs to me.

Screenshot 2023-11-27 at 3 10 30 PM
Screenshot 2023-11-27 at 2 58 24 PM

Can we try to capture this work in this PR as well? Do you have the bandwidth to include this improvement? 

> [!IMPORTANT]
> This is NOT a regression from this work. This occurs in `develop` as well.

packages/usa-tooltip/src/index.js
```diff
@@ -171,7 +171,7 @@ const showToolTip = (tooltipBody, tooltipTrigger, position) => {
     e.style.left = `${
       tooltipTrigger.offsetLeft + tooltipTrigger.offsetWidth + TRIANGLE_SIZE
     }px`;
-    e.style.margin = `-${topMargin / 2}px 0 0 0`;
+    e.style.margin = `${topMargin / -2}px 0 0 0`;
```
mahoneycm, line 174: ## OPTIONAL - Negative denominator

Should we move the negative number back to the numerator for consistency with the other calculations?

```suggestion
    e.style.margin = `${-topMargin / 2}px 0 0 0`;
```

mejiaj: @mahoneycm if it's inconsistent I wouldn't consider the request optional. Unless I'm misunderstanding the before/after.

mejiaj (changes requested):
Based on @mahoneycm [findings](https://github.com/uswds/uswds/pull/5264#pullrequestreview-1751305104) it makes sense that this should work for all sides.

packages/usa-tooltip/src/index.js
mahoneycm: Updated my request to remove "optional"

aduth:
Thanks for the reviews, @mahoneycm and @mejiaj . I've been able to circle back to this work and address your comments.

>top and right tooltips seem to be working as expected but I noticed left and bottom tooltips are still showing offset bugs to me.

@mahoneycm I wasn't able to reproduce the issue with the left margin, but I was able to reproduce the problem with the bottom margin. This should be fixed in e0b7d23, where I mostly mirrored the logic in `positionTop` to apply to `positionBottom`.

I also added some new examples in ddc53f0 to show offsets in all directions, which should hopefully simplify testing.

packages/usa-tooltip/src/test/test-patterns/test-usa-tooltip-utilities.twig
aduth: > For simpler testing/review you could use the utility name as the text. _It wasn't immediately clear to me until I looked at the utils what was supposed to happen._

I incorporated this into the new examples in ddc53f0.

packages/usa-tooltip/src/index.js
aduth: > We should move the negative number back to the numerator for consistency with the other calculations.

I updated this in 2596c38.

ryparker:
@coderabbitai full review

PR #5274, opened 2023-05-05
USWDS - Tooltip: Fix tooltip component impacting content layout

### LINKED ISSUE #5273: USWDS - Bug: Tooltip affects content layout
### Describe the bug

When applying a tooltip to a button, if that button has its own layout styles (e.g. full-width), the tooltip will unexpectedly impact those layout styles.

1. Add `full-width` utility class to a button which has `.usa-tooltip`
2. Observe that the button does not display at full width

### Expected Behavior

The button layout styles are not impacted by the existence of a tooltip

### Related code

Sample code:

```html
<button class="usa-button usa-tooltip width-full" data-position="top" title="Tooltip">
  Full-Width Button with Tooltip
</button>
```

Demo: https://codepen.io/aduth/pen/bGmaLRG 

### Screenshots

Via https://codepen.io/aduth/pen/bGmaLRG demo playground:

image

### System setup

- USWDS v3.4.1
- Chrome 113
- macOS 12.6.5

### Additional context

The root problem is that the Tooltip component JavaScript wraps the button in a span element with `display: inline-block` styling, which can conflict with the original button's own styling, if the button is expected full-width or block display layout.

https://github.com/uswds/uswds/blob/29eebc4abe3d9c9d968bd3ef54cff9e8add449eb/packages/usa-tooltip/src/styles/_usa-tooltip.scss#L8

The [demo linked above](https://codepen.io/aduth/pen/bGmaLRG) experimented with a few potential solutions using `display: contents` or `display: inline`, but each has its own issues:

- `display: contents` does not work with the expected behavior of `position: relative` to contain the tooltip's absolute positioning
- `display: inline` (a `span`'s default `display`) causes the tooltip to show slightly offset from the bounds of the button

It may be worth rethinking how the tooltip is positioned:

- Either avoiding the wrapping element altogether and computing the absolute positioning based on whatever closest bounding container exists for the button
- Or using the button itself as the relative container and moving the tooltip contents within the button

### Code of Conduct

aduth:
# Summary

**Improve tooltip compatibility with different content layouts.** Tooltips should no longer affect the layout appearance of its content, such as full-width content.

Fixes #5273

## Related pull requests

This incorporates the changes from #5263, which would ideally be merged separately.

## Preview link

## Problem statement

See #5273

## Solution

This implements the second of the two "rethinking how the tooltip is positioned" suggestions mentioned in the "Additional context" of #5273:

>Or using the button itself as the relative container and moving the tooltip contents within the button

## Testing and review

Test in local preview: http://localhost:6006/?path=/story/components-tooltip--test

Verify no regressions in:

- Different positions
- Constrained viewport width where not enough room for horizontal tooltip ([see "Show on Right" example in Test Tooltip preview](http://localhost:6006/?path=/story/components-tooltip--test))

Additionally, confirm the new support for full-width buttons on the [Test Tooltip preview](http://localhost:6006/?path=/story/components-tooltip--test).

### DIFF
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -5,7 +5,6 @@ const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
 const isElementInViewport = require("../../uswds-core/src/js/utils/is-in-viewport");
 
 const TOOLTIP = `.${PREFIX}-tooltip`;
-const TOOLTIP_TRIGGER = `.${PREFIX}-tooltip__trigger`;
 const TOOLTIP_TRIGGER_CLASS = `${PREFIX}-tooltip__trigger`;
 const TOOLTIP_CLASS = `${PREFIX}-tooltip`;
 const TOOLTIP_BODY_CLASS = `${PREFIX}-tooltip__body`;
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -307,6 +307,26 @@ const hideToolTip = (tooltipBody) => {
 };
 
 /**
+ * Handles events to activate tooltip.
+ * @param {MouseEvent|KeyboardEvent} event Mouseover or focusin event initiating activation.
+ */
+const onTooltipActivate = (event) => {
+  const { trigger, body } = getTooltipElements(event.currentTarget);
+
+  showToolTip(body, trigger, trigger.dataset.position);
+};
+
+/**
+ * Handles events to deactivate tooltip.
+ * @param {MouseEvent|KeyboardEvent} event Mouseout or focusout event initiating deactivation.
+ */
+const onTooltipDeactivate = (event) => {
+  const { body } = getTooltipElements(event.currentTarget);
+
+  hideToolTip(body);
+};
+
+/**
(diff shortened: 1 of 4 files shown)


ryparker:
@coderabbitai full review

ryparker:
The full-width tooltip fix in `ee887755` changes markup that existing styles and scripts rely on: `.usa-tooltip__trigger` and the wrapper structure disappear, and the tooltip body moves inside the trigger. That's why this has the breaking-change label.

Two other cases need attention. [Custom `data-classes` are added to a wrapper](https://github.com/uswds/uswds/blob/ee887755c8aaf7630dd27badcb4b35d1f14aff97/packages/usa-tooltip/src/index.js#L344) that never enters the document. With two triggers, the second also finds the first tooltip's body because [lookup still searches their parent](https://github.com/uswds/uswds/blob/ee887755c8aaf7630dd27badcb4b35d1f14aff97/packages/usa-tooltip/src/index.js#L19).

Can we fix the layout while keeping the existing classes and wrapper, applying `data-classes` to the attached element? If the markup needs to change, existing integrations still need to work and each trigger needs to find its own tooltip before browser and screen-reader review.

PR #5324, opened 2023-06-07
USWDS - Button Group: Improve appearance of segmented button text wrapping

aduth:
# Summary

**Improve appearance of segmented button text wrapping.** When segmented button group text wraps, the height of all buttons will now update to match.

## Preview link

Local storybook link: http://localhost:6006/?path=/story/components-button-group--test-text-wrapping

## Problem statement

Currently, if a segmented Button Group includes buttons with text long enough to wrap, only the button with the longer text will wrap, which avoids the segmented group from appearing as a single grouped collection.

## Solution

Add flex and grid styling to force buttons to grow together and align text as vertically centered.

Before|After
---|---
localhost_6006__path=_story_components-button-group--test-text-wrapping|localhost_6006__path=_story_components-button-group--segmented

### DIFF
--- a/packages/usa-button-group/src/usa-button-group--test-text-wrapping.twig
+++ b/packages/usa-button-group/src/usa-button-group--test-text-wrapping.twig
@@ -1,0 +1,11 @@
+<ul class="usa-button-group usa-button-group--segmented">
+  <li class="usa-button-group__item">
+    <button type="button" class="usa-button">Button</button>
+  </li>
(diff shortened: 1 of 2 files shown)


aduth:
Hey @mahoneycm , apologies for the delayed response. Just wanted to drop a note that I've seen your feedback, but have been rather busy to take a look. I'll plan to try to find a fix for large viewport styling when time permits. (Or I'm also totally okay with pushes directly to my branch!)

Issue #5322, opened 2023-06-05
USWDS - Bug: Bundling JS causes init issues for Banner/Accordion


mejiaj:
On bundling, Browserify adds `void(0)` to the `hasInit` flag. This flag checks if Accordion has already init, either Banner or Accordion.

Work originally done in [USWDS - Banner: Add accordion dependency #5242](https://github.com/uswds/uswds/pull/5242).

### Expected

Components should work individually (regardless of order) and bundled in `uswds.js` and `uswds.min.js`.

- [ ] Importing `uswds.js` components should init without issues.
- [ ] Importing Banner JS should init accordion.
- [ ] Importing Accordion should init accordion (including accordion in Banner).
- [ ] Mixing order of Banner & Accordion JS should not cause conflicts.

### Preview

- [Error on site](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/dw-changelog-button-outline-whc/components/accordion/). Open up devtools and confirm `hasInit` is undefined.

mejiaj:
Closing in favor of [#5179](https://github.com/uswds/uswds/issues/5179). PR was reverted and issue reopened.

PR #5326, opened 2023-06-07
Revert "Merge pull request #5242 from uswds/jm-banner-init"

mejiaj:
This reverts commit 10ef49f8d8cae3ca20c818371a14f19ac291e6d0. 

Work was reverted to look into:
- Issues when bundling.
- Event listener conflicts.
- Complications when using `uswds-init`.

Work will continue in #5322.

### DIFF
--- a/packages/usa-accordion/src/index.js
+++ b/packages/usa-accordion/src/index.js
@@ -82,18 +82,10 @@ const accordion = behavior(
   },
   {
     init(root) {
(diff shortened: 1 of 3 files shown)


thisisdano:
I'm re-opening https://github.com/uswds/uswds/issues/5179 and assigning it to `3.6.0` for now so we can continue to track it

PR #5344, opened 2023-06-21
USWDS - Fieldset: Resolve legend margin error

### LINKED ISSUE #4153: USWDS - Bug: Vertical spacing doesn't work for some form input components
### Describe the bug
Input components that use `legend` instead of `label` do not have the expected vertical spacing between them. Instead, they are rendered with no top margin.
This impacts Radio Buttons, Date Inputs, and other components that utilize fieldsets.

1. Open Chrome
2. Go to https://designsystem.digital.gov/components/radio-buttons/
3. "Inspect" the page
4. Duplicate the fieldset element in the component preview
5. Notice the lack of space between the legend and the last radio button

### Expected behavior
There should be margin of 3 units between the top of a form control and the previous element.

### Screenshots
**Current Behavior:**
Screen Shot 2021-04-14 at 1 47 24 PM
**Expected Behavior:**
Screen Shot 2021-04-14 at 2 24 45 PM

### System setup
- **USWDS version:** 2.11.1
- **Device:** 
- **Operating system:** MacOS 10.14.6
- **Browser and version:** Chrome 89.0.4389.114

### Additional context
I believe the source of the issue is a webkit issue that prevents margins from rendering properly for legend elements.

Looking at the [HTML specification](https://html.spec.whatwg.org/multipage/rendering.html#the-fieldset-and-legend-elements) I believe the styles as they currently exist _should_ work correctly, so I believe this is a browser compatibility issue.

See this old post describing others encountering the same issue: https://stackoverflow.com/questions/2568591/cannot-add-margin-to-legend-element-in-safari-chrome-webkit?noredirect=1&lq=1
And the tracking ticket for chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=895846&q=component%3ABlink%3EForms%3EFieldset&can=2

This should be resolvable by moving the top margin from the legend to the fieldset, but that might have other unintended consequences.

mahoneycm:
# Summary

Resolved `usa-fieldset` & `usa-legend` margin conflict. `usa-fieldset` now appropriately respects the margin set by child legend elements.

## Breaking change

This is _potentially_ a breaking change.

> **Warning**
> In Chrome and Safari, anywhere there is `usa-legend` nested beneath `usa-fieldset`, a larger margin is rendered than before. This comes from the margin set by `usa-legend` which was hidden by this bug.

## Related issue

Closes #4153 

## Related pull requests

[Form change log item →](https://github.com/uswds/uswds-site/pull/2192)

## Preview link
[Checkbox →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=components-form-inputs-checkbox--default)
[Mailing Address →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=patterns-forms--mailing-address)
[Create Account Page →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=pages-create-account--create-account-page)

## Problem statement

`fieldset` does not respect margins set by `usa-legend` or `usa-legend--large`.

## Solution

Add `display: contents` to `usa-fieldset` styles. 

> [display: contents](https://drafts.csswg.org/css-display/#propdef-display) simply removes their [principal box](https://drafts.csswg.org/css-display/#principal-box), and their contents render as normal. [^1]

[^1]: https://drafts.csswg.org/css-display/#unbox:~:text=display%3A%20contents%20simply%20removes%20their%20principal%20box%2C%20and%20their%20contents%20render%20as%20normal.

## Testing and review
1. Inspect components which use `usa-fieldset` and `usa-legend`. Ex:
   - [Checkbox →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=components-form-inputs-checkbox--default)
   - [Mailing Address →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=patterns-forms--mailing-address)
   - [Create Account Page →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=pages-create-account--create-account-page)
2. Confirm the margin set by `usa-legend` or `usa-legend--large` is respected and visable.

It may be helpful to duplicate the `fieldset` element in the preview and make sure there is a visible margin between them

### Testing Checklist
- [ ] Confirm margin is respected
- [ ] Removing `display: contents` from `usa-fieldset` causes margin to be disregarded
- [ ] Confirm there is no unintended visual regression in components using `usa-fieldset`

### Screenshots from issue
**Current Behavior:**
Screen Shot 2021-04-14 at 1 47 24 PM
**Expected Behavior:**
Screen Shot 2021-04-14 at 2 24 45 PM

---


amyleadem (changes requested):
@mahoneycm This is a nice simple fix and it looks like this solution works as expected in Chrome and Safari. I noticed that Firefox is mostly unchanged, but I believe that is because it was already respecting the margin. 

I caught some visual changes that made me think that we should consider this a visually breaking change. Any `usa-fieldset` with a `usa-legend` element will now have a different top margin in Chrome and Safari. It looks like Firefox was already respecting the margin so the only changes I expect there are if an additional top margin is added via utility classes, etc (see an example in the notes below). 

We should think through the ramifications of this change and flag this as a breaking change if needed. If it is, we should move this to the 3.7.0 release and provide a helpful description for users so they can know what to look for and how to fix the changes.

Change requested:
- In the [form story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/patterns-forms--disabled-form-elements) the fieldsets have an added `margin-top-2` utility class. Could you remove those? They will not be needed any longer with this PR and it will make for easier testing. 

Some browser notes:

- In Chrome, Safari, and Firefox the elements are now evenly spaced vertically in the form story ([Before](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/patterns-forms--disabled-form-elements&args=disabled_state:none) | [After](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/?path=/story/patterns-forms--disabled-form-elements&args=disabled_state:none)). This is expected. I just wanted to note that the reasons they are evenly spaced are a bit different:
- In Chrome and Safari, the margins grew:

    image
- In Firefox, the margins shrank when the `margin-top-2` utility class was present:
       
    image
- And there was no change in Firefox when the margin utility class was removed:
    
    image
- In Chrome and Safari, I noticed that the mailing address example, the fieldset shifts down ([Before](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=patterns-forms--mailing-address&viewMode=story) | [After](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=patterns-forms--mailing-address&viewMode=story)). This also happens in the create account page ([Before](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=pages-create-account--create-account-page&viewMode=story) | [After](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=pages-create-account--create-account-page&viewMode=story)). 
- In Chrome, I also see that the component preview on our site now has slightly larger vertical margins than before.

    image


I’ll keep poking around at individual components and will report back if I see anything else.

hobbez:
While testing the proposed solution of .usa-fieldset{display:contents;}, I came across an odd behavior – using the fieldset id as a URL anchor does not scroll the page to the fieldset. Without the additional CSS rule, the scroll works fine.

hobbez:
As workaround to the anchor issue, setting an ID for the fieldset's legend tag and using that as the anchor instead works fine.

mahoneycm:
@amyleadem `margin-top-2` removed and pushed up!

I apologize, I should have noted that this change would affect anywhere there is `usa-legend` nested within `usa-fieldset`! That was why I added the [mailing address](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=patterns-forms--mailing-address) and [create account](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fieldset-legend-margin/iframe.html?args=&id=pages-create-account--create-account-page) pages as examples. 

Good to note that this is considered a potentially breaking change. I've updated the PR description with both of those notes!

packages/usa-fieldset/src/styles/_usa-fieldset.scss
```diff
@@ -3,6 +3,7 @@
 .usa-fieldset {
   @extend %block-input-general;
   border: none;
+  display: contents;
   margin: 0;
   padding: 0;
```
amyleadem, line 8: Applying `display:contents` to the fieldset element appears to cause all other fieldset styles to be ignored. This can be ok on a base level because the USWDS styles are basically declaring a reset to 0. However, I get concerned that this will cause confusion with users who apply utility classes or styles to the fieldset element. It also looks like anchor ids aren't recognized. 

Wanted to share this explanation for `display:contents` because I found the explanation helpful: 

> display: contents causes an element's children to appear as if they were direct children of the element's parent, ignoring the element itself. This can be useful when a wrapper element should be ignored when using CSS grid or similar layout techniques. [^1]

I do think that prioritizing `usa-legend` styles over `usa-fieldset` make sense because they are more substantive. I just wanted to check if this was the only path forward. If so, we should find a way to document this very clearly. 

[^1]: https://caniuse.com/css-display-contents

packages/usa-form/src/test/test-pattern/test-usa-form.twig
```diff
@@ -50,7 +50,7 @@
 </fieldset>
 
 
-<fieldset class="usa-fieldset margin-top-2">
+<fieldset class="usa-fieldset">
```
amyleadem, line 53: @mahoneycm Can you also remove the margin on the checkbox component on this page?

amyleadem (changes requested):

packages/usa-fieldset/src/styles/_usa-fieldset.scss
mahoneycm: One issue with moving the margin to `usa-fieldset` is that we offer two different legend options; `usa-legend` and `usa-legend--large`, which features larger text and a smaller margin. 

https://github.com/uswds/uswds/blob/e8f22e087f7b91364adb24f4385ed206385655c3/packages/usa-legend/src/styles/_usa-legend.scss#L1-L19

I wonder if we could take a similar approach to prose and and set the margin using something like

```scss
.usa-fieldset {
   ...
   * + .usa-legend {
      margin-top: units(3);
   }
    
    * + .usa-legend--large {
       margin-top: units(2);
   }
}
```

What do you think of that?

mahoneycm: Ooh sorry I forgot there are limitations selecting a parent based on children. I'll look into using `has()` or other newer selectors to see if it's possible to set the margin on fieldset based on it's child element but we may be stuck there

mahoneycm: Found my old notes on this issue and `postcss` rendering breaks when we attempt to target classes using `:has()`.

Alternatively, we could consider creating classes for `fieldset` to handle the different legend sizes, but that would be a breaking change for users as well. 

To me, this was the path of least resistance for users to utlize the margins set via `usa-legend` classes.

packages/usa-form/src/test/test-pattern/test-usa-form.twig
mahoneycm: Resolved in [06d1cfa](https://github.com/uswds/uswds/pull/5344/commits/06d1cfa7c7501b67bcf5e25234f4c25040011c31)

packages/usa-fieldset/src/styles/_usa-fieldset.scss
mahoneycm: As far as documenting this change, do you thing we need to add guidance to our forms page or form elements that use fieldset such as checkbox / radio that users should be mindful when adding styles to fieldset/legend?

mejiaj: > Ooh sorry I forgot there are limitations selecting a parent based on children. I'll look into using `has()` or other newer selectors to see if it's possible to set the margin on fieldset based on it's child element but we may be stuck there

This quirk is limited (or should be) to SASS mixins/functions.

mejiaj: @amyleadem @mahoneycm caniuse lists a resource [^1] for partial support. It has an important accessibility warning:

>The display property has no effect on an element’s semantics: these are defined by the document language and are not affected by CSS. Aside from the none value, which also affects the aural/speech output and interactivity of an element and its descendants, the display property only affects visual layout: its purpose is to allow designers freedom to change the layout behavior of an element without affecting the underlying document semantics.
>
> Looking at our sponsor list example, it means that the item is no longer seen as a list, but as something else ([test case in CodePen](https://codepen.io/hidde/pen/gzbMeL)).

It _seems_ mostly resolved, but safari [^2] still has issues.

[^1]: https://hidde.blog/more-accessible-markup-with-display-contents/#heading-4
[^2]: https://adrianroselli.com/2022/07/its-mid-2022-and-browsers-mostly-safari-still-break-accessibility-via-display-properties.html

mahoneycm:
## Accessibility concerns

After much consideration, we've decided to hold off on this bug fix for the time being.

[Caniuse](https://caniuse.com/css-display-contents#:~:text=Feedback-,1,Partial,-support%20refers%20to) reports the proposed solution as only partially supported by browsers due to accessibility concerns. Adding `display: contents` appears to have a history of causing browsers to misinterpret the `role` attribute. Though some of these concerns have been addressed by the web browsers, it appears there are still be bugs surrounding the use of the display value.

Since this is a result of a browser bug, which can be worked around by adding margin to the fieldset, we're closing this PR for further discussion.

> **Note**
> In my voiceover testing, I did not personally run into issues with changing the `display` of fieldset. We should look into further implications of the potentially affected `role` attribute and the Accessibility Tree.

PR #5413, opened 2023-08-01
USWDS - Range: Remove redundant ARIA attributes

### LINKED ISSUE #5038: USWDS - Feature: [Update the Range slider aria-valuenow attribute in real time when the slider is changed by a user]

I noticed that when using the Range Slider component, https://designsystem.digital.gov/components/range-slider/, the `aria-valuenow` attribute does not update in real time as a user changes the position of the slider. In addition, there should be an initial value set that I would think would be the average of the min and max values. 

Here is code that sets a basic average starting value and then updates in real time as the user changes the slider position. 

```javascript
      // Calc avg / mean function.
      function calcAverage(array) {
        return (
          array.reduce(
            (previousValue, currentValue) => previousValue + currentValue
          ) / array.length
        );
      }

      // Query all usa-range items.
      const form_ranges = document.querySelectorAll('.usa-range');

      if (form_ranges) {
        // Array reduce for the above, there could be more than one on a page.
        const form_range_item = [...form_ranges];
        // Loop all range sliders.
        form_range_item.forEach((range_item) => {

          // Define min / max and calc mean with the function.
          // Use Math.trunc to convert the "number" from a string to a number.
          const min = Math.trunc(range_item.getAttribute('min'));
          const max = Math.trunc(range_item.getAttribute('max'));
          const range_start_value = (calcAverage([min, max]))

          // Set the starting value when the page loads
          // as the mean between min and max.
          range_item.setAttribute('aria-valuenow', range_start_value)

          // Listen for a change on the slider and update it in real time.
          range_item.addEventListener("change", function () {
            this.setAttribute("aria-valuenow", this.value)
          })
        });
      }
```

None at this time

### Additional context

_No response_

### Code of Conduct

mahoneycm:
# Summary

**Removed redundant ARIA attributes to improve the screen reader experience.** To incorporate these changes, update your range component markup.

## Breaking change

## Related issue

Closes #5038 

## Related pull requests

[Changelog →](https://github.com/uswds/uswds-site/pull/2220)

## Preview link

Preview link:
[Range slider →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-range-slider-sr/iframe.html?args=&id=components-form-inputs-range--range)

## Problem statement

Range slider keyboard navigation does not work with **VoiceOver**. This is caused by redundant aria attributes that are reserved for elements that are more or less "faking" being an `input` with `type="range"`.

<details>

<summary>Example</summary>

Slider can only move one increment and does not update read out correctly:
_VoiceOver - Video has audio if you unmute_

https://github.com/uswds/uswds/assets/25211650/e4bed68f-fd87-408f-b489-6dc647d4bb62

</details>

**Note: Some more advanced screen readers such as JAWS may intuitively navigate around this issue**

## Solution

Remove the following attributes:
-  `aria-valuemin`
- `aria-valuemax`
- `aria-valuenow`
- `role="slider"`

### Sources

[MDN Docs: Range slider best practices →](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role#best_practices)

<details>

<summary>Example</summary>

Slider has full range and updates read out correctly

https://github.com/uswds/uswds/assets/25211650/c7395539-04ed-425e-8542-895bbaf1b37a

</details>

## Testing and review

1. Test [range slider on develop branch →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=components-form-inputs-range--range)
    1. Confirm keyboard navigation works without screen reader
    2. Activate the screen reader and attempt to use keyboard controls 
2. Open [range slider fix preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-range-slider-sr/iframe.html?args=&id=components-form-inputs-range--range)
3. Confirm screen reader keyboard navigation is possible
4. Confirmed read outs are improved
5. Confirm no visual or usability regression

### 
- [ ] Confirm screen reader navigation is possible
- [ ] Confirm screen reader readouts work as intended
- [ ] Confirm no regression in component

---

We have discussed potential screen reader callout enhancements which are captured in #5431


amycole501 (approved):
This sounds good in JAWS. I can hear "10, 20, 30" etc, and can use the arrow keys to move up and down the range slider. I recommend that whatever the range is indicating be written into the help text. For example "ten percent" instead of just "ten" if the slider is indicating a percentage range.

In NVDA the action is identical to JAWS. I used Chrome to test both.

mahoneycm:
It looks like we can accomplish the help text that @amycole501 mentions using `aria-valuetext`: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuetext

I'll do a bit more research on this and we can discuss if we want this to be included with this PR, or separated out into it's own issue. My primary concern is if we should maintain `aria-valuenow` to be used with `aria-valuetext`.

amyleadem (commented):
This works well for me in VoiceOver in Safari and VoiceOver Gestures in iOS Safari. I ran into some hiccups with the screen reader performance in Firefox and Chrome, but these issues don’t seem isolated to USWDS (see notes below for more details).

I tested the following:
- [x]  Confirmed the issue exists when using VoiceOver in desktop browsers
- [ ]  Confirmed that the updated component is operable and provides appropriate feedback in VoiceOver. (Values from 0-100 are read out in increments of 10 as you navigate forwards and backwards through the range.)
- In Safari, the component works well with VoiceOver.
- In Firefox I am not able to move backwards in value (ie, moving from 40 to 20) with VoiceOver. However, this issue doesn’t seem isolated to USWDS because I noticed that the same behavior occurs in the “Cowbell” example on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) in Firefox as well.
- In Chrome, I ran into intermittent issues where it will read the same value twice and read 90 and 10 as the end points, rather than 0 and 100. This is not consistent though and I have been able to get a successful readout in Chrome.  This also doesn’t seem isolated to USWDS because I noticed that the same behavior occurs in the “Cowbell” example on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) as well.
- Both issues in Firefox and Chrome are preferable to the current readout on `develop`.
- Interestingly, the issue does not exist in iOS Safari with VoiceOver Gestures and reads out the percentage (”10%”, “20%”, etc). It works well in both `develop` and this PR branch.

Do you get the same results?

amycole501:
The component sounds great in JAWS, NVDA, and Narrator on PC desktop in Chrome and Firefox.
The only functionality improvement may be to tell people not to press and hold the arrow keys as nothing is announced if you do that. It's silent. So you have to click then listen. That may be something we can add in ARIA or instructions?

mahoneycm:
@amyleadem So I _was_ able to replicate the Chrome bug after entering a bunch of inputs to try to trip up VoiceOver. During my testing, I found if you press the left and right arrow keys at the same time it would enable something called "Quick nav". It was only while quick nav was turned on that I would experience this bug! 

By pressing both of the arrow keys again and disabled quick nav, the bug went away! I was also able to replicate the Firefox bug you mentioned only after enabling quick nav!

Could you try testing this again when you have the time? I'm curious if you experience the bugs after toggling quick nav.

amyleadem (approved):
@mahoneycm good catch! My system did appear to have "quick nav" turned on. When I turned it off, I did not experience the strange behavior in Chrome and Firefox anymore. Nice discovery!

mejiaj (approved):

mejiaj:
Code looks good. 

@mahoneycm since this and #5472 have changes to the component. Are there special considerations we need to be aware of when merging?

mahoneycm:
@mejiaj This PR should be merged before #5472. This work captures a bug fix and #5472 is a feature enhancement that builds off of this change.

Beyond that, I don't believe any other accommodations need to be made!

thisisdano (approved):

Issue #5077, opened 2022-12-21
USWDS - Feature: Ability to have <Table/> headers sticky to the top


etanb:
### Is your feature request related to a problem? Please describe.

For usage of `<Table/>` components with massive data sets, it is hard to remember what each column denotes. Scrolling back up isn't a good solution for 100s of rows of data as you'll lose your place, and converting everything to `stacked` is a huge design change.

### Describe the solution you'd like

An additional CSS utility class like `usa-table--sticky-header` that would have the header information of a table sticky to the top of the page as a user scrolls.

### Additional context

I've created this working solution within the USWDS for review: https://github.com/uswds/uswds/pull/5074

Would love to make this a reality!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Feature+Request%22) for duplicate feature requests.

amycole501:
Sample link created by @amyleadem [https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-sticky-table-header/iframe.html?args=&id=components-table--sticky-header&viewMode=story](url) 

I used JAWS to listen to the example sticky header table and before I would say it's accessible, it needs to have the table header row declared as the header row. When I listen to it it doesn't tell me which column the data cells are in. It just reads the cell data out loud and if I couldn't see the table I wouldn't know what the meaning of the data cell is. 
It does say aloud "sticky header" which for someone using a screen reader may not be relevant.

PR #5420, opened 2023-08-02
USWDS - Table: Add sticky headers

amyleadem:
# Summary

**New table variant.** Created the `.usa-table--sticky-header` variant to allow sticky positioning on table headers. 

> [!CAUTION]
> Sticky headers are not compatible with the scrollable and stacked table variants.

Thanks to @etanb for originating this work in PR https://github.com/uswds/uswds/pull/5074.

## Breaking change

## Related issue

Closes https://github.com/uswds/uswds/issues/5077

## Related pull requests

[Changelog and documentation PR](https://github.com/uswds/uswds-site/pull/2229)

## Preview link

- [Sticky table](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-sticky-table-header/?path=/story/components-table--sticky-header)
- [Default table](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-sticky-table-header/?path=/story/components-table--default)
- [Borderless table](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-sticky-table-header/?path=/story/components-table--borderless)
- [Striped table](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-sticky-table-header/?path=/story/components-table--striped)
- [Sortable table](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-sticky-table-header/?path=/story/components-table--sortable)

## Problem statement

For very long tables, it can be difficult to remember the associated header content. Creating the option to have table headers stick to the top of the page as the user scrolls would help keep tabular data in context.

## Solution

This change creates a `.usa-table--sticky-header` variant that gives the table header sticky positioning. 

> **Note**
> At this time, sticky headers are not compatible with the scrollable and stacked table variants. For the scrollable variant, this incompatibility is related to a known issue ([on W3C CSS repo](https://github.com/w3c/csswg-drafts/issues/865)) with sticky positioning and scrolling overflow. If we want to pursue a sticky header that is compatible with scrollable tables, I recommend we open a separate issue because it will require a decent lift to find a workaround. 

### New variant

| Variant name              | Description | 
| ---------------------------- | :-------------- |
| `.usa-table--sticky-header` |   Allows the table header element to stick to the top of the page as the user scrolls. Preferred for long tables with many rows. This variant is not compatible with the scrollable and stacked table variants. |   

### New settings

| Setting name              | Description | 
| ---------------------------- | :-------------- |
| `$theme-table-background-color` |   Defines the table background color. A value of `default` uses the body background color. |   
| `$theme-table-sticky-top-offset` |  Defines the top offset for sticky headers. |

## Testing and review
- Confirm that sticky positioning works with the default, borderless, striped, and sortable table variants.
- Test on a variety of browsers
- Confirm that sticky headers are compatible with multiple header rows
- Confirm that the new settings work as expected
- Confirm no visual regressions

To test in Storybook:
1. In a variety of browsers, open each table variant story from the list of preview links
1. In each story, turn on the sticky header control 
1. Confirm that the table header sticks to the top of the page when you scroll down
    > **Note** 
    > Sticky headers will not stick in scrollable and stacked variants. If the header is not sticking, confirm that the `scrollable` control is turned off.  
1. Turn on the control for multiple header rows
1. Confirm that both header rows stick to the top of the page when you scroll down and confirm that there are no visual irregularities (like table content showing behind the header on scroll)


packages/uswds-core/src/styles/mixins/general/table.scss
```diff
@@ -356,6 +362,46 @@ $table-unsorted-icon-color: get-color-token-from-bg(
...
+      border-right: none;
+    }
+  }
+
+  tbody tr:last-child {
+    td,
+    th {
+      border-bottom: none;
+    }
+  }
```
amyleadem, line 384: > **Note**
> To prevent an issue with content showing behind the sticky table headers on scroll, I added `border-collapse: separate`. This caused the cell borders to essentially double up, which created a need to reconfigure how the borders were styled.

packages/uswds-core/src/styles/mixins/general/table.scss
```diff
@@ -356,6 +362,46 @@ $table-unsorted-icon-color: get-color-token-from-bg(
...
+
+  thead {
+    position: sticky;
+    top: $theme-table-sticky-top-offset;
```
amyleadem, line 388: > **Warning**
> This offset is useful for when there is another element that already sticks to the top of the page, like a sticky header navigation. However, I suspect that most users will need to write custom rules if their sticky elements changes at different breakpoints (For example, our sticky header navigation goes away in the mobile view). Do we want to keep this setting with the understanding that additional style overrides might often be necessary?

mahoneycm: I feel like we should keep it. Like most of our work, we provide the tools, but it's up to the users to implement them in a way that works for them.

packages/usa-table/src/usa-table.stories.js
```diff
@@ -20,4 +39,19 @@ Borderless.args = BorderlessContent;
 export const Striped = Template.bind({});
 Striped.args = StripedContent;
 
+export const StickyHeader = Template.bind({});
+StickyHeader.args = {
+  ...DefaultContent,
+  ...StickyHeaderContent,
+  stickyHeader: true,
```
mahoneycm, line 46: Just curious, any reason why you chose to add `stickyHeader: true` here instead of in the `StickyHeaderContent` json? My initial thought was that would be more consistent with existing variant/content patterns.

mahoneycm (changes requested):
This is looking great! A couple questions and a potential compilation error. Let me know if I can be of any assistance!

packages/usa-table/src/usa-table.stories.js
amyleadem: Good catch! I moved it to JSON in [Move 'stickyheader:true' to JSON](https://github.com/uswds/uswds/pull/5420/commits/0d8a54dbfe4b4610a8bcd0d89a0b5fb5746ab939).

mahoneycm (approved):
Looks great! Happy to have this variant in our library!

- [x] Confirm sticky position works with the default, borderless, striped, and sortable table variants.
- [x] Tested across browsers
- [x] New settings work as expected and are error free
- [x] No visual regressions

alex-hull:
Listened to sticky header table in JAWS and NVDA. All column header names are announced properly when navigating with both. @amyleadem

mejiaj:
@amyleadem pulled latest from `develop` because `npm start` wasn't running with node v20 (expected 16).

**Error displayed**

```bash
ModuleBuildError: Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: error:0308010C:digital envelope routines::unsupported
```

mejiaj (changes requested):
PR description looks good. Thanks for creating improving the stories too.

Added suggestions in review. Moving forward I'd like to simplify our templates as much as possible, so as not to rely so much on logic. Some of my comments reflect that. We should improve what we can here, but have this in our minds for future templating work.

packages/usa-table/src/usa-table--sortable/usa-table--sortable.twig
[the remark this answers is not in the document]
amyleadem: You have to first turn off the scrollable variant in the control since it is incompatible with sticky headers. I added a note to the scrollable control in [7ef482c](https://github.com/uswds/uswds/pull/5420/commits/7ef482cbc2f00fcd0e8df30a149e56b56fae9135) to clarify. 

Let me know if that helps.

packages/usa-table/src/usa-table.twig
```diff
@@ -14,24 +26,28 @@
...
-          {% if not cell.scope %}
-            <td>{{ cell.title }}</td>
-          {% else %}
-            <th scope="{{ cell.scope }}">{{ cell.title }}</th>
-          {% endif %}
+        {% for th in thead %}
+          <th scope="{{ th.scope|default('col') }}">{{ th.title }}</th>
         {% endfor %}
       </tr>
     {% endfor %}
```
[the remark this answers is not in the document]
amyleadem: This tests that all `thead` content is visible when there are multiple rows in a sticky header. Wanted to include a test for this because initially only the first row in the `thead` would stick and the other row(s) would move with  the rest of the unsticky content. 

To minimize confusion in the future, I've added a note to the control in [58d0a6a](https://github.com/uswds/uswds/pull/5420/commits/58d0a6afef50b8ddf24e55f0dc3eb2ff57fbd855).

amyleadem (commented):
@mejiaj  I believe I have addressed all your comments related to USWDS code. I am going to look into the guidance in [USWDS-Site - Changelog: add entry and doc updates for sticky table headers](https://github.com/uswds/uswds-site/pull/2229#top) to make sure it addresses [this comment](https://github.com/uswds/uswds/pull/5420#discussion_r1447561047). Let me know if you need anything else.

packages/usa-table/src/usa-table.stories.js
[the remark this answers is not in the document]
mejiaj: My bad. Somehow skipped a word — did we test the HTML template output on site to ensure there weren't regressions.

Looks like you did, thank you.

packages/usa-table/src/usa-table--sortable/usa-table--sortable.twig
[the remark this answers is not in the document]
mejiaj: @amyleadem is it worth making these controls conditional[^1]? For example, only show these controls if sticky headers are disabled.

**Example**
```js
  argTypes: {
    scrollable: {
      name: "Scrollable (Turning this on will disable sticky headers)",
      control: { type: "boolean" },
      defaultValue: false,
      if: { arg: "sticky_header", truthy: false } // Hide this control if sticky
    },
  },
```

---

If the control for multiple header rows is strictly for sticky header testing — wouldn't a separate test be better?

[^1]: https://storybook.js.org/docs/essentials/controls#conditional-controls

packages/usa-table/src/usa-table.twig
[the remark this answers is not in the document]
mejiaj: If the control for multiple header rows is strictly for sticky header testing — a separate test might be clearer.

packages/usa-table/src/usa-table--sortable/usa-table--sortable.twig
```diff
@@ -1,5 +1,7 @@
+{% if scrollable %}
 <div class="usa-table-container--scrollable" tabindex="0">
-  <table class="usa-table">
+{% endif %}
+  <table class="usa-table {{sticky_header ? 'usa-table--sticky-header' : ''}}">
```
mejiaj, line 4: It'd be easier to update the class name if you define it once and re-use later.

```twig
{% if sticky_header %}
  {% set sticky_header_class = "usa-table--sticky-header" %}
{% endif %}

  <!-- Allows you do do something like this. -->
  <table class="usa-table {{ sticky_header ? sticky_header_class }}">
```

packages/uswds-core/src/styles/settings/_settings-components.scss
```diff
@@ -213,6 +214,7 @@ $theme-table-sorted-header-background-color: "accent-cool-light" !default;
 $theme-table-sorted-background-color: "accent-cool-lighter" !default;
 $theme-table-sorted-stripe-background-color: "blue-cool-10v" !default;
 $theme-table-sorted-icon-color: default !default;
+$theme-table-sticky-top-offset: -1px !default;
```
mejiaj, line 217: Small concern here is this is similar to the current hero setting where it's not very flexible or re-usable.

Not a blocker for this work.

mejiaj (changes requested):
Thanks @amyleadem, two more minor comments. One for improving clarity in settings and another for making the template class easier to update.

packages/usa-table/src/usa-table.twig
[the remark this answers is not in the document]
amyleadem: Good idea! Removed the universal control and created a test pattern for it instead in [1e447af](https://github.com/uswds/uswds/pull/5420/commits/1e447afdeb30cc95a231c6c6831104fd0dcd6403)

packages/usa-table/src/usa-table--sortable/usa-table--sortable.twig
amyleadem: Updated in [cae00e5](https://github.com/uswds/uswds/pull/5420/commits/cae00e59eec298dd295a202d63d3cc280a56efea)

packages/uswds-core/src/styles/settings/_settings-components.scss
amyleadem: Yeah, the setting is fairly limited, but I am not sure if there is a way around it. I would imagine a lot of projects will need to write some custom style rules for the top offset.

packages/usa-table/src/usa-table--sortable/usa-table--sortable.twig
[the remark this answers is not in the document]
amyleadem: I played with this a bit but lean away from adding it in this instance since neither is actually the dependency of the other. Instead they are just incompatible. I ended up feeling like I would be confused about which options are available for the component, so I left it as is. 

Let me know if you object.

amyleadem:
@mejiaj I believe I've addressed your concerns. Let me know if you have any questions.

packages/uswds-core/src/styles/settings/_settings-components.scss
mejiaj: @amyleadem does this setting accept units too or only pixels? Do we have any guidance for this?

mejiaj (approved):
Code looks good, thank you.

thisisdano (approved):
NICE

PR #4385, opened 2021-11-04
USWDS - Accessibility: Added button functionality to usa-button links

mahoneycm:
# Summary

**Added spacebar activation to links styled as buttons**. This allows users to trigger these links the same way they would expect to trigger buttons. `usa-card` buttons have been replaced with links styled as buttons.

This PR has two main goals based on @thisisdano's comments in #3690:
- Links that are styled as buttons should be triggered by the spacebar without using `role="button"` [^1]
- `usa-card` buttons should be replaced with links [^2]

[^1]: https://github.com/uswds/uswds/issues/3690#issuecomment-951091169
[^2]: https://github.com/uswds/uswds/issues/3690#issuecomment-951095032

## Breaking change

:warning: This is a breaking change.

`usa-card` buttons have been replaced with links styled as buttons.

Before: 

```html
<div class="usa-card__footer">
    <button type="button" class="usa-button">Visit Florida Keys</button>
</div>
```

After:
```html
<div class="usa-card__footer">
    <a href="#" class="usa-button">Visit Florida Keys</a>
</div>
```

## Related issue

Closes #3690 

## Related PRs
[Changelog PR →](https://github.com/uswds/uswds-site/pull/2085)

## Preview link
[usa-card w/ links styled as buttons](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-button-group-a11y/iframe.html?args=&id=components-card--default&viewMode=story)

Preview link:

## Problem statement

Links styled as buttons should be triggered as such

## Solution

Create a simple JS file to target all anchors with the `usa-button` class and allow them to be triggered with the spacebar

Replace buttons in `usa-card` with `usa-button` anchor tags

## Testing and review

1. Visit `button-group` storybook preview
2. Turn on screen reader
3. Tab to "Back" link
4. Verify the screen reader reads it as a link
5. Press spacebar key
6. Verify that the link is followed
7. Visit usa-card
8. Verify that all buttons are now links with `usa-button` class
9. Verify spacebar & enter button trigger the links
   - It may help to update the links `href` to verify the page updates

Before opening this PR, make sure you’ve done whichever of these applies to you:
- [x] Confirm that this code follows the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
[...]

PR #5435, opened 2023-08-14
USWDS - Button: Create story for links styled as buttons

### LINKED ISSUE #5254: USWDS - Create story for links styled as buttons
## Description

Currently the only links with the `usa-button` class in storybook are the "back" button on [button-group](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=components-button-group--default) & the [card buttons](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-button-group-a11y/iframe.html?args=&id=components-card--default) (after #4385) 

It would be helpful to give them their own story, potentially nexted within `usa-button` to showcase links styled as buttons

mahoneycm:
# Summary

Created a storybook preview for links styled as buttons. The story matches existing button storybook previews.

## Breaking change

## Related issue

Closes #5254 

## Related pull requests

~[Changelog →](https://github.com/uswds/uswds-site/pull/2242)~

This is an internal change.

## Preview link

[Links styled as buttons →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-button-link-story/?path=/story/components-button--test-links-as-buttons)

## Problem statement

Currently the only links with the `usa-button` class in storybook are the "back" button on [button-group](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?args=&id=components-button-group--default) & the [card buttons](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-button-group-a11y/iframe.html?args=&id=components-card--default) (after [#4385](https://github.com/uswds/uswds/pull/4385))

## Solution

Create a storybook preview for links styled as buttons.

### Notes:

- Removed `disabled` button from link preview
    - Anchor elements cannot be disabled via the attribute.
    - Users must use our disabled classes + remove pointer events to disable links
- Excluded `type` storybook control from Link story
    - I excluded the `type` attribute from the Links as Buttons template as well as the control. Anchor elements do not have `type` functionality.
- `aria_disabled` control
    - Control appears to be outdated and is not mentioned in any button templates. Submitted a ticket to remove. (#5436) 
    

## Testing and review

1. Inspect storybook preview
2. Confirm controls work
3. Confirm storybook includes all necessary elements to showcase links styled as buttons

### Testing Checklist

- [ ] Story appropriately showcases links styled as buttons
- [ ] Story and controls are configurated as we would prefer
- [ ] Test template matches existing patterns
- [ ] Variants apply correctly

## Further Considerations

Should we add links styled as buttons to the component preview?

<details>
<summary>Image</summary>

image
</details>


amyleadem (changes requested):
Thanks for setting this up, @mahoneycm. It's looking good. I had just a couple thoughts:

- It would be helpful to also have  dropdown control that lets us include a button variant (like `usa-button--accent-cool`) in the modifiers.
- I think we should clearly label this as a test pattern (unless we decide to consider it differently). I added a few notes below.

packages/usa-button/src/usa-button.stories.js
```diff
@@ -35,6 +37,7 @@ export default {
 };
 
 const Template = (args) => Component(args);
+const LinkTemplate = (args) => LinkTest(args);
```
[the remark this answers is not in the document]
mahoneycm: Resolved in https://github.com/uswds/uswds/pull/5435/commits/01641ba693157901517b1fb73e22fbda129c1e2b

packages/usa-button/src/usa-button.stories.js
```diff
@@ -62,3 +65,13 @@ Secondary.args = SecondaryContent;
 
 export const Unstyled = Template.bind({});
 Unstyled.args = UnstyledContent;
+
+export const LinksAsButtons = LinkTemplate.bind({});
+LinksAsButtons.args = LinkContent;
+LinksAsButtons.argTypes = {
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [01641ba](https://github.com/uswds/uswds/pull/5435/commits/01641ba693157901517b1fb73e22fbda129c1e2b)

packages/usa-button/src/content/usa-button~link.json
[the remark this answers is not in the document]
mahoneycm: Resolved in [01641ba](https://github.com/uswds/uswds/pull/5435/commits/01641ba693157901517b1fb73e22fbda129c1e2b)

mahoneycm:
@amyleadem Updated the story to follow test patterns in [01641ba](https://github.com/uswds/uswds/pull/5435/commits/01641ba693157901517b1fb73e22fbda129c1e2b) !

As for the dropdown, I had the same idea actually! I ended up going against it though because then you could only assign one modifier at a time. This would mean we couldn't do things like `usa-button--outline usa-button--inverse` or `usa-button--accent-warm usa-button--big` for example.

If we don't think that's necessary, I definitely think it could be a cool enhancement for the button story.

Honestly, maybe we could have a secondary dropdown for additional modifier classes?

amyleadem (changes requested):
@mahoneycm That’s a good point about the modifier control. I think the text box control type works well given the circumstances.

Everything is looking good in Storybook here. However, I noticed that after moving the JSON to the `test-patterns` directory I now get an error when I run the `build:html` command. I recently encountered this issue on another PR as well. Here is the error:

```html
[webpack-cli] Failed to load '/Users/aleadem/Dev/github/uswds-workspace/uswds/webpack.twig.config.js' config
[webpack-cli] Error: ENOENT: no such file or directory, open './packages/usa-button/src/test/test-patterns/usa-button.json'
```

From what I've seen, it appears to happen when the `test-patterns` directory contains both a twig and a JSON file. Since the JSON is so simple here, I am wondering if it makes sense to just pass the data as args in `stories.js` or if we want to dive deeper into what is happening in the webpack config.

mahoneycm:
@amyleadem Thanks for catching that! I wonder what the issue with having both file types is...

I went ahead and moved the data into args within `stories.js` for the time being. 👍

packages/usa-button/src/usa-button.stories.js
```diff
@@ -62,3 +64,17 @@ Secondary.args = SecondaryContent;
 
 export const Unstyled = Template.bind({});
 Unstyled.args = UnstyledContent;
+
+export const TestLinksAsButtons = TestTemplateLinks.bind({});
+TestLinksAsButtons.args = {
+  modifier: "",
+  text: "Link styled as button",
+  is_demo: true,
```
amyleadem, line 72: Small tweak, but it should be safe to remove this line since we declare it as the defaultValue in [line 26](https://github.com/uswds/uswds/pull/5435/files#diff-33b264a45949b5dbda9916d0c19947e50ba5bfe9feb66f9aa46562488872e73eL25)
```suggestion
```

packages/usa-button/src/test/test-patterns/test-button-links.twig
```diff
@@ -0,0 +1,14 @@
+{% if is_demo and 'usa-button--big' not in modifier %}
```
amyleadem, line 1: I noticed that `usa-button.twig` [adds an additional check](https://github.com/uswds/uswds/blob/develop/packages/usa-button/src/usa-button.twig#L1-L3) to provide a dark background for the `usa-button--outline usa-button--inverse` classes. Was there a reason it was left off here? 

```suggestion
{% if 'usa-button--outline usa-button--inverse' in modifier %}
  <div class="bg-base-darkest padding-1" style="max-width: fit-content">
{% endif %}

{% if is_demo and 'usa-button--big' not in modifier %}
```

amyleadem (changes requested):
Looking good! Just a couple of small notes:

packages/usa-button/src/usa-button.stories.js
mahoneycm: Ooh thanks for pointing that out! Removed in [fb698eb](https://github.com/uswds/uswds/pull/5435/commits/fb698eb16744f410b4615af245f56604ae5caf24)

packages/usa-button/src/test/test-patterns/test-button-links.twig
mahoneycm: Originally, I left it out because the modifiers have to be manually entered via text strings. Users would have to put in exactly `usa-button--outline usa-button--inverse` for it to work. So swapping them for example doesn't show the black background ` usa-button--inverse usa-button--outline`

That said, I'd rather have it so we _can_ experiment with it than not have it! 

Included the check in [814f746](https://github.com/uswds/uswds/pull/5435/commits/814f746eb4b3ae7fd75567fc088c7fc3f3bcc79d)

amyleadem (approved):
Looks good to me!

packages/usa-button/src/usa-button.stories.js
```diff
@@ -62,3 +64,16 @@ Secondary.args = SecondaryContent;
 
 export const Unstyled = Template.bind({});
 Unstyled.args = UnstyledContent;
+
+export const TestLinksAsButtons = TestTemplateLinks.bind({});
+TestLinksAsButtons.args = {
+  modifier: "",
+  text: "Link styled as button",
+};
+TestLinksAsButtons.argTypes = {
```
mejiaj, line 73: Usually the name for tests should be just `test` and it tests more than just the component.

A description, if its a true test, as to _what_ we're testing would be helpful. 

Otherwise, it could be moved out of tests and alongside the other variants. If we don't have reasons to validate it as a "true" component, we can add a disclaimer.



image

mejiaj (changes requested):
Thanks for addressing this issue.

### Notes

Typically we use test stories to see a component within another context (_like its behavior with other components or alongside other variants_).

This is simply the same component, but in link form. It might make more sense to either:

1. Make it a regular story variant.
2. Make it a true test and add real buttons alongside.

Either could include a disclaimer that this is experimental or for prototyping.

packages/usa-button/src/usa-button.stories.js
mahoneycm: I updated it to follow standard component story patterns! Since it's used throughout our site and the design system, it deserves to be a real story with the other button variants!

mejiaj (changes requested):
Looks better as a story, thank you. Can you look into the build errors before approving?

<details><summary> CircleCI build error</summary>
<p></p>

```bash
ERROR in Conflict: Multiple assets emit different content to the same filename usa-button.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-button~accent-cool.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-button~accent-warm.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-button~base.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-button~big.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-button~outline-inverse.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-button~outline.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-button~secondary.html

ERROR in Conflict: Multiple assets emit different content to the same filename usa-button~unstyled.html

webpack 5.76.0 compiled with 9 errors in 4524 ms
```
</details> 

https://app.circleci.com/pipelines/github/uswds/uswds/5592/workflows/24c8302a-f207-4ec9-b0dd-f9d05ce20126/jobs/10050

mahoneycm:
@mejiaj build failure was a result of having multiple twig files with imported contend in the same directory it seems. Followed the pattern set by other components and moved the twig file into it's own directory and it resolved the issue! 👍

mejiaj (approved):
LGTM, thanks for the adjustments

thisisdano (approved):
Good conversation, good outcome

Issue #5427, opened 2023-08-07
USWDS - Bug: Disabled color settings do not check for contrast


amyleadem:
## Summary
When customizing the disabled state color settings, the system does not send a warning when the selected colors do not meet required contrast ratios. 

For example, the following configuration results in text that does not meet color contrast requirements with its background :

```
@use "uswds-core" with (
  $theme-show-notifications: false,
  $theme-color-disabled-lighter: "gray-20",
  $theme-color-disabled-light: "gray-20",
  $theme-color-disabled: "gray-20",
  $theme-color-disabled-dark: "gray-20",
  $theme-color-disabled-darker: "gray-20",
);
```

Here is a screenshot of the resulting display (note that the text is not visible in the button and select components and that the outline button does not meet contrast requirements):
![image](https://github.com/uswds/uswds/assets/93996430/586d8251-e70c-4068-9e87-d36599cc3f9e)

In contrast, here is what it looks like in the default configuration:
![image](https://github.com/uswds/uswds/assets/93996430/d4695918-b1e4-4b0a-b769-91045f4d0675)

## Expected behavior
The system should send a warning when settings values do not result in the required color contrast ratio.

mejiaj:
Moving to next release (3.8.0) as a potential enhancement.

maggiewachs:
@amyleadem @mahoneycm Hello! I'm a web accessibility engineer with CivicActions; we have a few projects that use the USWDS. 

I was wondering what's the reason for this fix? Disabled elements [don't need to meet contrast minimum](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html), and that's by design because they're meant to look inactive/unusable. Making them higher contrast could actually be confusing to users, especially those with low vision.

jaclinec:
@maggiewachs Thanks for reaching out!

Over the last several months, we have been researching and re-evaluating the role of disabled elements in USWDS. Our research pointed us to pursue the following actions:

1) Draft recommendations against using disabled elements at all (this is a work in progress), and 
2) Update the styles of USWDS disabled elements to meet color contrast requirements. 

We updated our disabled styles because our research pointed to disabled elements having meaning. Including disabled elements on a page indicates to users that an action must be taken before the element can be enabled. By providing styles that failed to meet color contrast requirements, we were excluding certain audiences from being able to access and understand that meaning. We understand that this deviates from the low-contrast presentation that is the common standard, but it felt important to include all audiences here.

[...]

PR #5402, opened 2023-07-26
USWDS: Fix disabled theme color settings

amyleadem:
# Summary

- **Fixed configuration errors with disabled color settings.** All disabled color settings should now be configurable and accept hex color values.
- **Updated disabled settings names to match other state colors.** Updated default disabled color settings values and added `$theme-color-disabled-lighter` and `$theme-color-disabled-darker`.
- **Added "disabled-lighter" and "disabled-darker" color tokens.** Also updated the default values for all disabled tokens.

## Breaking change

:warning: This is a breaking change. The names and values of disabled settings and tokens have changed and will need to be updated according to the table below if they are used in your project. All disabled states must meet minimum color contrast requirements for text (4.5:1).

## Related issue

Closes #5334

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2214)

## Preview link

[Disabled form elements page](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-fix-disabled-theme-colors/?path=/story/patterns-forms--disabled-form-elements)

## Problem statement
1. `$theme-color-disabled-text`, `$theme-color-disabled-text-reverse`, and `$theme-color-text-on-disabled` settings return a module loop error for any custom configuration.

```
$theme-color-disabled-light: "magenta-10", //compiles
$theme-color-disabled: "magenta-10", //compiles
$theme-color-disabled-dark: "magenta-10", //compiles
$theme-color-disabled-text: "magenta-10", //module loop error
$theme-color-disabled-text-reverse: "magenta-10", //module loop error
$theme-color-text-on-disabled: "magenta-10", //module loop error
```

2. Unlike other state color settings like `$theme-color-success` and `$theme-color-error`, the disabled color settings do not accept hex values. 
```
$theme-color-disabled-light: #fffff, //compiles (but not currently used in the system)
$theme-color-disabled: #333333, //error: not a token
$theme-color-disabled-dark: #333333, //compiles (but not currently used in the system)
$theme-color-disabled-text: #333333, //error: not a token
$theme-color-disabled-text-reverse: #333333, //error: not a token
$theme-color-text-on-disabled: #333333, //error: not a token
```

## Solution
1. Added the `!default` flag to these settings to make them configurable in projects.
2. Created new disabled color shortcodes and used them to replace the `$theme-color-disabled-` references in the Sass. Using shortcodes enables the system to accept hex code values for these settings.
[...]

PR #5372, opened 2023-07-11
USWDS - Memorable date: Wrap elements at narrow widths

amyleadem:
# Summary

Updated memorable date styles to allow elements to wrap to multiple lines at narrow widths.

## Breaking change

This is not a breaking change.

## Related issue

Closes #5135

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2170)

## Preview link

Preview link: [Memorable date component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-memorable-date-wrap/?path=/story/components-form-inputs-memorable-date--memorable-date)

## Problem statement

The memorable date fields get visibly cut off at screen widths less than ~430px or when using screen magnification.

image

## Solution

Allowing the memorable date fields to wrap to multiple lines will let users access all elements of the component at narrow screen widths and when using screen magnification.

## Testing and review

- Confirm that all elements of the memorable date component are visible at screen widths <400px 
- Confirm that all elements of the memorable date component are visible when using screen magnification
- Confirm that the component is still intuitive and usable when broken into multiple lines.

PR #5455, opened 2023-08-21
USWDS - Disabled mixin: Test for color contrast

mahoneycm:
# Summary

**Added color contrast checks for disabled tokens.** On compilation, USWDS will test disabled element color contrast and provide a fallback color if minimum contrast is not met. If the fallback does not meet minimum requirements, a warning will be provided.

## Breaking change

## Related issue

Closes https://github.com/uswds/uswds/issues/5427

## Related pull requests

[Changelog PR →](https://github.com/uswds/uswds-site/pull/2254)

## Preview link

[Disabled form elements →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-disabled-color-contrast/iframe.html?args=&id=patterns-forms--disabled-form-elements)

[Button outline →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-disabled-color-contrast/iframe.html?args=&id=components-button--outline)

[Button outline inverse →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-disabled-color-contrast/iframe.html?args=&id=components-button--outline-inverse)

## Problem statement

#5402 normalized disabled tokens for users. Users could unintentionally set the tokens to colors with poor contrast, causing disabled elements to lose the necessary context.

## Solution

Utilize `set-bg-and-color()` mixin which includes a contrast check, fallback colors, and warning message if fallback colors are unable to resolve contrast issue.

> [!IMPORTANT]
> These tests can only be conducted when background color is known. Elements with disabled colors without a defined background color are _still_ at risk for breaking color contrast.

More information in **************************Further Considerations************************** below
> 

### Color contrast failsafe testing:

By default, `u-disabled` is set to the following:

```scss
@mixin u-disabled($bg-color: "disabled-lighter", $text-color: "disabled-dark")
```

To test, we can update these values to have non-compliant color contrast

1. Visit `_settings-color.scss`
2. Update `$theme-color-disabled-dark` to match `$theme-color-disabled-lighter`
    - `gray-40`
3. Inspect disabled elements in Disabled Forms pattern (`http://localhost:6006/?path=/story/patterns-forms--disabled-form-elements`).
4. Color should revert to `ink` fallback
5. Update `$theme-disabled-lighter` to a darker color than `$theme-disabled-lighter`
    - `gray-60`
6. Inspect warning in terminal

### Testing checklist

- [ ]  Code matches USWDS standard
- [ ]  Disabled styles are tested properly for color contrast
    - [ ]  Fallbacks to `ink token`
    - [ ]  Displays a warning when fallback fails color contrast
- [ ]  No visual regression

---

## Further Considerations

<details>
<summary>Toggle</summary>

### Unknown background colors

For elements with an undefined background color, we ****could**** use `set-text-from-bg` and assume `$theme-body-background-color` as the color to test against

This would work for (potentially) most cases but is a big assumption.

It could benefit components like checkbox labels or outline buttons, but if those elements are placed in an area with a different background color, we would be testing against the wrong color.

Food for thought. Maybe we can discuss ways to enhance these color contrast tests further down the line.

### No mixin parameters used
Now that we've removed the `u-disabled` parameters from `_usa-button`, no styles use the optional parameters and the default is used across the board. Its probably fine to maintain the current functionality of the parameters, but flagging in case we want to remove.

### Secondary fallback value

I’d like to investigate a secondary fallback for contrast checks. 

Currently, if the `$preferred-color-token` fails contrast checks, it uses the `$preferred-fallback-token` which is `ink` by default.

When that token fails the contrast check, we still use the token but provide a warning in the build pipeline.

I’d like to further investigate adding a secondary fallback to test white against the background color. In my mind, if black still fails color contrast, the next reasonable test would be to use white rather than leave black on a dark background.

We do something similar specifically when the background grade is `45`

```scss
// When $bg-color = 45, serve alternate tokens with color grades of 0 and 100
  // This provides sufficient contrast to pass WCAG AA standards
  @if $bg-grade == 45 and $wcag-target == "AA" {
    $preferred-text-token: "white";
    $fallback-text-token: "black";
  }
```

This will probably be a second issue/enhancement
</details>


packages/usa-combo-box/src/styles/_usa-combo-box.scss
```diff
@@ -68,7 +68,6 @@ $close: map-merge(
     // Placeholder overrides to ensure color contrast compliance accross browsers
     &::placeholder {
       opacity: 1;
-      color: color("disabled-dark");
```
mahoneycm, line 71: Browsers set the color for the placeholder via the `-webkit-text-fill-color` above. Removing this color definition because it seems unnecessary.

mahoneycm:
@amyleadem Pulled in the latest and it should be ready for review!

mejiaj (approved):
Code looks good.  I was able to test:
- [x] Colors are set using `disable-lighter` (background) and `disable-dark` (text).
- [x] Warning appears on compile when contrast is not met.

amyleadem (changes requested):
Thanks for taking this on! Generally this looks good but I’ve flagged some issues in the list below. I consider some of these items to be out of scope and have added a note when I think they don’t need to be blockers for this PR. Curious to hear if you disagree!

- [ ] **Confirm color contrast with light settings.** Set all disabled settings to “gray-20”. Check contrast ratios for components:
    - [x] Button
    - [ ] Button outline
        - Outline buttons can be set to failing contrast ratios (For example, set $theme-color-disabled: “gray-20” and $theme-body-background-color: "white”)
    - [x] Character count
    - [ ] Checkbox/radio
        - Checkbox and radio can be set to failing contrast ratios (For example, set $theme-color-disabled: “gray-20” and $theme-body-background-color: "white”)
    - [x] Combo box
    - [ ] Date picker
        - The icon color ends up being inconsistent with other disabled icons when the fallback is in place. This is because the icon color is defined with opacity instead of a disabled color. This might not be avoidable without breaking changes, but wanted to flag the inconsistency. I don’t think this needs to be a blocker for this PR.
        image
    - [x] File input
    - [ ] Input mask
        - `.usa-input-mask--content` is hard-coded to “gray-50”. Is it possible to run this through our contrast mixins as well?
    - [x] Input prefix/suffix
    - [x] Memorable date
    - [x] Range slider
    - [x] Select
    - [x] Text input
    - [ ] Time picker
        - Same issue as date picker. I don’t think it needs to be a blocker
- [ ] **Confirm color contrast with dark settings.** Set all disabled settings to “gray-80”. Check contrast ratios for components.
    - All components failed to meet color contrast except checkbox/radio and button outline. I did receive appropriate warnings about the failures during compilation, though. For this reason, I think it meets requirements.
    - Changing the text color ($theme-color-disabled-dark) to a light contrasting color against a dark background ($theme-color-disabled-lighter) fixes most of contrast issues, but the icons do not adjust, which can make them hard to see. I did not receive any warnings about this.
        - I consider this a failure of the test, but to keep things in scope, I think we can address this later (if at all). Dark disabled colors seem to me to be more of an edge case, potentially used in dark mode designs(?). I am curious to learn how disabled colors are typically handled in dark modes. I don’t think this needs to be a blocker for this PR, but flagging it in case anyone else has thoughts.
- [x] Check high contrast mode

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -22,6 +22,7 @@
   $input-type: null
 ) {
   $input-text-color: get-color-token-from-bg($bg-color);
+  $input-text-color-disabled: get-color-token-from-bg($bg-color, "disabled", "gray-50");
```
mahoneycm, line 25: Compares theme `disabled` color against form background colors. 

Falls back to standard `gray-50` if `disabled` doesn't match color contrast. The goal here is to provide the standard disabled color if the user-updated color does not pass color contrast. 

This seemed to me like a better solution than defaulting to `ink` which doesn't "read" as disabled.

A warning is provided if both colors fail color contrast checks.

### Dark backgrounds

Changing the form background color to a dark color will fail a color contrast check, leaving low contrast between disabled content and its background BUT this matches the current fallback behavior of the color "ink" which is the default fallback for most components.

Are we ok relying on the compilation warning in these situations?

mahoneycm (commented):
Thanks @amyleadem ! Here are my thoughts for each of your concerns:

## Button outline

> Outline buttons can be set to failing contrast ratios (For example, set $theme-color-disabled: “gray-20” and $theme-body-background-color: "white”)

This is related to my note about this only working for components with defined background colors. There’s no way for us to calculate the background color behind transparent element.
  
> [!IMPORTANT]  
>  These tests can only be conducted when background color is known. Elements with disabled colors without a defined background color are *still* at risk for breaking color contrast.
    
We could only ******assume****** the background color is the `$theme-body-background-color` but if the button is placed in a different element/section with a different background, then the check would be invalid. 
    
Consider our [Sign In page examples](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-disabled-color-contrast/iframe.html?args=&id=pages-sign-in--sign-in-page) where the background color is set to `base-lightest`.
    
What do you think about setting a warning to check the color contrast when it would otherwise fail against `$theme-body-background-color`.

## Checkbox/radio labels

> Checkbox and radio can be set to failing contrast ratios (For example, set $theme-color-disabled: “gray-20” and $theme-body-background-color: "white”)

Resolved in [9bb417c](https://github.com/uswds/uswds/pull/5455/commits/9bb417c8238c415a5e867e12b4ded83989dcc3d1) and [96a1c21](https://github.com/uswds/uswds/pull/5455/commits/96a1c21a534884ee648667ab04d70ed5a3a4d719)

Additional notes in file comments.

amyleadem (changes requested):
This is looking good! Thanks for the thoughtful responses, @mahoneycm. I had one question about some checkbox/radio styles. I think my other concerns have been addressed. Here are my thoughts:

- Button outline - That explanation makes sense. With the transparent background color, I don’t see a way to automate this check. I think we’ll just have to move forward without a contrast check here.
- Checkbox/radio - This fix looks good. I confirmed the following:
    - [x] There are no visual regressions after consolidating duplicate styles
    - [x] I receive a warning when neither “disabled” nor “gray-50” meet contrast requirements
- Input mask - To keep things moving, I think we can consider this out of scope.
- Component icons - To keep things moving, I think we can consider this out of scope. I vote we open a follow-up issue to consider later.

I'd like to have @mejiaj confirm that the recent changes and discussions.

packages/uswds-core/src/styles/mixins/utilities/_disabled.scss
mejiaj, line 1: Please add SASSDoc style comments to these two mixins.

```
/// A description of the mixin.
///
/// @param {String} $bg-color - Description of param.
/// @output TBD
///
/// @example scss - Description
///   .foo {
///       @include u-disabled("a", "b");
///   }
```

[SASSDoc Annotations](http://sassdoc.com/annotations)

mejiaj (changes requested):
No visual regressions found with disabled styles. No issues with code. 

Added suggestions for improving documentation.

packages/usa-button/src/styles/_usa-button.scss
```diff
@@ -202,7 +202,13 @@ $button-stroke: inset 0 0 0 units($theme-button-stroke-width);
...
 .usa-button--outline-inverse[aria-disabled="true"] {
-  @include u-disabled("transparent", "disabled");
+  &,
+  &:hover,
+  &:active,
+  &:focus {
+    background-color: transparent;
+    color: color("disabled");
+  }
 }
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [987177b](https://github.com/uswds/uswds/pull/5455/commits/987177b4a455ce5bde957af2bc3db7bbb583f308)

packages/uswds-core/src/styles/mixins/utilities/_disabled.scss
mahoneycm: Resolved in [987177b](https://github.com/uswds/uswds/pull/5455/commits/987177b4a455ce5bde957af2bc3db7bbb583f308)

amyleadem (approved):
Looks good to me!

mejiaj (approved):
I was able to test different values in following settings:

```scss
$theme-color-disabled-family
$theme-color-disabled-lighter
$theme-color-disabled-light
$theme-color-disabled
$theme-color-disabled-dark
$theme-color-disabled-darker
```

Got a warning on colors that weren't meeting contrast.

```bash
Warning: Neither the specified preferred color token (`disabled-dark`) nor the fallback color token (`ink`) have AA contrast on a `disabled-lighter` background. Using `ink`. Please check your source code and project settings.
    packages/uswds-core/src/styles/functions/color/get-color-token-from-bg.scss 65:7  get-color-token-from-bg()
    packages/uswds-core/src/styles/mixins/helpers/set-text-from-bg.scss 14:28         set-text-from-bg()
    uswds-core/src/styles/mixins/helpers/set-text-and-bg.scss 16:3                    set-text-and-bg()
    packages/uswds-core/src/styles/mixins/utilities/_disabled.scss 28:5               u-disabled()
    packages/usa-textarea/src/styles/_usa-textarea.scss 9:5                           @forward
    packages/usa-textarea/src/styles/_index.scss 4:1                                  @forward
    _index.scss 5:1                                                                   @forward
    _index.scss 22:1                                                                  @forward
    packages/uswds/_index.scss 51:1                                                   @forward
    src/stylesheets/uswds.scss 3:1                                                    root stylesheet
```

### Re: input mask

I noticed that issue too and confirmed the value is hardcoded. We should look into ways to improve this in a follow-up issue.

mahoneycm:
Hey there, @Miskeyceh94 

Did you mean to leave that comment tagging #5372 ? Was there any additional context you wanted to leave?

Thanks!

thisisdano (approved):

Miskeyceh94:
🥴

Pada Sel, 27 Feb 2024 07.47, Dan O. Williams ***@***.***>
menulis:

Issue #5431, opened 2023-08-09
USWDS - Bug: Range slider screen reader experience lacks context


mahoneycm:
### Describe the bug

When using the range slider with a screen reader, users are left without context. Because sliders rely on context from their implementation, we should allow devs to provide additional context to for screen readers.

### Steps to reproduce the bug

1. Visit the [range slider](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-range-slider-sr/iframe.html?args=&id=components-form-inputs-range--range) component page _// This link is from #5413 which improved screen reader functionality of the component_
2. Turn on screen reader
3. Increase/decrease slider
4. Listen to readouts

There currently is no way to add units of measurement to these readouts

### Expected Behavior

Devs should be able to customize callouts to fit the need of their component.

Examples:
- Percentages
- Satisfaction rating (4 out of 5)
- Temperature

Really any context that visual users might have, the screen reader should readout on input changes

An additional potential enhancement discussed with @amycole501 would be to also read out the position of the slider in terms of how far along the range it is. 

For example, "4 out of 5 stars (80% of available range)".

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

amyleadem:
@mahoneycm Wondering if there is overlap here with this issue: https://github.com/uswds/uswds/issues/4504

PR #5472, opened 2023-08-29
USWDS - Range: Enhance SR callouts

mahoneycm:
# Summary

Added optional `data-text-unit` and `data-text-preposition` attributes to range slider. It now reads `{{ value }} of {{ max }}` when calling out the range slider value. The `of` can be changed using `data-text-preposition` as needed for use cases or different languages. 

Additionally, this PR enhances the storybook controls to easily allow users to set the min, max, step, and newly added data attribute values.

> **Important**
> The new JS file does not contain a teardown method at this time. Compared to other components, it appears we only teardown dynamic components, remove event listeners, and reset logic-related values.
> 
> It did not seem necessary to add a teardown to remove the `aria-valuetext` attribute at this time.

## Breaking change

Existing users will not need to change anything by default but will need to edit markup if they’d like to include the new unit feature.

## Related issue

Closes https://github.com/uswds/uswds/issues/5431

## Related pull requests

[Changelog and guidance update →](https://github.com/uswds/uswds-site/pull/2310)

[USWDS - Range: Remove redundant ARIA attributes #5413 →](https://github.com/uswds/uswds/pull/5413)

## Preview link

[Range slider →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-range-slider-sr-unit/?path=/story/components-form-inputs-range--range)

## Problem statement

When using the range slider with a screen reader, users are left without context. Because sliders rely on context from their implementation, we should allow devs to provide additional context to for screen readers as well as some way to let users know how far along the slider the handle is.

## Solution

This PR improves SR callouts in two ways.

The first, is adding the max value to the callout. Now when the slider moves, it will read the current value, then ******“of max value”****** where `max value` is the numerical max value.

Next, is adding an additional `data-text-unit` attribute. This attribute can be set to any unit the user can choose to include. Some examples may include:

- Percent
- Temperature
- Stars ********(rating)********
- Decibles
- etc…

Essentially dependent on the use case of the slider. 

When included, the screen reader will update to read out along the following examples:

- *******“30 percent of 100”*******
- ***************“72 degrees Fahrenheit of 90”***************
- ******************“3.5 stars of 5”******************
- **********************“3 pizza slices of 8”**********************
- etc.

With the addition of the `data-text-preposition` attribute, devs can change the language between the units and the total to fit whatever language they are developing for.

For example:
| Attribute | Value |
|--------|--------|
| data-text-unit | Percent |
| data-text-preposition |`null` |
| Output | `40 percent of 100` | 

Spanish: 

| Attribute | Value |
|--------|--------|
| data-text-unit | por ciento |
| data-text-preposition | de |
| Output | `40 por ciento de 100` | 

>**Note**
>Screen readers will still rely on system settings to read numbers in correct language

### New data attributes

`data-text-unit` - Set the `data-text-unit` attribute to include a unit with screen reader callouts. This provides additional context to assistive technology users. It is empty by default. Adding a value will be read as: _"x **units** of y"_.

`data-text-preposition` - Set the preposition between units and total. By default, we'll use `of` (i.e. 40 percent `of` 100). This will allow devs to update the verbiage depending on language.

### Further consideration

Devs should still provide proper context before users reach the slider. It is important for both visual, and screen-reader users, that the usability [guidance](https://designsystem.digital.gov/components/range-slider/#guidance) is followed.

This is also a JS solution. JS is required to set update the `aria-valuetext` attribute, which is standard across various accessible slider components researched during this work.

You can view some [examples from W3C WAI](https://www.w3.org/WAI/ARIA/apg/patterns/slider/). 

> **Note**
> The examples at the above link are all nonsemantic examples, and are built using divs, ARIA attributes, and JS
 

## Testing and review

1. Visit [Range slider →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-range-slider-sr-unit/?path=/story/components-form-inputs-range--range)
2. In controls, add a `data-text-unit`
3. Leave `data-text-prepostion` empty
4. Refresh page
5. Activate screen reader
6. Interact with the range slider
7. Confirm callouts have been improved
8. Confirmed preposition defaults to `of`
9. In controls, add a `data-text-preposition`
10. Interact with the range slider
11. Confirm `of` is replaced with your new value
12. If desired, play around with the unit to listen to other use cases
13. Feel free to play with min, max, and step if you want to customize the range slider for different use cases

### Testing checklist
- [ ] If unit is present, screen reader reads out unit with value
- [ ] If unit is not present, screen reader still reads value and max value
- [ ] Screen reader callouts are justifiably enhanced
- [ ] Context is improved for screen reader users

> **Note**
> Sometimes storybook has trouble initializing JS. If changes aren’t taking place or if readout is not reading properly, try refreshing the page and listening again.


amycole501 (commented):
Really clear audio descriptions happening for the slider. In JAWS in Chrome I hear "forty percent of one hundred" for example which is great. The only issue I foresee is someone pressing and holding an arrow key and thereby hearing nothing until the audio catches up to the keyboard command.

mahoneycm:
I was able to get the unit test working! I feigned an input using:

```js
el.dispatchEvent(new KeyboardEvent("input", { bubbles: true }));
```

@amyleadem @mejiaj Let me know what you both think! The test listens for an input the same way the component does. If we decide to add further event listeners for keyboard inputs we can dig further, but may not be necessary thanks to native keyboard input interaction.

amyleadem:
Hi @mahoneycm, 

I ran into some screen reader issues during testing:

- [ ] Confirm the value updates as expected in Chrome
    - With keyboard navigation in Chrome with VoiceOver, I see that the `aria-valuetext` is updating as expected, but sometimes the readout does not happen as expected. Sometimes it will get stuck on a value, skip a value, or read the wrong value.
    - Generally the Chrome VoiceOver experience is greatly improved over `develop` with this PR, but users would likely be confused by the incorrect readouts.
    - This behavior matches what I found in **[USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) PR**
- [ ] Confirm the value updates as expected in Safari
    - With keyboard navigation in Safari with VoiceOver, I cannot change the value from “20 out of 100” even though visually the value appears to change
    image
    - This is a regression because the values are read out as expected in `develop`
    - This is also a regression from the **[USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) PR**
- [ ] Confirm the value updates as expected in Firefox
    - With keyboard navigation in Firefox with VoiceOver, I can advance the value successfully, but I run into trouble when I try to move backwards. Sometimes the value doesn’t change at all, sometimes it jumps all the way to 0.
    - Generally the Firefox VoiceOver experience is greatly improved with this PR, but the quirks still prevent it from being fully usable
    - This behavior matches what I found in **[USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) PR**
- [ ] Confirm the value updates as expected in iOS Safari
    - With touch navigation in iOS Safari with VoiceOver gestures, it will only read “20 percent of 100” when I swipe up or down as instructed (Similar to macOS safari).
    - This is a regression. In `develop`, the value changes as expected with up/down swipe.
    - This is also a regression from the **[USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) PR**

I also had some additional questions about Storybook and the GitHub issues:

- For Storybook, I am a bit confused about the unit test story. Is it necessary? Is it possible to consolidate it with the main range story?
- I am wondering if it makes sense to focus on [USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) first and then look at [USWDS - Range: Enhance SR callouts](https://github.com/uswds/uswds/pull/5472). Either that or we combine the PRs (Which I know makes for some messier testing). It just seems like this PR builds on the other one and there is a lot of overlap with testing results. Curious what you think!

I haven't yet gotten to take a deeper dive into the code. I will look more closely once the behavioral issues are figured out. Let me know if you have any questions!

amyleadem (changes requested):
Hi @mahoneycm, 

I like the possibilities of this improvement but I ran into some screen reader issues during testing:

- [ ] Confirm the value updates as expected in Chrome
    - With keyboard navigation in Chrome with VoiceOver, I see that the `aria-valuetext` is updating as expected, but sometimes the readout does not happen as expected. Sometimes it will get stuck on a value, skip a value, or read the wrong value.
    - Generally the Chrome VoiceOver experience is greatly improved over `develop` with this PR, but users would likely be confused by the incorrect readouts.
    - This behavior matches what I found in **[USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) PR**
- [ ] Confirm the value updates as expected in Safari
    - With keyboard navigation in Safari with VoiceOver, I cannot change the value from “20 out of 100” even though visually the value appears to change
    image
    - This is a regression because the values are read out as expected in `develop`
    - This is also a regression from the **[USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) PR**
- [ ] Confirm the value updates as expected in Firefox
    - With keyboard navigation in Firefox with VoiceOver, I can advance the value successfully, but I run into trouble when I try to move backwards. Sometimes the value doesn’t change at all, sometimes it jumps all the way to 0.
    - Generally the Firefox VoiceOver experience is greatly improved with this PR, but the quirks still prevent it from being fully usable
    - This behavior matches what I found in **[USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) PR**
- [ ] Confirm the value updates as expected in iOS Safari
    - With touch navigation in iOS Safari with VoiceOver gestures, it will only read “20 percent of 100” when I swipe up or down as instructed (Similar to macOS safari).
    - This is a regression. In `develop`, the value changes as expected with up/down swipe.
    - This is also a regression from the **[USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) PR**

I also had some additional questions about Storybook and the GitHub issues:

- For Storybook, I am a bit confused about the unit test story. Is it necessary? Is it possible to consolidate it with the main range story?
- I am wondering if it makes sense to focus on [USWDS - Range: Remove redundant ARIA attributes](https://github.com/uswds/uswds/pull/5413) first and then look at [USWDS - Range: Enhance SR callouts](https://github.com/uswds/uswds/pull/5472). Either that or we combine the PRs (Which I know makes for some messier testing). It just seems like this PR builds on the other one and there is a lot of overlap with testing results. Curious what you think!

I haven't yet gotten to take a deeper dive into the code. I will look more closely once the behavioral issues are figured out. Let me know if you have any questions!

mahoneycm:
@amyleadem I still need to conduct some more iOS testing but I did swap the `input` event listener for `change` which should resolve the Safari bug.

I was not able to replicate the other browser bugs but I believe that is due to the "Quick nav" mode we discussed in #5413.

RE: iOS VoiceOver, I did find this note on the W3C a11y patterns range slider page.
>  **Warning**
> Some users of touch-based assistive technologies may experience difficulty utilizing widgets that implement this slider pattern because the gestures their assistive technology provides for operating sliders may not yet generate the necessary output. To change the slider value, touch-based assistive technologies need to respond to user gestures for increasing and decreasing the value by synthesizing key events. This is a new convention that may not be fully implemented by some assistive technologies. Authors should fully test slider widgets using assistive technologies on devices where touch is a primary input mechanism before considering incorporation into production systems.

So we _may_ need to include swipe specific event listeners to trigger key events. Have you worked on any other tickets that have had to have specific swipe inputs?

mahoneycm:
@amyleadem I'm able to confirm the regression that I cannot use the single drag functionality to increase or decrease the slider on this branch, but I was able to on `develop`.

I was able to use the `double tap, hold, then drag to increase or decrease` on both branches but it didn't hear callouts on either.

I checked out Android TalkBack while I was at it and luckily that screen reader was able to handle the increases and callouts!

Since this is a regression on iOS, it's probably a good idea to get this sorted before merge. Do you have any experience with JS changes preventing iOS VoiceOver commands?

mahoneycm:
@amyleadem Ok so had a weird hunch about it being a Storybook JS initialization issue and I think it might be. I tested on an iPad again today and found the same issue, then refreshed the page, and tried again and it worked as expected!

It's odd that it prevented the input though so definitely something to keep an eye out for. Let me know if it works for you after a refresh and we can discuss next steps!

amyleadem (changes requested):
@mahoneycm, I am still working on reviewing the code but wanted to report that after testing the component with quick nav turned off, the component is now working beautifully across browsers with VoiceOver! :tada: :tada: :tada:

One note: I feel like it is safe to remove the unit test story. The unit test and default stories seem to be redundant. What do you think?

Here is what I tested:
- [x] Confirm the value updates as expected in Chrome
- [x] Confirm the value updates as expected in Safari
- [x] Confirm the value updates as expected in Firefox
- [x] Confirm the value updates as expected in iOS Safari
- [x] All controls work as expected
    - Was able to successfully customize unit, min, max, and step
- [x] Ran `npm start` `npm run build:html`, and `test:unit` without error

I'll come back later to review the code.

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,52 @@
+const selectOrMatches = require("../../uswds-core/src/js/utils/select-or-matches");
+const behavior = require("../../uswds-core/src/js/utils/behavior");
+
+const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
+
+const RANGE_CLASSNAME = `${PREFIX}-range`;
+const RANGE = `.${RANGE_CLASSNAME}`;
+
+/**
+ * Called when range slider has optional unit data attribute.
```
amyleadem, line 10: It would be helpful to be a bit more specific here.
```suggestion
 * Called when range slider includes the data-optional-unit attribute.
```

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+    },
+  },
+};
+
+const range = behavior(rangeEvents, {
+  init(root) {
+    selectOrMatches(RANGE, root).forEach((rangeSlider) => {
+      enhanceSRContext(rangeSlider);
+    });
+  },
```
amyleadem, line 49: Do we want to export `updateCallout`?

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+ * Collects necessary attributes, then constructs and appends aria-valuetext attribute for screen reader context.
+ *
+ * @param {HTMLInputElement} targetRange - The range slider input element
+ */
+const updateCallout = (targetRange) => {
+  const rangeSlider = targetRange;
+  const unit = rangeSlider.getAttribute("data-optional-unit");
+  const val = rangeSlider.value;
+  const max = rangeSlider.getAttribute("max");
+  const callout = `${val} ${unit || ""} of ${max}`;
```
amyleadem, line 20: Wondering if we could add multi-lingual support here by adding an optional custom text data attribute for "of"

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+  rangeSlider.setAttribute("aria-valuetext", callout);
+};
+
+/**
+ * Checks for optional unit data attribute. If attribute is present, updates aria-valuetext and adds input event listener.
```
amyleadem, line 26: This comment feels like it describes `updateCallout`. Maybe it could more directly describe `enhanceSRContext`?

amyleadem (changes requested):
This is looking good, @mahoneycm 👍 

I had just a few notes: 
- I added a few questions and comments to `index.js`. They all are pretty small items. 
- Unit tests look good. Way to figure it out!
- In the top of the PR description, can you add details about the new `data-optional-unit` attribute? It would be helpful to explicitly spell out a description of the attribute as well as default values, etc.

packages/usa-range/src/index.js
mahoneycm: Resolved in [9e0d349](https://github.com/uswds/uswds/pull/5472/commits/9e0d349c6e591069db7a4e237b3ffdb9c19d2257)

packages/usa-range/src/index.js
mahoneycm: That's probably for the best!
Resolved in [c540d35](https://github.com/uswds/uswds/pull/5472/commits/c540d3560069eacd2a6a01376d6efa5bdd863344)

packages/usa-range/src/index.js
mahoneycm: Tbh the `enhanceSRContext` became a bit redundant. Simply calling `updateCallout` during initialization gets the job done as expected. I've removed `enhanceSRContext`. 

Let me know if you see any issue with that!

Resolved in [5d2afeb](https://github.com/uswds/uswds/pull/5472/commits/5d2afebc1f75c5b931e13619a15e123a098b0b1e)

packages/usa-range/src/index.js
mahoneycm: I went ahead and added this in [cbfe406](https://github.com/uswds/uswds/pull/5472/commits/cbfe406e881c73ca35466a6b681b468c31425e41)

The language for the numbers will still rely on users system settings, but this will allow devs to update the language as expected!

mahoneycm:
@mejiaj Made a couple updates on this PR. Updated the description and testing instructions as well. Bumping for review when you have the time!

amyleadem (changes requested):
This works well, @mahoneycm! I tested this successfully on Firefox, Chrome, and Safari with VoiceOver. I had a few small notes/questions for you before approval:

- [x] Confirmed that the data attributes correctly update the unit and preposition text in the `aria-valuetext` readout
- [x] Confirmed that the `aria-valuetext` value updates when the slider value changes in Chrome, Firefox, and Safari
- [ ] Confirmed that the data attributes follow a predictable naming structure
    - I’d like to consider removing “optional” from the naming structure, unless we are using it elsewhere. I’ve added a comment below with more details
- [ ] Confirmed that the story controls are logical and work as expected
    - The controls work well, but I am wondering if we can add a name for the preposition control. I’ve added a comment below with more details.
- [ ] Confirmed unit tests as logical and work as expected
    - The existing unit tests work well. Could we also have a test confirming the preposition data attribute works?

packages/templates/usa-base/_base.twig
```diff
@@ -1,4 +1,4 @@
-{% set lang = lang ? lang : 'en' %}
+{% set lang = lang ? lang : 'es' %}
```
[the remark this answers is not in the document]
mahoneycm: This appears to be an unintentional change while I was doing some language testing. Thank you for flagging!

Resolved in [80ce058](https://github.com/uswds/uswds/pull/5472/commits/80ce058b3fb2e26160fcc9759ff7df3bc1a29dd4)

packages/usa-range/src/usa-range.stories.js
```diff
@@ -8,12 +8,35 @@ export default {
       control: { type: "radio" },
       options: ["none", "disabled", "aria-disabled"],
     },
+    optional_unit: {
+      name: "Unit",
+      control: { type: "text" },
+    },
```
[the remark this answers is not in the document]
mahoneycm: I like this! Implemented in [aa25daa](https://github.com/uswds/uswds/pull/5472/commits/aa25daab8abf7203b269ccf214b753fa0ffd1fc0)

packages/usa-range/src/usa-range.twig
```diff
@@ -10,6 +10,7 @@
     value="20"
     role="slider"
     {% if optional_unit %}data-optional-unit="{{ optional_unit }}"{%- endif %}
+    {% if optional_preposition %}data-optional-preposition="{{ optional_preposition }}"{%- endif %}
```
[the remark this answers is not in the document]
mahoneycm: That's a good point. Looking at other components, it looks like the data- attributes are largely optional in the first place. I've updated the name as you recommended! I'll be sure to update the guidance as well!

Data attributes updated in [aa25daa](https://github.com/uswds/uswds/pull/5472/commits/aa25daab8abf7203b269ccf214b753fa0ffd1fc0)

mahoneycm (commented):
@amyleadem Great ideas. Changes made and ready for re-review! Updating guidance next 👍

mejiaj (changes requested):
Nice work, especially appreciate the unit test and controls for testing. Added comments for code clarity and improvements.

### What I tested

- [x] Correct information displays on initial load when setting data attributes.
- [x] Readout is correct when updating range via click/drag.
- [x] StorybookJS controls work.

### How I tested

- VoiceOver on Chromium 118
- Initially **only** set `data-text-unit`.
- Later on, added `data-text-preposition`.
- Changed `step` from `10→1`.

### Notes

Only noticed that the readout still uses plural (on single item - _example: test one in range slider_).

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+const behavior = require("../../uswds-core/src/js/utils/behavior");
+
+const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
+
+const RANGE_CLASSNAME = `${PREFIX}-range`;
+const RANGE = `.${RANGE_CLASSNAME}`;
+
+/**
+ * Called when range slider includes the data-optional-unit attribute.
+ * Collects necessary attributes, then constructs and appends aria-valuetext attribute for screen reader context.
```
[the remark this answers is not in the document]
mahoneycm: Thanks for this recommendation! Clear to see the benefit here. I'll keep this in mind for future comments 👍 

Resolved in [944db02](https://github.com/uswds/uswds/pull/5472/commits/944db02390bb73bacc057a7fdbd323fa5f0a7da4)

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+const updateCallout = (targetRange) => {
+  const rangeSlider = targetRange;
+  const optionalPrep = rangeSlider.getAttribute("data-text-preposition");
+  const unit = rangeSlider.getAttribute("data-text-unit");
+  const val = rangeSlider.value;
+  const max = rangeSlider.getAttribute("max");
+
+  let prep;
+
+  if (rangeSlider.getAttribute("data-text-preposition")) {
```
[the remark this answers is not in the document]
mahoneycm: This is resolved by the short-circuit OR operator recommendation.

Resolved in [944db02](https://github.com/uswds/uswds/pull/5472/commits/944db02390bb73bacc057a7fdbd323fa5f0a7da4)

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+    prep = optionalPrep;
+  } else {
+    prep = "of";
+  }
```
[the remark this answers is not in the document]
mahoneycm: I'm a fan of this solution! Resolved in [944db02](https://github.com/uswds/uswds/pull/5472/commits/944db02390bb73bacc057a7fdbd323fa5f0a7da4)

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+  },
+};
+
+const range = behavior(rangeEvents, {
+  init(root) {
+    selectOrMatches(RANGE, root).forEach((rangeSlider) => {
+      updateCallout(rangeSlider);
+    });
+  },
+  updateCallout,
```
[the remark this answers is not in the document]
mahoneycm: I feel like giving users the ability to update the callout via JS could be useful in a few situations.
- Could be used when/if the range value is updated via JS
   - `aria-valuetext` would need to be updated to this new value.
   - Alternatively, users would have to call the `change` event as well.
   - Consider the unit test example:
    ```js
        slider.value = "30";
        EVENTS.change(slider);
     ```
    - Instead, users can update the value then call the `updateCallout` method.
- `updateCallout` could also be used if the range needs to change what it represents.
   - For example, if the form had a toggle to change between a percentage-based slider or a numbered unit. 

As for teardown, should I create a teardown function to remove the appended `aria-valuetext` value?

packages/usa-range/src/usa-range.stories.js
```diff
@@ -8,12 +8,39 @@ export default {
...
       options: ["none", "disabled", "aria-disabled"],
     },
+    text_unit: {
+      name: "Unit",
+      control: { type: "text" },
+    },
+    text_preposition: {
+      name: "Preposition",
+      control: { type: "text" },
+    },
```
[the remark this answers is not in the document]
mahoneycm: I wasn't sure the _best_ way to do this within the controls. I added examples to the control label. Let me know what you think!

Resolved in [c63d900](https://github.com/uswds/uswds/pull/5472/commits/c63d90022000e3d636f975333e3549eb988fcf5e)

packages/usa-range/src/test/sr-callout--preposition.spec.html
```diff
@@ -0,0 +1,14 @@
...
+  <input
+    id="usa-range"
+    class="usa-range"
+    type="range"
+    min="0"
+    max="100"
+    step="10"
+    value="20"
+    role="slider"
+    data-text-unit="por ciento"
```
amyleadem, line 11: Since this is a separate test, I'm wondering if it makes sense to only test the `data-text-preposition` attribute here.

```suggestion
```

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,53 @@
...
+const updateCallout = (targetRange) => {
+  const rangeSlider = targetRange;
+  const defaultPrep = "of";
+  const optionalPrep = rangeSlider.dataset.textPreposition;
+  const prep = optionalPrep || defaultPrep;
+  const unit = rangeSlider.dataset.textUnit;
+  const val = rangeSlider.value;
+  const max = rangeSlider.getAttribute("max");
+
+  const callout = `${val} ${unit || ""} ${prep} ${max}`;
```
amyleadem, line 31: This creates an extra empty space when there is no unit. For example, `20  de 100`. This doesn't appear to affect the readout in VoiceOver. I don't think this needs to be a blocker but wanted to flag it.

amyleadem (changes requested):
Everything is working well! Just a couple of small notes in the comments below. There are also some merge conflicts that need to be cleaned out.

packages/usa-range/src/index.js
mahoneycm: Ooh that's a little annoying. Do you think we should update the unit tests to match? Or do you think I should revert to an `if/else` ?

amyleadem: My preference is to edit the script so it outputs the expected number of spaces, but I don't want it to be a blocker. So if you can make it work, let's do it!

mahoneycm: I agree, I'd rather not have the extra space. My only other thought was adding a space to the `unit` value but that seemed messy... 

Resolved in [30e1d2b](https://github.com/uswds/uswds/pull/5472/commits/30e1d2b65d0389b012c4d00c52a46e985fe12592)

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+ * @param {HTMLInputElement} targetRange - The range slider input element
+ */
+const updateCallout = (targetRange) => {
+  const rangeSlider = targetRange;
+  const defaultPrep = "of";
+  const optionalPrep = rangeSlider.dataset.textPreposition;
+  const prep = optionalPrep || defaultPrep;
+  const unit = rangeSlider.dataset.textUnit;
+  const val = rangeSlider.value;
+  const max = rangeSlider.getAttribute("max");
```
amyleadem, line 29: One more thought: if no `max` attribute is present on the component, the readout will give the value of `null`. For example, "20 of null". Wondering if we could put a default value there (the default value for the max attribute appears to be 100 - [Source MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#validation)) or a console error.

packages/usa-range/src/test/sr-callout--preposition.spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+const rangeSliderSelector = () => document.querySelector(".usa-range");
+
+const tests = [
+  { name: "document.body", selector: () => document.body },
+  { name: "range slider", selector: rangeSliderSelector },
+];
+
+tests.forEach(({ name, selector: containerSelector }) => {
+  describe(`Range slider component initialized at ${name}`, () => {
```
amyleadem, line 28: All range unit tests have the same description. It can be hard to parse which test is for which. Can you make these more descriptive? It might be helpful to follow the model in the unit tests for date picker.

amyleadem (changes requested):
@mahoneycm the component is working well. Thanks for removing the extra space from the readout.

I have resolved the comments that have been addressed and left a comment on the remaining comment. I also found a few more small items. Let me know if you have any questions.

packages/usa-range/src/index.js
mahoneycm: This is a great catch! Added a default value in [b06a84f](https://github.com/uswds/uswds/pull/5472/commits/b06a84f772290ec6f15aa5645256c762199467cc) 👍

packages/usa-range/src/test/sr-callout--preposition.spec.js
mahoneycm: Added some clarity to unit test names in [39cca2b](https://github.com/uswds/uswds/pull/5472/commits/39cca2ba6addbabb78c58cc4728800d1beea3fbe). Thanks for the reference!

packages/usa-range/src/test/sr-callout--preposition.spec.js
```diff
@@ -0,0 +1,74 @@
...
+      });
+
+      afterEach(() => {
+        body.textContent = "";
+      });
+
+      it("adds screen reader callout on initialization", () => {
+        assert.ok(
+          valueText,
+          "adds aria-valuetext attribute with updated preposition"
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [39cca2b](https://github.com/uswds/uswds/pull/5472/commits/39cca2ba6addbabb78c58cc4728800d1beea3fbe)

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+ * @param {HTMLInputElement} targetRange - The range slider input element
+ */
+const updateCallout = (targetRange) => {
+  const rangeSlider = targetRange;
+  const defaultPrep = "of";
+  const optionalPrep = rangeSlider.dataset.textPreposition;
+  const prep = optionalPrep || defaultPrep;
+  const unit = rangeSlider.dataset.textUnit;
+  const val = rangeSlider.value;
+  const max = rangeSlider.getAttribute("max") || 100;
```
amyleadem, line 29:
```suggestion
  // Note: 100 is the native default value for the max attribute on range inputs
  const max = rangeSlider.getAttribute("max") || 100;
```

packages/usa-range/src/index.js
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Update range callout for screen readers using the optional data attributes.
+ *
+ * Get optional data attributes, construct and appends aria-valuetext attribute.
+ *
+ * @example
+ *
+ * <input id="usa-range" class="usa-range" type="range" min="0" max="100" step="10" value="20" role="slider" data-text-unit="degrees">
+ *
+ * Callout returns "20 degrees of 100."
```
amyleadem, line 18:
```suggestion
 * <input id="usa-range" class="usa-range" type="range" min="0" max="100" step="10" value="20" data-text-unit="degrees">
 *
 * Callout returns "20 degrees of 100."
 * 
 * <input id="usa-range" class="usa-range" type="range" min="0" max="100" step="10" value="20" data-text-preposition="de">
 * 
 * Callout returns "20 de 100."
```

amyleadem (approved):
This is great, @mahoneycm! I tested this successfully on Firefox, Chrome, and Safari with VoiceOver. 

@amycole501 was also able to re-test this on NVDA and JAWS today and confirmed that the readout is still working as expected.

I resolved all review comments after confirming that they have been addressed. There is only one [open comment](https://github.com/uswds/uswds/pull/5472#discussion_r1380666425) at this time related to teardown, but I don’t believe any action is needed other than considering if we do in fact need a teardown method here. 

I confirmed the following:

- [x] Confirmed that the data attributes correctly update the unit and preposition text in the `aria-valuetext` readout
- [x] Confirmed that the data attributes follow a predictable naming structure
- [x] Confirmed that the `aria-valuetext` value updates when the slider value changes in Chrome, Firefox, and Safari
- [x] Confirmed that the story controls are logical and work as expected
- [x] Confirmed unit tests as logical and work as expected
- [x] Incorporates the attribute updates from https://github.com/uswds/uswds/pull/5413

mahoneycm:
@mejiaj Bumping for re-review when you have time!

packages/usa-range/src/index.js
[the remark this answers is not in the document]
mejiaj: Right, I was just curious to see if there were enough DOM changes to merit a teardown.

mejiaj (approved):
I appreciate the thorough code comments. No issues found in functionality. 

Thank you!

thisisdano (approved):
Nice work!

PR #4062, opened 2021-03-02
USWDS - Tooltip: positioning fix #3963

scottqueen-bixal:
# USWDS - Tooltip: positioning fix
closes Tooltip positioning #3915

## Description
- JS, not enough case logic to handle top and bottom fall back position. ie. if the content is not visible on top or bottom, it should allow for left or right positioning if possible
- Styles, the constrained class that gets added before the position update occurs is not strict enough, it shouldn't be able to have a width that is too small.

## This work;

- improve the positioning logic by checking all available directions first, then compressing only if it does not fit on the screen
- updates `.usa-tooltip_body--wrap` to include a max-width to keep the pseudo-element from being too small for the content

# Test Coverage
Because these updates require multiple/various calculations of "client" values, (ie. clientHeight, offsetWidth) they would be best covered with e2e web-based functional testing, rather than mocked through our current JSDOM driven unit tests. [See this request for JSDom](https://github.com/jsdom/jsdom/issues/135) also reference [Unimplemented parts of the web platform](https://github.com/jsdom/jsdom#unimplemented-parts-of-the-web-platform) via JSDom documentation.

1. Navigate to : [test/tooltip-utilities](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/gscottqueen-gsq-tooltip-fix/components/detail/tooltip-utilites.html)
2a. Hover/tab over the "RSS" icon to display the tooltip, ensure that it is in line with the description across screen sizes
2b. Hover/tab over the "Show on Right "`<button>` icon to display the tooltip, ensure that it is in line with the description across screen sizes

Before you hit Submit, make sure you’ve done whichever of these applies to you:

- [ ] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [x] Run `npm test` and make sure the tests for the files you have changed have passed.
    -  currently failing a snyk test because of del ^6.0
- [x] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [x] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #5475, opened 2023-08-29
USWDS - Tooltip: Add opacity:0 to initial state

### LINKED ISSUE #4458: USWDS - Bug: Tooltip opacity is not set to 0
### Describe the bug
When implementing the [ReactUSWDS](https://github.com/trussworks/react-uswds) version of the Tooltip component, I came across what I believe to be a regression bug. The functionality for positioning & showing the Tooltip is [described](https://github.com/uswds/uswds/blob/main/src/js/components/tooltip.js#L36) as setting opacity to 0 so the correct position can be calculated, and then adding the `is-visible` class to set opacity to 1 once ready - this prevents any flickering while the tooltip UI is being shown.

However, while working on the React component I encountered flickering and while debugging, I looked in the USWDS CSS for where the tooltip body gets set to 0 opacity and couldn't find anything.

It looks like this CSS declaration used to exist, but was maybe removed by mistake when refactoring how the Tooltip positioning works (see https://github.com/uswds/uswds/pull/4062/files#diff-db6545d240cf59c3fa7f1d5249e4c31f85afba05625594e10a762044b98c54fbL31).

It's difficult to reproduce this bug since it may only happen if the Tooltip body needs to be re-positioned, which can happen very quickly. However, you can turn off the `is-visible` class in a dev environment and see that it has no effect on the visibility of the Tooltip.

### Expected behavior
I would expect the Tooltip body to have `opacity: 0` so that adding the `is-visible` class toggles its visibility.

### Screenshots
n/a

### System setup
- **USWDS version:** since 2.11.0
- **Device:** Desktop
- **Operating system:** Mac OS 12
- **Browser and version:** Chrome

### Additional context
n/a

amyleadem:
# Summary

**Restored the `opacity: 0` style rule to the tooltip body's initial state.** This prevents the component from flickering if its position needs to be recalculated. 

## Breaking change

## Related issue

Closes #4458

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2265)

## Preview link

[Tooltip component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-tooltip-opacity-0/?path=/story/components-tooltip--tooltip)

## Problem statement
To prevent flickering in the tooltip component, the tooltip is programmed to receive an opacity shift slightly after the tooltip positioning is set. [This comment](https://github.com/uswds/uswds/blob/develop/packages/usa-tooltip/src/index.js#L36-L37) explains that the initial state should be set to `opacity: 0` so that any position changes will occur when the component is visibly hidden. 

However, the tooltip does not receive an initial style of 0 opacity, which can result in flickering.

It looks like this CSS declaration used to exist, but was possibly removed by mistake when refactoring how the tooltip positioning works (see the code changes in [this PR](https://github.com/uswds/uswds/pull/4062/files#diff-db6545d240cf59c3fa7f1d5249e4c31f85afba05625594e10a762044b98c54fbL31)).

## Solution

Reinstating the initial `opacity: 0` style rule should allow the `is-visible` class to perform as expected and prevent flickering in the tooltip. 

## Testing and review
- Open the tooltip component and confirm that the tooltip toggles visibility as expected.
- Confirm that toggling the `is-visible` class in web inspector shows and hides the component as expected.

### DIFF
--- a/packages/usa-tooltip/src/styles/_usa-tooltip.scss
+++ b/packages/usa-tooltip/src/styles/_usa-tooltip.scss
@@ -26,7 +26,7 @@ $triangle-size: 5px;
   color: color($theme-tooltip-font-color);
   display: none;
   font-size: size("ui", $theme-tooltip-font-size);
-  opacity: 0;
+  opacity: 0; // Required for recalculating position.
   padding: units(1);
   pointer-events: none;
   width: auto;
--- a/packages/usa-tooltip/src/styles/_usa-tooltip.scss
+++ b/packages/usa-tooltip/src/styles/_usa-tooltip.scss
@@ -26,6 +26,7 @@ $triangle-size: 5px;
   color: color($theme-tooltip-font-color);
   display: none;
   font-size: size("ui", $theme-tooltip-font-size);
+  opacity: 0;
   padding: units(1);
   pointer-events: none;
   width: auto;


mahoneycm (approved):
Code is working great!

I had to add the class `is-set` in addition to `is-visible` while testing the class additions, but hover toggling worked as expected!

packages/usa-tooltip/src/styles/_usa-tooltip.scss
```diff
@@ -26,6 +26,7 @@ $triangle-size: 5px;
   color: color($theme-tooltip-font-color);
   display: none;
   font-size: size("ui", $theme-tooltip-font-size);
+  opacity: 0;
```
mejiaj, line 29: We could add a note to avoid this in the future.

```suggestion
  opacity: 0; // Required for recalculating position.
```

mejiaj (commented):
LGTM, just a minor note on documenting why we need the style rule.

mejiaj (approved):

thisisdano (approved):
Simple and effective!

Issue #5434, opened 2023-08-10
USWDS - Bug: Responsive stacked table cannot be parsed by VoiceOver on Safari


devowhippit:
### Describe the bug

I was looking into using the [responsive stacked table](https://designsystem.digital.gov/components/table/#responsive-stacked-table-2) this week, and I found that I couldn't use VoiceOver on Safari to read the table contents at all. I was wondering if I was using voice-over controls wrong, but after looking into it a bit more, it seems like a documented issue (I was not aware of before) that modifying the display properties of tables makes them inoperable with this SR+Browser combination.

Using Safari v16.5.2 on MacOS.

[Bug and screen-reader support reference and documentation](https://adrianroselli.com/2022/07/its-mid-2022-and-browsers-mostly-safari-still-break-accessibility-via-display-properties.html).

### Steps to reproduce the bug

1. Navigate to a [responsive stacked table example](https://designsystem.digital.gov/components/table/#responsive-stacked-table-2) using a MacOS + Safari
2. Use keyboard commands VO (ctrl + opt) + Left Arrow to navigate to the table example. VoiceOver will announce the table caption and display a visible focus indicator on the table.
3. VoiceOver will instruct to use VO (ctrl + opt) + Shift + Down Arrow to enter into the table. 
4. When entering into the table, however, VoiceOver will announce the table as empty, and the cells can't be navigated. 

### Expected Behavior

At step 4, the user should be able to navigate the table cells using VO (ctrl + opt) + arrow keys.

### Screenshots

![Screenshot 2023-08-10 at 2 19 10 PM](https://github.com/uswds/uswds/assets/3826974/7a898a2c-38b1-41ce-afac-7fca9f51683d)
![Screenshot 2023-08-10 at 2 19 41 PM](https://github.com/uswds/uswds/assets/3826974/9444cc11-29b5-4c26-882a-72988f124825)

### System setup

- Using Safari v16.5.2 on MacOS
- USWDS v3.5
- USWDS public documentation site

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

mejiaj:
[...]

PR #5477, opened 2023-08-30
USWDS - Table: Allow stacked table to be readable with VoiceOver + Safari combo

### LINKED ISSUE #5464: USWDS maturity model page updates
The [USWDS maturity mode](https://designsystem.digital.gov/maturity-model/)l page needs some updates, including around evaluating and updating/removing the [USWDS maturity assessment worksheet](https://designsystem.digital.gov/files/uswds-maturity-assessment-worksheet_v0.3.pdf).

Make the appropriate updates to this page before the September monthly call, so that users who visit the page as a result of the call have a good experience.

Work to be done as part of this issue:

- Evaluate page for timeliness and make appropriate updates
- Update links and headers to remove ambiguity
- Review for plain language and take opportunity to improve where possible

mahoneycm:
# Summary

Fixed a bug that prevented VoiceOver from reading stacked table content while using Safari.

## Breaking change

This is not a breaking change

## Related issue

Closes https://github.com/uswds/uswds/issues/5434

## Related pull requests

[Changelog →](https://github.com/uswds/uswds-site/pull/2264)

## Preview link

[Stacked table →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-stacked-table-voice-over-fix/?path=/story/components-table-stacked--default)

## Problem statement

Specifically when using **VoiceOver** on **Safari**, tables with `display: block` on the <tr> are skipped or read as empty.

This is caused by a well known bug[^1] which causes tables with `display` properties to break accessibility with screen readers.

[^1]: https://adrianroselli.com/2022/07/its-mid-2022-and-browsers-mostly-safari-still-break-accessibility-via-display-properties.html

This bug used to affect most browsers, but in recent years has been resolved. Safari seems to be the only browser still affected by this. At the time of this PRs creation the referenced article was last updated: `August 14th, 2023.`

Additional mdn doc PR with a well organzied list of associated bugs: https://github.com/mdn/browser-compat-data/issues/19994

## Solution

I was able to improve the screen reader experience by removing `display: block` from stacked table `tr` tags.

Now, VoiceOver + Safari can properly navigate through the stacked table’s content and read it’s values

### Differences

There is one slight difference between the way VO + Safari read the table data compared to other browsers.

Other browsers will read cells data labels with “column # of #” depending on how many data cells there are in each row. In our example, this would be “column 1 of 3” for example.

Safari now reads this as “Column 1 of 1”. While this differs from the other browsers, it is considered accurate because there is only one column.

The downside is that it doesn’t let the user know how many “columns” (if table wasn't stacked) are within the said “row”.

VoiceOver ****will**** correctly read out how many “rows” (sections) of the stacked table there are.

This is caused by a Safari bug 
> Tables with display: block on the <td> and <th> give the wrong column count.[^2]
[^2]: https://github.com/mdn/browser-compat-data/issues/19994

### Confirm existence of bug

1. Visit [develop branch stacked table variant **********on Safari********** →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-table-stacked--default)
2. Activate VoiceOver
3. Attempt to navigate table using keyboard inputs
   - Note that VO key + up/down arrows will [jump between rows](https://dequeuniversity.com/screenreaders/voiceover-keyboard-shortcuts#vo-mac-tables:~:text=Move%20up%20or%20down%20in%20a%20column)
5. Visit the stacked table variant on ****another browser****
6. Confirm proper readouts

### Confirm improvement

1. Visit [preview **********************on Safari →**********************](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-stacked-table-voice-over-fix/?path=/story/components-table-stacked--default)
2. Confirm no visual regression
3. Attempt to navigate the table with VoiceOver and keyboard inputs
4. Confirm you are able to access all cell information
5. Repeat for ******************************another browser******************************
7. Confirm no unexpected changes

### Testing checklist

- [ ]  Stacked table is now accessible using ********VoiceOver + Safari********
- [ ]  Confirm we’re comfortable with the readout addressing each row as a single column
- [ ]  No unexpected changes to other browsers
- [ ]  No visual regressions

### DIFF
--- a/packages/uswds-core/src/styles/mixins/general/table.scss
+++ b/packages/uswds-core/src/styles/mixins/general/table.scss
@@ -148,7 +148,6 @@ $table-unsorted-icon-color: get-color-token-from-bg(
   tr {
     border-bottom: units(0.5) solid color($theme-table-border-color);
     border-top-width: 0;
-    display: block;
     width: 100%;
     th:first-child,
     td:first-child {


mahoneycm:
@amycole501 Yeah if you don't mind comparing to the develop builds with other browsers/screen readers to confirm no accidental readout changes that would be great!

amyleadem (approved):
@mahoneycm, this is looking good! I like that there is a clean, one-line answer to an important bug fix. 

I don’t see a big issue with Safari reading the content out as one column. It is not ideal but it is better than what we have now and does match the visual presentation. 

I tested on Chrome, Safari, and Firefox with VoiceOver in MacOS. I also tested in iOS Safari with VoiceOver Gestures. I tested the following: 

- [x] Confirmed existence of the bug with VoiceOver in Safari on MacOS and VoiceOver Gestures Safari on iOS
- [x] Confirm that the stacked table content is now accessible with VoiceOver in Safari VoiceOver Gestures in iOS Safari
- [x] Confirm no visual regression in Safari and other browsers
- [x] Confirm the readout makes sense
- [x] Confirm no visual or screen reader regression in other browsers

amycole501 (commented):
The table sounds great in JAWS in Chrome. No issues detected and all rows were read correctly. I'm currently having issues with NVDA reading the table. Am employing some help from colleagues to see if it's just an issue with my copy of NVDA or if it's the table causing nothing to be read.

mahoneycm:
@amycole501 Glad to hear it! Would you mind approving the PR if it looks good to you? Whenever you get the chance 😊

thisisdano (approved):
Clean fix, great documentation!

PR #5491, opened 2023-09-07
USWDS - Identifier: Update "An official website" screen reader text

### LINKED ISSUE #5362: USWDS - Enhancement: Improve "An official website" text in identifier
## Summary
In the identifier component, we recommend using the text "_An official_ website of the Parent agency". There is a risk that this text could lead to confusion for screen reader users, who might hear "_un_official" instead of "an official". 

## Related issue
We changed the same "An official" text in the banner component in issue https://github.com/uswds/uswds/issues/4419 and PR https://github.com/uswds/uswds/pull/4925.

amyleadem:
# Summary

**Updated the screen reader readout to say "Official" instead of "An official".** When read out on a screen reader, the statement "An official website of [Agency name]" can sound like "UNofficial website of [Agency name]". To minimize confusion, we hid the word "An" from screen readers with `aria-hidden`. 

## Breaking change

:warning: This requires a markup change for English-language implementations of the identifier. 

In English versions of the identifier component, users should update their markup to hide the word "An" from screen readers by wrapping the word in a span with an  `aria-hidden` attribute.

Before
```html
<p class="usa-identifier__identity-disclaimer">
    An official website of the <a href="">[Parent agency]</a>
</p>
```

After
```html
<p class="usa-identifier__identity-disclaimer">
    <span aria-hidden="true">An </span>official website of the <a href="">[Parent agency]</a>
</p>
```

## Related issue

Closes #5362

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2266)

## Preview link

- [Identifier component (Default)](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-identifier-text/?path=/story/components-identifier--default)
- [Identifier (Multiple parents)](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-identifier-text?path=/story/components-identifier--multiple-parents-and-logos)
- [Identifier (No logo)](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-identifier-text?path=/story/components-identifier--no-logos)
- [Identifier (Taxpayer disclaimer)](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-identifier-text/?path=/story/components-identifier--taxpayer-disclaimer)

## Problem statement
When read out on a screen reader, the statement "An official website of the [Agency name]" can sound like "UNofficial website of [Agency name]". 
 
## Solution
To minimize the impact on visual presentation, we created a screen reader-only solution that hides "An" from the readout. Instead the screen reader will now read out "official website of the [Agency name]".

## Testing and review

- Using a variety of browsers and screen readers, confirm that the screen reader readout is intuitive and makes sense.
- Confirm that the new aria change is added only to English versions of the component.
- Consider: Is it confusing that it is only for English implementations? Anything we should do to add clarity?

### DIFF
--- a/packages/usa-identifier/src/content/usa-identifier.json
+++ b/packages/usa-identifier/src/content/usa-identifier.json
@@ -5,7 +5,8 @@
   "masthead": {
     "aria_label": "Agency identifier,",
     "description": "Agency description,",
-    "content": "An official website of the",
+    "content_prefix": "An",
+    "content": "official website of the",
     "parent": {
       "name": "<Parent agency>",
       "shortname": "<Parent shortname>",
--- a/packages/usa-identifier/src/content/usa-identifier~multiple-logos.json
+++ b/packages/usa-identifier/src/content/usa-identifier~multiple-logos.json
@@ -1,10 +1,12 @@
 {
   "modifier": "",
   "domain": "domain.gov",
+  "lang": "en",
   "masthead": {
     "aria_label": "Agency identifier,,,",
     "description": "Agency description,,,",
-    "content": "An official website of the",
+    "content_prefix": "An",
+    "content": "official website of the",
     "parent": {
       "name": "<Parent agency>",
       "shortname": "<Parent shortname>",
--- a/packages/usa-identifier/src/content/usa-identifier~no-logos.json
+++ b/packages/usa-identifier/src/content/usa-identifier~no-logos.json
@@ -5,7 +5,8 @@
   "masthead": {
     "aria_label": "Agency identifier,,",
     "description": "Agency description,,",
-    "content": "An official website of the",
+    "content_prefix": "An",
+    "content": "official website of the",
     "parent": {
       "name": "<Parent agency>",
       "shortname": "<Parent shortname>",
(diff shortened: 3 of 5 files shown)


mahoneycm (approved):
Lgtm! I like the direction you took with the fix!

- [x] Sounds great with VoiceOver across browsers
- [x] Spanish variant is unexpected

If we include a guidance update as to why we left it out, we can maybe add instructions to consider all languages used by the site and to reduce confusion for screen reader users where possible

mejiaj:
Have we discussed the possibility of removing it all together so that it's `Official website…`? That way the language would be consistent regardless of translation.

CC @amycole501 @thisisdano

thisisdano:
We have discussed changes to the boilerplate, but banner language is a big question that we'll need to address comprehensively and consistently. I do think we'll want to ask this question, and build evidence toward a better solution, but I also think we'll want to leave the discussion until after the Web Components version is out so we can consider the implementation angle at the same time.

mejiaj (approved):
Code looks good, I compared component to Identifier on [develop storybookJS](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-identifier--default).

thisisdano (approved):

PR #5493, opened 2023-09-08
USWDS - Modal: Remove padding style rules after modal close

### LINKED ISSUE #5361: USWDS - Bug: Modal removes existing padding on close
In storybook we have a right padding set on body. When modal checks for `body.style.paddingRight` we get an empty string. This causes a visual regression where the incorrect padding is set and then removed all together.

### Related code

https://github.com/uswds/uswds/blob/aceec3346a3e8f8b40acb4af043bccca498c60c4/packages/usa-modal/src/index.js#L129-L132

## Examples

<details><summary>Initial padding</summary>
<p></p>

image

</details> 

<details><summary>Padding gone on modal close</summary>
<p></p>

image
</details> 

## Steps to reproduce
1. Go to [default modal example](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?p=all&args=&id=components-modal--default&viewMode=story) in develop.
2. Resize browser to `400px`.
3. Confirm `16px` padding on right side.
4. Open and close modal.
5. Confirm right padding is now gone.

## Additional context

Originally found when testing PR #5315.

amyleadem:
# Summary
**Updated how the modal JavaScript resets body padding.** Now the component will rely on the CSS to reset the body padding rather than injecting an inline style with JavaScript. 

**Improved body padding detection in modal JavaScript.** The JS now checks for body padding after the modal has been set up. 

## Breaking change

## Related issue

Closes #5361

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2273)

## Preview link

- [Modal component in Storybook](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-storybook-modal-padding/iframe.html?args=&id=components-modal--default&viewMode=story)
- [uswds-sandbox test integration](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/al-test-modal-padding/pages/modal/)
- [uswds-site test integration](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-test-modal-padding/components/modal/)

## Problem statement
When the modal opens, it adds a temporary padding to the body to account for the missing scrollbar. When the modal closes, it is expected that the padding resets to the body's initial padding definition. For example, the Storybook body has an initial padding value of `1rem`, and the expectation is that the body styles would return to `1rem` when the modal is closed. However, in the current state in Storybook, the modal JS changes the body padding to `0px` when the modal is closed.

This appears to only happen in Storybook and could not be replicated in `uswds-sandbox` or `uswds-site`. I believe it is related to how Storybook initializes. 

## Solution

1. The JavaScript now reverts back to the original CSS styling when the modal is closed, rather than declaring another inline style. Here is a description of the before and after (note the change in the last step of the process):
    - In `develop`, the component changes the body padding using the following methods:
        - The initial body padding is set with CSS (For example, 12px)
        - Once the modal is activated, the JS adds an inline style rule to the body element that updates the padding to include the width of the scrollbar (For example, 27px).
        - Then when the modal is closed the **_JS resets the definition of the inline style rule to the the initial value of 12px_**. 
    - In this PR, the component changes the body padding using the following methods:
        - The initial body padding is set with CSS (For example, 12px)
        - Once the modal is activated, the JS adds an inline style rule to the body element that updates the padding to include the width of the scrollbar (For example, 27px).
        - Then when the modal is closed the JS **_removes the padding style rule that it added and the padding is defined with the original CSS again._**
2. The JavaScript now explicitly checks for body padding style during the modal is set up. This allowed the JS to recognize the body padding in Storybook.
    - :warning: I am not sure if this move comes with any downsides. I'd appreciate input!

## Testing and review
- Confirm that the body padding behaves as expected when the modal is opened/closed.
- Confirm no regressions in modal behavior. 
- Consider if there is any risk to this change.

In [Storybook](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-storybook-modal-padding/iframe.html?args=&id=components-modal--default&viewMode=story):
1. Resize the window to 450px
2. Check the initial padding on the body element (Should be 1em, defined in the CSS)
3. Open the modal
4. Check the updated padding on the body element (Should be 32px, defined in an inline style rule)
    - Note: The 32px value is created by adding 17px (the computed padding on the body) and 15px (the computed width of the scrollbar)
6. Close the modal
7. Confirm that the inline body padding style was removed and that the body padding is restored to the original value, defined by the CSS (1em)

In the [sandbox demo](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/al-test-modal-padding/pages/modal/) or [USWDS-site demo](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-test-modal-padding/components/modal/):
1. Check the initial padding on the body element (Should be 38px, defined in the CSS)
2. Open the modal
4. Check the updated padding on the body element (Should be 53px, defined in an inline style rule)
6. Close the modal
7. Confirm that the body padding is restored to the original value (Should be 38px, defined in the CSS)

### DIFF
--- a/packages/usa-modal/src/index.js
+++ b/packages/usa-modal/src/index.js
@@ -125,10 +125,11 @@ function toggleModal(event) {
   // Account for content shifting from body overflow: hidden
   // We only check paddingRight in case apps are adding other properties
   // to the body element
-  body.style.paddingRight =
-    body.style.paddingRight === TEMPORARY_PADDING
-      ? INITIAL_PADDING
-      : TEMPORARY_PADDING;
+  if(body.style.paddingRight === TEMPORARY_PADDING){
+    body.removeAttribute("style");
+  } else{
+    body.style.paddingRight = TEMPORARY_PADDING;
+  }
 
   // Handle the focus actions
   if (safeActive && openFocusEl) {
@@ -165,7 +166,6 @@ function toggleModal(event) {
     returnFocus.focus();
     modal.focusTrap.update(safeActive);
   }
-
   return safeActive;
 }
 
--- a/packages/usa-modal/src/index.js
+++ b/packages/usa-modal/src/index.js
@@ -134,7 +134,7 @@ function toggleModal(event) {
   // Temporarily increase body padding to include the width of the scrollbar.
   // This accounts for the content shift when the scrollbar is removed on modal open.
   if (body.style.paddingRight === TEMPORARY_BODY_PADDING) {
-    body.removeAttribute("style");
+    body.style.removeProperty("padding-right");
   } else {
     body.style.paddingRight = TEMPORARY_BODY_PADDING;
   }
--- a/packages/usa-modal/src/index.js
+++ b/packages/usa-modal/src/index.js
@@ -96,11 +96,11 @@ function toggleModal(event) {
     // If it doesn't have an ID, make one
     // Store id as data attribute on modal
     if (clickedElement.hasAttribute(OPENER_ATTRIBUTE)) {
-      if (clickedElement.getAttribute("id") === null) {
+      if (this.getAttribute("id") === null) {
         originalOpener = `modal-${Math.floor(Math.random() * 900000) + 100000}`;
-        clickedElement.setAttribute("id", originalOpener);
+        this.setAttribute("id", originalOpener);
       } else {
-        originalOpener = clickedElement.getAttribute("id");
+        originalOpener = this.getAttribute("id");
       }
       targetModal.setAttribute("data-opener", originalOpener);
     }
--- a/packages/usa-modal/src/index.js
+++ b/packages/usa-modal/src/index.js
@@ -28,13 +28,6 @@ let modal;
 
 const isActive = () => document.body.classList.contains(ACTIVE_CLASS);
 const SCROLLBAR_WIDTH = ScrollBarWidth();
-const INITIAL_PADDING = window
-  .getComputedStyle(document.body)
-  .getPropertyValue("padding-right");
-const TEMPORARY_PADDING = `${
-  parseInt(INITIAL_PADDING.replace(/px/, ""), 10) +
-  parseInt(SCROLLBAR_WIDTH.replace(/px/, ""), 10)
-}px`;
 
 /**
  *  Is bound to escape key, closes modal when
@@ -49,7 +42,7 @@ const onMenuClose = () => {
  * @param {KeyboardEvent} event the keydown event
  * @returns {boolean} safeActive if mobile is open
  */
-function toggleModal(event) {
+function toggleModal(event, temporaryBodyPadding) {
   let originalOpener;
   let clickedElement = event.target;
   const { body } = document;
@@ -87,11 +80,11 @@ function toggleModal(event) {
     // If it doesn't have an ID, make one
     // Store id as data attribute on modal
     if (clickedElement.hasAttribute(OPENER_ATTRIBUTE)) {
-      if (this.getAttribute("id") === null) {
+      if (clickedElement.getAttribute("id") === null) {
         originalOpener = `modal-${Math.floor(Math.random() * 900000) + 100000}`;
-        this.setAttribute("id", originalOpener);
+        clickedElement.setAttribute("id", originalOpener);
       } else {
-        originalOpener = this.getAttribute("id");
+        originalOpener = clickedElement.getAttribute("id");
       }
       targetModal.setAttribute("data-opener", originalOpener);
     }
@@ -122,13 +115,12 @@ function toggleModal(event) {
     body.classList.toggle(PREVENT_CLICK_CLASS, safeActive);
   }
 
-  // Account for content shifting from body overflow: hidden
-  // We only check paddingRight in case apps are adding other properties
-  // to the body element
-  if (body.style.paddingRight === TEMPORARY_PADDING) {
+  // Temporarily increase body padding to include the width of the scrollbar.
+  // This accounts for the content shift when the scrollbar is removed on modal open.
+  if (body.style.paddingRight === temporaryBodyPadding) {
     body.removeAttribute("style");
   } else {
-    body.style.paddingRight = TEMPORARY_PADDING;
+    body.style.paddingRight = temporaryBodyPadding;
   }
 
   // Handle the focus actions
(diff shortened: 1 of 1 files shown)


packages/usa-modal/src/index.js
```diff
@@ -87,11 +80,11 @@ function toggleModal(event) {
     // If it doesn't have an ID, make one
     // Store id as data attribute on modal
     if (clickedElement.hasAttribute(OPENER_ATTRIBUTE)) {
-      if (this.getAttribute("id") === null) {
+      if (clickedElement.getAttribute("id") === null) {
```
amyleadem, line 83: I started getting an error related to `this` once I passed a parameter to the click event in line 308. I updated this to `clickedElement` and wanted to confirm that this was the correct replacement. I see that it still creates and assigns an opener `id` as needed.

packages/usa-modal/src/index.js
```diff
@@ -28,13 +28,6 @@ let modal;
 
 const isActive = () => document.body.classList.contains(ACTIVE_CLASS);
 const SCROLLBAR_WIDTH = ScrollBarWidth();
-const INITIAL_PADDING = window
-  .getComputedStyle(document.body)
-  .getPropertyValue("padding-right");
-const TEMPORARY_PADDING = `${
-  parseInt(INITIAL_PADDING.replace(/px/, ""), 10) +
-  parseInt(SCROLLBAR_WIDTH.replace(/px/, ""), 10)
-}px`;
```
amyleadem, line 37: > **Note**
> I moved these items so that the styles would be captured after the modal is set up in line 278.

packages/usa-modal/src/index.js
```diff
@@ -284,6 +277,18 @@ modal = {
...
+      // Value is created by checking for initial body padding
+      // and adding the width of the scrollbar.
+      const initialBodyPadding = window
+        .getComputedStyle(document.body)
+        .getPropertyValue("padding-right");
+      const temporaryBodyPadding = `${
+        parseInt(initialBodyPadding.replace(/px/, ""), 10) +
+        parseInt(SCROLLBAR_WIDTH.replace(/px/, ""), 10)
+      }px`;
+
```
mahoneycm, line 291: I'm curious... have we tried removing this entirely?

The padding on my body didn't match up with your values in the testing. On storybook it was 16px instead of 32px and sandbox stayed at 38px.

I noticed that there weren't any visible changes so I did some light testing on removing the initial padding check and update. I didn't notice any content shift or any unintended behaviors!

I was curious if you saw something different than I do or if you've tried removing this portion of the javascript in general 🤔

mahoneycm:
Flagging that we might have to update #5315 to include changes made by this PR

packages/usa-modal/src/index.js
amyleadem: @mahoneycm Could you try logging the values for `SCROLLBAR_WIDTH`, `temporaryBodyPadding`, and `initialBodyPadding` after the modal initializes and let me know what you see? I am curious if it is registering `SCROLLBAR_WIDTH` as 0.

mahoneycm: You've got good intuition!

Screenshot 2023-09-14 at 2 02 44 PM

packages/usa-modal/src/index.js
mejiaj: Leaving this at the top as-is or via a function helps with clarity. I'd avoid adding more to `init()` if we can avoid it.

mejiaj (changes requested):
I was able to confirm the original issue is specific to StorybookJS and its known init issues.

### How I confirmed

- **Using StorybookJS:** Confirming padding was reported as `0` initially. We call `start.js` in `.storybook/preview.js` to initialize components. In this file we're checking for two DOM states; one for DOMReady and another for DOM _already_ loaded [best practice outlined in [Document: DOMContentLoaded event - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete)].
- **USWDS Sandbox:** Added padding and got the correct values on toggle.

---

It seems like a better experience to move the padding check closer to the modal toggle. Is it possible to handle padding when `toggleModal()` is called?

amyleadem:
@mejiaj I moved the padding check to `setUpModal()` because we really just want to capture and reuse the initial value. Curious what you think!

mahoneycm (approved):
Values are differing for me but it IS working as expected!

- [x] Initial padding matches up
- [x] Opening the modal does not cause content shift
- [x] Opening modal updates body's padding via inline styles
- [x] Closing the modal reverts padding to initial state!

packages/usa-modal/src/index.js
```diff
@@ -122,13 +131,13 @@ function toggleModal(event) {
...
-  // We only check paddingRight in case apps are adding other properties
-  // to the body element
-  body.style.paddingRight =
-    body.style.paddingRight === TEMPORARY_PADDING
-      ? INITIAL_PADDING
-      : TEMPORARY_PADDING;
+  // Temporarily increase body padding to include the width of the scrollbar.
+  // This accounts for the content shift when the scrollbar is removed on modal open.
+  if (body.style.paddingRight === TEMPORARY_BODY_PADDING) {
+    body.removeAttribute("style");
```
mejiaj, line 137: Is it possible to only reset `paddingRight` instead of removing styles altogether?

mejiaj (changes requested):
Small question on the removal of _all_ styles on body, otherwise LGTM.

packages/usa-modal/src/index.js
amyleadem: Updated `removeAttribute` to `removeProperty` in https://github.com/uswds/uswds/pull/5493/commits/3fac64c4b43b71242f4b5f87e11c27725a944071

mejiaj (approved):
Nice fix, thanks!

thisisdano (approved):

PR #5499, opened 2023-09-12
USWDS - uswds-core: Font normalization not matching USWDS specifications

nickjg231:
# Summary

We want to normalize the Bitter header font in the VA Design System to reflect the normalization of fonts already in the USWDS. The USWDS has documentation on how one may apply tokens to get the correct normalization settings. However, when trying to apply those settings to existing fonts so that they match the USWDS specifications, the fonts render to the page with different values than expected.

See the example below:
image

For the first three columns, we have existing fonts. In the first section titled USWDS, these are the `line-height` and `font-size` values that should be generated when using a `size` token of `10` and a line-height token of 3, according to the [Normalization and line height table defined in designsystem.digital.gov](https://designsystem.digital.gov/design-tokens/typesetting/overview/#normalization-and-line-height-2). 

When these fonts were rendered using those token values however, the Computed section shows the actual output to the page. The `line-height` values were all off by small, but noticeable amounts. 

After some investigation, it seems that the issue is that in the line-height normalization calculations, we are rounding line-height to a tenths place instead of a hundredths place. By rounding with too little precision, we are altering the values that are rendered to the page by a noticeable amount. This pull request fixes the discrepancy between the USWDS defined values for normalized font-size and line-height versus the values that are rendered to the page.

## Breaking change

## Related issue

Closes #_[issue_no]_

## Related pull requests

_Indicate if there are other pull requests necessary to complete this issue._

## Problem statement

This pull request fixes the line-height value calculated in the normalization logic so that it more closely matches the USWDS defined values.

## Solution

## Major changes

## Testing and review

### DIFF
--- a/packages/uswds-core/src/styles/functions/utilities/line-height.scss
+++ b/packages/uswds-core/src/styles/functions/utilities/line-height.scss
@@ -38,8 +38,8 @@
     (math.div($system-base-cap-height, $this-cap))
   );
   $normalized-line-height: math.div(
-    math.round($normalized-line-height * 10),
-    10
+    math.round($normalized-line-height * 100),
+    100
   );
 
   @return $normalized-line-height;


amyleadem:
We are leaving this in triage until we can confirm and understand the related issue.

mahoneycm:
Hello @nickjg231 !

Thank you so much for your contribution and for reporting these inconsistencies!

After looking into the history of the line height function, it appears to have always rounded to the first decimal place instead of the second. Since changing the function results in a visual change (albeit relatively small) we've decided to update the table to reflect the actual values USWDS is currently outputting, rather than changing USWDS to match the table.

Closing out this PR in favor of that solution.

Thanks again!

PR #5513, opened 2023-09-21
USWDS - Header: Allow mobile nav to stay open during new page load

### LINKED ISSUE #5058: USWDS - Bug: Selection of a link in mobile menu is not smooth
### Describe the bug

When a user chooses a page in the mobile menu, the expectation is for the next view that appears to the user to be the new, chosen page. Instead, the menu closes. The current (old) page appears for a time. Then, the new, chosen page loads. The result is flashy and clunky. On a website like USA.gov, where each page is so similar in appearance, it can also be confusing.
The current www.usa.gov has the expected behavior. Check it out! (The screenshot .gif attached below is for a test version of the next usa.gov on federalist. It uses the mobile menu closing behavior of USWDS).
You can also see the flashy mobile menu closing experience on the USWDS website, but I didn't screenshot a gif of that.
Our mobile menu gets used a lot, and this behavior is not a smooth experience.

Use a mobile phone or mobile replicator in dev tools.
Go to USWDS website like https://designsystem.digital.gov/design-principles/.
Open Mobile Menu.
Choose a new page.
Observe that the mobile menu closes, revealing old page. Then the new page loads.

### Expected Behavior

Mobile menu stays open when new page is clicked. Then new page loads.

### Related code

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

mahoneycm:
# Summary

**Improved mobile nav behaviors.** The mobile nav menu will now stay open while new pages load to reduce confusion caused by content updating after the menu closes.

## Breaking change

## Related issue

Closes #5058

## Related pull requests

[Changelog →](https://github.com/uswds/uswds-site/pull/2282)

## Preview link

[Storybook link →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-mobile-nav-close-behavior/?path=/story/components-header--default) 

[Site demo →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-5513-demo/)

[Sandbox demo →](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/cm-5513-demo/)

## Problem statement

After clicking a link within the mobile nav, the nav menu closes and there is a pause before the new page is loaded. This can cause confusion, especially when pages are similar in appearance.

From #5058:

> When a user chooses a page in the mobile menu, the expectation is for the next view that appears to the user to be the new, chosen page. Instead, the menu closes. The current (old) page appears for a time. Then, the new, chosen page loads. The result is flashy and clunky. On a website like [USA.gov](http://usa.gov/), where each page is so similar in appearance, it can also be confusing. It seems like the request for the new page has not happened.
> 

<details><summary>Gif example</summary>
<p>

_Credit to @DK1999_

</p>
</details> 

## Solution

Enhance header logic to compare clicked links targets to the current location `href`.

- **href paths are the same**: immediately close the navigation window.
- **href paths are different**: keep the navigation open while the page loads.

Changes affect mobile nav only and desktop view should remain unaffected. Accordions _should_ still close when links, page body, or other targets are clicked.

### Mobile nav calculation

This solution uses `!isActive()` to check that the active mobile nav class is not present. This allows us to prevent accordion buttons from closing when their child link is clicked and the menu stays open. 

### isChildSection calculation

Using the [URL web API](https://developer.mozilla.org/en-US/docs/Web/API/URL), we can compare `pathnames` of the current page and the clicked link. If they are the same, it deems the link as a subsection of the current page.

### Unit tests

I’ve updated the nav unit tests to include tests for the new mobile interactions and to confirm previous desktop only interactions.

I also removed a test that no longer fit the expected behaviors due to this enhancement.

## Major Changes

- Mobile navigation now remains open during new page loads
- Behaviors refactored into individual functions
- Accordion is no longer imported into `header`

## Testing and review

> [!TIP]
> When testing links to new pages, it may be helpful to throttle your connection to more easily view the mobile nav staying open during page load.

### USWDS Site demo

In this demo I recommend the card page because it has nested subsections in the mobile nav, but feel free to click through other areas of the site to test functionality after confirming intended behaviors through the steps below.

1. Visit [site demo build →](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-5513-demo/)
2. Reduce window size to allow nav to break into mobile version.
3. Using the mobile nav, navigate to the **card** component page.
4. Confirm that the nav stays open while the page loads.
5. Once on the card component page, reopen nav and select one of the subsection links (i.e. preview).
6. Confirm that the nav immediately closes and page properly jumps to the hash target.

### Sandbox Demo

Sandbox features the default nav with dropdown sections. We want to ensure the desktop dropdown functionality remains the same!

1. Visit [sandbox demo build →](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/cm-5513-demo/)
2. Confirm desktop actions are unchanged.
    1. Open accordion link dropdown.
    2. Confirm clicking body, in-page link, new page link, or other elements on the page closes dropdown menu.
3. Confirm mobile nav actions are updated.
    1. Open menu.
    2. Open the accordion link dropdown.
    3. Select subsection and confirm nav closes.
    4. Re open nav and click either the in-page or new page link.
    5. Confirm menu stays open while page loads.
    6. Reopen the nav menu and the accordion link dropdown.
    7. Repeat steps for the nested page link and confirm accordion and nav does not close.

### Testing Checklist

- [ ]  Change is considered a valid enhancement to header/nav.
- [ ]  Code is up to USWDS standards.
    - [ ]  Approve mobile nav check is valid.
    - [ ]  Approve child subsection check.
    - [ ] Approve removal of `accordion` package import.
- [ ]  Desktop interactions are unchanged.
- [ ]  Mobile menu collapses when links to subsections are clicked.
- [ ]  Mobile menu stays open when links to new pages are clicked.
- [ ]  New unit tests are approved.
- [ ]  All unit tests pass.


packages/usa-header/src/index.js
```diff
@@ -158,19 +160,41 @@ const handleEscape = (event) => {
...
+          if (navActive !== this) {
+            hideActiveNavDropdown();
+          }
+          // store a reference to the last clicked nav link element, so we
+          // can hide the dropdown if another element on the page is clicked
+          if (!navActive) {
+            navActive = this;
+            toggle(navActive, true);
+          }
         }
```
mahoneycm, line 198: I was hoping to refactor this into a function, but doing so seemed to break functionality.

packages/usa-nav/src/test/template.html
```diff
@@ -24,13 +24,13 @@
           </button>
           <ul id="basic-nav-section-one" class="usa-nav__submenu">
             <li class="usa-nav__submenu-item">
-              <a href="#">Navigation link</a>
+              <a href="javascript:void(0)">Navigation link</a>
```
mahoneycm, line 27: Changing to `javascript:void(0)` allowed me to simulate links to pages without having the test leave the current page.

packages/usa-nav/src/test/template.html
```diff
@@ -44,19 +44,19 @@
...
             </li>
           </ul>
         </li>
         <li class="usa-nav__primary-item">
-          <a class="usa-nav__link" href="javascript:void(0)">
-            <span>Simple link</span>
+          <a id="section-link" href="#">
+            Link to section header
```
mahoneycm, line 59: Using `#` allowed me to simulate subsection header navigation

mejiaj (changes requested):
Thanks for the tests on sandbox and site, they were helpful!

### PR Description

I updated the formatting in the **Solution** section for readability. 

**Suggestions**
1. Can you add the release note? 
2. Would you mind adding a plain language pass to the problem statement?

### Code

No issues with functionality. Added suggestions to improve readability and clarity in logic.

packages/usa-header/src/index.js
```diff
@@ -188,57 +189,83 @@ const handleEscape = (event) => {
...
-        // can focus on whatever they've just selected.
-
-        // Some navigation links are inside accordions; when they're
-        // clicked, we want to collapse those accordions.
-        const acc = this.closest(accordion.ACCORDION);
-
-        if (acc) {
-          accordion.getButtons(acc).forEach((btn) => accordion.hide(btn));
-        }
-
```
mahoneycm, line 224: ## Removed accordion import

- Accordion was required to use the `hide` function on nav link clicks.
- New functions properly collapse accordions without the import.
- Removing the `accordion` import did not cause any regressions.
- Removing `accordion` from `uswds-core` module exports did not cause any regressions not found on current `develop` branch.

Are there any other concerns with this change that I may have missed? Should accordion behaviors be imported by default in case a user does not import the accordion package?

packages/usa-header/src/index.js
```diff
@@ -188,57 +189,83 @@ const handleEscape = (event) => {
...
+ */
+const isChildSection = (targetLink) => {
+  const baseURL = new URL(CURRENT_PAGE);
+  const linkDestination = new URL(targetLink.href);
+  if (baseURL.pathname === linkDestination.pathname) {
+    return true;
+  }
+
+  return false;
+};
```
mahoneycm, line 206: ## isChildSection() logic

This test now uses the `pathname` property to compare the URL path before any hash symbols [^1]

[^1]: https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname

mahoneycm (commented):
## Update 1/12/24

- Removed accordion import.
- Renamed `NAV_CONTROL` → `NAV_ACCORDION`
    - In my opinion this is more descriptive
- Separated behaviors into individual functions
    - Each has JS Comments providing context
- Added **Major Changes** section to PR description

### Storybook preview links

Currently, the twig template sets header nav links to `javascript:void(0)`. Clicking these links will no longer close the mobile navigation because it doesn't pass "child section" test.

We can simulate links closing the mobile nav by switching the `href` to `#`. The most noticeable change that would come from this is that it would affect the component code preview on site.

packages/usa-header/src/index.js
```diff
@@ -210,21 +234,25 @@ navigation = behavior(
...
+          }
 
-        // If the mobile navigation menu is active, we want to hide it.
-        if (isActive()) {
-          navigation.toggleNav.call(navigation, false);
+          // If the mobile navigation menu is active, we want to hide it.
+          if (isActive()) {
+            navigation.toggleNav.call(navigation, false);
+          }
         }
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [132f322](https://github.com/uswds/uswds/pull/5513/commits/132f322796ce2b57dfac68b98d0d9ea315fabe52)

packages/usa-header/src/index.js
```diff
@@ -188,19 +190,41 @@ const handleEscape = (event) => {
   focusNavButton(event);
 };
 
+/**
+ * Takes clicked link and current url, removes any hash from the link, and compares them.
+ * If link destination is determined to be the same page; return true.
+ * If link destination is determined to be on a separate page; return false.
+ *
+ * @param {HTMLAnchorElement} targetLink - Target link clicked in header.
+ * @returns {boolean}
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [132f322](https://github.com/uswds/uswds/pull/5513/commits/132f322796ce2b57dfac68b98d0d9ea315fabe52)

packages/usa-header/src/index.js
```diff
@@ -188,19 +190,41 @@ const handleEscape = (event) => {
...
+          // If another nav is open, close it
+          if (navActive !== this) {
+            hideActiveNavDropdown();
+          }
+          // store a reference to the last clicked nav link element, so we
+          // can hide the dropdown if another element on the page is clicked
+          if (!navActive) {
+            navActive = this;
+            toggle(navActive, true);
+          }
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [132f322](https://github.com/uswds/uswds/pull/5513/commits/132f322796ce2b57dfac68b98d0d9ea315fabe52)

packages/usa-header/src/index.js
```diff
@@ -188,57 +189,84 @@ const handleEscape = (event) => {
...
-        if (navActive !== this) {
-          hideActiveNavDropdown();
-        }
-        // store a reference to the last clicked nav link element, so we
-        // can hide the dropdown if another element on the page is clicked
-        if (!navActive) {
-          navActive = this;
-          toggle(navActive, true);
-        }
+      [NAV_ACCORDION]() {
```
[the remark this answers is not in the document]
mejiaj: Normally I'd agree, but the class doesn't have anything to do with accordion. That _could_ cause some confusion.

https://github.com/uswds/uswds/blob/66f99928a0fb802a922f3ac33323f208e243042a/packages/usa-header/src/index.js#L17

mejiaj (changes requested):
I'm not able to see the feature anymore using your site demo link: [Accordion](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-5513-demo/components/accordion/#accordion-package)

[USWDS - Header- Allow mobile nav to stay open during new page load #5513.webm](https://github.com/uswds/uswds/assets/3385219/a579a023-9066-47f5-9e04-e6a116f1795d)

I would recommend adding somewhere the example URLs/links where you'd want to leave nav open.

packages/usa-header/src/index.js
```diff
@@ -188,57 +189,83 @@ const handleEscape = (event) => {
...
+ * @returns {boolean} Return `true` if target is same page || Return `false` if target is on a different page.
+ */
+const isChildSection = (targetLink) => {
+  const baseURL = new URL(CURRENT_PAGE);
+  const linkDestination = new URL(targetLink.href);
+  const currentURLPath = baseURL.origin + baseURL.pathname;
+  const linkURLPath = linkDestination.origin + linkDestination.pathname;
+
+  return currentURLPath === linkURLPath;
+};
```
mahoneycm, line 206: ## Improve is child section logic

RE: @mejiaj comment:

>Can you share an example use case?
>
>If we need to check that the links href is part of the current page there are a few ways to do this. For example, you could also use includes() to compare strings.
>
>Pseudocode example
 ```js
const isChildSection = (targetLink) => (window.location.href).includes(targetLink.href);
```
> If you only need to read the values you can do baseURL.pathname and targetLink.pathname.

I found that using `include` causes unintentional behavior when the link path is deeper but contains the previous path.

Take the patterns page on our site for example:

```
designsystem.digital.gov/ → /patterns/ → /patterns/intro
```

Since each page includes the previous path, each time we click into a nested page from the mobile nav, it will close despite it loading a new page. 

Instead we should check that the base url and path before `#` match exactly to indicate it is in fact the same page.

I improved the check to include the link origin (base url).

Let me know if you have any other questions or concerns! 👍

**Comment by mahoneycm** [comment]
> I'm not able to see the feature anymore using your site demo link: [Accordion](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/cm-5513-demo/components/accordion/#accordion-package)
> 
>  USWDS.-.Header-.Allow.mobile.nav.to.stay.open.during.new.page.load.5513.webm 
> I would recommend adding somewhere the example URLs/links where you'd want to leave nav open.

@mejiaj  You're actually viewing intended behavior. In this video, you are clicking subsection links which should immediately close the mobile nav.

If you instead click a mobile nav link to a new page, you'll notice the nav stays open while the page loads. 

> [!TIP]
> It may help to throttle your connection to slow the new page load and view the nav behaviors.

File: packages/usa-header/src/index.js (hunk 9 above)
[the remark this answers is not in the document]
Comment (mahoneycm) [reply packages/usa-header/src/index.js:251]: My thought here was that it was a more semantic name given the fact that it is targeting accordion buttons. While we’re using the `button.usa-nav__link` class to target the element, the element shares the `usa-accordion__button class`.

In my opinion, the name `NAV_CONTROL` is more indicative of the mobile menu button than the accordion/dropdown elements it’s targeting. 

What do you think of `NAV_DROPDOWN` or `NAV_DROPDOWN_BUTTON` instead?

File: packages/usa-header/src/index.js (hunk 4 above)
Comment (mahoneycm) [reply packages/usa-header/src/index.js:224]: I don't believe there is any functional change after my additions. The accordions still collapse as expected! This is still accomplished by the `hideActiveNavDropdown` and `toggle` functions.

https://github.com/uswds/uswds/blob/59192241426774230f6d0a86b6218b1bc3cbf779/packages/usa-header/src/index.js#L166-L173

File: packages/usa-header/src/index.js (hunk 9 above)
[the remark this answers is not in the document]
Comment (mejiaj) [reply packages/usa-header/src/index.js:251]: I see, `NAV_DROPDOWN` makes sense. I'll let others chime though before we commit.

**Review by mejiaj — COMMENTED** [review]
Only a question on the use of `new URL()`.

File: packages/usa-header/src/index.js [hunk 11]
```diff
@@ -188,57 +189,83 @@ const handleEscape = (event) => {
...
+/**
+ * Takes current url and the url from the clicked link, then compares link origin and path without any hash to
+ * deterimine if the link is to a subsection of current page.
+ *
+ * @param {HTMLAnchorElement} targetLink - Target link clicked in header.
+ * @returns {boolean} Return `true` if target is same page || Return `false` if target is on a different page.
+ */
+const isChildSection = (targetLink) => {
+  const baseURL = new URL(CURRENT_PAGE);
+  const linkDestination = new URL(targetLink.href);
```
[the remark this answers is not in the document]
mahoneycm: Great catch! I was unaware that I could grab the `origin` and `pathname` from the `window.location` object but it worked like a charm. I went ahead and swapped that out 👍

packages/usa-nav/src/test/navigation.spec.js
```diff
@@ -161,13 +176,6 @@ describe("navigation toggle", () => {
     assert.strictEqual(document.activeElement, menuButton);
   });
 
-  it("collapses dropdowns when focus leaves nav", () => {
-    menuButton.click();
-    navLink.click();
-    EVENTS.focusOut(navLink);
-    assert.strictEqual(isVisible(nav), false);
-  });
```
amyleadem, line 169: We should consider restoring this

packages/usa-nav/src/test/navigation.spec.js
```diff
@@ -163,9 +180,10 @@ describe("navigation toggle", () => {
 
   it("collapses dropdowns when focus leaves nav", () => {
     menuButton.click();
-    navLink.click();
-    EVENTS.focusOut(navLink);
-    assert.strictEqual(isVisible(nav), false);
```
mahoneycm, line 168: ## Invalid test

This test would not properly test dropdowns closing on focus out for a few reasons

### 1. `navLink.click()` closes the nav

Clicking the navigation link closes the entire navigation, not just the dropdowns (consider mobile views). 

To illustrate this, removing `EVENTS.focusOut(navLink)`still causes test to pass.

### 2. Incorrect testing instructions

Instead of testing that the entire nav is closed, we should be checking if the navigation dropdown/accordion is closed.

I was able to emulate this by instead focusing out of `navPrimary` (emulating focusing the search bar for example) and checking the `aria-expanded` attribute.

packages/usa-nav/src/test/navigation.spec.js
```diff
@@ -163,9 +180,10 @@ describe("navigation toggle", () => {
...
+    accordionButton.click();
+    assert.strictEqual(accordionButton.getAttribute("aria-expanded"), "true");
+    EVENTS.focusOut(navPrimary);
+    assert.strictEqual(accordionButton.getAttribute("aria-expanded"), "false");
```
mahoneycm, line 186: ## Valid test

This test properly:

1. Opens navigation
2. Opens navigation dropdown/accordion
3. Confirms dropdown is expanded
4. Focuses out of the `usa-nav__primary` section, which properly emulates current focus change states
5. Confirms the dropdown is now closed

amyleadem (changes requested):
I am seeing some unexpected behavior when I install this branch locally on uswds-site.

When I click on a nested link to a new page in the mobile menu, the accordion that the link is nested inside of closes immediately while the rest of the menu stays open waiting for the page to load. This is different from what I see on the federalist previews. I've added a recording of the behavior below . 

I also see this behavior in Storybook when I add a new page url to a nested menu link. 

Note: the network is throttled to slow 3G in this screen capture:

https://github.com/uswds/uswds/assets/93996430/acf0496f-8bf7-4aa6-bbd3-e4b529173912

mahoneycm:
@amyleadem Can you do me a favor and try running `npm run build` and then testing this again?

I saw some odd behavior on my [test branch](https://github.com/uswds/uswds-site/pull/2281) testing locally but they cleared up after running `npm run build`. I did the same on `main` and the accordions stayed open as expected.

thisisdano:
My gut reaction is that I do think this should block the PR. I'm hesitant to bundle a new distraction along with a subtle (but meaningful) refinement

mejiaj:
@mahoneycm can you create a new issue for the bug discovered so we can address that first?

I'll bump this work out until that's addressed.

ryparker:
@coderabbitai full review

Issue #4191, opened 2021-05-06
Flag icon is low resolution


rfultz:
**Is your feature request related to a problem? Please describe.**

When trying to optimize our site, the flag icon on the left of the usa-banner is setting off 'serves images with low resolution' warnings. The png is only 16 pixels wide in 2021 when every device is ultra high resolution.

It's the first image on every page of our site and represents our country and it's the lowest resolution, blurriest image on the site.

**Describe the solution you'd like**

Could we get larger flag image options? At this point, even exporting the original source at double the current width would help.

**Describe alternatives you've considered**

Should we just use our own image that's as close to this one (and maybe a more accurate blue)? I don't like the idea of only our site using a different flag.

Would it be out of bounds for us to create one/some and submit as pull request(s)?

**Additional context**

There was previously an [issue](https://github.com/uswds/uswds/issues/2190) filed for this but it was over three years ago.

jAigret-Bix:
PRs are welcome, feel free to create one.

aduth:
I had trouble tracking down a higher-resolution source version of this image. I can't find one in this repository, nor in the [Sketch design assets](https://designsystem.digital.gov/download/).

Oddly enough, I did find one in [our project source](https://github.com/18F/identity-idp/blob/main/app/assets/images/us-flag.png):

![image](https://user-images.githubusercontent.com/1779930/124150126-43d96100-da5f-11eb-8837-b6da757516ee.png)

While this could work to scale down and appear nicer for higher pixel-density displays, I think ideally this would be updated to use an SVG rather than PNG. Not sure if that exists anywhere, or if a new one could be created.

aduth:
> Oddly enough, I did find one in [our project source](https://github.com/18F/identity-idp/blob/main/app/assets/images/us-flag.png):

This isn't quite the same graphic; there are more stars and stripes in this version, and the colors are slightly different.

But based the original comment from #2190, I was able to rebuild what I think is the "current version":

<img src="https://gist.githubusercontent.com/aduth/08f3fb95aeb63c63dddc9171062c033d/raw/159987d5bc217f5317436f745e99d3f188f0494d/flag.svg" width="160" height="110">

SVG source:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 160 110">
<rect width="160" height="110" fill="#fff"/>
<rect width="160" height="10" fill="#DD0E07"/>
<rect y="100" width="160" height="10" fill="#DD0E07"/>
[...]

PR #5542, opened 2023-10-04
USWDS - Banner: Optimize banner flag icon, provide vector option

aduth:
# Summary

Replaces the banner flag image with a losslessly compressed version, and adds a new vector (SVG) version of the image.

## Related issue

Continues from #4191, closed by #4792.

## Related PR
[Changelog PR](https://github.com/uswds/uswds-site/pull/2327)

## Problem statement

As an end-user, I expect that pages using the U.S. Web Design system load quickly, so that I don't have to wait for assets to load.

## Solution

Optimizes the existing `us_flag_small.png` image. It was optimized using [Oxipng](https://github.com/shssoichiro/oxipng), via [Squoosh](https://squoosh.app/). The choice to use Oxipng was based on a desire for lossless compression, so there should be no degradation in the quality of the image. Lossy compression is possible, and would result in an even smaller file size (I was able to shrink to 168 bytes using [TinyPNG](https://tinypng.com/)), but this would have some marginal effect on the quality of the image. tl;dr "Lossless" seemed least controversial 😄 

**Result:**

Before: 545 bytes
After: 244 bytes
Diff: -301 bytes (-55%)

This also proposes to add a hand-optimized vector version of the image, originally proposed at https://github.com/uswds/uswds/issues/4191#issuecomment-878580597, and adjusted for this pull request using colors from the higher-quality image added in #4792.

**Why?**

- So that a project can have the option to use a vector image
- Vector images are scalable, and could produce better image appearance at certain sizes or on specific devices
- While the `.svg` is larger size uncompressed than the `.png` file, it is still very small (380 bytes), and would compress to the smallest possible size (195 bytes gzipped, 180 bytes brotli'd)
- Could allow for inlining of the image in markup

## Testing and review

Visually inspect the revised assets.


mejiaj (approved):
Thanks for doing this and the detailed explanation of the solution, very helpful!

Found no visual regressions in Banner component.

**Local (left) | Develop (right)**
image

New flag SVG is included in built `dist` directory. _Tested with `npx gulp build`_.

<details><summary>Screenshot</summary>
<p></p>

image
</details>

mahoneycm (approved):
Lgtm!

- [x] No visual regression
- [x] File size noticeably smaller 
- [x] Both images are correctly compiled into the `dist` directory

thisisdano (approved):
Thank you!

PR #5548, opened 2023-10-06
USWDS - Bump jsdom: Faster unit tests

deckar01:
# Summary

Upgrade to the latest version of jsdom to run the unit tests ~30% faster.

## Breaking change

## Related issue

https://github.com/jsdom/jsdom/issues/3234

## Problem statement

`getComputedStyle` is used in several test utilities and it requires an expensive global traversal of all elements and style sheets. These calls account for ~50% of the runtime of the unit tests.

## Solution

https://github.com/jsdom/jsdom/pull/3482 optimized this operation to cache parent traversals so that subsequent calls in the same DOM run ~75% faster.

## Testing and review

Run the tests locally and compare the execution time of `npx gulp unitTests`.

## Dependency updates

| jsdom |     19.0.0      |   22.1.0   |


amyleadem (changes requested):
@deckar01 Thank you for this contribution! The time to run `npx gulp unitTests` has been noticeably improved. Here are the results on my machine:

- Before: 17s
- After: 10s

This update is looking good, but it looks like a couple of merge conflicts have popped up. Would you be able to resolve those? Once that is complete I can approve the PR. 

Here are the things I tested:

- [x] Confirmed I can successfully run `npm install`, `npm run start`, `npm run test` without error
- [x] Confirmed that 22.1.0 is the most recent version of `jsdom`
- [x] Confirmed that the updated in-page nav unit test runs as expected

amyleadem (approved):
Thanks for the quick update! Looks good to me.

mahoneycm (approved):
Lgtm!

Speed before change: **22s**
Speed after change: **10s**

- [x] No new vulnerabilities introduced
- [x] No errors on install or test
- [x] In-page nav test works as expected

mejiaj (approved):
Thanks for improving this!

thisisdano (approved):
Great improvement, thank you!

PR #5550, opened 2023-10-08
USWDS - Site Alert - Repair top margin, broken by #4922

lpsinger:
# Summary

**Removed the top margin from the site alert component.**

PR #4922 added a top margin to the site alert, which looks wrong when a site alert immediately follows a gov banner.

## Breaking change

## Related PR
[Changelog PR](https://github.com/uswds/uswds-site/pull/2324)

## Preview link

This affects https://test.gcn.nasa.gov.

### USWDS 3.1.0

uswds-3 1 0

### USWDS 3.6.1, before this patch

before

### After this patch

after

## Problem statement

The site alert should not have any top margin. It didn't before USWDS 3.6.1. The margin looks like a glitch when a site alert immediately follows a gov banner.

## Solution

Move the top margin back from the mixin to _usa-alert.scss so that applies to alerts but not site alerts.

## Testing and review

I tested this locally to produce the screenshots above.

## Dependency updates

None.

### DIFF
--- a/packages/usa-alert/src/styles/_usa-alert.scss
+++ b/packages/usa-alert/src/styles/_usa-alert.scss
@@ -13,6 +13,10 @@ $alert-icons: (
 
 .usa-alert {
   @include alert-styles;
+
+  * + & {
+    margin-top: units(2);
+  }
 }
 
 @each $name, $icon in $alert-icons {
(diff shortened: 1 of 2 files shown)


thisisdano:
Great patch. We're going to work to get this into 3.7.0.

amyleadem (approved):
Thank you for this fix! I confirmed that this PR removes the top margin from the site alert component and preserves the top margin on the alert component. 

### Before

Site alert
image

Alert
image

### After
Site alert
image

Alert
image

mahoneycm (approved):
Lgtm! Thanks for taking care of this @lpsinger 👏 

- [x] Site alert no longer has a margin between banner and alert
- [x] Standard alerts are unaffected by this change 
- [x] No visual regressions or build errors

thisisdano (approved):
Thank you!

PR #5567, opened 2023-10-18
USWDS - Section / Dark Background: Use consistent styling for dark background links

aduth:
# Summary

**Use consistent link styling for links in dark background content.** Previously, `usa-section--dark` and `usa-dark-background` did not have consistent styling for links.

## Related discussion

See related discussion in our application at https://github.com/18F/identity-dev-docs/pull/367#discussion_r1364004777

## Solution

Provides the complete set of link pseudo classes between `usa-section--dark` and `usa-dark-background`:

Before:

State|`usa-section--dark`|`usa-dark-background`
---|---|---
Initial|✅|✅ 
`:hover`|✅|✅
`:active`|✅|❌
`:visited`|❌|✅

After:

State|`usa-section--dark`|`usa-dark-background`
---|---|---
Initial|✅|✅ 
`:hover`|✅|✅
`:active`|✅|✅
`:visited`|✅|✅

### DIFF
--- a/packages/usa-dark-background/src/styles/_usa-dark-background.scss
+++ b/packages/usa-dark-background/src/styles/_usa-dark-background.scss
@@ -1,5 +1,7 @@
 @use "uswds-core" as *;
 
+$background-context: "Background";
+
 .usa-dark-background {
   background-color: color("base-darker");
 
@@ -9,6 +11,10 @@
   }
 
   a {
-    @include set-link-from-bg("base-darker");
+    @include set-link-from-bg(
+      "base-darker",
+      $theme-link-reverse-color,
+      $context: $background-context
+    );
   }
 }
--- a/packages/usa-dark-background/src/styles/_usa-dark-background.scss
+++ b/packages/usa-dark-background/src/styles/_usa-dark-background.scss
@@ -9,18 +9,6 @@
   }
 
   a {
-    color: color($theme-link-reverse-color);
-
-    &:hover {
-      color: color($theme-link-reverse-hover-color);
-    }
-
-    &:active {
-      color: color($theme-link-reverse-active-color);
-    }
-
-    &:visited {
-      color: color($theme-link-reverse-color);
-    }
+    @include set-link-from-bg("base-darker");
   }
 }
--- a/packages/usa-dark-background/src/styles/_usa-dark-background.scss
+++ b/packages/usa-dark-background/src/styles/_usa-dark-background.scss
@@ -11,12 +11,16 @@
   a {
     color: color($theme-link-reverse-color);
 
-    &:visited {
-      color: color($theme-link-reverse-color);
-    }
-
     &:hover {
       color: color($theme-link-reverse-hover-color);
+    }
+
+    &:active {
+      color: color($theme-link-reverse-active-color);
+    }
+
+    &:visited {
+      color: color($theme-link-reverse-color);
     }
   }
 }
--- a/packages/usa-section/src/styles/_usa-section.scss
+++ b/packages/usa-section/src/styles/_usa-section.scss
@@ -1,5 +1,7 @@
 @use "uswds-core" as *;
 
+$section-context: "Section";
+
 .usa-section {
   @include border-box-sizing;
   @include u-padding-y($theme-site-margins-width);
@@ -33,6 +35,10 @@
   }
 
   a {
-    @include set-link-from-bg("primary-darker");
+    @include set-link-from-bg(
+      "primary-darker",
+      $theme-link-reverse-color,
+      $context: $section-context
+    );
   }
 }
--- a/packages/usa-section/src/styles/_usa-section.scss
+++ b/packages/usa-section/src/styles/_usa-section.scss
@@ -33,18 +33,6 @@
   }
 
   a {
-    color: color($theme-link-reverse-color);
-
-    &:hover {
-      color: color($theme-link-reverse-hover-color);
-    }
-
-    &:active {
-      color: color($theme-link-reverse-active-color);
-    }
-
-    &:visited {
-      color: color($theme-link-reverse-color);
-    }
+    @include set-link-from-bg("primary-darker");
   }
 }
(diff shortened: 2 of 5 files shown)


amyleadem (approved):
Thank you for this fix! Everything looks good to me. 

I tested by adding the following code to the [Section - dark story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-section--dark) and toggling the `:visited` element state in web inspector: 

```html
<a class="usa-link" href="http://designsystem.digital.gov">Test link</a>
```

Before:
image

After: 
image

It would also be great to see a link added to the `packages/usa-section/src/usa-section.twig` file for testing, but that can be done in a follow-up PR if needed.

aduth:
> It would also be great to see a link added to the `packages/usa-section/src/usa-section.twig` file for testing, but that can be done in a follow-up PR if needed.

Sure! Not sure if you have a preference, but I adopted some content from alert stories which included a link. See 2b0bbb48196b5274e2c93d90ca5137b8faee0d3f.

packages/usa-section/src/styles/_usa-section.scss
```diff
@@ -42,5 +42,9 @@
     &:active {
       color: color($theme-link-reverse-active-color);
     }
+
+    &:visited {
+      color: color($theme-link-reverse-color);
+    }
   }
```
mejiaj, line 49: We could use the `set-link-from-bg` mixin here too.

mejiaj (changes requested):
Thanks for the PR and updating the story!

No issues or regressions found. Added suggestions for using the mixin to set the colors consistently.

packages/usa-dark-background/src/styles/_usa-dark-background.scss
```diff
@@ -11,12 +11,16 @@
...
     }
+
+    &:active {
+      color: color($theme-link-reverse-active-color);
+    }
+
+    &:visited {
+      color: color($theme-link-reverse-color);
+    }
   }
```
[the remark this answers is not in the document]
mejiaj: Resolved in 6db5629.

packages/usa-section/src/styles/_usa-section.scss
mejiaj: Resolved in 6db5629.

mejiaj (approved):
Thanks for the PR. 

Hope you don't mind, the change was so small I pushed up a commit to use the color setting mixin in 6db5629.

aduth:
> Hope you don't mind, the change was so small I pushed up a commit to use the color setting mixin in [6db5629](https://github.com/uswds/uswds/commit/6db5629a788f00bef71c9793478489b0244f6f90).

I don't mind at all Thanks for jumping in to make those changes, and apologies for being slow to respond.

packages/usa-dark-background/src/styles/_usa-dark-background.scss
```diff
@@ -9,14 +9,6 @@
...
-    color: color($theme-link-reverse-color);
-
-    &:visited {
-      color: color($theme-link-reverse-color);
-    }
-
-    &:hover {
-      color: color($theme-link-reverse-hover-color);
-    }
+    @include set-link-from-bg("base-darker");
```
aduth, line 12: @mejiaj Out of curiosity, does switching from `$theme-link-reverse-color` have any impact on the effectiveness of that theme configuration? Or is that theme configuration being deprecated?

mahoneycm, line 12: We could add `$theme-link-reverse-color` as the preferred link color for the mixin. This way it maintains the theme setting when appropriate but will resort to a fallback if need be

```suggestion
    @include set-link-from-bg("base-darker", $theme-link-reverse-color);
```
This results in a lighter link text than what's currently displayed

Current:
Screenshot 2023-12-05 at 3 18 07 PM

With preferred theme token:

Screenshot 2023-12-05 at 3 17 41 PM

mejiaj: Not deprecated, the mixin uses `$theme-link-color` (primary)  and `$theme-link-reverse-color` (base-lighter) by default.

mejiaj: @mahoneycm, I was unable to reproduce this. The mixin uses theme link colors by default and I was still getting `base-lighter` as the value (_for both passing and leaving out theme link reverse color_).

For clarity and consistency I've added:
1. A `$context` variable for future debugging.
2. Explicitly passing `$theme-link-reverse-color` to avoid any confusion.

mahoneycm (approved):
Lgtm! Thanks @aduth and @mejiaj !

amyleadem (approved):
Looks good to me. I had a small note about defining the active state with `set-link-from-bg`, but I don't think it should be a blocker because all link states are now legible.

packages/usa-dark-background/src/styles/_usa-dark-background.scss
```diff
@@ -9,14 +11,10 @@
...
-    }
-
-    &:hover {
-      color: color($theme-link-reverse-hover-color);
-    }
+    @include set-link-from-bg(
+      "base-darker",
+      $theme-link-reverse-color,
+      $context: $background-context
+    );
```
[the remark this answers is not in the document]
mejiaj: @amyleadem good find. Yes, please create an issue.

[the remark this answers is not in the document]
amyleadem: Opened https://github.com/uswds/uswds/issues/5659 for follow-up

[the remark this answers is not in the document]
thisisdano: I believe the reason why this ignores `active` is to allow the original link color to be lighter (since there are fewer lightening steps to account for without `active`).

thisisdano (approved):
Nice! Thank you!

PR #5577, opened 2023-10-25
USWDS - Core: Avoid usa-prose override styles when usa-prose not loaded

aduth:
# Summary

**Reduced output size for projects optimized using component packages that do not include `usa-prose`.**

## Problem statement

As a developer, I expect that if I choose to [optimize my installation to use only packages for components I'm using](https://designsystem.digital.gov/documentation/migration/#6-optimize-your-installation), the compiled output will not include styles for unused components, so that the compiled output is as optimized as possible.

## Solution

The general idea is that we only want to include the `.usa-prose`-wrapped content if `usa-prose` is actually loaded. After referencing [`sass:meta` documentation](https://sass-lang.com/documentation/modules/meta/) and what's [defined in the `usa-prose` styles themselves](https://github.com/uswds/uswds/blob/develop/packages/usa-prose/src/styles/_usa-prose.scss), I landed at `meta.mixin-exists("prose-test")`. Another option might be to define a new mixin or variable explicitly for this check.

### Results:

Results will vary depending on which packages someone chooses to include. The example below uses `usa-process-list`, which includes a [non-insignificant amount of prose-overriding styles](https://github.com/uswds/uswds/blob/48bfdf2960d7cbd043dac5bf71a9616ec1defa65/packages/usa-process-list/src/styles/_usa-process-list.scss#L14-L39).

Test file:

```scss
@use "uswds-core" with ($theme-show-notifications: false);
@forward "usa-process-list";
```

Sass-compressed, raw:

```
sass --load-path=packages --style=compressed example.scss | wc -c
```

**Before:** 5612 bytes
**After:** 5004 bytes
**Diff:** -608 bytes (-10.8%)

Sass-compressed, [Brotli compression](https://en.wikipedia.org/wiki/Brotli) (using [brotli-size-cli](https://www.npmjs.com/package/brotli-size-cli)):

```
sass --load-path=packages --style=compressed example.scss | brotli-size
```

**Before:** 759 bytes
**After:** 735 bytes
**Diff:** -24 bytes (-3.2%)

### DIFF
--- a/packages/uswds-core/src/styles/mixins/helpers/override-prose.scss
+++ b/packages/uswds-core/src/styles/mixins/helpers/override-prose.scss
@@ -1,6 +1,11 @@
+@use "sass:meta";
+
 @mixin override-prose {
   @content;
-  .usa-prose {
-    @content;
+
+  @if meta.mixin-exists("prose-test") {
+    .usa-prose {
+      @content;
+    }
   }
 }


mahoneycm (commented):
This looks like an acceptable approach to me!

`prose-test` was created while refactoring our prose styles to resolve an error that occurred when using it as a placeholder. It essentially just tests to ensure that children of `usa-prose` get the proper styles. 

We questioned if this test was still necessary after the test was resolved but in the nature of catching errors when they occur, I see value in maintaining it. Therefore, I think using the meta `mixin-exists()` function using `prose-test` is valid.

@mejiaj @amyleadem how do you all feel about that? Do you find any value in creating a mixin _specifically for_ checking the existence of prose?

### Testing checklist

- [x] Test is logical and would return true anytime `usa-prose` is included.
- [x] The compiled scss file is noticeably smaller.

mejiaj:
@aduth if we moved `prose-test` to a more appropriate place (like `tests.scss`) this would break, correct?

Would it be more accurate to frame the problem statement more like:
_Don't load prose override styles if I'm not loading usa-prose in my project._

If so, I think we should create an issue where we can address via a setting or some other method.

aduth:
@mejiaj It would break only if it wasn't being loaded in the main stylesheet load path (i.e. if you decided to only load it for tests). And yes, I think you've framed the problem statement correctly.

Personally, I prefer that the generated stylesheets are optimized by default without having to tinker with additional settings, though I take your point to create an issue to discuss further. It'd be fair to say this was proposed mostly as a proof-of-concept, though it could also serve as a temporary solution pending a more "durable" approach 🤷. Happy to close this in favor of an issue if it'd be better to discuss that way.

I do think that even if `prose-test` was meant as a "test" mixin, the general principle here is mostly concerned with finding _something_ that's defined within the prose module that we can test for as an indicator of it being loaded. I could even imagine a new dedicated mixin specifically for this test that doesn't do anything aside from exist for the sole purpose of being tested for.

e.g.

```scss
// packages/usa-prose/src/styles/_usa-prose.scss

@mixin prose-loaded {}
```

ryparker:
I'm closing this during the backlog cleanup because we don't have capacity to finish the review now. The discussion above about detecting whether `usa-prose` is loaded is useful context for reducing unused CSS.

If you'd like to continue, comment here or reopen the PR. I'll help with the rebase and work through the remaining approach with you. Thanks for the contribution, and sorry we left it waiting.

PR #5587, opened 2023-10-30
USWDS - Banner: Remove invalid baseline styling from banner button

aduth:
# Summary

**Removes invalid styles from the banner component.** These styles had no effect and were safely removed.

## Problem statement

As a developer, I expect that USWDS ships valid CSS, so that (a) it is correct, (b) my CSS bundles are optimized only to include what has a meaningful effect on styles, and (c) my code will not be flagged by automated validation tools.

The application of `vertical-align: baseline` is invalid here because the component is also styled as `display: block`

https://github.com/uswds/uswds/blob/48bfdf2960d7cbd043dac5bf71a9616ec1defa65/packages/usa-banner/src/styles/_usa-banner.scss#L239

References:

>Note that vertical-align only applies to inline, inline-block and table-cell elements: you can't use it to vertically align [block-level elements](https://developer.mozilla.org/en-US/docs/Glossary/Block-level_content).

Source: https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align

>Applies to: | inline-level and `table-cell` elements

Source: https://drafts.csswg.org/css2/#propdef-vertical-align

Validation warning via VSCode:

Screenshot 2023-10-30 at 8 56 32 AM

## Solution

Removes the `vertical-align: baseline` styles.

### DIFF
--- a/packages/usa-banner/src/styles/_usa-banner.scss
+++ b/packages/usa-banner/src/styles/_usa-banner.scss
@@ -230,7 +230,7 @@ $banner-icon-close: (
   @include button-unstyled;
   @include u-pin("left");
   @include u-pin("y");
-  @include u-text("primary", underline, baseline);
+  @include u-text("primary", underline);
   @include set-link-from-bg(
     $theme-banner-background-color,
     $theme-banner-link-color,


mejiaj (approved):
Good find, thanks for submitting.

No issues in any screen size with removing this. Tested on [Develop preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-banner--default&p=all) branch.

amyleadem (approved):
Thank you for this fix! Everything looks good to me. 

I checked the following items:
- [x] Confirmed that `vertical-align` is invalid on `block` elements
- [x] Confirmed that this change removes `vertical-align` from `.usa-banner__button`
- [x] Confirmed that there is no visual change in positioning for `.usa-banner__button`

thisisdano (approved):
Thank you!

Issue #5619, opened 2023-11-16
USWDS - Bug: Redundant role attribute on USA banner icons with empty alt text


jhancock532:
### Describe the bug

This is a minor accessibility issue raised by HTML validation tools such as NuHTMLValidator.

The following error is logged twice by NuHTMLValidator when using the USWDS Banner component.

```
Error: An img element which has an alt attribute whose value is the empty string must not have a role attribute.

<img class="usa-banner__icon usa-media-block__img" src="..." role="img" alt="" aria-hidden="true">
```

These errors refer to the two banner icons used in this component, [found on lines 30 and 37 of `usa-banner.twig`](https://github.com/uswds/uswds/blob/3175c0258a63e6e464ea8284bb90f452f46463a1/packages/usa-banner/src/usa-banner.twig#L30).

### Steps to reproduce the bug

1. Visit the HTML validator site: https://validator.w3.org/nu/
2. Copy the URL to the USWDS website page for the banner component into the 'Enter the URL...' text box - `https://designsystem.digital.gov/components/banner/`
3. Click validate, and note that several errors of the above description are raised in the errors list, each relating to the multiple uses of the banner component on this page.

### Expected Behavior

No error should appear, as img elements should not need a `role="img"` attribute when used with empty alt text.

### Related code

I have searched the codebase for other uses of `role="img"` and none were found on an image element with empty alt text.

### Screenshots

<img width="981" alt="Screenshot 2023-11-16 at 15 11 06" src="https://github.com/uswds/uswds/assets/18164832/05a53852-a63e-4381-b6e4-b4c41eead90a">

### System setup

- USWDS version: 3.7.0
- Device: Laptop
- OS: MacOS
- Browser: Chrome

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

brunerae:
For more information, see comments in #5620. We'll consider this as we approach future markup changes. Thank you!

PR #5620, opened 2023-11-16
USWDS - Banner: Remove redundant banner icon role attribute

jhancock532:
# Summary

**Removed unnecessary role attributes from banner icon images.** A minor accessibility improvement to HTML markup that prevents screen readers from announcing the role of an image that has no content to relay to users.

## Breaking change

Please update the banner component HTML template in your project to remove the `role="img"` attributes for the image elements with the class `usa-banner__icon`.

## Related issue
Closes [#_5619_](https://github.com/uswds/uswds/issues/5619)

## Related pull requests

No related pull requests.

## Preview link

No preview link generated.

## Problem statement

We should avoid confusing screenreader systems by announcing the role of an element that doesn't have any information to announce. This is confusing and is a minor annoyance for screenreader users. 

This problem is minor as it likely doesn't affect many screenreader systems, which are generally smart enough to handle invalid HTML / odd attribute pairings. However, as the banner is used consistently across all US government sites, it would be good to fix regardless.

## Solution

Remove the extra unneeded role attributes.

## Major changes

No major changes made.

## Testing and review

Please test using https://validator.w3.org/nu/. If you don't have a live webpage to test against, you can also test using [the manual text input box](https://validator.w3.org/nu/?#textarea) by copying and pasting the HTML source code there.

### DIFF
--- a/packages/usa-banner/src/usa-banner.twig
+++ b/packages/usa-banner/src/usa-banner.twig
@@ -27,14 +27,14 @@
     <div class="usa-banner__content usa-accordion__content" id="{{ banner.id }}">
       <div class="grid-row grid-gap-lg">
         <div class="usa-banner__guidance tablet:grid-col-6">
-          <img class="usa-banner__icon usa-media-block__img" src="./img/icon-dot-gov.svg" role="img" alt="" aria-hidden="true">
+          <img class="usa-banner__icon usa-media-block__img" src="./img/icon-dot-gov.svg" alt="" aria-hidden="true">
           <div class="usa-media-block__body">
             {# Will cause error without trim because it doesn't know what to do with new lines #}
             <p><strong>{{ domain.heading }}</strong><br/>{{ domain.text | trim | raw }}</p>
           </div>
         </div>
         <div class="usa-banner__guidance tablet:grid-col-6">
-          <img class="usa-banner__icon usa-media-block__img" src="./img/icon-https.svg" role="img" alt="" aria-hidden="true">
+          <img class="usa-banner__icon usa-media-block__img" src="./img/icon-https.svg" alt="" aria-hidden="true">
           <div class="usa-media-block__body">
             <p><strong>{{ https.heading }}</strong><br/>{{ https.pretext | trim | raw }} ({{ lock }}) {{ https.posttext | trim | raw }}</p>
           </div>


mahoneycm:
Thanks for your contribution @jhancock532! 

During my testing with VoiceOver, I found no difference between the callouts from this PR and what's currently on `develop`, probably due to the `img` element's `aria-hidden` attribute. 

I also ran both through the validator you linked and both passed without any issues.

I've created a test repo to make the change easily viewable. @amycole501 or @alex-hull, could either of you check these with other screen readers to see if there are any noticeably different callouts on the banner? This change targets the `.gov` building icon and the green lock icon.

[Banner [develop] →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?id=components-banner--default)
[Banner [removed role="img"] →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/fix-redundant-banner-icon-role-attribute-test/iframe.html?args=&id=components-banner--default)

amycole501:
I used JAWS and NVDA in Chrome, Edge, and Firefox (on PC)
I didn't hear any difference in any combination of AT and browser.
ANDI didn't seem to treat either version any differently either.

mahoneycm:
## Additional information on attributes

After digging a bit deeper it appears that all of the attributes were intentionally set to resolve other accessibility bugs and to provide a consistent experience across all screen readers at the time.

Thanks to @amyleadem for digging deeper into this!

### Identifying an SVG as an image

Due to a [VoiceOver bug](https://webkit.org/b/216364), VoiceOver does not correctly announce SVG images as images. Include [role="img"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role) to all  elements with SVG source files to ensure assistive technologies correctly announce the SVG as image content. [^1]
 
`MDN`

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#identifying_svg_as_an_image

### aria-hidden

Some additional information from https://github.com/uswds/uswds/pull/3892 where we added the `aria-hidden` attribute:

>Adding role="img" fixes a webkit bug that causes images with an SVG source be be announced as a group, not an image. However, adding role="img" and alt="" result in an object that's selectable by AT but not announced (at least in our testing on VoiceOver and Chrome). Adding aria-hidden="true" results in a properly hidden element.
>
>Adding aria-hidden="true" only would be equally effective (it renders role="img" moot) but we choose to add all three to be as consistent as possible with , even if it is redundant. Thus, each of the following can always be true for :
>
>Use alt="" for decorative image
>Use role="img" for images with an SVG source
>Use aria-hidden="true" for decorative images with an SVG source

jhancock532:
It looks like there's no accessibility issue related to this change as it's been tested across supported platforms and behaves as expected. I think that this is just a validation error / won't fix if we want to err on the side of caution, assuming that VoiceOver might still play up in the future and announce images with SVGs incorrectly? That webkit bug thread didn't look like the error had conclusively been resolved for all users...

Apologies for any confusion @mahoneycm, the HTML validation error not appearing for you may be because when testing a Storybook instance with NuHTMLValidator directly (pasting in the URL) it doesn't seem to wait for the component to load before analysing the page's HTML.

<details>
<summary>Evidence of the validator not waiting for the component to load</summary>
Checking the `source` checkbox, then loading the page...
image

then scrolling down to view the contents of the body tag; the component isn't in the source code.
image
</details>

However, when copying and pasting the component HTML into the validator, the error shows. The error also appears when validating the USWDS documentation site, which doesn't load in components via JS - https://validator.w3.org/nu/?doc=https%3A%2F%2Fdesignsystem.digital.gov%2Fcomponents%2Fbanner%2F

Thank you for investigating @amycole501, @mahoneycm!

mejiaj:
Thank you for the PR, unfortunately we cannot accept at this time since it was an intentional change. 

This is outlined in https://github.com/uswds/uswds/pull/3892. Also referencing the original [webkit bug [webkit bugzilla]](https://bugs.webkit.org/show_bug.cgi?id=216364) as reference.

We'll keep the original open so that we can revisit the need for this role and address it in upcoming web components work (_since we treat markup changes as breaking change_).

PR #5617, opened 2023-11-15
USWDS - Nav: Fix compilation error when max header width is set to "none"

mahoneycm:
# Summary

Resolved a compilation error when `$theme-max-header-width` was set to `none`. 

## Breaking change

This is not a breaking change

## Related issue

Closes #5579

## Related pull requests

[Changelog →](https://github.com/uswds/uswds-site/pull/2356)

[USWDS: Header: Add fix for Safari-only header bug #5443](https://github.com/uswds/uswds/pull/5443#top)

## Preview link

[Header →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-max-width-none-fix/?path=/story/components-header--default)

[Documentation page →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-max-width-none-fix/?path=/story/pages-documentation-page--documentation-page)

## Problem statement

A compilation error occurs when setting `$theme-max-header-width` to `none`.

This is a regression which stems from our Safari header navigation bug fix #5543. The `calc()` method cannot take a value of `none` so it breaks the calculation.

## Solution

Use sass `@if` at-rule to check if `$theme-max-header-width` value is `none` before enabling the Safari bug fix logic.

- Since there is no max width, the Safari header bug does not occur
- Therefore, we can bypass the additional logic when max-width is set to `none`

### Test compilation error

1. Set `$theme-header-max-width` to `"none"`
2. Ensure there are no compilation errors

### Safari header bug

Follow Amy’s testing steps from #5543 to ensure there are no regressions:

Use Storybook to test that the mobile menu opens in Safari:

1. Open the Documentation page in Safari
2. Resize the window to somewhere between 1009px-1024px
3. Click the "Menu" button to open the mobile navigation panel
4. Confirm that the mobile header successfully opens as expected

Use Storybook to test that this does not affect other browsers:

1. Open the Documentation page in other browsers
2. Resize the window to somewhere between 1009px-1024px
3. Click the "Menu" button to open the mobile navigation panel
4. Confirm that the main page content is not scrollable

### Testing checklist

- [ ] Confirm there are no compilation errors when `$theme-header-max-width` is set to `"none"`.
- [ ] Confirm use of `@if` is appropriate in this situation.
- [ ] Confirm there are no regressions with header in Safari after this change.

PR #5624, opened 2023-11-17
USWDS - Header: Use $theme-header-min-width for safari fix calc

### LINKED ISSUE #5579: USWDS - Bug: Setting $theme-header-max-width: "none" causes error when build usa-nav styles
### Describe the bug

We have used the setting $theme-header-max-width: "none" to get full width headers. With the latest release, 3.6.1 we are having a build error when building the style sheets:

```
CssSyntaxError: postcss-csso: /Users/mbucknel/wdfn/wdfn-viz/node_modules/@uswds/uswds/packages/usa-nav/src/styles/_usa-nav.scss:486:0: Number, dimension, ratio or identifier is expected
```
Prior to version 3.6.1, we had no build issues. If I set the theme variable to something like "desktop", the style sheets build as expected.

1. Set $theme-header-max-width: "none".  For example:
'''
@use 'uswds-core' with (
  $theme-font-path: './fonts',
  $theme-image-path: './img',
  $theme-header-max-width: "none",
);
'''
2. Build the style sheets with the recommended build procedure for the sass style sheets 

### Expected Behavior

Build works as expected

### Related code

_No response_

### Screenshots

_No response_

### System setup

Using USWDS 3.6.1
Running the command on Mac OS 13.6, using sass: 1.68.0, postcss 8.4.31 and postcss-csso 6.0.1

### Additional context

_No response_

### Code of Conduct

amyleadem:
# Summary

Fixed a bug that prevented `$theme-max-header-width` from accepting a value of "none".

## Breaking change

## Related issue

Closes #5579 

## Related pull requests

This PR replaces the fix proposed in https://github.com/uswds/uswds/pull/5617.

[Changelog PR](https://github.com/uswds/uswds-site/pull/2356)

## Preview link

Preview link: [Documentation page](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-safari-bug-fix-regression/iframe.html?args=&id=pages-documentation-page--documentation-page&viewMode=story)

## Problem statement

A compilation error occurs when setting `$theme-max-header-width` to `"none"`.

This is a regression which stems from our Safari header navigation bug fix in  https://github.com/uswds/uswds/pull/5543. The calc() method cannot take a value of none so it breaks the calculation.

## Solution

The calculation mistakenly used `$theme-header-max-width` instead of `$theme-header-min-width`. The settings have two distinct purposes:

- `$theme-header-min-width` is the setting that determines where the mobile nav should change to the desktop nav.
- `$theme-header-max-width`  is the setting that determines the maximum width of the desktop navigation. 

For the Safari fix to work as expected, the fix needs to be triggered within 15px of the mobile/desktop breakpoint, which is `$theme-header-min-width`.

By removing `$theme-header-max-width` from the calculation, there should no longer be errors when the setting's value is set to "none". 

## Testing and review

- Confirm that there is no compilation error when you set `$theme-header-max-width` to "none" 
- Confirm that the Safari bug fix from https://github.com/uswds/uswds/pull/5617 still works by checking if the mobile menu opens as expected in Safari:
  1. Open the [Documentation page](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-safari-bug-fix-regression/iframe.html?args=&id=pages-documentation-page--documentation-page&viewMode=story) in Safari
  1. Resize the window to somewhere between 1009px-1024px
  1. Click the "Menu" button to open the mobile navigation panel
  1. Confirm that the mobile header successfully opens as expected
- Confirm that the mobile menu opens as expected in other browsers:
   1. Open the [Documentation page](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-safari-header-bug/iframe.html?args=&id=pages-documentation-page--documentation-page&viewMode=story) in other browsers
   1. Resize the window to somewhere between 1009px-1024px
   1. Click the "Menu" button to open the mobile navigation panel
   1. Confirm that the main page content is not scrollable
- Confirm that the bug fix works with different values for `$theme-header-min-width`
  1. In a local build, change the value of `$theme-header-min-width` to expected values like "widescreen" or "tablet"
  1. Open the [documentation page](http://localhost:6006/iframe.html?args=&id=pages-documentation-page--documentation-page&viewMode=story) in Safari
  1. Resize the window to within 15px of the value set for `$theme-header-min-width`
  1. Confirm that the menu
- Confirm that we don't need to account for any special values for `$theme-header-min-width` that could cause an error (like "none")

### DIFF
--- a/packages/usa-nav/src/styles/_usa-nav.scss
+++ b/packages/usa-nav/src/styles/_usa-nav.scss
@@ -479,15 +479,18 @@ $expand-less-icon: map-merge(
 //
 // Note: 15px is the current width of the Safari scrollbar.
 // More details in https://github.com/uswds/uswds/pull/5443
-$safari-header-bug-min-width: calc(
-  units($theme-header-min-width) - px-to-rem(15px)
-);
 
-@media (min-width: $safari-header-bug-min-width) {
-  .usa-js-mobile-nav--active.is-safari {
-    overflow-y: scroll;
-    position: fixed;
-    // --scrolltop set with JS with zero as fallback.
-    top: var(--scrolltop, 0);
+@if $theme-header-min-width != "none" {
+  $safari-header-bug-min-width: calc(
+    units($theme-header-min-width) - px-to-rem(15px)
+  );
+
+  @media (min-width: $safari-header-bug-min-width) {
+    .usa-js-mobile-nav--active.is-safari {
+      overflow-y: scroll;
+      position: fixed;
+      // --scrolltop set with JS with zero as fallback.
+      top: var(--scrolltop, 0);
+    }
   }
 }
--- a/packages/usa-nav/src/styles/_usa-nav.scss
+++ b/packages/usa-nav/src/styles/_usa-nav.scss
@@ -479,7 +479,8 @@ $expand-less-icon: map-merge(
 //
 // Note: 15px is the current width of the Safari scrollbar.
 // More details in https://github.com/uswds/uswds/pull/5443
-@if $theme-header-min-width != "none" {
+@if $theme-header-min-width != "none" and not $theme-header-min-width == "auto"
+{
   $safari-header-bug-min-width: calc(
     units($theme-header-min-width) - px-to-rem(15px)
   );
--- a/packages/usa-nav/src/styles/_usa-nav.scss
+++ b/packages/usa-nav/src/styles/_usa-nav.scss
@@ -475,12 +475,12 @@ $expand-less-icon: map-merge(
 }
 
 // Safari-only fix that forces a vertical scrollbar when mobile menu is open.
-// Only needed in the 15px immediately preceding $theme-header-max-width.
+// Only needed in the 15px immediately preceding $theme-header-min-width.
 //
 // Note: 15px is the current width of the Safari scrollbar.
 // More details in https://github.com/uswds/uswds/pull/5443
 $safari-header-bug-min-width: calc(
-  units($theme-header-max-width) - px-to-rem(15px)
+  units($theme-header-min-width) - px-to-rem(15px)
 );
 
 @media (min-width: $safari-header-bug-min-width) {
--- a/packages/usa-nav/src/styles/_usa-nav.scss
+++ b/packages/usa-nav/src/styles/_usa-nav.scss
@@ -1,3 +1,4 @@
+@use "sass:map";
 @use "uswds-core" as *;
 
 // Navigation helpers
@@ -478,20 +479,23 @@ $expand-less-icon: map-merge(
 // Only needed in the 15px immediately preceding $theme-header-min-width.
 //
 // Note: 15px is the current width of the Safari scrollbar.
-// More details in https://github.com/uswds/uswds/pull/5443
-@if $theme-header-min-width != "none" and $theme-header-min-width != "auto" {
+// Note: This fix only applies when $theme-header-min-width is defined with a system breakpoint
+//   because the header visually breaks with other values.
+//   This bypass prevents compilation errors with values like "none" or 1px.
+
+$our-breakpoints: map-deep-get($system-properties, breakpoints, standard);
+
+@if map-has-key($our-breakpoints, $theme-header-min-width) {
   $safari-header-bug-min-width: calc(
     units($theme-header-min-width) - px-to-rem(15px)
   );
 
-  @if $safari-header-bug-min-width > 0 {
-    @media (min-width: $safari-header-bug-min-width) {
-      .usa-js-mobile-nav--active.is-safari {
-        overflow-y: scroll;
-        position: fixed;
-        // --scrolltop set with JS with zero as fallback.
-        top: var(--scrolltop, 0);
-      }
+  @media (min-width: $safari-header-bug-min-width) {
+    .usa-js-mobile-nav--active.is-safari {
+      overflow-y: scroll;
+      position: fixed;
+      // --scrolltop set with JS with zero as fallback.
+      top: var(--scrolltop, 0);
     }
   }
 }
--- a/packages/usa-nav/src/styles/_usa-nav.scss
+++ b/packages/usa-nav/src/styles/_usa-nav.scss
@@ -477,26 +477,21 @@ $expand-less-icon: map-merge(
 // Safari-only fix that forces a vertical scrollbar when mobile menu is open.
 // Only needed in the 15px immediately preceding $theme-header-min-width.
 //
-// Note: To prevent error, this fix is currently scoped to only work
-//  when $theme-header-min-width receives expected values. The setting is used
-//  primarily in at-media(), which only accepts breakpoints from $system-properties.
 // Note: 15px is the current width of the Safari scrollbar.
 // More details in https://github.com/uswds/uswds/pull/5443
-$our-breakpoints: map-deep-get($system-properties, breakpoints, standard);
-
-@if map-has-key($our-breakpoints, $theme-header-min-width) {
+@if $theme-header-min-width != "none" {
   $safari-header-bug-min-width: calc(
     units($theme-header-min-width) - px-to-rem(15px)
   );
 
-  @media (min-width: $safari-header-bug-min-width) {
-    .usa-js-mobile-nav--active.is-safari {
-      overflow-y: scroll;
-      position: fixed;
-      // --scrolltop set with JS with zero as fallback.
-      top: var(--scrolltop, 0);
+  @if $safari-header-bug-min-width > 0 {
+    @media (min-width: $safari-header-bug-min-width) {
+      .usa-js-mobile-nav--active.is-safari {
+        overflow-y: scroll;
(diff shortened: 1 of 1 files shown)


mahoneycm (approved):
Great catch! Looks like this slipped through since `theme-header-max-width` and `theme-header-min-width` both shared the same value.

- [x] No visual regression 
- [x] Fix is logical based on the origin of the bug
- [x]  Safari bug fix works as expected
- [x] Calc error is no longer present

packages/usa-nav/src/styles/_usa-nav.scss
```diff
@@ -475,19 +475,22 @@ $expand-less-icon: map-merge(
...
+    units($theme-header-min-width) - px-to-rem(15px)
+  );
+
+  @media (min-width: $safari-header-bug-min-width) {
+    .usa-js-mobile-nav--active.is-safari {
+      overflow-y: scroll;
+      position: fixed;
+      // --scrolltop set with JS with zero as fallback.
+      top: var(--scrolltop, 0);
+    }
```
amyleadem, line 494: Quick note that I was also playing with this to make it a bit more resilient. Might be overkill for the need though since the Sass spits out warnings for non-named breakpoint values. This would just prevent negative values in the media query if someone were to set the value for `$theme-header-min-width` to something like 0. 
```suggestion
@if map-has-key($project-spacing-named, $theme-header-min-width) {
  $safari-header-bug-min-width: calc(
    units($theme-header-min-width) - px-to-rem(15px)
  );

  @media (min-width: $safari-header-bug-min-width) {
    .usa-js-mobile-nav--active.is-safari {
      overflow-y: scroll;
      position: fixed;
      // --scrolltop set with JS with zero as fallback.
      top: var(--scrolltop, 0);
    }
  }
}

```

thisisdano (changes requested):
Now that I'm digging in, I still see the error that James mentions

amyleadem:
Ok, I'll fix it up and will tag you both for re-review when it's ready.

thisisdano:
Sorry I jumped into this too soon!

packages/usa-nav/src/styles/_usa-nav.scss
```diff
@@ -475,19 +475,23 @@ $expand-less-icon: map-merge(
...
-  units($theme-header-max-width) - px-to-rem(15px)
-);
-
-@media (min-width: $safari-header-bug-min-width) {
-  .usa-js-mobile-nav--active.is-safari {
-    overflow-y: scroll;
-    position: fixed;
-    // --scrolltop set with JS with zero as fallback.
-    top: var(--scrolltop, 0);
+@if $theme-header-min-width != "none" {
```
mahoneycm, line 482: ## Special Tokens

I'm getting a build error when setting `$theme-header-min-width: "auto"`

```js
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
Undefined operation "calc(auto - 0.94rem) > 0".
    ╷
487 │   @if $safari-header-bug-min-width > 0 {
    │       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ╵
  packages/usa-nav/src/styles/_usa-nav.scss 487:7  @forward
  packages/usa-nav/src/styles/_index.scss 4:1      @forward
  packages/uswds/_index.scss 38:1                  root stylesheet
  ```
  
- This is a regression from `develop`.
- But it should be noted that setting `auto` breaks the layout of the header.
- Would the error be preferred? If so, should we update the acceptable tokens to exclude values that break the layout? (negative values, 0,) 
- Potentially out of scope

> [!IMPORTANT]
> I began experiencing inconsistent builds and build errors during my testing. There is a chance my broken layouts for certain special tokens _might_ have been a local issue.

**Review by mahoneycm — COMMENTED** [review]
This is looking good to me! One comment about special characters but expected values look great!

- [x]  Confirm that there is no compilation error when you set `$theme-header-max-width` to "none"
- [x]  Confirm that the Safari bug fix from [USWDS - Nav: Fix compilation error when max header width is set to "none" #5617](https://github.com/uswds/uswds/pull/5617) still works by checking if the mobile menu opens as expected in Safari:
- [x]  Confirm that the mobile menu opens as expected in other browsers:
- [x]  Confirm that the bug fix works with different values for `$theme-header-min-width`
- [ ]  Confirm that we don't need to account for any special values for `$theme-header-min-width` that could cause an error (like "none")

File: packages/usa-nav/src/styles/_usa-nav.scss (hunk 2 above)
Comment (amyleadem) [reply packages/usa-nav/src/styles/_usa-nav.scss]: @mahoneycm I updated the Safari fix to only run if the value of `$theme-header-min-width` fits a _realistic_ value for the setting. 

The header visually breaks and causes warnings when the value of `$theme-header-min-width` is outside of the defined breakpoints in our system (card, card-lg, mobile, mobile-lg, tablet, tablet-lg, desktop, desktop-lg, widescreen). This is because the setting is used primary with our `at-media` mixin, which only accepts the breakpoints defined in `$system-properties`.

For this reason, I decided to bypass the Safari fix unless the setting was defined with the same values that are acceptable to `at-media`.

> [!WARNING]
> There is some risk with this approach. If we ever update the accepted values for the `at-media` mixin to be something other than the system breakpoints, the Safari fix might not be applied in all the right places. However, I feel like the breakpoints in $system-properties will be a stable and reliable place to look for realistic header breakpoints. In my mind, this reduces the risk of future issues.

<details><summary>More on the thought process</summary>
<p>
<ol>
<li>First I tried bypassing only the "special" values of “auto” and “none”. This worked in most cases but would cause an error with 1px/2px values because of the unit mismatch. </li>
<li>This caused me to think about how it would be better to only apply the safari fix to *realistic* values, rather than *possible* values. The header visually breaks and causes warnings when the value of `$theme-header-min-width` is outside of the defined breakpoints in our system (card, card-lg, mobile, mobile-lg, tablet, tablet-lg, desktop, desktop-lg, widescreen). For this reason, I decided to bypass the Safari fix unless the setting was defined with one of those values. </li>
<ul>
<li>Note: there is some risk with this because if we ever update the accepted values for the `at-media` mixin, the Safari fix might not be applied in all the right places. However, I feel like the breakpoints in $system-properties will be a stable and reliable place to look, which reduces the risk of future issues in my mind. 
</ul>
<li> Alternatively, we could create a custom map that includes all possible reasonable values, which could mean including the "medium" spacing token set. I ultimately moved away from this because it seemed very rare that a mobile nav would be viewed in safari <120px, especially if the header is broken in other ways by this value.  </li>
</ol>
</p>
</details> 

Can you confirm the following?
- Setting `$theme-header-min-width` to any breakpoint value (card, card-lg, mobile, mobile-lg, tablet, tablet-lg, desktop, desktop-lg, widescreen)
  -  does NOT cause a compile error 
  -  DOES generate an `is-safari` class in the CSS 
- Setting `$theme-header-min-width` to any non-breakpoint value (Ex: "none", "auto", 15, "hamburger")
  -  does NOT generate an `is-safari` class in the CSS 
  -  does NOT cause a compile error. However, this warning is still expected for non-breakpoint values: 
```
Warning: `none` is not a valid USWDS project breakpoint. Valid values: card, card-lg, mobile, mobile-lg, tablet, tablet-lg, desktop, desktop-lg, widescreen
```

File: packages/usa-nav/src/styles/_usa-nav.scss (hunk 1 above)
Comment (amyleadem) [reply packages/usa-nav/src/styles/_usa-nav.scss]: Another update:  see [this comment](https://github.com/uswds/uswds/pull/5624#discussion_r1419315542)

File: packages/usa-nav/src/styles/_usa-nav.scss (hunk 2 above)
Comment (mahoneycm) [reply packages/usa-nav/src/styles/_usa-nav.scss]: I'm still seeing `is-safari`. The logic looks good and the build warning are appearing, but the safari class is added. I've tried rebuilding the project multiple times.

The confusing thing is when I run a debug, it returns `false` as expected and should _not_ be running the block of scss 🤔 

```scss
@debug map-has-key($our-breakpoints, $theme-header-min-width); // Output: false
```

- [x]  Standard breakpoint generates `is-safari` class
- [x]  No compilation errors
- [ ]  Non-breakpoint values bypass safari fix
- [x]  Non-breakpoint values cause compilation warning, but no error

Is the `is-safari` class being omitted for you?

amyleadem:
@mejiaj This is ready for review. You can find details about the most recent update in [this comment](https://github.com/uswds/uswds/pull/5624#discussion_r1419315542) as part of a previous discussion. Curious to hear what you think!

mejiaj (approved):
Nice fix. I tested:

- `$theme-header-min-width: "none";` 
- `$theme-header-min-width: "tablet-lg";`
- `$theme-header-min-width: "desktop"`

No errors or invalid CSS compiled.

thisisdano (approved):
🎉

PR #5631, opened 2023-11-21
USWDS - Button: Collapse unstyled button width at small viewport sizes

aduth:
# Summary

**Unstyled buttons no longer appear as full width at narrow viewport sizes.** This brings the button into alignment with expectation of it being used in plain text contexts.

## Breaking change

This may be considered a breaking change if someone had come to rely on the previous behavior. However, I'd argue that this behavior should have never been considered expected, and therefore fits the definition of a bug fix.

## Related PR
[Changelog PR](https://github.com/uswds/uswds-site/pull/2795)

## Preview link

Preview link: http://localhost:6006/?path=/story/components-button--unstyled

Before|After
---|---
Screenshot 2023-11-21 at 9 04 46 AM|Screenshot 2023-11-21 at 9 05 20 AM

## Problem statement

We often use unstyled buttons as visually equivalent to a link. Therefore, we've encountered issues where the unstyled button occupies full width at mobile viewport sizes, causing line breaks in content.

See screenshots at https://github.com/18F/identity-idp/pull/9632 for example effect.

## Solution

The solution proposed here resets the `width: 100%` applied to buttons by default at mobile viewport sizes, specifically for the unstyled button.

### DIFF
--- a/packages/uswds-core/src/styles/mixins/general/button-unstyled.scss
+++ b/packages/uswds-core/src/styles/mixins/general/button-unstyled.scss
@@ -14,6 +14,7 @@
   text-align: left;
   margin: 0;
   padding: 0;
+  width: auto;
 
   &:hover,
   &.usa-button--hover,


mejiaj:
### For discussion

This is a legacy style. Should unstyled buttons be inline at small screen sizes too (_inline on desktop_)?

| Mobile | Desktop |
| :----- | :---- |
| image [Source →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-button--default&p=all) | image [Source →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-button--default&p=all) |

mahoneycm:
@mejiaj If the goal of these buttons are to replicate the look and feel of links, it makes sense to me that these would be inline. 

From what I can tell in the codebase and our site, there shouldn't be any visual regressions. We used the unstyled button for the unstyled language selector but it appears to be unaffected by this change

mahoneycm:
Hey there @aduth 

Thank you for contributing this work and providing the screenshots in the related issue you linked. It really helped to showcase the current state of the button and why we'd benefit from updating.

We've added this to the `3.9.0` and will review shortly!

amyleadem (approved):
Thanks for this submission @aduth! The change looks good to me.

A few of notes for follow-on tasks with this PR: 
1. We should mark this as a breaking change in the PR description and release notes. This PR will change the default visual styles of "unstyled" buttons, and we should note that users should confirm the display of this variant in their projects. 
1. We should _consider_ adding a note to the in the PR description, release notes, and/or button component page that the pattern of using buttons styled as links is currently under review with our accessibility and UX teams, and might not be a recommended practice.
1. We should create follow-on issue that tests the target height of unstyled buttons to make sure they are compliant with [WCAG 2.5.8](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html).

mejiaj (approved):
A small, but impactful change. Thanks for the PR!

Wanted to note that the unstyled button is also used in [language selector [develop]](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-language-selector--unstyled-example&p=all).

thisisdano (approved):

thisisdano:
@amyleadem Can you create the issue you mentioned in your review?

> We should create follow-on issue that tests the target height of unstyled buttons to make sure they are compliant with [WCAG 2.5.8](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html).

amyleadem:
Good catch @thisisdano! Opened #6091 to follow-up on the touch target size of the unstyled button variant.

PR #5636, opened 2023-11-24
USWDS - Alert and Site alert: Adjust left alignment calculations

### LINKED ISSUE #5583: USWDS - Alert/Site alert: Fix content alignment issues
## Summary
When `$theme-site-margins-width` receives custom values, it can cause the the content inside the alert component to overlap or become misaligned with the banner and other components. For example, setting `$theme-site-margins-width: 1px` results in the following:

image

Setting the value to `$theme-site-margins-width: `15` results in the following:

image

And `$theme-site-margins-width: `6`:

image

We should recalculate the padding and margins on this component so that they consistently align with the other components that use  `$theme-site-margins-width`.

amyleadem:
# Summary

**Fixed a bug that caused `$theme-site-margins-width` to adjust alignment inside the alert and site alert components.** 

> [!Important]
> At desktop widths, the site alert and alert components will no longer share left and right alignment. If users want alerts to align with site alert, they should use site alert markup. 

## Breaking change

Alignment on the alert and site alert components will likely shift from this change. Confirm that your implementation aligns as expected. 

## Related issue

Closes #5583

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2759)

## Preview link

- [Test - Alert comparison test story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-alert-icon-alignment/?path=/story/components-alert--test-alert-site-alert-comparison)
- [Test - Alerts in template story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-alert-icon-alignment/?path=/story/components-alert--test-alerts-in-template)
- [Demo in uswds-sandbox](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/al-test-5636-alert-align/pages/prose/)
    - Related demo PR in case you want to test changing settings values: uswds/uswds-sandbox#149
 
## Problem statement

When `$theme-site-margins-width` receives custom values, it can cause the the content inside the alert component to overlap or become misaligned with the banner and other components. For example, setting `$theme-site-margins-width: 1px` results in the following:

image

Setting the value to `$theme-site-margins-width: `15` results in the following:

image

And `$theme-site-margins-width: `6`:

image

## Solution

This PR rebuilds the alignment in the alert and site alert components so that they consume settings values as expected.

#### Mobile
image

#### Desktop
image

### Generally
1. Confirm code meets USWDS standards
1. Confirm that it is reasonable to ask users to use site-alert markup when they want the content to align with site margins.

### Default state
1. Open up the Storybook previews ([Test - Alert comparison test story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-alert-icon-alignment/?path=/story/components-alert--test-alert-site-alert-comparison) |  [Test - Alerts in template story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-alert-icon-alignment/?path=/story/components-alert--test-alerts-in-template)) of this component:
    - At mobile, confirm that all alert and site alert variants have the same left and right alignment
    - At desktop, confirm that site alert content shares left and right alignment with banner and footer
    - At desktop, confirm that all alert variants share the same left alignment

### With settings adjustments

> [!Important] 
> All [alert component settings](https://designsystem.digital.gov/documentation/settings/#alert) apply to site alert _except_ `$theme-alert-padding-x`. This is because the horizontal alignment of the site alert is fundamentally different from the alert component. We will update the documentation to clarify this in the [changelog PR](https://github.com/uswds/uswds-site/pull/2759). The conversation about this decision is outlined in [this comment thread](https://github.com/uswds/uswds/pull/5636#discussion_r1752344906?)

1. Run a local build of uswds from this branch 
1. Open the "Alert components comparison" story (http://localhost:6006/?path=/story/components-alert--alert-components-comparison)
1. Set `$theme-site-margins-width` to 15 and 0. For each instance, confirm in the test stories that:
    1. At mobile, all alerts and site alerts share the same left and right alignment
    1. At desktop, all site alerts share the same left and right alignment as the banner
    1. At desktop, all alerts share the same left and right alignment as each other
    1. The gap between the icon and the alert text does not change from default
1. Set `$theme-alert-bar-width` to 15 and 0. For each instance, confirm in the test stories that:
    1. The left colored bar adjusts to match the declared width
    1. Text content does not overlap with the colored bar
1. Set `$theme-alert-padding-x` to 15 and 2. For each instance, confirm in the test stories that:
    1. At desktop and mobile, the left and right alignment on the alerts adjusts to match the new setting value
    1. At desktop, all site alerts maintain the same left and right alignment as the banner
    1. At mobile, all site alerts share the same left and right alignment of 2.5 units
1. Set `$theme-alert-icon-size` to 6 and 2. For each instance, confirm in the test stories that:
    1. The standard icon size adjusts to match the declared value (this should not affect the slim icon implementations)
    1. At all window widths, the text for alert and site alert components adjusts left alignment to match the new icon size. The gap between the icon and the text should not change, and the elements should not overlap. 
1. Set `$theme-alert-padding-y` to 6 and 0. For each instance, confirm in the test stories that:
    1. The vertical padding adjusts to match the declared value (this does not apply to the slim variant)
    3. All other elements align as expected

### Compare to 3.1.0
Additionally, you can compare this PR with the alerts in the [release-3.1.0](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/release-3.1.0/?path=/story/components-alert--emergency) branch to see the components _before_ the alignment adjustments in [3.2.0](https://github.com/uswds/uswds/releases/tag/v3.2.0). 

The main differences that I found between the 3.1.0 and this PR are:
- Adjusting `$theme-alert-padding-x` adjusted the gap between the icon and text in 3.1.0. It no longer does that in this PR, which I think is an improvement. (The gap is now consistent regardless of x padding size)
    image 
- The [site alert in 3.1.0](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/release-3.1.0/?path=/story/components-site-alert--info) _does NOT_ have a bar on left side of the component. It does in [develop](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-site-alert--info) and in [this PR](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-alert-icon-alignment/?path=/story/components-site-alert--info).


packages/usa-alert/src/styles/_usa-alert.scss
```diff
@@ -2,15 +2,6 @@
...
 
-// Alert variables ---------- //
-$alert-icons: (
-  success: "check_circle",
-  warning: "warning",
-  error: "error",
-  info: "info",
-  emergency: "error",
-);
-
```
amyleadem, line 13: Moved these icons to `alert-status-styles.scss` so that both alert and site alert have the same variants. This enables us to recommend users use the site alert markup structure for alerts that should share site margins.

packages/usa-alert/src/test/test-patterns/test-alert-comparison.twig
```diff
@@ -1,21 +1,3 @@
...
-    'heading': 'Official websites use .gov',
-    'text': 'A <strong>.gov</strong> website belongs to an official government organization in the United States.'
-  },
-  'https': {
-    'heading': 'Secure .gov websites use HTTPS',
-    'pretext': 'A <strong>lock</strong>',
-    'posttext': 'or <strong>https://</strong> means you’ve safely connected to the .gov website. Share sensitive information only on official, secure websites.'
-  }
-} %}
-
```
amyleadem, line 18: > [!Note]
> Replaced this custom data with the default banner JSON

packages/usa-alert/src/test/test-patterns/test-alert-comparison.twig
```diff
@@ -78,44 +73,14 @@
...
-        'Twitter',
-        'YouTube',
-        'Instagram',
-        'RSS'
-      ]
-    }
-  }
-} %}
-
-{% include '@components/usa-footer/src/usa-footer.twig' with footer %}
```
amyleadem, line 121: > [!Note]
> Replaced this custom data with the default footer JSON

packages/usa-site-alert/src/styles/_usa-site-alert.scss
```diff
@@ -1,24 +1,25 @@
 @use "sass:math";
 @use "uswds-core" as *;
 
-// Alert variables ---------- //
-$site-alert-icons: (
-  info: "info",
-  emergency: "error",
-);
-
```
amyleadem, line 9: > [!Note]
> Moved these icons to `alert-status-styles.scss` so that both alert and site alert have the same variants. This enables us to recommend users use the site alert markup structure for alerts that should share site margins.

packages/uswds-core/src/styles/variables/project-alerts.scss
amyleadem, line 1: > [!Note]
> I changed the alignment to use margins instead of padding. This eliminated the need for the optical variables here. Since we no longer needed these `optical` variables, I went ahead and moved `$alert-slim-icon-size` to `alert-status-styles.scss`. Please let me know if this goes against our standards.

packages/usa-alert/src/usa-alert.stories.js
```diff
@@ -1,6 +1,11 @@
 import Component from "./usa-alert.twig";
 import TestComponent from "./test/test-patterns/test-usa-alert-lists.twig";
 import ComparisonComponent from "./test/test-patterns/test-alert-comparison.twig";
+import onPageComponent from "./test/test-patterns/test-alert-in-template.twig";
+import BannerContent from "../../usa-banner/src/content/usa-banner.json";
+import FooterContent from "../../usa-footer/src/usa-footer.json";
+import SiteAlertContent from "../../usa-site-alert/src/content/usa-site-alert~info.json";
+import SidenavContent from "../../usa-sidenav/src/usa-sidenav.json";
```
mahoneycm, line 8: **quibble (non-blocking):** I'm curious if there's a way to collect this data in the twig template without defining each line how we were before. Since this is just for the test story, I'd be interested in excluding it from the `stories.js` file.

That said, I toyed around with importing the JSON directly to the template with no luck.

```twig
{% set bannerContent %}
  {% include "@components/usa-banner/src/content/usa-banner.json" %}
{% endset %}

{% include '@components/usa-banner/src/usa-banner.twig' with bannerContent %}
```

Do you know of anyway we could accomplish this?

packages/usa-alert/src/usa-alert.stories.js
```diff
@@ -51,4 +57,16 @@ Warning.args = WarningContent;
 
 export const Test = TestTemplate.bind({});
 
-export const AlertComparison = ComparisonTemplate.bind({});
+export const AlertComponentsComparison = ComparisonTemplate.bind({});
```
mahoneycm, line 60: **polish:** Since these are test stories, we should add test to their names.
```suggestion
export const TestAlertComponentsComparison = ComparisonTemplate.bind({});
```

Let me know what you think!

packages/usa-site-alert/src/styles/_usa-site-alert.scss
```diff
@@ -1,19 +1,24 @@
...
     @include alert-styles;
+
+    .usa-alert__body {
+      // Align content with site margins at desktop
+      @include u-maxw($theme-site-alert-max-width);
+      @include at-media($theme-site-margins-breakpoint) {
+        @include u-margin-x("auto");
+        left: calc(
+          units($theme-site-margins-width) - units($theme-alert-bar-width) / 2
+        );
```
mahoneycm, line 15: **note:** Confirming that this matches current site alert specificity 👍

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -84,10 +95,13 @@
...
-  @include at-media($theme-site-margins-breakpoint) {
-    padding-left: units($theme-site-margins-width) * 2;
-    padding-right: units($theme-site-margins-width) * 2;
+  .usa-alert__heading,
+  .usa-alert__text {
+    margin-left: $alert-x-align-text-with-icon;
+
+    @include at-media($theme-site-margins-breakpoint) {
+      margin-left: $alert-x-align-text-with-icon;
+    }
```
mahoneycm, line 104: **Question:** Was there a reason for this media-query? It applies the same margin-left that is set in the line above it. 

image

Removing it doesn't appear to affect either alert.

**Suggestion:** Remove media query.

```suggestion
  .usa-alert__heading,
  .usa-alert__text {
    margin-left: $alert-x-align-text-with-icon;

    @include at-media($theme-site-margins-breakpoint) {
      margin-left: $alert-x-align-text-with-icon;
    }
```

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -134,38 +141,41 @@
...
-    );
-    @include at-media($theme-site-margins-breakpoint) {
-      padding-left: 2 * $alert-icon-optical-padding;
+    .usa-alert__heading,
+    .usa-alert__text {
+      margin-left: 0;
+
+      @include at-media($theme-site-margins-breakpoint) {
+        margin-left: 0;
+      }
```
mahoneycm, line 150: **Question:** Same as before, do we need this media query if we're applying the same margin as the line above?

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -113,17 +127,10 @@
...
     position: absolute;
-    top: units($theme-alert-padding-y) * 0.75;
-    @include at-media($theme-site-margins-breakpoint) {
-      left: calc(
-        units($theme-site-margins-width) - units($theme-alert-bar-width)
-      );
-    }
+    top: calc(
+      units($theme-alert-padding-y) - units($theme-alert-icon-size) / 6
+    );
```
mahoneycm, line 133: **Question:** Could you speak to this calculation? Why are we dividing by 6?

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -134,38 +141,41 @@
...
 // Variant - slim
 @mixin add-slim-alert-icon {
   &:before {
-    background-size: $alert-slim-icon-size;
-    height: $alert-slim-icon-size;
-    top: units($theme-alert-padding-y) * 0.5;
-    width: $alert-slim-icon-size;
     @supports (mask: url("")) {
-      mask-size: $alert-slim-icon-size;
+      mask-size: units($alert-slim-icon-size);
```
mahoneycm, line 159: **praise:** Thanks for catching this and adding units!

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -134,38 +141,41 @@
...
     }
+    height: units($alert-slim-icon-size);
+    width: units($alert-slim-icon-size);
+    top: units($alert-slim-icon-padding);
```
mahoneycm, line 163: **quibble (non-blocking):** Can we swap these to be in alphabetical order

```suggestion
    top: units($alert-slim-icon-padding);
    width: units($alert-slim-icon-size);
```

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -134,38 +141,41 @@
...
-      padding-left: calc(
-        units($theme-site-margins-width) + $alert-slim-icon-size
-      );
+    .usa-alert__heading,
+    .usa-alert__text {
+      margin-left: $alert-x-align-text-with-slim-icon;
+
+      @include at-media($theme-site-margins-breakpoint) {
+        margin-left: $alert-x-align-text-with-slim-icon;
+      }
```
mahoneycm, line 178: **Question:** Same as before, do we need this media query if we're applying the same margin as the line above?

mahoneycm (commented):
Overall this is looking good! Left a couple questions to clear up and then we should be good. Thanks @amyleadem !

packages/usa-alert/src/usa-alert.stories.js
amyleadem: Good point - updated in [83a5fb3](https://github.com/uswds/uswds/pull/5636/commits/83a5fb3cb786d19ee92922ba0bf179aea26ac23b)

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
amyleadem: Good catch - removed the redundant/unnecessary media queries in [8e5295c](https://github.com/uswds/uswds/pull/5636/commits/8e5295cd51d7fdd232d96f514025210d6d987b5f)

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
amyleadem: Good catch - removed the redundant/unnecessary media queries in [8e5295c](https://github.com/uswds/uswds/pull/5636/commits/8e5295cd51d7fdd232d96f514025210d6d987b5f)

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
amyleadem: Good catch - removed the redundant/unnecessary media queries in [8e5295c](https://github.com/uswds/uswds/pull/5636/commits/8e5295cd51d7fdd232d96f514025210d6d987b5f)

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
amyleadem: Updated in [8e5295c](https://github.com/uswds/uswds/pull/5636/commits/8e5295cd51d7fdd232d96f514025210d6d987b5f)

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
amyleadem: Thanks for asking this. I think it is a good question.

This calculation creates some visual accommodation for the vertical alignment in the icon. It is more resilient than the previous top calculation because it ties it more closely to the header position, rather than relying solely on some percentage of the vertical padding. 

I found that subtracting 1/6 of the icon size gave the most favorable visual results at a mix of icon sizes _greater than or equal to_ the default size of 4 units (combined with the default header size). However, the top position would likely need to be customized if the icon size is less than 4 units. The 1/6 number was inspired by `$alert-icon-optical-factor` in [project-alerts.scss](https://github.com/uswds/uswds/pull/5636/files#diff-a629fe0eb65a35c445fc909bde2aa36d7fe8dd17bdd35bfbb6b7da4583365efdL8). 

I figured making the icon larger than default would be the more common icon size customization, assuming that users that wanted the smaller icon would choose the slim variant or do other manual CSS customizations. This might not be the appropriate assumption. 

I can see that this number might feel arbitrary. Let me know if you think I should revisit it.

packages/uswds-core/src/styles/variables/project-alerts.scss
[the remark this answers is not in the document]
amyleadem: Good note. I realize the test instructions were not clear. I've updated the PR description to read "At mobile, alerts and site alerts share the same x alignment". 

**question:** After looking deeper into this test, I am thinking that site alert and alert actually probably _shouldn't_ necessarily align at mobile. I am thinking now that site alert should not be influenced at all by `$theme-alert-padding-x`, even in mobile. (And to be fair, I'm also wondering if `$theme-alert-padding-x` should even affect alert at mobile).

 I took a look at release 3.1.0, which was the release before the original alert alignment PR in [3.2.0](https://github.com/uswds/uswds/releases/tag/v3.2.0), and it looks like the alert component _was_ affected by `$theme-alert-padding-x` at all screen widths, but site alert _was not_ affected by `$theme-alert-padding-x` at any screen width. I am thinking of restoring that behavior here. Any objections?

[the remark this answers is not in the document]
amyleadem: **follow-up question:** After looking deeper into this test, I am thinking that site alert and alert actually probably _shouldn't_ necessarily align at mobile. I am thinking now that site alert should not be influenced at all by `$theme-alert-padding-x`, even in mobile. (And to be fair, I'm also wondering if `$theme-alert-padding-x` should even affect alert at mobile).

 I took a look at release 3.1.0, which was the release before the original alert alignment PR in [3.2.0](https://github.com/uswds/uswds/releases/tag/v3.2.0), and it looks like the alert component _was_ affected by `$theme-alert-padding-x` at all screen widths, but site alert _was not_ affected by `$theme-alert-padding-x` at any screen width. I am thinking of restoring that behavior here. Any objections?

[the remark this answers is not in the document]
mahoneycm: Looking at the settings page, this setting is only found for `alert` and _not_ `site-alert` so that might indicate there's less chance for visual regression for users who have used this setting.

You are more familiar with this component and it's styles + behaviors so I trust your judgment here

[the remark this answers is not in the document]
amyleadem: I updated the Sass so that site alert _does NOT_ use `$theme-alert-padding-x` at any screen width. Alert will use this setting at all screen widths. 

[5b8e2d6](https://github.com/uswds/uswds/pull/5636/commits/5b8e2d64b28cc685a709eedadc0c32ce9a9b40cd)

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
amyleadem: Wanted to return to this with a bit more of a math focus.

It gets a bit complicated to calculate this since we are trying to align the top with a normalized font size and line height, but generally, we can figure out that our circular icons have a padding of 1/12 _built into the icon itself_:

Path size of 20x20
image

SVG canvas size of 24x24
image

Based on this, to align the icon to the top of the headline text, we can reasonably assume that we should adjust the icon by at least 1/12 (2px / 24px). This is what the standard icon would look like if we divided by 12 rather than 6 here:
image

You can see the top of the headline is generally aligned with the top of the icon here. But visually, since the standard icon is slightly larger than the headline, an additional negative adjustment is helpful for the icon to appear visually centered. Because the font size is set with typeset and `$theme-alert-font-family`, it is not easy to glean the rendered font-size to include in the math. So manually poking at it, I landed on 6 as the best division option here across variants. In the end, it is a bit arbitrary, but it is also very simple to override in a project if needed. 

image

At the moment, I can't think of a better way to position this while setting the icon with a pseudo element. Open to other ideas if you have them!

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -1,16 +1,32 @@
...
 @use "../../functions" as *;
-@use "./at-media" as *;
 @use "./border-box-sizing" as *;
 @use "./set-text-and-bg" as *;
 @use "../general/icon" as *;
 @use "../utilities" as *;
 @use "../typography/typeset.scss" as *;
 @use "../helpers/set-link-from-bg" as *;
 
+$alert-icon-gap: 1.5;
```
amyleadem, line 11: > [!Note]
> This is the gap between the icon and the text. The gap is currently the same regardless of icon size. From a design perspective, it might be beneficial to adjust this size proportionally to the icon size. I personally have decided the gap looks reasonable as is across variants, but flagging in case it bothers anyone else.

packages/usa-site-alert/src/styles/_usa-site-alert.scss
```diff
@@ -1,19 +1,26 @@
...
-  emergency: "error",
-);
-
 .usa-site-alert {
   .usa-alert {
     @include alert-styles;
+
+    .usa-alert__body {
+      @include u-maxw($theme-site-alert-max-width);
+      @include u-margin-x(2.5);
```
amyleadem, line 10: > [!Note]
> This is set to match the default value for `$theme-alert-padding-x`.

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -1,16 +1,34 @@
...
+// shared icon styles
+$alert-x-align-text-with-icon: calc(
+  units($theme-alert-icon-size) + units($alert-icon-gap)
+);
+$alert-x-align-text-with-slim-icon: calc(
+  units($alert-slim-icon-size) + units($alert-icon-gap)
+);
+$alert-icon-top-align: calc(
+  units($theme-alert-padding-y) - units($theme-alert-icon-size) / 6
+);
```
amyleadem, line 31: > [!Note]
> This value was moved in the code, but there is active discussion about it in this [comment](https://github.com/uswds/uswds/pull/5636#discussion_r1761688893).

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -83,11 +94,13 @@
   );
   @include add-alert-icon($icon, $banner-text-color-token, $bgcolor);
   @include set-text-and-bg($bgcolor);
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  min-height: calc(units($theme-alert-icon-size) + $alert-icon-top-align * 2);
```
amyleadem, line 100: > [!Note]
> These lines were added to vertically center the text content when the icon size is taller than the content.

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -113,17 +126,10 @@
...
-      );
     position: absolute;
-    top: units($theme-alert-padding-y) * 0.75;
-    @include at-media($theme-site-margins-breakpoint) {
-      left: calc(
-        units($theme-site-margins-width) - units($theme-alert-bar-width)
-      );
-    }
+    // Adjusts the left position to accommodate the empty space built into the icon svgs
+    left: calc(units($theme-alert-icon-size) * -1 / 12);
```
amyleadem, line 131: > [!Note]
> This adjustment was added because the alert icons have some empty padding built into the icon svgs. When the svg size is set to 24x24 pixels, the icon path is only 20x20 pixels. That means there are 2px of empty space on all sides of the svg, which means it needs a 1/12 (2px/24px) adjustment to allow the icon path to align with other content.

image
image

mejiaj (commented):
@amyleadem thanks for all your work untangling this. The tests you've added have helped a lot too.

Tested the following:

### Defaults

- [x] Alert looks good on mobile & desktop
- [x] site alerts look good on mobile & desktop

### Customization

Via settings there were too many combinations, notes, and questions. I've added my comments in the relevant lines of code to be able to use threads to discuss and resolve. If I were to post all my findings in the review comments it'd be much harder to have a discussion.

packages/usa-alert/src/test/test-patterns/test-alert-in-template.twig
```diff
@@ -0,0 +1,87 @@
...
+<div class="usa-section">
+  <div class="grid-container">
+    <div class="grid-row grid-gap">
+      <div class="usa-layout-docs__sidenav desktop:grid-col-3">
+        {% include '@components/usa-sidenav/src/usa-sidenav.twig' %}
+      </div>
+
+      <main class="usa-layout-docs__main desktop:grid-col-9 usa-prose usa-layout-docs" id="main-content">
+
+        {% for alert in alerts %}
```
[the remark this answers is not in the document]
amyleadem: If I understand correctly, this is expected. The component is designed to share left alignment at the left edge of the icons. The text itself is not expected to share the same left alignment across slim and standard variants. This is the pre-existing behavior in [3.1.0](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/release-3.1.0/?path=/story/components-alert--info):

image

Please let me know if I misunderstood.

amyleadem (commented):
@mejiaj  I have addressed a lot of your comments, but I still have a few things remaining to investigate. I just wanted to post the comments I could. Here are my remaining tasks:
1. I still need to go back through the template code and see if I can reuse any of it across templates. 
1. I need to go back in and confirm that the alignment method here is the most logical.

The comments are a bit of a tangled mess, so I've pulled out the open conversations here:

1. What role should `$theme-alert-padding-x` play in site alert? [Related comment](https://github.com/uswds/uswds/pull/5636#discussion_r1769322148)
1. Should we make site alert variants for warning, error, etc? [Related comment](https://github.com/uswds/uswds/pull/5636#discussion_r1742829000)
1. Is there an issue with the alignment of the text in slim variants vs standard variants? [Related comment](https://github.com/uswds/uswds/pull/5636#discussion_r1769260302)

We can discuss these in our meeting later today and decide if we want to move forward with this in 3.9.0.

packages/uswds-core/src/styles/variables/project-alerts.scss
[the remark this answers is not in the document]
amyleadem: Note: More context for this conversation can be found in the [related comment thread](https://github.com/uswds/uswds/pull/5636#discussion_r1767643159) from @mejiaj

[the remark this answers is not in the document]
amyleadem: I have added some negative left alignment to accommodate the empty space in the svgs. More details in [this comment](https://github.com/uswds/uswds/pull/5636#discussion_r1767649986).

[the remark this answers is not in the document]
amyleadem: Note: During offline conversation, we decided that if we go with option 3, we should add documentation stating that this setting in particular does not apply to site alert. Add clarity that site alert margins are determined by site margins. We will bring this conversation to dev sync.

packages/usa-alert/src/test/test-patterns/test-alert-in-template.twig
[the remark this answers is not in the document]
amyleadem: > [!Note]
> During offline discussion, we agreed that this is ok because it is the expected display.

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -134,38 +136,33 @@
...
-      mask-size: $alert-slim-icon-size;
+      mask-size: units($alert-slim-icon-size);
     }
+    height: units($alert-slim-icon-size);
+    top: units($alert-slim-icon-padding);
+    width: units($alert-slim-icon-size);
   }
 }
 
 @mixin alert-styles-slim {
```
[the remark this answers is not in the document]
amyleadem: > [!Note]
> Resolving this comment because we discussed offline that this is expected display.

packages/usa-alert/src/styles/_usa-alert.scss
amyleadem: Resolving this comment because this change was reverted in [d3a2a08](https://github.com/uswds/uswds/pull/5636/commits/d3a2a083b4adb5545711acaccfcbaeeaaf278c82)

packages/usa-site-alert/src/styles/_usa-site-alert.scss
amyleadem: In the interest of performance, this change was reverted in [d3a2a08](https://github.com/uswds/uswds/pull/5636/commits/d3a2a083b4adb5545711acaccfcbaeeaaf278c82). Site alert and alert will keep the same variants they currently have in 3.8.2

packages/usa-alert/src/styles/_usa-alert.scss
```diff
@@ -2,18 +2,13 @@
...
 .usa-alert {
   @include alert-styles;
 
+  .usa-alert__body {
+    @include u-margin-x($theme-alert-padding-x);
```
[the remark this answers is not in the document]
amyleadem: Yes, good note. I updated the alignment to use padding instead of margin in [069cb7b](https://github.com/uswds/uswds/pull/5636/commits/069cb7b745ebdffa0cbd2fc744ba698061d1c766) and [914ca8a](https://github.com/uswds/uswds/pull/5636/commits/914ca8abfc332bd4655b4a09510caef531bd3055)

packages/usa-alert/src/test/test-patterns/test-alert-in-template.twig
```diff
@@ -0,0 +1,87 @@
...
+{% include '@components/usa-site-alert/src/usa-site-alert.twig' %}
+
+<div class="usa-section">
+  <div class="grid-container">
+    <div class="grid-row grid-gap">
+      <div class="usa-layout-docs__sidenav desktop:grid-col-3">
+        {% include '@components/usa-sidenav/src/usa-sidenav.twig' %}
+      </div>
+
+      <main class="usa-layout-docs__main desktop:grid-col-9 usa-prose usa-layout-docs" id="main-content">
```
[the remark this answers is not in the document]
amyleadem: After diving into this, there is more of a difference here between the templates than just the grid. I want to showcase different components and content in different regions. That said, I was able to share the alert test content across templates. 

For clarity and ease of maintenance, I prefer to leave the templates as is, but let me know if you have a strong preference to rework it.

packages/usa-alert/src/test/test-patterns/test-alert-comparison.twig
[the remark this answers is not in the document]
amyleadem: Not sure if this is what you were suggesting, but I created `test-alert-content.json` to be shared across `test-alert-comparison.html` and `test-alert-in-template.html` in [92d864f](https://github.com/uswds/uswds/pull/5636/commits/92d864f87c47bcf91d6dc71ce8211c79d4d5b031).

packages/uswds-core/src/styles/mixins/helpers/alert-status-styles.scss
```diff
@@ -83,11 +94,13 @@
...
+  flex-direction: column;
+  justify-content: center;
+  min-height: calc(units($theme-alert-icon-size) + $alert-icon-top-align * 2);
 
-  padding-left: units($theme-alert-icon-size) + ($alert-icon-optical-padding);
-  @include at-media($theme-site-margins-breakpoint) {
-    padding-left: units($theme-site-margins-width) * 2;
-    padding-right: units($theme-site-margins-width) * 2;
+  > * {
+    margin-left: $alert-x-align-text-with-icon;
```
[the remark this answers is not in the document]
amyleadem: Update: Because the variable was no longer reused anywhere, I went ahead and removed the variable and just defined the value directly in the style rule. 

Additionally, I updated a lot of the methods for horizontal alignment, but think I want to keep this as is instead of moving it to body. The body is no longer oriented around just the text but the icon as well. To me, this is clearer and easier to work with but let me know if you want something different. Happy to discuss!

amyleadem (commented):
@mejiaj I believe I have addressed all your comments and made the updates requested in our discussion today. I did the following:

1. Reverted the change that created additional variants for site alert. Site alert and alert will continue to have the same variants that they did in 3.8.2 and before. 
1. Updated the alignment methods so that it uses `padding` rather than `left` and `margin`. This should improve clarity in the setting names and will make styling the components a little more intuitive.
1. Made no change regarding the role of `$theme-alert-padding-x` in site alert. After our group discussion, we agreed that all other alert settings will affect site alert but not this one. This is because of the core horizontal alignment differences across components. As a next step, I will enhance the related documentation to make this clear. 
1. Created a shared JSON file (`test-alert-content.JSON`) for alert test content and removed the content from the test twig files.
1. Confirmed that the template markup is different enough to keep them separate (but let me know if you disagree)

I _think_ that is it. Let me know if you have any questions or want to walk through this in real time together.

packages/usa-alert/src/content/test-usa-alert-content.json
amyleadem, line 1: > [!Note]
> I originally placed this file in the `/test-patterns` directory, but it caused an error when running `npm run build:html`. I moved it to a couple of different spots, but the only place it didn't cause an error is in this `/content` directory. If you see a better place for it, please let me know!

packages/usa-alert/src/test/test-patterns/test-alert-in-template.twig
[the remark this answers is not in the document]
mejiaj: Just wanted to make sure I wasn't missing anything to compare against. Thanks

packages/uswds-core/src/styles/variables/project-alerts.scss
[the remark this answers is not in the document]
amyleadem: Additional note: During dev sync, we decided it was appropriate to move forward with option 3. This means that `$theme-alert-padding-x` should have no effect on site alert, while all the other alert settings should.

mejiaj (commented):
@amyleadem I was able to see the new defaults and modify settings in your sandbox branch [`al-test-5636-alert-align`] without any issues. Thanks for creating that!

Settings I tested:
```scss
// _uswds-theme.scss

// Tested with varying combinations.
@use "uswds-core" with (
  $theme-site-margins-width: 4,
  $theme-alert-bar-width: 2,
  $theme-alert-icon-size: 2,
  $theme-alert-padding-x: 4,
  $theme-alert-padding-y: 4,
);
```

Added a few comments to improve readability and code re-use templates.

packages/usa-alert/src/test/test-patterns/test-alert-comparison.twig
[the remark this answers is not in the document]
amyleadem: I have updated the test stories to use includes in [76be26f](https://github.com/uswds/uswds/pull/5636/commits/76be26ff9d1d0c59872ceba0b90903d10760ff83)

packages/usa-alert/src/test/test-patterns/test-alert-in-template.twig
```diff
@@ -0,0 +1,33 @@
...
+          <div class="usa-alert {{ alert.alertModifier }}">
+            <div class="usa-alert__body">
+              {% if alert.title %}<h4 class="usa-alert__heading">{{ alert.title }}</h4>{% endif %}
+              <p class="usa-alert__text">
+                Lorem ipsum dolor sit amet,
+                <a class="usa-link" href="javascript:void(0);">consectetur adipiscing</a>
+                elit, sed do eiusmod.
+              </p>
+            </div>
+          </div>
```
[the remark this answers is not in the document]
amyleadem: I have updated the test stories to use includes in [76be26f](https://github.com/uswds/uswds/pull/5636/commits/76be26ff9d1d0c59872ceba0b90903d10760ff83)

packages/usa-alert/src/content/test-usa-alert-content.json
```diff
@@ -0,0 +1,102 @@
+{
+  "alerts": [
+    {
+      "header": "Info alert",
+      "title": "Info alert",
+      "modifier": "usa-alert--info",
+      "text": "Lorem ipsum dolor sit amet sed do eiusmod."
```
mejiaj, line 7: **polish:** You can set and forget in the template if it's all the same. 

Makes the data more scannable because you only have to worry about the differences.

**Example**
```suggestion
```

Then set `text` in the test template because it'll load the template context and then what you've set.

```twig
{# test-alert-comparison.twig #}

{# Shared text for Alert and Site alert. #}
{% set text = "Lorem ipsum dolor sit amet sed do eiusmod." %}
```

**Output**
image

mejiaj (commented):
@amyleadem one suggestion for polish, but otherwise LGTM.

packages/usa-alert/src/content/test-usa-alert-content.json
amyleadem: Oooo...nice! Updated in [d42d4d3](https://github.com/uswds/uswds/pull/5636/commits/d42d4d390d28680add1c1a26c7e3c1982f68bec7)

mejiaj (approved):

RachelCorsino (commented):
## Default Settings
- [X] At mobile, confirm that all alert and site alert variants have the same left and right alignment
- [X] At desktop, confirm that site alert content shares left and right alignment with banner and footer
- [X] At desktop, confirm that all alert variants share the same left alignment

## Settings Adjustments
### Set `$theme-site-margins-width` to 15 and 0
- [X] At mobile, all alerts and site alerts share the same left and right alignment

_Maybe I'm misunderstanding what I should be seeing from this test, but I see the site-alert alignment different from the alert alignment in mobile view? This screenshot is with `theme-site-margins-width` set to 0._
Screenshot 2024-10-01 at 3 09 13 PM

- [X] At desktop, all site alerts share the same left and right alignment as the banner
- [X] At desktop, all alerts share the same left and right alignment as each other
- [X] The gap between the icon and the alert text does not change from default

_Everything else looks good here!_

### Set `$theme-alert-bar-width` to 15 and 0
- [X] The left colored bar adjusts to match the declared width
- [X] Text content does not overlap with the colored bar

_Everything looks good here!_

### Set `$theme-alert-padding-x` to 15 and 2. For each instance, confirm in the test stories that
- [X] At desktop and mobile, the left and right alignment on the alerts adjusts to match the new setting value
- [X] At desktop, all site alerts maintain the same left and right alignment as the banner
- [X] At mobile, all site alerts share the same left and right alignment of 2.5 units

_Everything looks good here!_

### Set `$theme-alert-icon-size` to 6 and 2
- [X] The standard icon size adjusts to match the declared value (this should not affect the slim icon implementations)
- [X] At all window widths, the text for alert and site alert components adjusts left alignment to match the new icon size. The gap between the icon and the text should not change, and the elements should not overlap.

_Everything looks good here!_

### Set `$theme-alert-padding-y` to 6 and 0
- [X] The vertical padding adjusts to match the declared value (this does not apply to the slim variant)
- [X] All other elements align as expected

_Everything looks good here!_

## Other
- [X] Adjusting $theme-alert-padding-x adjusted the gap between the icon and text in 3.1.0. It no longer does that in this PR

_Confirmed that the gap between icon and text in this PR does not change based on `$theme-alert-padding-x` setting_

- [X] The [site alert in 3.1.0](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/release-3.1.0/?path=/story/components-site-alert--info) does NOT have a bar on left side of the component. It does in [develop](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-site-alert--info) and in [this PR](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-alert-icon-alignment/?path=/story/components-site-alert--info).

_Confirmed that the bar on the left side of the component appears in this PR_

amyleadem:
@RachelCorsino 
I consider the screenshot you shared to be expected behavior in `uswds-sandbox` because the alert and other main content lives inside a `grid-container` that has some extra x-padding, whereas the site alert does not. So the difference in alignment isn't coming from the component itself. 

If you look at the [Test alert site alert comparison story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-alert-icon-alignment/?path=/story/components-alert--test-alert-site-alert-comparison), you should see that the two components align in mobile. Please let me know if you have any questions!

RachelCorsino:
@amyleadem - I see it working as expected in the [Test alert site alert comparison story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-alert-icon-alignment/?path=/story/components-alert--test-alert-site-alert-comparison).

Everything looks good from my tests - Approved!

RachelCorsino (approved):

thisisdano (approved):
A lot of work here! It will be nice to move this gnarly little issue off your plate!

PR #5654, opened 2023-12-04
USWDS - Header: Vertically center icon in nav

mahoneycm:
# Summary

Fixed the vertical alignment of the expand icon on header navigation items.

## Related issue 
Closes #5660 

## Preview link

[Header component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/feature-vertically-center-nav-elements/?path=/story/components-header--default)

## Related PR
This work originated in https://github.com/uswds/uswds/pull/5267 (Thanks @aduth!)
[Changelog →](https://github.com/uswds/uswds-site/pull/2389)

## Problem statement

When navigation links include text which breaks to multiple lines (particularly in mobile navigation), the expander icon is shown at the bottom right of the link, which makes it more difficult to identify.

## Solution

Vertically center the icon.

Before|After
---|---
Screen Shot 2023-05-02 at 10 51 40 AM|Screen Shot 2023-05-02 at 10 52 04 AM

## Testing and review

1. Visit header variant pages
2. Inspect mobile styles
3. Create large link span for accordion buttons
4. Confirm expander icon maintains vertical alignment

### Testing checklist

- [ ] Mobile navigation expander icons maintain vertical alignment
- [ ] No visual regression in desktop variants

PR #5655, opened 2023-12-04
USWDS - Nav: Resolve padding overflow bug

### LINKED ISSUE #5656: USWDS - Header: Line breaks cause text to overflow into padding area
## Description

When navigation button spans cause a line break, the span overflows into the padding area.

1. Visit [header story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-header--default)
2. Edit span text to cause a line break
3. Inspect element

## Screenshot
image

mahoneycm:
# Summary

Fixed a bug that caused long nav button spans to overflow into its padding. Now, nav buttons properly break into two lines and maintain center alignment.

## Breaking change

## Related issue

Closes #5656

## Related pull requests

#5267
[Changelog →](https://github.com/uswds/uswds-site/pull/2387)

## Preview link

[All headers demo →](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/cm-demo-5655/)

[Header storybook preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-vertical-center-nav-padding-fix/?path=/story/components-header--default)

## Problem statement

When navigation button spans cause a line break, their text flows into their padding instead of breaking onto a new line

image

## Solution

- Set the `span` element to `display: inline-block`
- Update `usa-nav__primary` child item flex alignment to `center`

## Testing and review

1. Visit the [testing repo](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/cm-demo-5655/) or [Header storybook preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-vertical-center-nav-padding-fix/?path=/story/components-header--default) story
2. In desktop views:
    1. Confirm larger spans do not overflow into padding area
    2. Confirm icons maintain center alignment
    3. Confirm nav buttons maintain equal size
    4. Confirm appropriate default and hover states
3. In mobile views:
    1. Confirm there is no visual regression
    2. Confirm icons are vertically aligned

### Testing checklist

- [ ]  Long nav button spans do not break center alignment
- [ ]  Buttons maintain equal size when one span breaks into two lines
- [ ]  Mobile styles are unaffected
- [ ]  Hover state is unaffected
- [ ]  No visual regressions

## Further considerations

> [!NOTE]
> There seems to be a CSS behavior that causes a width calculation quirk and leaves empty space between the span and the icon. This behavior is present on `develop`.
>I briefly tried resolving but didn't find any quick solutions. Do we think this is out of scope or should we dig deeper?

Screenshot 2023-12-05 at 1 39 36 PM
Screenshot 2023-12-05 at 2 20 21 PM


mejiaj (approved):
Looks good, thank you. Tested mobile and desktop for regressions on chromium.

amyleadem (changes requested):
@mahoneycm I noticed that changing the display to `inline-block` caused the nav items to change height, which results in the submenu to jump up and down if one nav item is taller than another: 

image

I quickly played with a possible fix and found that adding the following restores the height. The following code renders the screenshot below. 

```
.usa-nav__primary {align-items:stretch}
.usa-nav__primary button {height:100%}
```

image

It might also be worth exploring alternatives to `inline-block` as well.

Once this issue is addressed, I'll test some more to make sure it doesn't have a risk for inadvertent changes.

mahoneycm:
@amyleadem Thank you so much for spotting that! Definitely an improvement and will prevent any weird click area issues.

After some testing, I noticed that the last nav item (the link) could become uncentered if another button's text breaks into three lines. This seems unlikely at the current font size but possible if a user were to override the nav link sizes.

I toyed with it for a while but was unable to get the link to maintain its vertical alignment without updating markup. I was hoping to find a solution to center the items while maintaining the full height but was unable to, unfortunately.

Do you think this is enough of an edge case to not worry about? It seems unlikely to me but I was hoping to cover all our bases.

image

image

packages/usa-nav/src/styles/_usa-nav.scss
```diff
@@ -224,6 +226,8 @@ $expand-less-icon: map-merge(
 
     &[aria-expanded] {
       span {
+        display: inline-block;
+
         &::after {
           position: absolute;
           top: 50%;
```
mahoneycm, line 233: Im curious if it would make more sense if we moved these lines to `usa-nav__primary.usa-accordion__button.span` on line 300 instead of `[aria-expanded]. Seems like it might be a bit more descriptive but it is further away from the other `aria-expanded=true/false` styles which are relevant to this work.

https://github.com/uswds/uswds/blob/05e4ee234628ca6f45a3ca76ef1e470a02a0add3/packages/usa-nav/src/styles/_usa-nav.scss#L307-L315

packages/usa-nav/src/styles/_usa-nav.scss
mahoneycm, line 1: ## Redundant media query

Unrelated to this work but it looks like we have a duplicate media query on line `291`. 

https://github.com/uswds/uswds/blob/05e4ee234628ca6f45a3ca76ef1e470a02a0add3/packages/usa-nav/src/styles/_usa-nav.scss#L286-L300

packages/usa-nav/src/styles/_usa-nav.scss
```diff
@@ -161,15 +162,23 @@ $expand-less-icon: map-merge(
...
           color: color("primary");
         }
       }
     }
+
+    > button,
+    > a {
+      @include at-media($theme-header-min-width) {
+        height: 100%;
+      }
```
mejiaj, line 180: > [!CAUTION]
> Selectors like this are heavy handed and could cause breaking changes. Acceptable because this component heavily relies on this pattern. We should use this with a lot of caution.

mejiaj (approved):

amyleadem (approved):
Tested the following in Chrome, Safari, and Firefox at desktop widths. 

- [x] Navbar items align to vertical center when one item spans multiple lines and other items span one line
    - [x] Test when current item is two lines, other items one line
    - [x] Test one non-current item is two lines, others are one line
    - [x] Test when simple link is two lines, others are one line
- [x] Navbar items align to vertical center when all items are one line
    - No changes from `develop` except a small shift in icon location
- [x] Navbar items align to vertical center when all items span multiple lines
- [x] The code affects only the mobile presentation.
    - All changes are inside the $theme-header-min-width breakpoint
- [x] Everything still works after pulling in #5654

thisisdano (approved):

PR #5289, opened 2023-05-16
USWDS - Footer: Remove grid dependency.

danbrady:
# Summary

**Removed grid dependency from Footer package.** Now you no longer need to use the usa-layout-grid package with usa-footer resulting in a much smaller package size.

## Breaking change

Not a breaking change. Only CSS refactored. 

## Related issue

Closes #5287

## Related pull requests

TBD

_Note: Remove layout-grid from the [dependency lists](https://designsystem.digital.gov/components/footer/#package) from doc site._

## Problem statement

The footer currently has the layout-grid dependency although very little of it is used. This adds additional, unnecessary file weight.

## Solution

Removed the grid layout dependency and brought layout styles into the footer directly. 

Possible limitation: I only brought in the column sizes that are needed to support the three variations of the footer (default, slim, big) at all the breakpoints. If alternate layouts are  desired, addition styles would need to be added. For example, changing the number of columns or how wide the link columns span in the big footer. 

## Testing and review

Existing unit test still completes successfully. 
Compared the unminified CSS size of the Footer before and after changes: 63kb -> ~40kb 

### Isolate Footer package
- In `uswds/_index.scss` remove all packages besides `usa-banner`
- Check `dist/css/uswds.css` filesize
- Validate appearance of all footer variations at all breakpoints are acceptable

~~Looking for feedback, please:~~

- ~~Existing footer styles has reference to "disclosure button styles". I didn't remove them, but don't see them in any template markup. Are these needed?~~
- ~~I updated existing styles to use utility mixins as opposed to native CSS properties. It was mixed before, is there a preference?~~  
- ~~As mentioned in the possible limitations above, I only brought in the columns that are used for the footer variations. If this is the right approach, I can make a mixin to output these in a more concise, readable way.~~ 
- ~~I make use of the 'gap' property for flexbox items. This is not supported in IE11 but all other supported browsers. USWDS v3 states IE11 support was dropped. Is this ok?~~
- ~~FYI: I also separated the logo lockup and address into includes as the markup was the same for multiple footer variations.~~

~~FYI: @mejiaj~~

PR #5675, opened 2023-12-13
USWDS - Footer: Grid layout hotfix

### LINKED ISSUE #5676: USWDS - Bug: Removing layout-grid dependency causes style conflicts.
### Describe the bug

Changes in `3.7.1` (#5289) cause visual regressions in customized footer components due to specificity conflicts with grid-layout classes.

The grid classes defined in the footer scss take precedence over standard grid classes.

[Slack thread](https://gsa-tts.slack.com/archives/C3F14AHSQ/p1702323288819739) with additional context

To reproduce the bug, we'll need to update the footer's grid column value.

1. Visit this [footer demo](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/cm-footer-grid-test/) on desktop widths.
   - The first footer link uses `desktop:grid-col-8`
3. Open devtools and inspect the first primary link `li` element
4. Review output styles.
   - `width` is set by `.usa-footer .grid-row .mobile-lg:grid-col-4` _// 33.33%_
   - `width` should be set by `desktop:grid-col-8` _// 66.667%_
   - `mobile-lg:grid-col-4` will take priority because it's definition in footer has higher specificity than layout-grid's `desktop:grid-col-8`

### Expected Behavior

Footer should accept customized layout grid styles when `layout-grid` package is included.

### Code of Conduct

mahoneycm:
# Summary

**Resolves style specificity issues between `usa-footer` and `usa-grid-layout`.** This resolves visual regressions potentially caused by #5289.

## Breaking change

## Related issue

Closes #5676 

## Related pull requests

Changes implemented in #5289

## Preview link

[Footer →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-footer-grid-hotfix/?path=/story/components-footer--default)

[Demo repo →](https://github.com/uswds/uswds-sandbox/tree/cm-testing-footer-layout-grid)

[Demo footer →](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/cm-testing-footer-layout-grid/)

## Problem statement

Layout grid styles defined in footer have higher specificity than other layout-grid styles. As a result, updated layout grid utility classes in footer markup do not affect layout as expected.

## Solution

> [!TIP]
> The files changed may look confusing, but it simply moves the styles above the `.footer{}` declaration instead of nesting them.

Move layout grid classes **above** `.footer` class.

**Example**
https://github.com/uswds/uswds/blob/9fe48ce5d022307a792c10e201b4497bff4bfb5d/packages/usa-footer/src/styles/_usa-footer.scss#L25-L45

## Testing and review

The demo repo below showcases how footer-defined grid styles have specificity over standard grid classes when both are imported to a project.

1. Checkout and start this [test repo](https://github.com/uswds/uswds-sandbox/tree/cm-testing-footer-layout-grid) on your local machine
    - By default, this test repo uses `3.7.1`
    - Imports both `footer` and `grid-layout`
2. Inspect markup for both footers
    - Top footer features a mobile utility class which is found in **footer** and a desktop utility class found in **grid-layout**
    - Bottom footer features two utility classes, neither of which are included in footer css
3. In **desktop** width viewports
    - Note that the top footer breaks only one link to second line.
    - The bottom footer has all links in one column.
5. Inspect `<li>` style changes
    - Note that the **footer** defined classes take priority, despite layout-grid being included
6. Install this branch
    - `npm install @uswds/uswds "https://github.com/uswds/uswds/tree/cm-footer-grid-hotfix" --save`
7. Trigger sass rebuild
8. Hard refresh the page
9. Resize window and view changes
   - Note that both footers now break links into two columns with two rows on _desktop_ widths
11. Inspect styles at breakpoints
12. Layout grid classes should now apply as expected

### Testing checklist

- [ ]  Layout grid utility classes work as expected
- [ ]  Footer styles no longer take priority
- [ ]  Confirm there is no issue with non-nested gird classes in `_usa-footer.scss`
- [ ]  `$theme-layout-grid-use-important: true` is not required for desired layout-grid styles

## Further consideration

I considered investigating a `sass:meta` test to check for the inclusion of `grid-layout` and nest the footer grid classes within that but figured this would be enough for a hotfix/mvp.

Alternatively, we could investigate using a theme setting to enable/disable these styles. I opt for a logic test _(if possible)_ so users are not required to take an extra step other than importing `grid-layout`

### DIFF
--- a/packages/usa-footer/src/styles/_usa-footer.scss
+++ b/packages/usa-footer/src/styles/_usa-footer.scss
@@ -22,11 +22,7 @@ $-chevron-expand-more: map-merge(
   )
 );
 
- // Grid styles ported over to remove dependency
- .grid-container {
-  @include grid-container($theme-footer-max-width);
-}
-
+// Grid styles ported over to remove dependency
 .grid-row {
   @include grid-row;
 }
@@ -119,6 +115,10 @@ $-chevron-expand-more: map-merge(
   @include border-box-sizing;
   @include typeset($theme-footer-font-family);
   overflow: hidden;
+
+  > .grid-container {
+    @include grid-container($theme-footer-max-width);
+  }
 }
 .usa-footer__return-to-top {
   @include u-padding-y(2.5);
--- a/packages/usa-footer/src/styles/_usa-footer.scss
+++ b/packages/usa-footer/src/styles/_usa-footer.scss
@@ -65,8 +65,11 @@ $-chevron-expand-more: map-merge(
     // Columns
     [class*="grid-col"] {
       @include u-position(relative);
-      @include u-width(full);
       box-sizing: border-box;
+    }
+
+    .grid-col {
+      @include u-width(full);
     }
 
     .grid-col-auto {
@@ -76,6 +79,10 @@ $-chevron-expand-more: map-merge(
     // Columns: Mobile Large
     /* stylelint-disable */
     @include at-media("mobile-lg") {
+      .mobile-lg\:grid-col {
+        @include u-width(full);
+      }
+
       .mobile-lg\:grid-col-auto {
         @include grid-col("auto");
       }
@@ -99,6 +106,10 @@ $-chevron-expand-more: map-merge(
 
     // Columns: Tablet
     @include at-media("tablet") {
+      .tablet\:grid-col {
+        @include u-width(full);
+      }
+
       .tablet\:grid-col-4 {
         @include grid-col(4);
       }
@@ -110,6 +121,10 @@ $-chevron-expand-more: map-merge(
 
     // Columns: Desktop
     @include at-media("desktop") {
+      .desktop\:grid-col {
+        @include u-width(full);
+      }
+
       .desktop\:grid-col-auto {
         @include grid-col("auto");
       }
--- a/packages/usa-footer/src/styles/_usa-footer.scss
+++ b/packages/usa-footer/src/styles/_usa-footer.scss
@@ -26,22 +26,22 @@ $-chevron-expand-more: map-merge(
 .grid-row {
   @include grid-row;
 
-  .grid-gap-1 {
+  &.grid-gap-1 {
     @include grid-gap(1);
   }
 
-  .grid-gap-2 {
+  &.grid-gap-2 {
     @include grid-gap(2);
   }
 
-  .grid-gap-4 {
+  &.grid-gap-4 {
     @include grid-gap(4);
   }
 
   // Gaps: Mobile Large
   /* stylelint-disable */
   @include at-media("mobile-lg") {
-    .mobile-lg\:grid-gap-2 {
+    &.mobile-lg\:grid-gap-2 {
       @include grid-gap(2);
     }
   }
--- a/packages/usa-footer/src/styles/_usa-footer.scss
+++ b/packages/usa-footer/src/styles/_usa-footer.scss
@@ -25,33 +25,33 @@ $-chevron-expand-more: map-merge(
 // Grid styles ported over to remove dependency
 .grid-row {
   @include grid-row;
+
+  .grid-gap-1 {
+    @include grid-gap(1);
+  }
+
+  .grid-gap-2 {
+    @include grid-gap(2);
+  }
+
+  .grid-gap-4 {
+    @include grid-gap(4);
+  }
+
+  // Gaps: Mobile Large
+  /* stylelint-disable */
+  @include at-media("mobile-lg") {
+    .mobile-lg\:grid-gap-2 {
+      @include grid-gap(2);
+    }
+  }
+  /* stylelint-enable */
 }
 
 // Gaps
 .grid-gap {
   @include grid-gap;
 }
-
-.grid-gap-1 {
-  @include grid-gap(1);
-}
-
-.grid-gap-2 {
-  @include grid-gap(2);
-}
-
-.grid-gap-4 {
-  @include grid-gap(4);
-}
-
-// Gaps: Mobile Large
-/* stylelint-disable */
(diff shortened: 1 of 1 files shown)


packages/usa-footer/src/styles/_usa-footer.scss
```diff
@@ -22,103 +22,102 @@ $-chevron-expand-more: map-merge(
...
+  @include typeset($theme-footer-font-family);
+  overflow: hidden;
 
-      .desktop\:grid-col-3 {
-        @include grid-col(3);
-      }
-    }
-    /* stylelint-enable */
+  > .grid-container {
+    @include grid-container($theme-footer-max-width);
```
mahoneycm, line 120: Moving `.grid-container` class to match what it was in `3.7.0` before changes were made.

https://github.com/uswds/uswds/blob/e670ce36a7ce6179fd789cf2ae1d1feae955f83a/packages/usa-footer/src/styles/_usa-footer.scss#L26-L34

mejiaj: @mahoneycm to confirm, the current state has them outside and this change brings them back under footer class?

mahoneycm: The current state has _all_ of the grid classes nested under the footer class. This change moves them outside of the footer class to resolve specificity issues.

I maintained `.usa-footer > .grid-container` to match `3.7.0`

mejiaj (changes requested):
@mahoneycm a few questions:

1. Does this create duplicate styles in the final CSS file?
2. If so, is there a negative impact on performance?
3. Is there a method that can help us avoid specificity issues or duplicating styles _(for example: `:where()` — baseline browser support or a flag to check if grid sass has loaded)_ ?

mahoneycm:
@mejiaj 

## Question response

### 1. Does this create duplicate styles in the final CSS file?
   - Yes
### 2. If so, is there a negative impact on performance?
   - It doesn't appear so
      - `develop` - styles.css size: 640.05kb
      - `Fix` - styles.css size: 639.56kb
### 3. Is there a method that can help us avoid specificity issues or duplicating styles _(for example: `:where()` — baseline browser support or a flag to check if grid sass has loaded)_ ?
   - `:where()`
      - Using `:where()` does seem to resolve style conflicts but it looks like we’ll have to use it on every grid class definition in footer.scss to negate specificity. How do you feel about that?
   - `sass:meta`
       - I tried looking into a few `sass:meta` logic checks but since the classes are still being included in the uswds package the checks return true
       - I believe a check like this would have to happen on the project level before being compiled into CSS. 

## Issues

### 1. The specificity negatively impacts users.
   - Precisely.
   - Ex.) If a user set a new desktop grid utility class, the mobile definition from footer would win the specificity check.
### 2. Guidance is needed to direct users to load `usa-layout-grid`.
   - Yes.
   - If users plan to use `layout-grid` classes that are not standard to our variants, they will need to load `usa-layout-grid`2

mejiaj:
@mahoneycm thanks for looking into my questions.

We should have guidance about the layout grid dependency on the website (if we don't have it already).

For immediate solutions, using `:where()` as a hotfix seems clearer in our intention to add layout support without needing to duplicate, override, or create specificity issues.

If you'd like to discuss with the team further please add the `Discussion` label and we can discuss in Slack or in Dev Sync.

mahoneycm:
Closing in favor of #5749

mahoneycm:
Reopened based on the discussion in [Dev sync - Release check-in 🔒](https://docs.google.com/document/d/172TU5BVS9kGZQDuPcAhGoZjaTd2ngzcTayq4WRRTBUs/edit#heading=h.pi5dvz1llmck) (only GSA emails).

mejiaj (approved):
Thanks for re-opening this. Tested in DigitalGov and locally. No issues found.

packages/usa-footer/src/styles/_usa-footer.scss
```diff
@@ -22,103 +22,102 @@ $-chevron-expand-more: map-merge(
...
-    @include grid-container($theme-footer-max-width);
-  }
+// Gaps
+.grid-gap {
+  @include grid-gap;
+}
 
-  .grid-row {
-    @include grid-row;
+.grid-gap-1 {
```
amyleadem, line 35: ## Optional
This is a small note, but I am wondering if we should force the output to match the other grid gap CSS to be attached to `.grid-row`: `.grid-row.grid-gap-[X]`. Then if anyone has a compiler that removes redundant code, it will remove these lines.

amyleadem (commented):
Looking good! Just one small note about aligning `.grid-gap` output with the current CSS. Not sure if it is necessary but wanted to flag for consideration.

- [x] Confirm no visual regressions with default view
- [x] Confirm that various `desktop` grid classes work as expected
- [x] Confirm that various `mobile`-lg grid classes work as expected
- [x] Confirm that various `tablet` grid classes work as expected

packages/usa-footer/src/styles/_usa-footer.scss
mahoneycm: Ok so I nested the classes using:

```scss
.grid-row {
   &.grid-gap-[x]
```

This seemed to properly duplicate the `.grid-row.grid-gap-[x]` classes when comparing the output css!

No observed visual regressions or style conflicts as a result 👍

amyleadem (approved):
Looks good to me!

thisisdano (approved):

PR #5681, opened 2023-12-15
USWDS - Pages: Avoid reordering content in Documentation example

### LINKED ISSUE #4594: Documentation template tab order should match visual order

#### As a developer
 Who uses USWDS in my project, I expect that my generated stylesheets only includes styles relevant for my site, so that I'm not shipping unnecessary code to my users, prolonging their load times.

#### As a user

As a user who navigates a site using USWDS, I expect that load times are reasonable, so that I can access the content I need in a timely manner.

As a user, [I expect that I can tab through content in a logical order](https://www.w3.org/TR/WCAG20-TECHS/H4.html), so that there is not a disconnect between the visual and mechanical presentation of the content.

Remove the file `src/stylesheets/components/_layout.scss`, or more specifically the `.usa-layout-docs__sidenav` and `.usa-layout-docs__main` style selectors.

Adjust [`components/layouts/docs-inner.njk`](https://github.com/uswds/uswds/blob/develop/src/components/layouts/docs-inner.njk) such that either (a) the placement of the side navigation is consistently before or after the main content, regardless of device size, or (b) the markup is duplicated and shown/hidden at the appropriate viewport size, but in such a way that ensures the [DOM order is always consistent with the visual order](https://www.w3.org/TR/WCAG20-TECHS/C27.html).

**Describe alternatives you've considered**

If USWDS needs styling for its own documentation, this should be compiled separate from what's included in USWDS.

**Additional context**

Related issue on `order` concerns: #4424

mejiaj:
# Summary

**Documentation page structure follows logical order.** Sidenav on page template now follows logical HTML order on mobile. Previously, the sidenav was shifted _after_ the content, which did not follow the HTML order.

## Breaking change

This is a potentially breaking change. 
**Removed template classes**
- `usa-layout-docs__main`: This is the class that had the styles that reordered element.
- `usa-layout-docs`: This selector unused in current styles.

> [!CAUTION]
> Users who rely on or extended the classes listed above would be negatively affected. See https://github.com/uswds/uswds/pull/5681#discussion_r1442055975.

## Related issue

Closes #4594.

## Related pull requests

Changelog entry: uswds/uswds-site#2495.

## Preview link

Preview links; _test all screen sizes_:

| Current (`develop`) | Feature branch |
|:--------|:--------|
| [Default] | [Preview] |

Sidenav now comes before content on small screens. Previously, it was reordered after content.

[Default]: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/pages-documentation-page--documentation-page&p=all

[Preview]: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-feature-doc-layout-order/?path=/story/pages-documentation-page--documentation-page

## Problem statement

In mobile, tab order doesn't match visual order in the documentation page example.

## Solution

Removed styling that re-ordered the layout.

## Testing and review

- Sidenav should come before content in small screens (currently after)
- Sidenav should be on left on larger screens.

### DIFF
--- a/packages/templates/usa-documentation/includes/_usa-docs-inner.twig
+++ b/packages/templates/usa-documentation/includes/_usa-docs-inner.twig
@@ -49,7 +49,7 @@
 
       </div>
 
-      <main class="usa-layout-docs__main desktop:grid-col-9 usa-prose usa-layout-docs" id="main-content">
+      <main class="usa-layout-docs__main desktop:grid-col-9 usa-prose" id="main-content">
         <h1>Page heading (h1)</h1>
         <p class="usa-intro">The page heading communicates the main focus of the page. Make your page heading descriptive and keep it succinct.</p>
         <h2 id="section-heading-h2">Section heading (h2)</h2>
--- a/packages/templates/usa-documentation/includes/_usa-docs-inner.twig
+++ b/packages/templates/usa-documentation/includes/_usa-docs-inner.twig
@@ -49,7 +49,7 @@
 
       </div>
 
-      <main class="usa-layout-docs__main desktop:grid-col-9 usa-prose" id="main-content">
+      <main class="desktop:grid-col-9 usa-prose" id="main-content">
         <h1>Page heading (h1)</h1>
         <p class="usa-intro">The page heading communicates the main focus of the page. Make your page heading descriptive and keep it succinct.</p>
         <h2 id="section-heading-h2">Section heading (h2)</h2>
--- a/packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
+++ b/packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
@@ -2,16 +2,9 @@
 
 // Flexbox positioning to move sidenav below main content on small screens
 .usa-layout-docs__sidenav {
-  order: 2;
-  padding-top: units(4);
+  padding-bottom: units(4);
 
   @include at-media("desktop") {
-    padding-top: 0;
-  }
-}
-
-.usa-layout-docs__main {
-  @include at-media("desktop") {
-    order: 2;
+    padding-bottom: 0;
   }
 }
--- a/packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
+++ b/packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
@@ -1,6 +1,5 @@
 @use "uswds-core" as *;
 
-// Flexbox positioning to move sidenav below main content on small screens
 .usa-layout-docs__sidenav {
   padding-bottom: units(4);
 


mejiaj:
> Hey @mejiaj, one thing that's not clear from the PR description that will resolve some of my questions is _why_ we're making this change.
> 
> Is this to match default `sidenav` behavior when implemented on other pages?

I've updated the summary. Feel free to edit for clarity. 

This change strictly affects the Documentation page example. We're making this change because in small screens the visual display does not match the tab order. Currently the sidenav is placed before the content, but then reordered to after on mobile.

packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
```diff
@@ -2,16 +2,9 @@
 
 // Flexbox positioning to move sidenav below main content on small screens
```
mahoneycm, line 3: We should update / remove this comment now that we're no longer moving the sidenav below main content

mahoneycm (changes requested):
Looks good to me! One small edit and this should be good to go.

## Testing checklist
- [x]  Compare sidenav on all screen sizes
- [x]  Sidenav comes before content on small screens
- [x]  Sidenav remains on the left for larger screens
- [x]  Tab order matches visual display
- [ ]  Comment needs to be updated to represent the change

packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
mejiaj: Updated in 219e6b3dc2dee006fccf198396d2da8f35d98214.

In the future, an update or a PR suggestion would be appreciated.

mahoneycm (approved):

packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
```diff
@@ -1,17 +1,9 @@
...
-  order: 2;
-  padding-top: units(4);
+  padding-bottom: units(4);
 
   @include at-media("desktop") {
-    padding-top: 0;
-  }
-}
-
-.usa-layout-docs__main {
```
amyleadem, line 13: This rule is no longer in the CSS but it still exists in the markup in [usa-docs-inner.twig](https://github.com/uswds/uswds/blob/develop/packages/templates/usa-documentation/includes/_usa-docs-inner.twig#L52). The markup there also adds an undefined `usa-layout-docs` class as well. Could you remove those as well just to keep things tidy? :broom:

packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
```diff
@@ -1,17 +1,9 @@
 @use "uswds-core" as *;
 
-// Flexbox positioning to move sidenav below main content on small screens
 .usa-layout-docs__sidenav {
```
amyleadem, line 3: Thinking about one of the points in the original issue about not having demo code leak into the system's CSS. I wonder if we could achieve the same results by using padding utility classes in the `usa-docs-inner.twig` file, rather than adding style rules via this class. This would enable us to remove these style rules entirely. 

_What do you think about switching to utility classes so that we can remove `usa-layout-docs.scss` entirely?_

This is likely more a philosophical question than a practical one, since the impact on the CSS is so small. But it feels worth considering. Curious what you think!

amyleadem (changes requested):
@mejiaj I was able to confirm that the tab order now matches the markup order.

Just as an aside, I think the sidenav looks strange in either location and I am curious about if we should provide guidance on where the sidenav should live in the markup and in mobile views, but those are questions for another day. No action needed on this other than possibly opening a follow-on issue. 

I added a couple of comments below with some suggestions. Can you take a look and let me know if you have any questions?

packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
mejiaj: Removed rule. 

Also, `usa-layout-docs` in 91213f6 because its placement/name seems off based on BEM standards.

packages/usa-layout-docs/src/styles/_usa-layout-docs.scss
mejiaj: That seems like a good follow-up issue/enhancement.

I hesitate to make it in the scope of this PR because new guidance would be a requirement.

mejiaj:
@amyleadem @mahoneycm addressed comments, thank you. PR is ready for re-review.

mahoneycm (approved):

amyleadem (changes requested):
This is looking good. I added one small request to remove an undefined class. Other than that, I think it is good!

packages/templates/usa-documentation/includes/_usa-docs-inner.twig
```diff
@@ -49,7 +49,7 @@
 
       </div>
 
-      <main class="usa-layout-docs__main desktop:grid-col-9 usa-prose usa-layout-docs" id="main-content">
+      <main class="usa-layout-docs__main desktop:grid-col-9 usa-prose" id="main-content">
```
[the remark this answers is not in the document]
mejiaj: @amyleadem I've removed the class in 66decc0. I've also bumped this up to _potentially breaking change_ and noted in the PR description.

Mainly because:
1. This is a markup change (even if minor)
2. And because classes have been removed it _could_ impact users that have relied on these classes for custom styling.

amyleadem (approved):
Looks good to me! Thanks @mejiaj

thisisdano:
Hm. I'm not sure this is something we can recommend. The outcome here — a page with a big ol' sidebar preceding, and blocking, the content is not good. This kind of page would be really bad for the majority of folks accessing a content page on mobile, and would be confusing. (You'd think that you somehow missed the content.)

From the original issue, I think we need to consider option B:

>(b) the markup is duplicated and shown/hidden at the appropriate viewport size, but in such a way that ensures the [DOM order is always consistent with the visual order](https://www.w3.org/TR/WCAG20-TECHS/C27.html).

As is, this is a solution that essentially breaks conventional functionality.

thisisdano (changes requested):
See my comment above, but I think this solution has too much of an impact on conventional functionality and reading experience

PR #5699, opened 2024-01-05
USWDS - Utilities: Avoid outputting styles for add-aspect if not configured for output

### LINKED ISSUE #5934: USWDS - Bug: Aspect utilities are always output even when excluded from settings
### Describe the bug

When using the `$output-these-utilities` setting to control utilities which are included in the built stylesheet, the `add-aspect` utilities are always included, even when not listed in the setting.

Example stylesheet:

```scss
@forward "uswds-core" with (
  $output-these-utilities: ()
);

@forward "uswds-utilities";
```

Compiling the above will produce a stylesheet with 148 lines of `.add-aspect-` utility styles.

### Expected Behavior

Nothing is output, since the setting is configured to include no utilities, and only the utilities package is forwarded.

### Related code

See above.

### Screenshots

_No response_

### System setup

- USWDS version: 3.8.0

### Additional context

_No response_

### Code of Conduct

aduth:
# Summary

**Generated styles will no longer include aspect ratio utility classes when `$output-these-utilities` theme setting is configured to exclude them**.

## Breaking change

## Related issue

Closes #5934

## Problem statement

As a developer, I expect that if I configure `$output-these-utilities` theme setting, the compiled output will only include those utilities, so that I can optimize my stylesheet bundles to include only what's being used.

## Solution

Normally, utility classes are generated programmatically through the `$utilities-package` list, generated by the `render-utility` mixin, which considers the `$output-these-utilities` theme setting. Since `add-aspect` styles are rendered ad-hoc, it requires that we explicitly check `$output-these-utilities` before those styles are rendered.

This adds a new `if-output-utility` mixin to consistently handle consideration of `"default"` setting as an alternative to the list of utilities.

## Testing and review

Test file:

```scss
@use "uswds-core" with (
  $output-these-utilities: ()
);
@forward "uswds-utilities";
```

Before:

```scss
.add-aspect-9x16 {
// ...
```

After:

```
```

(empty)

For review, it's easier to review with whitespace changes hidden: https://github.com/uswds/uswds/pull/5699/files?w=1


zachmargolis (approved):

ryparker:
@coderabbitai full review

Issue #5658, opened 2023-12-05
USWDS - Checkbox: Cleanup indeterminate story + run prettier


mahoneycm:
## Description

Work contributed in #5549 has a build error due to sass linting. Storybook previews should also be adjusted to isolate indeterminate previews and to match existing patterns.

Issue #5645, opened 2023-11-29
USWDS - Checkbox: Indeterminate checkboxes need high contrast styles


mahoneycm:
## Description

The indeterminate styles implemented in #5549 break in high contrast modes.

## Related PR
#5549

mahoneycm:
Completed with #5648

PR #5713, opened 2024-01-09
USWDS - Checkbox:  Add indeterminate styles

### LINKED ISSUE #918: Add indeterminate checkbox styling

amyleadem:
# Summary

**Added styles for indeterminate checkboxes.** Checkboxes will now display as indeterminate when you set `input.indeterminate = true` via JavaScript or add the `data-indeterminate` attribute. 

> [!TIP]  
> This is only a style addition and does not affect checkbox functionality. Originally contributed by @lpsinger in #5549.

## Breaking change

## Related issue

Resolves #918 

## Related pull requests

- #5549 
- #5658 
- #5645 
- **Changelog:** uswds/uswds-site#2482

## Preview link

[Default checkbox (w/ indeterminate toggle) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/checkbox-indeterminate-feature-branch/?path=/story/components-form-inputs-checkbox--default)
[Indeterminate checkbox →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/checkbox-indeterminate-feature-branch/?path=/story/components-form-inputs-checkbox--indeterminate)
[Indeterminate checkbox tile →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/checkbox-indeterminate-feature-branch/?path=/story/components-form-inputs-checkbox--indeterminate-tile)

## Problem statement

USWDS Checkboxes do not support the `:indeterminate` CSS pseudo-class which can be set via JS.

More info about indeterminate on [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate)

## Solution

Add USWDS styles for indeterminate checkboxes.

Note from #5549 
> The browser platform provides [built-in support for indeterminate checkboxes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes), with the JavaScript input.indeterminate property mapped to the CSS :indeterminate selector.
>
> However, since there is no HTML attribute to control the input.indeterminate property, I've also implemented a data-indeterminate attribute.

## Major changes

- Add indeterminate checkbox styles.
- Migrate selected input styles to mixin.

**Mixins were required to**
1. Ensure radio buttons were not affected
2. Reduced duplicate styles
3. Properly use `$input-active-color` within `-checkbox-and-radio-colors` mixin

### Checkbox testing
1. Visit checkbox story
4. Turn on indeterminate story control or visit indeterminate story
5. Inspect styles
6. Inspect styles in high contrast mode
7. Confirm indeterminate styles are properly displaying
8. Repeat steps for tile variant

### Radio testing

Radio groups without a selected item receive the `:indeterminate` pseudo-class, which caused visual regressions in early testing.

1. Visit radio story.
3. Inspect element and remove `checked` attribute from selected item.
4. Confirm no regression in visual styles.
5. Repeat steps for tile variant.

### Testing checklist

- [ ] Indeterminate styles display appropriately
- [ ] High contrast styles display appropriately in light and dark modes
- [ ] No visual regression in radio buttons


packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -131,8 +142,28 @@
           }
         }
       }
+      &:indeterminate,
+      &[data-indeterminate] {
+        &,
+        &:checked {
```
amyleadem, line 148: @mahoneycm Curious about this line. Is the intention to have indeterminate styles take precedent when both indeterminate and checked attributes are present?

amyleadem (commented):
This is looking good. I just had a question about a line in the Sass about when `checked` and `indeterminate` meet. 

I've checked the styles in Firefox, Safari, and Chrome.

- [x] Confirm indeterminate stories are present and meet standards
- [x] In default viewing mode, confirm styles are consistent with other checkbox states and meet standards
- [x]  In forced colors mode, confirm styles are consistent with other checkbox states and meet standards
- [x]  In disabled state, confirm styles are consistent with other checkbox states and meet standards
- [x] When `$theme-body-background-color` is set to a dark color (I tested “ink”), confirm styles are consistent with other checkbox states and meet standards
- [x] Confirm that running  `npm run prettier:js` , `npm run prettier:sass` and `npm run lint` do not generate new fixes
- [x] Confirm that styling `data-indeterminate` and `indeterminate` attributes make sense

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
mahoneycm: Precisely! This was a part of the original contribution but it allows consistent display of the indeterminate styles for checked or unchecked checkboxes.

amyleadem: After reading more about the spec for indeterminate state, my understanding is that a checkbox should be either `checked` OR `indeterminate`, not both. Indeterminate implies that it is _not_ checked, so I vote that we remove this explicit state of `:indeterminate:checked` from the Sass. 

Note that indeterminate styles will still take precedence over checked styles due to the order of the CSS, but I don't expect this to cause issues because the checkbox should have only one of these attributes. 

> ...radio and checkbox elements can be toggled between checked and unchecked states, but are sometimes in an indeterminate state, _neither checked nor unchecked_[^1]

[^1]: https://drafts.csswg.org/selectors/#indeterminate

mahoneycm: This sounds good to me! I'll remove and check for regressions 👍

amyleadem:
@mahoneycm  After further testing, I discovered this leaks into the radio component. If none of the radio buttons are checked, it results in the radio buttons displaying as indeterminate: 

image

This is what the same state looks like in `develop`:
image

mahoneycm:
Great catch... it looks like radio buttons receive `:indeterminate` when none are selected.

> Elements targeted by this selector are:
> ...
> `<input type="radio">` elements, when all radio buttons with the same name value in the form are unchecked [^1]
[^1]: [MDN docs](https://arc.net/l/quote/gwasmnjk)

This should be an easy fix by moving the styles out of the `checkbox-and-radio` styles to just `checkbox` styles. Let me know if you see any other concerns with this solution or the scope of the review!

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -16,6 +16,21 @@
...
+    border: ButtonText solid units("05");
+  }
+}
+
+@mixin selected-input-colors($input-active-color) {
+  background-color: color($input-active-color);
+  box-shadow: 0 0 0 units($theme-input-select-border-width)
+    color($input-active-color);
+}
+
```
mahoneycm, line 33: ## Mixins over extends

I was required to use a mixin over an extend for two reasons

### Cannot extend compound selectors

I attempted to extend the `usa-checkbox__label:checked` class but Sass no longer supports extending compound selectors. This caused a build error with an appropriate error message.

### Passing a parameter

Since `$input-active-color` is defined by parameters passed within the `-checkbox-and-radio-colors` mixin, I was only able to reference it by adding an argument. Mixins do not support this feature.

> Mixins are obviously necessary when you need to configure the styles using [arguments](https://sass-lang.com/documentation/at-rules/mixin/#arguments) [^1]

[^1]: [Sass documentation](https://arc.net/l/quote/jmwwgmqn)

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -16,6 +16,21 @@
   @include -checkbox-and-radio-colors($input-type: "radio", $args...);
 }
 
+@mixin selected-tile-colors($input-active-color) {
```
mahoneycm, line 19: ## Mixin name

I used the name `selected` tile/input colors because both `checked` and `indeterminate` reflect a sort of selected state.

Let me know if there is a better name we should use here.

mahoneycm (commented):
## Update 2/19/24

Resolved checked checkbox style conflicts and radio visual regressions.

- Reordered `indeterminate` styles so that `checked` styles will still take precedence over `indeterminant`
- Adjusted styles to not target `indeterminate` radio buttons
- Moved styles into mixins to maintain DRY code.
- Updated PR description to match additions to code and testing instructions

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -90,15 +108,7 @@
...
         @include format-label {
-          background-color: color.adjust(
-            color($input-active-color),
-            $alpha: -0.9
-          );
-          border-color: color($input-active-color);
-
-          @media (forced-colors: active) {
-            border: ButtonText solid units("05");
-          }
```
amyleadem, line 101: > [!Note]
> Confirming that all of these lines were moved to `@mixin selected-tile-colors`

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -62,9 +82,7 @@
   .usa-#{$input-type}__input {
     &:checked {
       @include format-input {
-        background-color: color($input-active-color);
-        box-shadow: 0 0 0 units($theme-input-select-border-width)
-          color($input-active-color);
```
amyleadem, line 67: > [!Note]
> Confirming that these lines were moved to `@mixin selected-input-colors`

amyleadem (commented):
Thanks for the helpful comments, @mahoneycm . I found some style issues with the indeterminate checkbox styles in the disabled state. I was also hoping we could have an indeterminate control or story for the radio component for easier testing. I've added details in the test notes below:

- [x] Check display on Chrome, Firefox, and Safari
- [x] Confirm checkbox indeterminate stories are present and meet standards
- [x] In standard mode, confirm styles are consistent with other checkbox states and meet standards
- [ ]  In forced colors mode, confirm styles are consistent with other checkbox states and meet standards’
    - I see an different border on indeterminate styles in the disabled state:
    image
    - It looks consistent in the default state:
    image
- [ ]  In disabled state, confirm styles are consistent with other checkbox states and meet standards
    - I see an extra blue border around the checkbox and tile in the disabled state. I see this in all browsers.
    image
    - I also see an extra border around the indeterminate checkbox for disabled with dark background
    image
- [x] Confirm no regressions in radio button component
- [ ] Confirm radio indeterminate stories are present and meet standards
    - I didn't see any controls or stories for the indeterminate state for the radio button. Could you build that for easier testing?
- [x] When `$theme-body-background-color` is set to a dark color (I tested “ink”), confirm styles are consistent with other checkbox states and meet standards
- [x] Confirm running  `npm run prettier:js` , `npm run prettier:sass` and `npm run lint` do not generate new fixes
- [x] Confirm that styling `data-indeterminate` and `indeterminate` attributes make sense
- [ ] Confirm code meets standards
    - I'll confirm this after the style issues listed above are fixed

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -118,6 +131,41 @@
...
+        &[aria-disabled="true"] {
+          @include format-input {
+            box-shadow: $tile-box-shadow--disabled;
+          }
+
+          @include format-label {
+            border-color: $tile-border-color--disabled;
+          }
+        }
+      }
```
mahoneycm, line 160: ## Disabled conflict fix

Moving the checkbox:indeterminate styles caused the newer `format-input` styles to take priority. These styles are required to maintain disabled box shadow and tile colors.

Alternatively, we can move these styles to the disabled styles defined in lines 182-191. I opted to move them here since they weren't required to be redefined for the disabled checkboxes without `indeterminate`.

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -132,7 +180,11 @@
         }
       }
       &:checked:disabled,
-      &:checked[aria-disabled="true"] {
+      &:checked[aria-disabled="true"],
+      &:indeterminate:disabled,
+      &:indeterminate[aria-disabled="true"],
+      &[data-indeterminate]:disabled,
+      &[data-indeterminate][aria-disabled="true"] {
         @include format-input {
```
mahoneycm, line 188: ## Indeterminate checkbox styles alternate location

As mentioned above, we could add the styles from lines 150-160 to this block but the redefinition for `checked:disabled, &:checked[aria-disabled="true"]` isn't necessary. 

Let me know if we'd prefer to move them here!

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -34,12 +49,19 @@
   $input-border-alpha: -0;
   $tile-border-color: color("gray-20");
   $tile-border-color--disabled: color("gray-10");
+  $tile-box-shadow--disabled: 0 0 0 units($theme-input-select-border-width)
```
mahoneycm, line 52: ## New variable

Created this new variable so we can update the disabled tile box shadow in one location. Affects checked + indeterminate checkboxes, ignores indeterminate radio

packages/usa-radio/src/usa-radio.stories.js
```diff
@@ -9,6 +9,11 @@ export default {
       control: { type: "radio" },
       options: ["none", "disabled", "aria-disabled"],
     },
+    indeterminate: {
+      name: "Indeterminate option (toggle to reset)",
+      control: { type: "boolean" },
+      defaultValue: false
+    }
```
mahoneycm, line 16: ## New storybook control

Added new storybook control to radio so it's easy to test `indeterminate` or "unselected` radio styles.

Toggle the control to remove selection!

packages/usa-radio/src/usa-radio.stories.js
```diff
@@ -38,3 +43,17 @@ export const TileAriaDisabled = TileTemplate.bind({});
...
+    defaultValue: true,
+  },
+};
+
+export const IndeterminateTile = TileTemplate.bind({});
+IndeterminateTile.argTypes = {
+  indeterminate: {
+    defaultValue: true,
+  },
+};
```
mahoneycm, line 59: ## Indeterminate radio stories

Added indeterminate radio stories since it was easy to do so using new control. 

I chose to not deactivate the control so that it's easy to reset after a selection is made!

mahoneycm (commented):
## Update 3/1/24

Ok! I resolved some of the style issues and added indeterminate radio stories + controls to make it easy to test in the future!

Left some comments with additional information to help get aquianted with new updates.

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
mahoneycm: Resolved in [71c1cb9](https://github.com/uswds/uswds/pull/5713/commits/71c1cb982c8ea3941e3c6a7ba8ce1dc5a31195b9)

amyleadem (commented):
Looks good to me! I added a couple of comments for optional changes to add clarity, but neither needs to be a blocker. 

- [x] Check display of radio and checkbox on Chrome, Firefox, and Safari
- [x] Confirm checkbox indeterminate stories are present and meet standards
- [x] In standard mode, confirm styles are consistent with other checkbox states and meet standards
- [x]  In forced colors mode, confirm styles are consistent with other checkbox states and meet standards
- [x]  In disabled state, confirm styles are consistent with other checkbox states and meet standards
- [x] Confirm no regressions in radio button component
- [x] Confirm radio indeterminate stories are present and meet standards
- [x] When `$theme-body-background-color` is set to a dark color (I tested “ink”), confirm styles are consistent with other checkbox states and meet standards
- [x] Confirm running  `npm run prettier:js` , `npm run prettier:sass` and `npm run lint` do not generate new fixes
    - Note: Found a small change and pushed it up
- [x] Confirm that styling `data-indeterminate` and `indeterminate` attributes make sense
- [x] Confirm code meets standards

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
```diff
@@ -124,6 +137,41 @@
 
   @if $input-type == "checkbox" {
     .usa-checkbox__input {
+      // Unchecked radio buttons receive :indeterminate pseudo-class.
+      // Because of this, we need to explicetly target checkbox :indeterminate styles.
```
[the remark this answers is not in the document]
mahoneycm: I thought this note would add clarity as to why we aren't including the styles below with lines 127-132 (which is where they were previously defined).

I wanted to document the decision here so that if future devs know to not move them in case they miss the `radio:indeterminate` regressions as we did in early testing.

If we don't think it's necessary, we can remove!

packages/usa-radio/src/usa-radio.stories.js
```diff
@@ -9,6 +9,11 @@ export default {
       control: { type: "radio" },
       options: ["none", "disabled", "aria-disabled"],
     },
+    indeterminate: {
+      name: "Indeterminate option (toggle to reset)",
```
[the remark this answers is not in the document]
mahoneycm: The reason I added it is because you can still select radio buttons while the `indeterminate` is turned on. Toggling the switch will then properly reset the `indeterminate` state.

Here's a quick demo:

[Kapture 2024-03-04 at 12.15.04.webm](https://github.com/uswds/uswds/assets/25211650/bc22bef5-cbb6-417d-9c41-913b79d19c98)

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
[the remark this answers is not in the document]
mejiaj: It'd be helpful to speak to that in the comment.

```scss
// Adding indeterminate styles here avoids style conflicts from setting in lines 127-129.
```

---

>I wanted to document the decision here so that if future devs know to not move them in case they miss the radio:indeterminate regressions as we did in early testing.

Does that mean moving the styles to `127-132` will cause similar regressions with checkbox? If so, we can just change `radio → checkbox` because it looks like one affects the other.

mejiaj (changes requested):
No issues with functionality, feature looks good. 

Made small edits to the PR description and suggestions to improve code clarity.

---

@amyleadem @mahoneycm can you resolve/hide comments that no longer apply?


<details><summary>For example, was this resolved?</summary>
<p></p>

image
</details> 

<details><summary>Is this captured in the PR description? Should we include it?
</summary>
<p></p>

image

</details>

packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss
mejiaj: Makes sense to leave as-is if `checked:disabled, &:checked[aria-disabled="true"]` don't apply.

packages/usa-radio/src/usa-radio.stories.js
```diff
@@ -9,6 +9,11 @@ export default {
       control: { type: "radio" },
       options: ["none", "disabled", "aria-disabled"],
     },
+    indeterminate: {
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [3c372c3](https://github.com/uswds/uswds/pull/5713/commits/3c372c3d0a6640c69d12d367de8574748d2d89ce)

packages/usa-radio/src/usa-radio--tile.twig
```diff
@@ -7,7 +7,8 @@
         id="historical-truth-2"  
         type="radio" 
         name="historical-figures-2" 
-        value="sojourner-truth" checked 
+        value="sojourner-truth" 
+        {% if indeterminate == false %} checked {% endif %}
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [3c372c3](https://github.com/uswds/uswds/pull/5713/commits/3c372c3d0a6640c69d12d367de8574748d2d89ce)

mahoneycm:
## Update 3/5/24

@mejiaj I updated and resolved the comments I was able to! Here are my responses to your other comments:

**Was indeterminate radio resolved?**

Yes! Radio styles now maintain the expected unselected/indeterminate state styles

**Is indeterminate radio + fix captured in PR description**?

Yes, it was captured in the in the "Radio Testing" section but I've moved it to "Solution" for better visibility:

> Radio groups without a selected item receive the :indeterminate pseudo-class, which caused visual regressions in early testing. To avoid this, styles are added for checkbox and not radio buttons

**Language surrounding `indeterminate option` control**

Quoted from [this comment](https://github.com/uswds/uswds/pull/5713#discussion_r1511726871):

> Perhaps updating the language in parenthesis might be better?
>
> Given a label like:
> a. Toggle single indeterminate option
> b. View  indeterminate option
> c. Test indeterminate option
>
> With this language we know a single field is available to test indeterminate.
> 
> Using only Indeterminate option I'd expect to toggle it and all options be indeterminate. Currently only the second field is available.

The thing I want to highlight is that the `:indeterminate` state is set for _all radio options_ if none are selected.

From MDN:
> Elements targeted by this selector are:
> ...
> • `<input type="radio">` elements, when all radio buttons with the same `name` value in the form are unchecked [^1]

[^1]: [MDN Docs](https://arc.net/l/quote/ikwjiqjk)

Because of this, I lean towards something like `Toggle indeterminate state` because it will remove the radio selection and enable indeterminate for all radio options.

Updated in [2636dbe](https://github.com/uswds/uswds/pull/5713/commits/2636dbeffbdf86b33f64eaa23540cb37b16888f2)

Let me know if you prefer something else.

**Mixin SASSDoc comments**

Added SASSDoc comments for the new mixins. I also renamed them to use `active` instead of `selected` to keep it consistent with the param/sass variable that is used to set the colors.

Resolved in [a399129](https://github.com/uswds/uswds/pull/5713/commits/a399129b56140f6f083fba211c7c31e60d59b3e7)

**Checkbox styles SASSDoc comments**

I updated the SASSDoc comment to focus on the class rather than the line numbers.

> // Adding indeterminate styles here avoids style conflicts from setting in .usa-#{$input-type}__input due to 
> // radio buttons receiving :indeterminate state when none are selected.


I opted for this because my previous changes caused the desired line range to move and I realized any references to the line numbers could easily be made incorrect. 

Resolved in [c31e6f9](https://github.com/uswds/uswds/pull/5713/commits/c31e6f99980b26054d564e2f4318fb32352dde56)

**Question about radio regression**

> Does that mean moving the styles to 127-132 will cause similar regressions with checkbox? If so, we can just change radio → checkbox because it looks like one affects the other. [^2]

[^2]: [Comment](https://github.com/uswds/uswds/pull/5713#discussion_r1511853596)

No, the regression is specifically found in radio buttons. It was caused due to unchecked radio groups receiving the indeterminate style by default. Therefore, we only want checkboxes to receive these styles to maintain the standard “unchecked” radio styles.

thisisdano (approved):
Nice work.

lpsinger:
Wonderful! I'm excited about this landing in main!

PR #5761, opened 2024-02-09
Prototype: Initial banner web component 

### LINKED ISSUE #5639: Web Components: Prototype native USA banner component
Prototype native banner component. 

## Done when

- [x] Understand architecture changes that need to be made
- [x] We know what code we can reuse
- [x] Required build dependencies (if any)

mejiaj:
> [!IMPORTANT]  
> This is native banner created for internal testing and insights.
 
# Summary

Initial native banner test.

## Breaking change

## Related issue

Closes #5639. 

## Related pull requests

n/a

## Preview link

[Preview link →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-native-wc-banner/?path=/story/components-banner--banner-wc-test)

## Problem statement

Native web components will allow us to:
1. Work closer to the browser.
2. Reduce the number of dependencies required.
3. Allow users to more easily stay up-to-date with the design system.

## Solution

Solution uses Light DOM. To allow existing USWDS styling and JS methods. This method was chosen to allow maximum re-use of existing code.

**Limitations of this approach**
1. Major changes to build requirements required to improve DX. For example, browser support for importing JSON is not widely supported.
    
    
    [MDN Source →](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#browser_compatibility)
1. Using light DOM allows us to re-use SASS & JS without major changes. But this puts components in a weird in-between state where we're not able to fully leverage native API's (especially in JS) to improve these components.
1. Additional development required to create methods for bundling and releasing.

An alternative solution is using a lighter framework based on web standards, like [Open Web Components](https://open-wc.org/), Lit, or StencilJS.

## Major changes

HTML moved to component JS. Concerned with how much markup there is for a component with no visual variants. This'll greatly impact maintainability when we start looking at components that have variants that require different markup.

## Testing and review

There shouldn't be any regressions to Banner functionality.

### DIFF
--- a/packages/usa-banner/src/Banner.js
+++ b/packages/usa-banner/src/Banner.js
@@ -1,0 +1,153 @@
+import * as content from "./content/usa-banner.json";
+
+const templateString = (data) => {
+  // Get the data object and destructure.
+  const { banner, domain, https } = data;
+
+  // Create the template with passed data.
+  /* html */ return `
+  <section
+    class="usa-banner"
+    aria-label="Official website of the United States government"
+  >
+    <div class="usa-accordion">
+      <header class="usa-banner__header">
+        <div class="usa-banner__inner">
+          <img
+            aria-hidden="true"
+            class="usa-banner__header-flag"
+            src="https://cdnjs.cloudflare.com/ajax/libs/uswds/3.3.0/img/us_flag_small.png"
+            alt=""
+          />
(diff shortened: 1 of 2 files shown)


mejiaj:
Closing as this work was experimental.

PR #5774, opened 2024-02-15
USWDS - Date picker: Prevent focus on mouseover

mahoneycm:
# Summary

**Fixed a bug that caused `mouseover` events to focus buttons and prevent keyboard navigation.** Now, hovering the mouse over date picker buttons will only cause the `hover` state.

## Breaking change

This is not a breaking change

## Related issue

Closes #5752

## Related pull requests

[Changelog →](https://github.com/uswds/uswds-site/pull/2496)

An alternative approach to #5778. This approach introduces a separate accessibility issue where screen-readers read the focused date every time a date button is hovered. This includes when a new date is selected via the keyboard while the mouse is hovering over a date. This would introduce confusion for screen reader users.

## Preview link

[Date Picker →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-date-range-remove-mouseover/?path=/story/components-form-inputs-date-picker--default)

[Date Range Picker →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-date-range-remove-mouseover/?path=/story/components-form-inputs-date-range-picker--default)

## Problem statement

Hovering the mouse over date picker calendar buttons cause them to gain focus. This is unexpected behavior which also prevents keyboard navigation through calendar buttons.

## Solution

Rework the `handleMouseoverFromDate()` function to target the date buttons of the current month and add the `--within-range` class. The function no longer calls the `renderCalendar()` function or sets focus on the hovered dates.

1. Follows expected `mouseover` behaviors.
2. Does not affect date picker or date range usage.

Additionally, I removed the `mouseover` event listeners and functions from month and year buttons. They did not appear to add any functionality beyond focusing the button on hover, which we want to prevent.

`handleMousoverFromDate` is maintained in order to properly render date range previews on hover.

### Accessibility improvements

This solution improves accessibility interaction in two ways.

**VoiceOver**

In its current iteration, when using the date picker with a mouse and VoiceOver, it would rapidly call out dates as your mouse moved over them. This, in my opinion, is unexpected behavior. VoiceOver should not read out other buttons or table content upon `mouseover` events.

With this update, VoiceOver will only call out the buttons when selected via the keyboard.

**High zoom magnification behaviors**

[...]

PR #5778, opened 2024-02-20
USWDS - Date picker: Prevent focus on `mouseover`

### LINKED ISSUE #5752: USWDS - Bug: Date picker - mouse hover preventing keyboard interaction on high zoom magnification
### Describe the bug

When testing the date picker for zoom magnification functionality, I found that at 400% and using a keyboard to select a date I was unable to use the arrow keys to move off a date. I was "stuck" on a date because my mouse was called over to that date automatically.

In essence, the mouse was jumping to focus on a date and wouldn't allow the keyboard to take control of functionality.

I was using Chrome on a PC.

1. Go to https://designsystem.digital.gov/components/date-picker/
2. Zoom into 400% magnification
3. Try tabbing into the date picker calendar icon; tab into a date to select
4. Then try using keyboard arrow keys to move around inside the calendar. Inconsistently it works; but sometimes if the mouse is anywhere inside the browser it'll "call" the cursor over to the date and create a keyboard trap.

### Expected Behavior

Tabbing into the calendar icon and navigating using a keyboard only should never trap the keyboard focus on a date even at a high magnification like 400%

> [!CAUTION]
> **This is a AA level failure**
> The intent of this Success Criterion is to support people with low vision who need to enlarge text and read it in a single column. When the browser zoom is used to scale content to 400%, it reflows - i.e., it is presented in one column so that scrolling in more than one direction is not necessary. [^1]

[^1]: [WCAG Reflow (Level AA) requirement](https://arc.net/l/quote/qblbjcul)

### Screenshots

_No response_

### System setup

PC, Chrome

### Additional context

This type of keyboard trap inside the calendar is an issue because some low vision users use a high level of magnification and a keyboard to navigate inside components.

### Code of Conduct

mahoneycm:
# Summary

**Fixed a bug that caused `mouseover` events to focus buttons and prevent keyboard navigation.** Now, hovering the mouse over date picker buttons will only cause the `hover` state.

## Breaking change

This is not a breaking change

## Related issue

Closes #5752

## Related pull requests

[Changelog →](https://github.com/uswds/uswds-site/pull/2496)

## Preview link

[Date Picker →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-date-range-update-mouseover/?path=/story/components-form-inputs-date-picker--default)

[Date Range Picker →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-date-range-update-mouseover/?path=/story/components-form-inputs-date-range-picker--default)

## Problem statement

Hovering the mouse over date picker calendar buttons cause them to gain focus. This is unexpected behavior which also prevents keyboard navigation through calendar buttons.

## Solution

Enhance logic to check for a focused calendar button. The calendar will apply the correct focus on render while permitting mouseover behaviors to exist cohesively.

This solution:

1. Follows expected `mouseover` behaviors.
2. Does not affect date picker usage.

Additionally, I removed the `mouseover` event listeners and functions from month and year buttons. They did not appear to add any functionality beyond focusing the button on hover, which we want to prevent.

`handleMousoverFromDate` is maintained in order to properly render date range previews on hover.

### Accessibility improvements

This solution improves accessibility interaction in two ways.

**VoiceOver**

In it’s current iteration, when using date picker with a mouse and VoiceOver, it would rapidly call out dates as your mouse moved over them. This, in my opinion, is unexpected behavior. VoiceOver should not read out other buttons or table content upon `mouseover` events.

With this update, VoiceOver will only call out the buttons when selected via the keyboard.

**High zoom magnification behaviors**

As depicted in #5752, using the date picker at a high magnification (400%, AA requirement) the mouse would sometimes jump to the selected button, causing a mouseover event, and locking the focus on that button. Keyboard interaction would be prevented by the hovered mouse. For users who rely on keyboard navigation, this would cause confusion as they would not move the mouse to identify the issue.

With this update, users will be able to use expected keyboard interactions without error.

### Alternatives considered

**blur**

I tested calling `blur()` on the selected element during keyboard interactions, but this caused accessibility issues at high zoom (400%). The window would jump unexpectedly and cause the mouse to move, hovering a new button, which then stole focus once again.

**Focus styles**

To prevent visual regression, I also considered adding focus styles to hover. In my opinion, this causes confusion because the actual focused element and the hovered button look the same and it’s hard to distinguish which is actually focused.

**Removing `mouseover` events altogether**

Originally, I had removed mouseover events all together. This caused a visual regression when selecting a date range for the first time where the range preview would not show. You can view this in action via this [draft PR](https://github.com/uswds/uswds/pull/5774) and testing Date Range Picker

### Additional concerns

Current date picker behaviors cause a new calendar to be rendered after every mouseover and keyboard event. It does so by cloning the existing calendar, setting appropriate classes and returning an HTML template literal. This pattern does not match modern USWDS practices and likely affects performance.

## Major changes

- Date picker no longer updates focus on `mouseover` events.
- Behavior changes help to meet WCAG AA requirement

> The intent of this Success Criterion is to support people with low vision who need to enlarge text and read it in a single column. When the browser zoom is used to scale content to 400%, it reflows - i.e., it is presented in one column so that scrolling in more than one direction is not necessary. [^1]
> [^1]: [WCAG Reflow (Level AA) requirement](https://arc.net/l/quote/qblbjcul)

## Testing and review

1. Visit date picker.
2. Test mouse interactions and confirm they work as expected.
3. Test keyboard interactions and confirm they work as expected.
4. Test VoiceOver and ensure callouts are expected.
5. Zoom to 400% magnification and use the keyboard to interact with the date picker. There should be no errors navigating throughout the calendar.
6. Repeat steps for all variants.
7. Repeat steps date range picker and it’s variants.

> [!TIP]
> When using keyboard navigation with a screenreader, use the screenreader’s **Table Navigation** controls (varies by screenreader). This can help prevent unexpected interactions. Use default arrow keys when jumping months via dates.

</aside>

### Testing checklist

- [ ]  Expected mouse behaviors work without error
- [ ]  Expected keyboard behaviors work without error
- [ ]  Screenreader behaviors work as expected without error
- [ ]  Component is usable at high magnification without error


amyleadem (commented):
Hi @mahoneycm, I tested keyboard navigation on the calendar at 400% zoom and I noticed an error where the calendar would stop reacting to keyboard input if my mouse cursor was hovering over any of the calendar dates. The mouseover was not causing it to focus, but it still interrupted keyboard interaction. Can you take a look at that?

packages/usa-date-picker/src/index.js
```diff
@@ -1695,17 +1695,25 @@ const handleEscapeFromCalendar = (event) => {
...
+  const { calendarEl, minDate, maxDate } = getDatePickerContext(event.target);
 
+  const focusedEl = document.activeElement;
+  const calendarDate = parseDateString(focusedEl.dataset.value);
   const date = adjustDateFn(calendarDate);
 
+  if (!calendarEl.contains(focusedEl)) {
+    return;
+  }
+
```
mahoneycm, line 1716: Early return to prevent console error when `focusedEl` is undefined

packages/usa-date-picker/src/index.js
```diff
@@ -2230,12 +2216,6 @@ if (!isIosDevice()) {
     [CALENDAR_DATE_CURRENT_MONTH]() {
       handleMouseoverFromDate(this);
     },
-    [CALENDAR_MONTH]() {
-      handleMouseoverFromMonth(this);
-    },
-    [CALENDAR_YEAR]() {
-      handleMouseoverFromYear(this);
-    },
```
mahoneycm, line 2238: Removed unnecessary `handleMousover` functions to prevent unexpected focus on mouseover behaviors

packages/usa-date-picker/src/index.js
```diff
@@ -1801,8 +1809,18 @@ const handleMouseoverFromDate = (dateEl) => {
...
+  // If the calendar contains a focused element, reapply focus after new calendar renders.
+  // Targeted using aria-label because they are unique and prevents error when
+  // focus is in the table header.
+  if (calendarEl.contains(focusedEl)) {
+    const focusElLabel = focusedEl.getAttribute("aria-label");
+    const newCalendar = renderCalendar(calendarEl, dateToDisplay);
+    newCalendar.querySelector(`[aria-label="${focusElLabel}"]`).focus();
+  } else {
+    renderCalendar(calendarEl, dateToDisplay);
+  }
```
mahoneycm, line 1823: From JS comments:

>   If the calendar contains a focused element, reapply focus after new calendar renders.
>  Targeted using aria-label because they are unique and prevents error when focus is in the table header.

packages/usa-date-picker/src/index.js
```diff
@@ -1801,8 +1809,18 @@ const handleMouseoverFromDate = (dateEl) => {
...
   const dateToDisplay = parseDateString(hoverDate);
-  const newCalendar = renderCalendar(calendarEl, dateToDisplay);
-  newCalendar.querySelector(CALENDAR_DATE_FOCUSED).focus();
+  const focusedEl = document.activeElement;
+
+  // If the calendar contains a focused element, reapply focus after new calendar renders.
+  // Targeted using aria-label because they are unique and prevents error when
+  // focus is in the table header.
+  if (calendarEl.contains(focusedEl)) {
+    const focusElLabel = focusedEl.getAttribute("aria-label");
```
mahoneycm, line 1818: ## activeElement vs. `--focused` class

When troubleshooting this issue, I decided I needed a way to identify what was actually `:focused` and what was hovered. It made sense to use `document.activeElement` to select the focused button natively while maintaining the focused class that is used to identify date range caps. 

This seemed like the path of least resistance though we should consider renaming the `--focused` class to `--hovered` or something similar for the sake of brevity

packages/usa-date-picker/src/index.js
```diff
@@ -1801,8 +1809,18 @@ const handleMouseoverFromDate = (dateEl) => {
...
+    const newCalendar = renderCalendar(calendarEl, dateToDisplay);
+    newCalendar.querySelector(`[aria-label="${focusElLabel}"]`).focus();
```
mahoneycm, line 1820: ## aria-label vs. dataset.value

In order to refocus the correct button, I used aria-label because they were unique and included on all buttons on the calendar. 

This also prevented an error when tabbing to navigation, month, and year buttons which did not include a `data-value` attribute

PR #5793, opened 2024-02-29
USWDS - Core: Replace receptor behavior with custom optimized implementation

aduth:
# Summary

**Reduced the JavaScript bundle size.** By optimizing the implementation of core component code, overall JavaScript sizes are reduced.

## Breaking change

This _could_ be considered a breaking change in the unlikely circumstance that someone is using USWDS' internal `behavior` module and expecting to leverage behaviors of the underlying [`receptor` library](https://github.com/shawnbot/receptor) that are not implemented here, namely:

1. ["A string in the form `types:delegate(selector)`"](https://github.com/shawnbot/receptor/tree/master?tab=readme-ov-file#-receptorbehaviorlisteners--properties:~:text=A%20string%20in%20the%20form%20types%3Adelegate(selector))
2. Support for [listener options](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#options), which is only [documented in passing](https://github.com/shawnbot/receptor/tree/master?tab=readme-ov-file#once:~:text=If%20you%20provide%20listener%20options%2C%20those%20will%20also%20be%20passed%20to%20removeEventListener().)

Overall, I'd expect it to be quite unlikely anyone would be using these features, since it would require familiarity with `receptor` such that I'd expect anyone interested in using them would use `receptor` directly rather than through USWDS's `behavior` abstraction.

Backwards-compatibility has been artificially preserved with the return value of `behavior` including additional function aliases `add` (alias for `on`) and `remove` (alias for `off`).

## Problem statement

As a developer interested in optimizing their installation of USWDS, I expect that USWDS will keep overhead of common code to an absolute minimum, so that I can maximize optimization in cases where I'm interested in using one or a few components.

* Example: At Login.gov, we include very few components in our critical-path JavaScript bundle ([source](https://github.com/18F/identity-idp/blob/fab35dc2accf4f34fbe194baf553abea8e3384d4/app/javascript/packs/application.ts#L1-L4))

As a user interacting with a site using USWDS, I expect that pages will load quickly, so that I can access information without delay.

As a USWDS maintainer, I expect that the project minimizes its use of third-party dependencies in cases where such code is already heavily customized, so that I can avoid the maintenance burden associated with keeping those dependencies (including transitive dependencies) up-to-date and free from security vulnerabilities.

## Solution

The changes proposed here reimplement `behavior` and `keymap`, preserving existing functionality while removing underlying use of [the third-party `receptor` library](https://www.npmjs.com/package/receptor).

**Why?**

- Reduces size of behavior by 66%, keymap by 65%, and minimal component import (banner) by 50%
- Narrows scope to features in use
- Avoid baked-in polyfills not necessary for current browser supports ([`Object.assign`](https://github.com/shawnbot/receptor/blob/de712fd134c5fb1f9c48671579245d58d3296bcd/src/behavior.js#L1), [`Element#closest`](https://github.com/shawnbot/receptor/blob/de712fd134c5fb1f9c48671579245d58d3296bcd/src/delegate.js#L2))
- Remove third-party dependencies to reduce maintenance overhead and security vulnerability attack surface area
   - Avoid using dependencies which aren't actively maintained (the last release of `receptor` was 6 years ago)
- Add TypeScript annotations for IDEs supporting code autocompletion (e.g. VS Code)

### Performance Impact

The following examples use [ESBuild](https://github.com/evanw/esbuild) and [brotli-size-cli](https://github.com/andrewiggins/brotli-size-cli) to simulate a real-world scenario of downloading a minified, bundled, and compressed JavaScript.

#### Scenario 1: Importing only `behavior`

```
echo "import './packages/uswds-core/src/js/utils/behavior'" | npx esbuild --bundle --minify | brotli-size
```

Before: 1.43 kB
After: 473 B
Diff: -957 B (-66.9%)

#### Scenario 2: Importing only `keymap`

```
# Before:
echo "import 'receptor/keymap" | npx esbuild --bundle --minify | brotli-size

# After:
echo "import './packages/uswds-core/src/js/utils/keymap'" | npx esbuild --bundle --minify | brotli-size
```

Before: 1.18 kB
After: 413 B
Diff: -767 B (-65%)

#### Scenario 3: Importing a single component ("Best Case")

```
echo "import './packages/usa-banner/src/index.js'" | npx esbuild --bundle --minify | brotli-size
```

Before: 1.84 kB
After: 896 B
Diff: -944 B (-51.3%)

#### Scenario 4: Importing the full JavaScript package ("Worst Case")

```
echo "import './packages/uswds-core/src/js/index.js'" | npx esbuild --bundle --minify | brotli-size
```

Before: 22.2 kB
After: 20.9 kB
Diff: -1.3 kB (-5.9%)

~**Note:** I believe the reason for lack of any difference here is that this pull request doesn't completely remove `receptor` yet, and there are imports referencing the package that may end up pulling in the whole library still ([example](https://github.com/uswds/uswds/blob/b741317537b0053a0c553188de349e35eb16298e/packages/uswds-core/src/js/utils/focus-trap.js#L2)). A follow-up pull request should seek to replace `receptor`'s `keymap` functionality and remove the dependency altogether, which should prove to have an impact on the bundle size for the full package.~

**Edit:** I addressed this in 9924be3 by updating the import semantics, which prevents the full library from being pulled in, a temporary fix until the library can be removed altogether.

**Edit 2:** In additional commits starting from d55ee9c1a, I included the full replacement/removal of the `receptor` library, which further reduces the overall bundle size.

## Testing and review

Verify all unit tests pass:

```
npm exec gulp unitTests
```

If full unit test coverage cannot be guaranteed, it may be wise to review experience for individual components within the Storybook preview at http://localhost:6006

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
       "dependencies": {
         "classlist-polyfill": "1.2.0",
         "object-assign": "4.1.1",
-        "receptor": "1.0.0",
         "resolve-id-refs": "0.1.0"
       },
       "devDependencies": {
@@ -15614,14 +15613,6 @@
       "integrity": "sha512-bT0jEz/Xz1fahQpbZ1D7LgmPYZ3iHVY39NcWWro1+hA2IvjiPeaXtfSqrQ+nXjApMvQRE2ASt1itSLRrebHMRQ==",
       "dev": true
     },
-    "node_modules/element-closest": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/element-closest/-/element-closest-2.0.2.tgz",
-      "integrity": "sha512-QCqAWP3kwj8Gz9UXncVXQGdrhnWxD8SQBSeZp5pOsyCcQ6RpL738L1/tfuwBiMi6F1fYkxqPnBrFBR4L+f49Cg==",
-      "engines": {
-        "node": ">=4.0.0"
-      }
-    },
     "node_modules/element-resize-detector": {
       "version": "1.2.4",
       "resolved": "https://registry.npmjs.org/element-resize-detector/-/element-resize-detector-1.2.4.tgz",
@@ -21466,11 +21457,6 @@
       "integrity": "sha512-cYofQu2Xpom82S6qD778jBDpwvvy39s1l/hrYij2u9AMdQcGRpaBu6kY4mVhuno5kJVi1DAz4aiphA2WI1/OAw==",
       "dev": true
     },
-    "node_modules/keyboardevent-key-polyfill": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/keyboardevent-key-polyfill/-/keyboardevent-key-polyfill-1.1.0.tgz",
-      "integrity": "sha512-NTDqo7XhzL1fqmUzYroiyK2qGua7sOMzLav35BfNA/mPUSCtw8pZghHFMTYR9JdnJ23IQz695FcaM6EE6bpbFQ=="
-    },
     "node_modules/keyv": {
       "version": "4.5.4",
       "resolved": "https://registry.npmjs.org/keyv/-/keyv-4.5.4.tgz",
@@ -22251,11 +22237,6 @@
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
       }
-    },
-    "node_modules/matches-selector": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/matches-selector/-/matches-selector-1.2.0.tgz",
-      "integrity": "sha512-c4vLwYWyl+Ji+U43eU/G5FwxWd4ZH0ePUsFs5y0uwD9HUEFBXUQ1zUUan+78IpRD+y4pUfG0nAzNM292K7ItvA=="
     },
     "node_modules/mathml-tag-names": {
       "version": "2.1.3",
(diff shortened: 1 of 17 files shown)


aduth:
The build failure seems to be unrelated to the changes here, starting in #5233 and affecting [all commits on `develop` since](https://github.com/uswds/uswds/commits/develop/) and [other pull requests](https://github.com/uswds/uswds/pull/5794).

brunerae:
This potentially resolves #6105

mejiaj (commented):
Apologies that we’ve taken so long to get this scheduled.

We do appreciate this contribution — managing big PR’s is a pain point we haven't been able to address! We’ve taken this long to consider it because it's a complex pull request that includes a lot of changes to core functionality. We approach these types of PRs with caution so we don’t disrupt downstream teams. Your additions to the unit tests have helped a lot. So, in short, thanks for submitting this.

To help us speed up reviews like this in the future, it would help if you could use:

- **Smaller PR's.** Breaking out the work helps us reliably and quickly test one thing at a time, which we may need to do over multiple sprints, in order to work it into the rest of our in-flight work.
- **JSDoc comments.** This would be more in line with our code conventions, and also improves readability.

aduth:
> * **Smaller PR's.** Breaking out the work helps us reliably and quickly test one thing at a time, which we may need to do over multiple sprints, in order to work it into the rest of our in-flight work.

Happy to do this. However, to be honest, I find that the turnaround time and process for reviews and merges makes it rather impractical to propose iterative changes. 

There's probably a few candidates here we could split out to make the overall changeset a little smaller, with minimal overlap and maybe a bit of follow-on cleanup to remove the dependency once everything else is replaced:

- Replace use of `receptor/keymap`
- Replace use of `receptor/once`
- Replace use of `receptor/ignore`
- Fix `tasks/test.js` to include all test files

I could go ahead and split out these items, if that'd make things easier.

aduth:
Closing in favor of smaller, targeted pull requests. I've also incorporated the review feedback into those pull requests.

It also occurred to me that one of the reasons for making this as a single monolithic change is that it's hard to see the true value (performance impact) of this effort until the `receptor` dependency is removed in its entirety.

- https://github.com/uswds/uswds/pull/6171
- https://github.com/uswds/uswds/pull/6172
- https://github.com/uswds/uswds/pull/6173
- https://github.com/uswds/uswds/pull/6174
- https://github.com/uswds/uswds/pull/6175

mejiaj:
Created #6291 to remove the dependency once everything is merged in.

PR #5811, opened 2024-03-11
Add notifications related to USWDS 3.8.0

thisisdano:
Added the following notes to capture the changes to the sidenav in our docs templates.

```
3.8.0
Release notes: github.com/uswds/uswds/releases/tag/v3.8.0

- Accessibility: [usa-layout-docs__sidenav] BREAKING: By default, we
  no longer support using CSS to reorder the sidenav position at mobile
  widths. We suggest that teams update their sidenav markup to better
  support tab order at mobile widths. See the release notes for more
  details on how to do this.

  To temporarily support the old CSS order functionality,
  set `$theme-sidenav-reorder: true` in your project settings.
- Settings: We added `$theme-sidenav-reorder`. When `true` this setting
  uses CSS to reorder the sidenav to follow the page text at mobile widths.
  This setting can introduce usability issues, so we suggest that teams
  update their sidenav markup instead. See the release notes for more detail.
```


mejiaj (approved):
LGTM

PR #5818, opened 2024-03-15
USWDS - JS Loading: Add JS loading hidden content component

### LINKED ISSUE #5817: USWDS - Feature: Improve accordion flash of content while JavaScript loads

Currently, accordion content will remain visible while the page is loading until the main JavaScript bundle has been loaded and evaluated, creating a visual flash of content very similar to the banner issue addressed in #3092.

Ideally, the accordion would automatically respect the `usa-js-loading` class added in the USWDS initializer script, and hide content while the main JavaScript is being loaded.

In practice, this may be difficult to implement, since the accordion content visibility is controlled by the associated button (a button which `aria-controls` the content by its `aria-expanded` state), which is not possible to implement via CSS selectors (AFAIK).

Possible solutions:

- (Likely not possible) Stylize the accordion associated with the button
- Lean into "usa-js-loading" as a proper component, with a CSS [BEM](https://css-tricks.com/bem-101/) "element" like `usa-js-loading__hidden-content` which could be added to elements expected to be hidden during the time that the main JavaScript bundle is being loaded
- Invert current behavior and hide all accordion content by default while JavaScript loads
   - This will still likely cause a flash of content for the accordion which is expected to be expanded by default
- Assume that the content follows immediately after the expander button and use new [CSS `:has` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) to stylize as hidden depending on button, during load
   - e.g. `.usa-js-loading .usa-accordion__heading:has([aria-expanded="false"]) + .usa-accordion__content { ... }`

_No response_

### Additional context

_No response_

### Code of Conduct

aduth:
# Summary

**Adds support for hiding content while JavaScript is loading.** This can be used on accordion content to avoid flashing content while JavaScript loads.

## Related issue

Closes #5817

Related to #3092

## Related pull requests

This is an alternative or complementary solution to #5826.

Even if #5826 is merged, this could still provide benefit if there was a use-case to want to hide specific content during the time that JavaScript is loading. That being said, I don't have any specific use-cases in mind, so would be fine to close this if #5826 is the preferred solution.

## Solution

This is the second solution proposed in #5817:

>An alternative solution could be to lean into "usa-js-loading" as a proper component, with a CSS [BEM](https://css-tricks.com/bem-101/) "element" like `usa-js-loading__hidden-content` which could be added to elements expected to be hidden during the time that the main JavaScript bundle is being loaded.

If this were to be accepted...

- ...it could make sense to move `uswds-init.js` source code from `uswds-core` into this component.
- ...it would require some corresponding changes to documentation site for component example markup to use the new class

## Testing and review

1. Go to http://localhost:6006/?path=/story/components-accordion--default
2. Observe hidden content unaffected by changes
3. In local copy of running code, comment out these lines: https://github.com/uswds/uswds/blob/98566ec5dc12807a2b0bc36699a033fe3a711d44/packages/usa-accordion/src/index.js#L86-L89
4. Refresh the page
5. Observe all content is visible
6. Using browser DevTools inspector, add `usa-js-loading` CSS class to some ancestor element of one of the accordions which would be collapsed without the changes implemented in Step 3
7. Observe that the accordion content becomes hidden

### DIFF
--- a/packages/usa-accordion/src/usa-accordion.twig
+++ b/packages/usa-accordion/src/usa-accordion.twig
@@ -6,7 +6,7 @@
         {{ item.title }}
       </button>
     </h4>
-    <div id="{{ id_prefix }}{{ item.id }}" class="usa-accordion__content usa-prose">
+    <div id="{{ id_prefix }}{{ item.id }}" class="usa-accordion__content usa-prose {% if item.expanded != true %}usa-js-loading__hidden-content{% endif %}">
       {{ item.content }}
     </div>
   {% endfor %}
--- a/packages/usa-js-loading/_index.scss
+++ b/packages/usa-js-loading/_index.scss
@@ -1,0 +1,4 @@
+// dependencies
+
+// src
+@forward "src/styles";
--- a/packages/usa-js-loading/src/styles/_index.scss
+++ b/packages/usa-js-loading/src/styles/_index.scss
@@ -1,0 +1,4 @@
+// dependencies
+
+// src
+@forward "usa-js-loading";
--- a/packages/usa-js-loading/src/styles/_usa-js-loading.scss
+++ b/packages/usa-js-loading/src/styles/_usa-js-loading.scss
@@ -1,0 +1,3 @@
+.usa-js-loading .usa-js-loading__hidden-content {
+  display: none;
+}
--- a/packages/uswds/_index.scss
+++ b/packages/uswds/_index.scss
@@ -2,6 +2,7 @@
 
 // Global
 // -------------------------------------
+@forward "usa-js-loading";
 @forward "uswds-global";
 
 // Helpers
--- a/src/stylesheets/packages/_usa-js-loading.scss
+++ b/src/stylesheets/packages/_usa-js-loading.scss
@@ -1,0 +1,4 @@
+// dependencies
+
+// src
+@forward "usa-js-loading";
--- a/src/stylesheets/packages/_uswds-global.scss
+++ b/src/stylesheets/packages/_uswds-global.scss
@@ -1,5 +1,6 @@
 // Global
 // -------------------------------------
+@forward "usa-js-loading";
 @forward "uswds-elements/lib/normalize";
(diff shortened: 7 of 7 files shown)


packages/usa-accordion/src/usa-accordion.twig
```diff
@@ -6,7 +6,7 @@
         {{ item.title }}
       </button>
     </h4>
-    <div id="{{ id_prefix }}{{ item.id }}" class="usa-accordion__content usa-prose">
+    <div id="{{ id_prefix }}{{ item.id }}" class="usa-accordion__content usa-prose {% if item.expanded != true %}usa-js-loading__hidden-content{% endif %}">
```
mejiaj, line 9: **question**: Is this to demo the feature or a new requirement for users?

mejiaj (commented):
Thank you, the approach looks good.

A small question on the accordion markup change. I can see this being complementary to #5826. Ideally, it'd be nice to save users from having to a markup change.

packages/usa-accordion/src/usa-accordion.twig
aduth: This pull request is an alternative to #5826 . With this approach, it'd require that users change their markup to include this new CSS class _as a progressive enhancement_ if they want to opt-in to the improved behavior. #5826 avoids this requirement, at the expense of having a gnarly CSS query (per your review feedback) and [an assumption about where in the markup the content lives](https://github.com/uswds/uswds/pull/5826#:~:text=Note%20by%20%22assumes,JavaScript%20has%20loaded.).

If #5826 is preferred, this pull request can be closed. If this pull request is preferred, I'd expect to close #5826 .

mejiaj: Both solutions look good. We try to minimize markup changes, wherever possible. If it's one or the other; I lean towards (#5826) with this PR as a possible option for the future.

Will defer to @thisisdano, @heymatthenry for final word.

PR #4243, opened 2021-06-23
DRAFT: Optimize initializer script

scottqueen-bixal:
This builds off https://github.com/uswds/uswds/pull/4233

- [ ] Follow the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [ ] Run `npm test` and make sure the tests for the files you have changed have passed.
- [ ] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.
- [ ] Title your pull request using this format: [Website] - [UI component]: Brief statement describing what this pull request solves.

PR #5819, opened 2024-03-15
USWDS - Core: Shrink size of uswds-init.js

aduth:
# Summary

**Optimized the size of the initializer script.** The download size is reduced by 64% with compression.

## Related pull requests

Related to (builds upon): #4233 (merged in #4243)

## Problem statement

The intent of the USWDS initializer script is to be loaded synchronously (blocking) at the beginning of page content to reduce the effect of flashing content. The fact that it is loaded in this way means it is very important for it to be as small as possible, to avoid the script itself preventing the page from loading in a timely manner.

Producing a small initializer script also opens makes it more feasible to be included as an inline script without worrying about the size (bytes transferred) of the inline script content.

## Solution

The proposed changes do the following, in order of effect:

- Remove the timer fallback class reversion
   - **Why?** This seems to have been implemented on the expectation that a `load` event would be called for each script being loaded into the page, and accommodating the scenario that none of those scripts would initialize USWDS. But the `window` `load` event is called ["when the whole page has loaded"](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event). Therefore, if USWDS is not present at the time that this event is fired, the class should be reverted immediately.
- Remove variables
   - **Why?** While this avoids some duplication resulting in an on-disk size reduction, the use of variables actually _increases_ the size of the asset after common compression, since compression (gzip, brotli) are [highly effective at compressing repeated content](https://en.wikipedia.org/wiki/Deflate#Duplicate_string_elimination:~:text=Within%20compressed%20blocks%2C%20if%20a%20duplicate%20series%20of%20bytes%20is%20spotted%20(a%20repeated%20string)%2C%20then%20a%20back%2Dreference%20is%20inserted%2C%20linking%20to%20the%20previous%20location%20of%20that%20identical%20string%20instead.). Eliminating variables also contributes toward being able to remove the IIFE, which is only needed in order to [create a closure context to avoid variables leaking into the global scope](https://developer.mozilla.org/en-US/docs/Glossary/IIFE#:~:text=This%20prevents%20accessing%20variables%20within%20the%20IIFE%20idiom%20as%20well%20as%20polluting%20the%20global%20scope.).
- Use arrow functions
   - **Why?** Fewer characters, at the cost of IE11 support. Even in IE11, the worst case scenario is that this code fails to run at all and would not add the `usa-js-loading` class, which can be considered a progressive enhancement.
- Remove `"use strict"`
   - **Why?** It likely doesn't have a meaningful effect given the constrained scope of the script. Additionally, if desired, strict mode can be enabled at script load by loading the script as ESM (`type="module"`), which is [strict mode by default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#:~:text=This%20is%20because%20modules%20use%20strict%20mode%20automatically.).
- Remove sourcemap
   - **Why?** Similar to above, a sourcemap for this file is not especially useful given its size, and it is already reasonably readable as-is, even when compressed.

Result (using [`brotli-size-cli`](https://www.npmjs.com/package/brotli-size-cli)):

```
npx gulp buildJS && brotli-size dist/js/uswds-init.min.js
```

**Before:** 188 bytes
**After:** 68 bytes
**Diff:** -120 bytes (-63.8%)

~If we wanted to really scrape every last optimization, we could probably also look to:~

1. ~Drop the `"use strict";` statement, since it likely doesn't have a meaningful impact on this script ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode))~
2. ~Reconfigure minification to avoid `//# sourceMappingURL=uswds-init.min.js.map` included in the output~

~Those two additional changes would bring the resulting Brotli size to 68 bytes (-63.8% overall).~ **Edit:** These are implemented with cc3cae646 and 80448a9da.


ryparker:
Thanks for working on reducing the initializer's size. I'm sorry this went without a review. I'm closing it during the backlog cleanup because we don't have capacity to review it now.

If you'd like to pick it back up, comment here or reopen the PR. I'll help with the review and any rebase it needs.

PR #5826, opened 2024-03-19
USWDS - Accordion: Hide collapsed content while JavaScript loads

### LINKED ISSUE #5817: USWDS - Feature: Improve accordion flash of content while JavaScript loads

aduth:
# Summary

**Collapsed accordion content will be hidden while JavaScript is loading.** This avoids the appearance of flashing content during page load.

## Related issue

Closes https://github.com/uswds/uswds/issues/5817

Related to https://github.com/uswds/uswds/issues/3092

## Related pull requests

This is an alternative or complementary solution to #5818.

## Preview link

Preview link: http://localhost:6006/?path=/story/components-accordion--default

## Problem statement

See #5817

## Solution

This is the fourth solution proposed in https://github.com/uswds/uswds/issues/5817:

>Assume that the content follows immediately after the expander button and use new [CSS :has pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) to stylize as hidden depending on button, during load

Note by "assumes" that the actual visibility toggle is based on the button's `aria-controls` reference, which technically could support accordion content occurring elsewhere in content. However, the proposed solution follows reference component example documentation. In the worst case, this only applies during the page load, and the corrected visibility state will always take effect once JavaScript has loaded.

The benefit of this approach over #5818 is that it "just works" for existing accordions and does not require additional markup changes to opt-in to benefitting from the behavior.

## Testing and review

1. Go to http://localhost:6006/?path=/story/components-accordion--default
2. Observe hidden content unaffected by changes
3. In local copy of running code, comment out these lines: https://github.com/uswds/uswds/blob/98566ec5dc12807a2b0bc36699a033fe3a711d44/packages/usa-accordion/src/index.js#L86-L89
4. Refresh the page
5. Observe all content is visible
6. Using browser DevTools inspector, add `usa-js-loading` CSS class to some ancestor element of one of the accordions which would be collapsed without the changes implemented in Step 3
7. Observe that the accordion content becomes hidden


aduth:
We're now running this code on https://www.login.gov/ as of https://github.com/GSA-TTS/identity-site/pull/1224, where it's helping avoid a flashing effect on our language picker, which is implemented as an accordion. You can see the effect in the recordings below:

Before:

https://github.com/GSA-TTS/identity-site/assets/1779930/a369d37f-4adc-4aa9-8652-2ca456f6100c

After:

https://github.com/GSA-TTS/identity-site/assets/1779930/cfab56f4-d533-46a5-b36e-c8d5ff5430e4

CTGM-Bixal (approved):
This looks good to me!

Confirmed that when `init` code is commented out and the `usa-js-loading` class is added to the root element containing accordion:
- [X] all the content collapses

Also tested:
- [X] When a user uses nested accordions and/or dynamic content within the accordion component, all accordion-buttons and dynamic content behaves as expected
- [X] Accordion-buttons that have `aria-expanded` set to `true` remain expanded while other's are collapsed

Although this css seems a little more complex in comparison to how we use `usa-js-loading` in other areas (such as banner and modal), this css seems relatively simple and efficient.

mejiaj:
@aduth can we scope this to the language picker accordion? It's at the top of the page, so would merit this treatment. 

The accordions in body don't need to be collapsed and the impact of flash of unstyled conntent wouldn't be as big.

aduth:
@mejiaj I'll take a look to see what it might look like to scope this down to applying to the language selector. We use the base accordion component for our language picker and not the [language selector component](https://designsystem.digital.gov/components/language-selector/), but it looks like the language selector internally reuses accordion so hopefully it should be able to apply the same.

(And, as an aside, we should probably update our implementation to use the language selector component)

mejiaj:
@aduth no worries, we can keep this work as-is.

We discussed in an internal meeting that we **do** want to resolve the flash of unstyled content anywhere it might be problematic to users.

Sorry for the confusion and thanks for the quick response!

packages/usa-accordion/src/styles/_usa-accordion.scss
```diff
@@ -196,3 +196,17 @@ $accordion-button-open-hc-icon: map-merge(
...
+    .usa-accordion__heading:has(
+        .usa-accordion__button:not([aria-expanded="true"])
+      ),
+    .usa-accordion__button:not([aria-expanded="true"]) {
+      & + .usa-accordion__content {
+        @include add-sr-only;
+      }
+    }
+  }
+}
```
mejiaj, line 212: **issue**: Can we simplify this selector? We could improve readability and clearly highlight intent with a simpler selector.

**For example**
```scss
// [uswds-init.js] Hide collapsed content offscreen while the JS loads.
.usa-js-loading {
  .usa-accordion__heading:has(.usa-accordion__button[aria-expanded="false"])
    + .usa-accordion__content {
    @include add-sr-only;  
  }
}
```

mejiaj (commented):
Overall looks good, added a comment for potentially simplifying the selector to clearly highlight intent.

---

To test, I had to add the init script to storybook's preview script.

```js
// preview.js
import "../packages/uswds-core/src/js/uswds-init";
```

Alternatively, and probably better, you can add the script to `.storybook/preview-head.html`, that'd be closer to our guidance in docs.

packages/usa-accordion/src/styles/_usa-accordion.scss
aduth: Part of the idea with the `:not([aria-expanded="true"])` is that an accordion can be marked up without `aria-expanded` and it's treated effectively the same as `aria-expanded=false`, but in those cases the `aria-expanded` attribute isn't assigned until the JavaScript finishes loading (the problem we're targeting), and the CSS selector wouldn't match if it was looking for `aria-expanded="false"`.

Then again, the documented code examples for accordion always include `aria-expanded`, so maybe that's not a use-case we want to support 🤷

aduth: Similarly, the idea with having separate selectors for `usa-accordion__heading` and `usa-accordion__button` is that it's also technically supported to mark up an accordion with a button only and not the heading element. 

```html
<!-- Valid: -->
<h4 class="usa-accordion__heading">
  <button type="button" class="usa-accordion__button" aria-controls="a1">
    Accordion Button
  </button>
</h4>
<div id="a1" class="usa-accordion__content usa-prose"></div>

<!-- Also valid: -->
<button type="button" class="usa-accordion__button" aria-controls="a2">
  Accordion Button
</button>
<div id="a2" class="usa-accordion__content usa-prose"></div>
```

Although again, the documented code examples for accordion always include a wrapping heading element, so maybe that's another use-case we don't want to support.

mejiaj: Good point, thanks for clarifying!

mejiaj (approved):

mejiaj:
Moving this to Fed Final Review since we decided this was the least invasive option and is in-line with the few components we have this on. 

If the need grows for a dedicated package, then we'll revisit #5818.

ryparker:
@coderabbitai full review

PR #5857, opened 2024-04-09
USWDS - Identifier: Fix content prefix bug in html-templates

### LINKED ISSUE #5853: USWDS - Bug: Incorrect text in Spanish language versions of the identifier component
### Describe the bug

The Spanish version of the [identifier component](https://designsystem.digital.gov/components/identifier/#identifier-preview) needs to be updated to reflect the following changes:

- change "Visite USA.gov en Español" (https://www.usa.gov/espanol/) to "Visite USAGov en Español" (https://www.usa.gov/es/)
- take out "An" from "An Un sitio web oficial de ..."

Text changes in the Identifier component - changes listed under "describe the bug"

### Expected Behavior

Requested changes in "describe the bug"

### Related code

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

amyleadem:
# Summary

Fixed the html-templates build for Spanish variants of the identifier component. Updated the Explicitly set `content_prefix` to `null` in JSON for Spanish language variants. This will remove the prefix "An" from the built html-templates.

## Breaking change

## Related issue

Closes #5853 

## Related pull requests

_Indicate if there are other pull requests necessary to complete this issue._

## Preview link

Preview link:

## Problem statement

## Solution

## Major changes

## Testing and review

### DIFF
--- a/packages/usa-identifier/src/content/usa-identifier~lang-es.json
+++ b/packages/usa-identifier/src/content/usa-identifier~lang-es.json
@@ -5,6 +5,7 @@
   "masthead": {
     "aria_label": "Identificador de la agencia,",
     "description": "Descripción de la agencia,",
+    "content_prefix": null,
     "content": "Un sitio web oficial de",
     "parent": {
       "name": "<Parent agency>",
--- a/packages/usa-identifier/src/content/usa-identifier~multiple-logos-lang-es.json
+++ b/packages/usa-identifier/src/content/usa-identifier~multiple-logos-lang-es.json
@@ -5,6 +5,7 @@
   "masthead": {
     "aria_label": "Identificador de la agencia,,,",
     "description": "Descripción de la agencia,,,",
+    "content_prefix": null,
     "content": "Un sitio web oficial de",
     "parent": {
       "name": "<Parent agency>",
--- a/packages/usa-identifier/src/content/usa-identifier~no-logos-lang-es.json
+++ b/packages/usa-identifier/src/content/usa-identifier~no-logos-lang-es.json
@@ -5,6 +5,7 @@
   "masthead": {
     "aria_label": "Identificador de la agencia,,",
     "description": "Descripción de la agencia,,",
+    "content_prefix": null,
     "content": "Un sitio web oficial de",
     "parent": {
       "name": "<Parent agency>",
--- a/packages/usa-identifier/src/content/usa-identifier~taxpayer-disclaimer-lang-es.json
+++ b/packages/usa-identifier/src/content/usa-identifier~taxpayer-disclaimer-lang-es.json
@@ -5,6 +5,7 @@
   "masthead": {
     "aria_label": "Identificador de la agencia,,,,",
     "description": "Descripción de la agencia,,,,",
+    "content_prefix": null,
     "content": "Un sitio web oficial de",
     "parent": {
       "name": "<Parent agency>",


packages/usa-identifier/src/content/usa-identifier~multiple-logos-lang-es.json
```diff
@@ -5,6 +5,7 @@
   "masthead": {
     "aria_label": "Identificador de la agencia,,,",
     "description": "Descripción de la agencia,,,",
+    "content_prefix": null,
```
mahoneycm, line 8: Is there a reason we aren't using `Un` as a prefix here to match the `An` in the English variants? To me it makes more since to keep this consistent rather than moving the prefix into the content and marking this `null`

mahoneycm (changes requested):
This is looking good to me! One question about using `Un` as the prefix instead of `null` but everything else is looking good.

Since the link works with a redirect in it's current state, I was leaning that this could be a patch fix, _however_, thinking about how widespread the typo could be we'll definitely want to flag it and that might necessitate a minor version.

Clicking through some showcase links, I was able to find a few identifiers whose identifiers didn't have the `An` and were displaying correctly, so maybe the issue isn't as  widespread and we might worry.
## Testing checklist

- [x]  Confirm the HTML in the `html-templates` directory renders as expected
    - [x]  Check that the component preview and component code on the demo [identifier page](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/al-identifier-html-fix/components/identifier/) display with the updated information
    - [x]  In this branch, run `npm run build:html` and confirm the HTML in the affected identifier files have the correct information
- [x]  Confirm that the Storybook build renders as expected
- [x]  Confirm there are no references to "Visite USA.gov en Español" or "https://www.usa.gov/espanol/" in uswds

packages/usa-identifier/src/content/usa-identifier~multiple-logos-lang-es.json
amyleadem: The reason we add "An" as a prefix here is so that we can isolate it and remove it from the screen reader readout, because "An official" sounded like "unofficial" when spoken (PR #5491). This understandably caused some confusion, so now the screen reader just reads out "Official website...". As far as I know, there is nothing in the Spanish phrase that causes the same type of confusion. 

I can see an argument for including it for consistency in structure, but I lean away from that because it feels like it adds an (albeit small) unnecessary burden to developers on Spanish sites to manipulate the structure like that.

mahoneycm: Ahh that makes sense! Thank you for clarifying! I'll give this my approval 👍

mahoneycm (approved):
LGTM!

packages/usa-identifier/src/content/usa-identifier~lang-es.json
```diff
@@ -5,6 +5,7 @@
   "masthead": {
     "aria_label": "Identificador de la agencia,",
     "description": "Descripción de la agencia,",
+    "content_prefix": null,
```
mejiaj, line 8: For consistency, it seems like the Spanish prefix here would be `Un`.

mejiaj (commented):
Minor comment, but not a blocker — just a suggestion.

--- 

Also, the breaking change section in PR description could use some clarity.

>This is not a breaking change.
>This is a markup change.

This means that there are two changes total and one is minor (template prefix) and the other breaking (markup) — correct?

>Important
>This might need to be bumped to 3.9.0 because this involves markup changes. We could also explore splitting this into two different PRs - one to fix the erroneous "An" in the component preview in 3.8.1 and one for the intentional text updates in 3.9.0.

Splitting it into two makes sense to me. Especially if we want to follow semver. Any preference @thisisdano?

packages/usa-identifier/src/content/usa-identifier~lang-es.json
mahoneycm: I brought up the same thing in [this thread](https://github.com/uswds/uswds/pull/5857#discussion_r1562625732) but Amy clarified that this attribute is used to ignore `An` in `an official` because it sounds like "unofficial". We shouldn't need to ignore the prefix in the Spanish variant!

amyleadem:
Update: I am going to work on splitting this into two separate PRs. This one will remain open for 3.8.1 and I will open a separate one for 3.9.0. Will post the additional PR here when it is ready.

amyleadem:
Update: I have created a new issue for this bug (https://github.com/uswds/uswds/issues/5893) and have broken out the related link update in PR #5892. 

@mahoneycm and @mejiaj This is ready for your re-review

mejiaj (approved):
Thanks for splitting up the PR. Helps a lot with testing.

Confirming the output Spanish HTML is no longer getting the English prefix.

```diff
<!-- On `npm run build:html` →
- <span aria-hidden="true">An </span>Un sitio web oficial de
+ Un sitio web oficial de
```

thisisdano (approved):
Thank you!

PR #5899, opened 2024-04-26
USWDS - Add missing angle bracket to package.json

skyf0l:
Just a small mistake


amyleadem (approved):
Looks good to me! Thanks for fixing this @skyf0l. 

**question:** Team, as a follow-on task, do we want to assess the purpose of this contributor list and assess if we should make any updates?

mejiaj (approved):
Thanks for fixing this! 

As you can see this section doesn't get many updates. We've discussed moving this elsewhere to clean up the `package.json` a little more + fit in more contributors.

thisisdano (approved):
Thank you!

Issue #5884, opened 2024-04-24
USWDS - Enhancement: Normalize form input error states.


mahoneycm:
# Description

While creating error state previews for #5879 I found a few visual discrepancies we should address before providing guidance on the site for uswds/uswds-site#2476

## Tasks

- [ ] Error border only visible below desktop widths
- [ ] Combobox elements lose `usa-input--error` on initialization
- [ ] #5517

amyleadem:
Adding a few more recommendations for future tasks:
- Standardize the structure for form element markup (For example, some components include a `usa-form` wrapper, others don't)
- Use includes to build the collected form stories
- Create a single collected form story that lets us toggle default, disabled, and error states

mahoneycm:
Hey there @jaclinec, we're trying to place the weight of these issues to decide if this should be approached in USWDS 3 or to tackle it in USWDS 4.

Do you have any opinion or insight into where this would fall into top tasks and how desired error state and guidance is?

Thanks!

jaclinec:
Hey @mahoneycm! So, using component guidance is a top task for users, but we didn't get more granular than that (in other words, we didn't ask about guidance around error states). I do think it's important, but AFAIK we don't have data yet on user needs for error states outside of anecdotal evidence. 

My 2 cents - forms are prolific on government sites, so I think it's safe to assume our users need robust support (code and guidance) for error states.

Issue #2794, opened 2018-09-27
Version 2 - Mobile menu oddity in Edge


stphnwlkr:
Seeing a very strange behavior in the V2 website. The desktop Edge browser, reduced to show the mobile menu, is showing bullets. This does not appear in Edge on iOS. This is a screenshot comparing Edge and Chrome.
![nav_issue](https://user-images.githubusercontent.com/1405187/46165371-7c203500-c25e-11e8-8888-95bb6ba64948.png)

thisisdano:
Hm! 🤔 Let me take a look at that. Are the bullets just in the mobile menu or do they appear in any other menu or unexpected location?

stphnwlkr:
Just in the mobile menu

thisisdano:
Unfortunately, I'm not able to easily troubleshoot Edge. Can you inspect this and let me know if there's something that looks amiss: particularly in `list-style-type` being attached to the wrong element or something like that?

stphnwlkr:
I am not sure either...compared Chrome to Edge and no difference.

stphnwlkr:
Interesting...apparently it is a known issue.  list-style-type:none not loading on initial page load. Refreshing the page fixes the problem. I have never noticed it.

https://github.com/webhintio/webhint.io/issues/210

thisisdano:
Whoa. Thank you for tracking that down!

stphnwlkr:
I submitted a bug to Microsoft (I think it might be a regression). 

Sort of related but not browser specific -  usa-accordion-bordered does not set list-style-type to none. See https://v2.designsystem.digital.gov/components/typography/#pairings-and-styles

thisisdano:
Awesome — and thanks. Fix is already underway for that accordion issue :)

stphnwlkr:
So will the accordion fix be like buttons where "bordered" is just and extension of usa-accordion ?

thisisdano:
Yes, in general we are moving to a classing model where variant classes are additive to the base class. As in `usa-button usa-button-outline` and `usa-accordion usa-accordion-bordered`

stphnwlkr:
Excellent!

maya:
Should this be closed since it's a Chrome bug?

thisisdano:
Edge bug, but yes :) 👋

PR #5908, opened 2024-05-01
USWDS - Character count: Enhance visual cue when limit is exceeded

### LINKED ISSUE #5851: USWDS - Character count: Implement a visual cue when the limit is exceeded
During summer 2023 usability testing, 3 users talked about how they prefer and expect to be prevented from typing or for some other indicator to let them know they have exceeded the limit. "That's the best way to alert me because I will no longer hear characters being typed. So that kind of lets me know that I've hit the end."

# Affected user groups

- All users
- Screen reader users
- Screen magnification users

# Research method
Usability testing with 5 visually impaired participants. See [findings report ](https://github.com/uswds/uswds/wiki/Usability-research-findings-for-Q3-2023)for details.

mahoneycm:
# Summary

Enhanced visual cue when `maxlength` is exceeded. Now, the component uses standard USWDS error styles to visually enhance the error state.

## Breaking change

## Related issue

Closes #5851

## Related pull requests

[Changelog →](https://github.com/uswds/uswds-site/pull/2648)

### New visual change only
This link contains only the changes from **this** PR
[Character count →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-char-count-error-state/iframe.html?args=&id=components-form-inputs-character-count--character-count&viewMode=story)

### Combined with additional styles
This preview combines this work with #5999. This is what the change will look like on release.
[Character count preview with enhanced form group error styles →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-test-repo-5908/iframe.html?args=&id=components-form-inputs-character-count--character-count&viewMode=story)

_[Link to demo branch](https://github.com/uswds/uswds/pull/6036)_

## Problem statement

Testing results show that users expect additional indication that they have exceeded the limit. Additionally, screen magnification users may be unable to see the character count status message beneath the textarea input.

## Solution

Add additional visual cues to let users know they have exceeded the character count limit.

## Major changes

- When users surpass the character limit, the relevant error classes are added to the form input, group, and label elements
- This class will be removed when the character count is beneath the limit

### Character count behaviors

1. Visit the [demo branch preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-test-repo-5908/iframe.html?args=&id=components-form-inputs-character-count--character-count&viewMode=story).
3. For both input variants, surpass relevant character limits.
4. Confirm label, input, and form group error classes and styles are applied correctly.
7. Backspace input to validate character count.
8. Confirm error class and styles are removed.
9. Run unit test and confirm there are no errors.

### Testing checklist

- [ ]  `npm run test` runs without error
- [ ]  `usa-input--error`classes are applied on character input over character limit
- [ ]  `usa-input--error`classes are removed on character input under character limit
- [ ]  `usa-form-group--error` classes are applied on character input over character limit
- [ ]  `usa-form-group--error` classes are removed on character input under character limit
- [ ]  `usa-label--error` classes are applied on character input over character limit
- [ ]  `usa-label--error` classes are removed on character input under character limit
- [ ]  Character count JS matches USWDS code standards.
- [ ]  ~Unit tests match USWDS code standards~ _// Unit tests removed_


packages/usa-character-count/src/index.js
```diff
@@ -28,13 +35,25 @@ const getCharacterCountElements = (inputEl) => {
...
+  if (!formGroupEl) {
+    throw new Error(`${CHARACTER_COUNT} is missing inner ${FORM_GROUP}`);
+  }
+
+  const labelEl = characterCountEl.querySelector(LABEL);
+
+  if (!labelEl) {
+    throw new Error(`${CHARACTER_COUNT} is missing inner ${LABEL}`);
+  }
+
```
mahoneycm, line 49: > [!NOTE]
> Added checks for form group and label elements. These elements are necessary for the overall structure of the component and will result in JS errors for `toggleErrorState` if they are missing.

packages/usa-character-count/src/index.js
```diff
@@ -150,9 +169,25 @@ const updateCountMessage = (inputEl) => {
...
+ * @param {HTMLInputElement|HTMLTextAreaElement} inputEl The character count input element
+ */
+const toggleErrorState = (inputEl) => {
+  const { labelEl, formGroupEl } = getCharacterCountElements(inputEl);
+  const errorState = inputEl.classList.contains(INPUT_ERROR_CLASS);
+
+  labelEl.classList.toggle(LABEL_ERROR_CLASS, errorState);
+  formGroupEl.classList.toggle(FORM_GROUP_ERROR_CLASS, errorState);
+};
+
```
mahoneycm, line 190: > [!NOTE]
> `toggleErrorState` adds/removes `form-group` and `usa-label` with relevant `--error` classes. Called on `focusout`

packages/usa-character-count/src/index.js
```diff
@@ -5,7 +5,14 @@ const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
 
 const CHARACTER_COUNT_CLASS = `${PREFIX}-character-count`;
 const CHARACTER_COUNT = `.${CHARACTER_COUNT_CLASS}`;
+const FORM_GROUP_CLASS = `${PREFIX}-form-group`;
+const FORM_GROUP_ERROR_CLASS = `${FORM_GROUP_CLASS}--error`;
+const FORM_GROUP = `.${FORM_GROUP_CLASS}`;
+const LABEL_CLASS = `${PREFIX}-label`;
+const LABEL_ERROR_CLASS = `${LABEL_CLASS}--error`;
+const LABEL = `.${LABEL_CLASS}`;
```
mahoneycm, line 13: > [!NOTE]
> Added classes and element used by `toggleErrorState`. Organized by when they appear in markup/DOM

packages/usa-character-count/src/index.js
```diff
@@ -178,11 +213,19 @@ const characterCount = behavior(
...
+      },
+    },
   },
   {
     init(root) {
       select(INPUT, root).forEach((input) => enhanceCharacterCount(input));
     },
+    FORM_GROUP_ERROR_CLASS,
+    LABEL_ERROR_CLASS,
+    INPUT_ERROR_CLASS,
```
mahoneycm, line 228: > [!NOTE]
> Exported for use in new unit tests

packages/usa-character-count/src/test/character-count.template.html
mahoneycm, line 1: ## Updated unit test templates

I added `usa-form-group` and `usa-label`'s to all character count unit tests for consistency and to prevent JS errors now that the component checks for these elements.

packages/usa-character-count/src/test/invalid-template-no-form-group.spec.js
mahoneycm, line 1: > [!NOTE]
> New unit test to check for the existence of form group element

packages/usa-character-count/src/test/invalid-template-no-label.spec.js
mahoneycm, line 1: > [!NOTE]
> New unit test to check for the existence of the label element

packages/usa-character-count/src/test/character-count.spec.js
```diff
@@ -139,12 +157,39 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
+      label.classList.add(LABEL_ERROR_CLASS);
+
+      EVENTS.focusout(input);
+
+      assert.strictEqual(
+        formGroup.classList.contains(FORM_GROUP_ERROR_CLASS),
+        false
+      );
+      assert.strictEqual(label.classList.contains(LABEL_ERROR_CLASS), false);
+    });
```
mahoneycm, line 191: > [!NOTE]
> New testing requirements to check for the addition / removal of error states on input and focusout

packages/usa-character-count/src/index.js
```diff
@@ -178,11 +213,19 @@ const characterCount = behavior(
         updateCountMessage(this);
       },
     },
+    focusout: {
+      [INPUT]() {
+        toggleErrorState(this);
+      },
+    },
```
mahoneycm, line 220: ## Add form group error classes on focusout

This idea came from a discussion between @jaclinec and I. We wanted to add the full error state classes but didn't want to have the content shift while the user was typing. She suggested we add these classes only after the user _leaves the field_.

packages/usa-character-count/src/index.js
```diff
@@ -150,9 +169,25 @@ const updateCountMessage = (inputEl) => {
     inputEl.setCustomValidity("");
   }
 
+  inputEl.classList.toggle(INPUT_ERROR_CLASS, isOverLimit);
```
mahoneycm, line 172: > [!NOTE]
> Adds / removes input error classes to match invalid status message class

mahoneycm (commented):
Due to the unit test updates, there are a handful of changed files with this PR. Added some comments to help add clarity to each of the changes:

amycole501:
@mahoneycm we think there might be an issue with the blue focus indicator and the red indicator not having enough contrast with one another. Is there a way to add a small space between the two (maybe a few pixels) to make a distinction there are two states happening (one focus, one error)? https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html

mahoneycm:
@amycole501 Here's what it looks like it with an offset of 2px

image

[Kapture 2024-05-03 at 15.59.41.webm](https://github.com/uswds/uswds/assets/25211650/e5f3c0d5-dfc3-4f7f-8dda-6dda5d665e15)


The input shrinks a bit but this is actually a result of the error border being added and not the outline offset. The same difference in input size occurs with or without the offset.

What do we think about something like this?

This issue _might_ be a bit out of scope for this specific issue but definitely something we should consider with #5884

CC: @jaclinec @mejiaj @amyleadem

amycole501:
I like it! Would like to know what @alex-hull thinks but the separation makes it much easier to see the two borders.

amyleadem (changes requested):
@mahoneycm 
I started testing the interactions on this and generally it looks good! I had one question below about what happens after you fix an error (shown in the comment below). Let me know if you want to discuss!

I'll return to the code review once we figure that out.

packages/usa-character-count/src/index.js
mahoneycm: @amyleadem this is the intentional behavior based on @jaclinec's recommendation to not cause content shift while the user is typing (see [comment](https://github.com/uswds/uswds/pull/5908#discussion_r1587818882) above and testing instructions)

If the content shift on _removal_ is acceptable, I can make it so that the error state is immediately removed rather than waiting for `focusout`.

Here's a quick example of adding/removing the classes on `input`. I think this would be particularly jarring for screen magnification users

 
[Kapture 2024-05-23 at 10.02.35.webm](https://github.com/uswds/uswds/assets/25211650/b35365c9-3e71-47e4-8692-aa7191e8292f)

mahoneycm:
> #5884

Currently, I'm leaning towards holding off on implementing this change since it has wider implications than just character count. I think it's a great candidate for #5884 which aims to normalize error states across all form components. I added a checklist item to that issue 👍

packages/usa-character-count/src/index.js
mahoneycm: Removed in [95a33e3](https://github.com/uswds/uswds/pull/5908/commits/95a33e323528e4c2310c487dc3fb9b3ea7b2e571)

packages/usa-character-count/src/test/character-count.spec.js
mahoneycm: Removed in [b7fa1f0](https://github.com/uswds/uswds/pull/5908/commits/b7fa1f0494c815f2c04032cf025d16944ea6912c)

packages/usa-character-count/src/index.js
mahoneycm: Behavior updated in [b7fa1f0](https://github.com/uswds/uswds/pull/5908/commits/b7fa1f0494c815f2c04032cf025d16944ea6912c)

Preview of new behaviors from the [demo repo](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-test-repo-5908/iframe.html?args=&id=components-form-inputs-character-count--character-count&viewMode=story). 

[Kapture 2024-08-23 at 12.16.41.webm](https://github.com/user-attachments/assets/2061a434-b21d-4e3f-85b6-b88cd4d34472)


There is a slight shift due to the increased border, but I don't think it's enough to cause concern!

mahoneycm:
## Update 8/23

- Restored group error state on input
- Removed `focusout` behaviors
- Created [demo repo](https://github.com/uswds/uswds/pull/6036) to showcase changes with updated form group styles from #5999

mejiaj:
@mahoneycm can you update the PR description preview links for clarity? For testing this PR, I expected to see the latest code _without_ the layout shift.

mejiaj (approved):
New visual changes look good and unit tests pass. Thanks!

**Before**

https://github.com/user-attachments/assets/a07db912-a85f-4a78-a06a-77a657d3492b


**After**

https://github.com/user-attachments/assets/7c47d7ae-3032-4691-922f-f591af2dc4b3

jaclinec (approved):
This looks great! Nice work!

amyleadem (commented):
@mahoneycm Behavior and styles are looking good @mahoneycm! 

I had one open question about the failure of the component when either `.usa-label` or `.usa-form-group` is missing. It feels a bit fragile to me. I'm wondering if we should at least mark it as a (potentially) breaking change so that users check that their implementations of the character count component have all the needed elements to work. 

Curious what others think!

packages/usa-character-count/src/index.js
mahoneycm: Behavior removed to account for new error state styles

amyleadem: This behavior makes more sense to me. Thanks @mahoneycm!

packages/usa-character-count/src/index.js
```diff
@@ -28,13 +35,25 @@ const getCharacterCountElements = (inputEl) => {
...
+
+  if (!formGroupEl) {
+    throw new Error(`${CHARACTER_COUNT} is missing inner ${FORM_GROUP}`);
+  }
+
+  const labelEl = characterCountEl.querySelector(LABEL);
+
+  if (!labelEl) {
+    throw new Error(`${CHARACTER_COUNT} is missing inner ${LABEL}`);
+  }
```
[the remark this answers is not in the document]
mahoneycm: **For form-group:**
This might be a good thing to discuss when normalizing things like form group and when we're using it. 

Since we include this error state as a behavior for character count, I see the importance of flagging that it should be included so users are properly updated that their form is in the error state. 

Maybe it would be better to issue a warning instead? I was matching the existing behaviors by adding the error.

**For label:**
If a user doesn't include a label for character count, it would be considered an accessibility failure so I think it's important to include _but_ looking at it again, I think I should have used the `for` attribute and the character count's ID instead of checking if it exists within the element tag.

That way if a user included a label but didn't nest it within the `usa-character-count` div, it'll still pass the test. 

How do you feel about that?

amyleadem:
@mahoneycm For next steps, can you change the error to a warning and create a test story that shows what happens when some elements of the form are not included in the build (usa-label, usa-form-group, etc)? We should also look at a warning in our guidance.

packages/usa-character-count/src/index.js
mahoneycm: Removed errors in [fe42118](https://github.com/uswds/uswds/pull/5908/commits/fe4211882a16f30e80c4559bb970e9b3119dfdf4)

packages/usa-character-count/src/test/invalid-template-no-form-group.spec.js
mahoneycm: **todo:** Remove unit test if we remove related errors

packages/usa-character-count/src/test/invalid-template-no-label.spec.js
mahoneycm: todo: Remove unit test if we remove related errors

mahoneycm:
## Update 9/12

**Changes**
- Removed error throws from `getCharacterCountElements()` function.
- Created two test stories to showcase character counts behaviors without a label and without the form-group div

**Todo**
- Remove label and form group test stories created to test error messages. I held off for now to confirm that we want to remove the errors.

**Didn't do**

I opted out of adding a `console.warn()` in place of the error message for two reasons

1. Too many warning messages

Currently, `getCharacterCountElements()` is called twice during initialization and on every input. This causes a lot of warning messages in the log and would be less helpful for the devs/users.

image

We could consider a slight refactor to limit when we call `getCharacterCountElements()` to resolve this issue. I held off because this felt like scope creep.

**note:** This is not an issue with the errors because the error prevents further JS from being ran.

2. Console warn isn't an existing pattern and it causes linting errors

There was some team hesitation around adding a console message because we don't do this anywhere else in our code and we would have to add linter ignore prompt. 

If we decide we would like to keep the warning, we can address these concerns and restore it.

**Question for team:** 

How do we feel about the necessity for labels or form group elements? 

Labels are required for accessibility compliance to provide instruction for users as well as the bolded error label styles. Form group is required for the left-side error border.

mejiaj (commented):
Confirming that I'm seeing the error styles when exceeding character limit. I've added some minor comments for code style.

### Re: Questions for the team

>How do we feel about the necessity for labels or form group elements?

Seems valid. Feels a little odd to check for valid form HTML, but it matches other patterns where we throw errors if the required component markup is missing.

**My opinion**
I'd prefer to rely on guidance, but I can also see a case for throwing an error.

packages/usa-character-count/src/usa-character-count.stories.js
```diff
@@ -12,6 +14,8 @@ export default {
 };
```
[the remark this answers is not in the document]
mahoneycm: @mejiaj  

**question:** Happy to do this here, but with the addition of #2794 it won't be necessary. 

Do you think I should go ahead and add it anyway?

packages/usa-character-count/src/index.js
```diff
@@ -16,10 +22,10 @@ const STATUS_MESSAGE_SR_ONLY = `.${STATUS_MESSAGE_SR_ONLY_CLASS}`;
 const DEFAULT_STATUS_LABEL = `characters allowed`;
 
 /**
- * Returns the root and message element for an character count input
+ * Returns the root, form group, label, and message elements for an character count input
  *
  * @param {HTMLInputElement|HTMLTextAreaElement} inputEl The character count input element
- * @returns {CharacterCountElements} elements The root and message element.
+ * @returns {CharacterCountElements} elements The root form group, label, and message element.
```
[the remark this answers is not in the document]
mahoneycm: Good catch! Resolved in [1fc6509](https://github.com/uswds/uswds/pull/5908/commits/1fc6509c8cd5f79ba65db7b9274d2893e584d077)

packages/usa-character-count/src/index.js
```diff
@@ -150,6 +162,12 @@ const updateCountMessage = (inputEl) => {
     inputEl.setCustomValidity("");
   }
 
+  inputEl.classList.toggle(INPUT_ERROR_CLASS, isOverLimit);
+  if (labelEl)
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [1fc6509](https://github.com/uswds/uswds/pull/5908/commits/1fc6509c8cd5f79ba65db7b9274d2893e584d077)

packages/usa-character-count/src/index.js
```diff
@@ -150,6 +162,12 @@ const updateCountMessage = (inputEl) => {
...
+    labelEl.classList.toggle(LABEL_ERROR_CLASS, isOverLimit);
+  if (formGroupEl)
```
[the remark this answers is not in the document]
mahoneycm: Resolved in [1fc6509](https://github.com/uswds/uswds/pull/5908/commits/1fc6509c8cd5f79ba65db7b9274d2893e584d077)

packages/usa-character-count/src/index.js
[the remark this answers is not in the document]
mahoneycm: Went ahead and removed the error messages and related unit tests! We can rely on guidance for the proper markup structure

mahoneycm:
>> How do we feel about the necessity for labels or form group elements?
>
> Seems valid. Feels a little odd to check for valid form HTML, but it matches other patterns where we throw errors if the required component markup is missing.
>
> My opinion
> I'd prefer to rely on guidance, but I can also see a case for throwing an error.

This makes sense to me! I went ahead and removed the related error messages and unit tests! We can rely on guidance and example code for proper markup 👍

packages/usa-character-count/src/usa-character-count.stories.js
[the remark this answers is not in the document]
mejiaj: The issue linked seems unrelated. This PR #5999 updates error styles that could address this.

packages/usa-character-count/src/index.js
```diff
@@ -150,6 +162,12 @@ const updateCountMessage = (inputEl) => {
     inputEl.setCustomValidity("");
   }
 
+  inputEl.classList.toggle(INPUT_ERROR_CLASS, isOverLimit);
+  if (labelEl)
+    labelEl.classList.toggle(LABEL_ERROR_CLASS, isOverLimit);
```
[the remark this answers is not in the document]
mejiaj: Addressed in 1fc6509.

packages/usa-character-count/src/index.js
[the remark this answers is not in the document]
mejiaj: Unit tests removed in fe4211882a16f30e80c4559bb970e9b3119dfdf4.

mejiaj (approved):
Confirming that this work displays the character count error state when markup is correct and user goes over limit.

Unit tests and builds pass. No console errors when interacting with the component.

[character-count-5908-2024-09-24 at 11.00.41.webm](https://github.com/user-attachments/assets/406c49be-f54b-4178-8cfe-5d8d7ccea396)

packages/usa-character-count/src/usa-character-count.stories.js
```diff
@@ -23,3 +27,29 @@ export const AriaDisabled = Template.bind({});
...
   disabled_state: "aria-disabled",
 };
+
+// Disabling a11y tests from `npm run test:a11y` because we're demo'ing the failure intentionally.
+export const TestNoLabel = TestNoLabelTemplate.bind({});
+TestNoLabel.parameters = {
+  axe: {
+    mode: "off",
+  },
+};
```
mejiaj, line 37: > [!TIP]
> This test intentionally fails a11y test to view the error state when markup is incorrect/missing.

Followed this guidance to disable a11y tests for those templates. [^1]

[^1]: https://github.com/chanzuckerberg/axe-storybook-testing?tab=readme-ov-file#disabledrules

amyleadem (approved):
Looks good to me! Confirmed that the updated styles have appropriate enhancements that make the error state more visible and consistent with other USWDS styles

Screenshots:
image

thisisdano (approved):
LGTM. Passing it over to @heymatthenry for his take.

heymatthenry (approved):
LGTM! And a great PR, to boot—clear and thoughtful.

PR #5909, opened 2024-05-02
USWDS - Tooltip: Make tooltip dismissible with escape key

### LINKED ISSUE #5901: USWDS - Bug: Tooltip needs "escape" key functionality
### Describe the bug

**Dismissible** (tooltip doesn't meet this criteria)
A [mechanism](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dfn-mechanism) is available to dismiss the additional content without moving pointer hover or keyboard focus, unless the additional content communicates an [input error](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dfn-input-error) or does not obscure or replace other content;
**Hoverable** (tooltip doesn't meet this criteria)
If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;
**Persistent** (tooltip DOES meet this criteria)
The additional content remains visible until the hover or focus trigger is removed, the user dismisses it, or its information is no longer valid.

According to W3C, the tooltip component interaction should be dismissible by hitting the "escape" key. 
https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/
https://www.w3.org/WAI/WCAG22/Techniques/client-side-script/SCR39.html

1. Go to https://designsystem.digital.gov/components/tooltip/
2. keyboard into one of the tooltip button examples
3. hit "escape"
Note how the tooltip remains visible

Additionally, the tooltip does not meet the criteria for "hoverable"
1. Go to https://designsystem.digital.gov/components/tooltip/
2. Hover over the tooltip with a mouse
3. Try to access the tooltip with the mouse. Note that it disappears. According to WCAG it shouldn't.

_If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;_

### Expected Behavior

When you click on the example button and hit 'esc' the tooltip should disappear.

### Screenshots

_No response_

### System setup

PC, Chrome, keyboard

### Additional context

_No response_

### Code of Conduct

amyleadem:
# Summary

**Updated tooltip component behavior to close active tooltips when the `escape` key is pressed.** This allows the component to meet the "dismissible" standard outlined in [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html).

> [!Note]
> This issue is slated to be in the same release as #5919, which also edits the tooltip JavaScript. We should confirm there are no conflicts between the two PRs.

## Breaking change

## Related issue

Closes #5901 

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2660)

## Preview link

[Tooltip component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-tooltip-escape/iframe.html?id=components-tooltip--tooltip&viewMode=story)

> [!important]
> In the [default storybook view](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-tooltip-escape/?path=/story/components-tooltip--tooltip), pressing `escape` will not close a tooltip that is triggered by `mouseover` until one of the trigger buttons receives focus. After that, it should work as expected.  However, this issue does not happen when the Story canvas is [opened in its own tab](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-tooltip-escape/iframe.html?id=components-tooltip--tooltip&viewMode=story). For that reason, I do not consider this to be an issue with the component, but rather Storybook. 

## Problem statement

According to W3C, the tooltip component interaction should be dismissible by hitting the "escape" key. 

## Solution

This PR creates the `handleEscape()` function to hide the tooltip when when the `escape` key is pressed. 

## Testing and review
1. Confirm that active tooltips close when you press `escape`. 
1. Confirm no regressions in behavior.
1. Confirm the code meet standards.

### DIFF
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -358,8 +358,10 @@ const setUpAttributes = (tooltipTrigger) => {
 };
 
 const handleEscape = (event) => {
-  const activeTooltips = document.querySelectorAll(".usa-tooltip__body.is-visible");
-  if ( event.key === 'Escape' || event.key === 'Esc') {
+  const activeTooltips = document.querySelectorAll(
+    ".usa-tooltip__body.is-visible"
+  );
+  if (event.key === "Escape" || event.key === "Esc") {
     activeTooltips.forEach((activeTooltip) => {
       hideToolTip(activeTooltip);
     });
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -373,7 +373,7 @@ const handleEscape = (event) => {
     `.${TOOLTIP_BODY_CLASS}.${VISIBLE_CLASS}`
   );
 
-  if (!activeTooltips) {
+  if (activeTooltips.length === 0) {
     return;
   }
 
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -1,9 +1,11 @@
 // Tooltips
+const keymap = require("receptor/keymap");
 const selectOrMatches = require("../../uswds-core/src/js/utils/select-or-matches");
 const behavior = require("../../uswds-core/src/js/utils/behavior");
 const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
 const isElementInViewport = require("../../uswds-core/src/js/utils/is-in-viewport");
 
+const BODY = "body";
 const TOOLTIP = `.${PREFIX}-tooltip`;
 const TOOLTIP_TRIGGER = `.${PREFIX}-tooltip__trigger`;
 const TOOLTIP_TRIGGER_CLASS = `${PREFIX}-tooltip__trigger`;
@@ -357,21 +359,27 @@ const setUpAttributes = (tooltipTrigger) => {
   return { tooltipBody, position, tooltipContent, wrapper };
 };
 
-
 /**
  * Hide all active tooltips when escape key is pressed.
  * @param event - Keydown event
  */
 
 const handleEscape = (event) => {
+  if (event.key !== "Escape") {
+    return;
+  }
+
   const activeTooltips = document.querySelectorAll(
     ".usa-tooltip__body.is-visible"
   );
-  if (event.key === "Escape" || event.key === "Esc") {
-    activeTooltips.forEach((activeTooltip) => {
-      hideToolTip(activeTooltip);
-    });
+
+  if (!activeTooltips) {
+    return;
   }
+
+  activeTooltips.forEach((activeTooltip) => {
+    hideToolTip(activeTooltip);
+  });
 };
 
 // Setup our function to run on various events
@@ -391,7 +399,6 @@ const tooltip = behavior(
         const { trigger, body } = getTooltipElements(e.target);
 
         showToolTip(body, trigger, trigger.dataset.position);
-        window.addEventListener("keydown", handleEscape);
       },
     },
     "mouseout focusout": {
@@ -399,8 +406,10 @@ const tooltip = behavior(
         const { body } = getTooltipElements(e.target);
 
         hideToolTip(body);
-        window.removeEventListener("keydown", handleEscape);
       },
+    },
+    keydown: {
+      [BODY]: keymap({ Escape: handleEscape }),
     },
   },
   {
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -361,8 +361,7 @@ const setUpAttributes = (tooltipTrigger) => {
 
 const handleEscape = () => {
   const activeTooltips = document.querySelectorAll(".is-set");
-  if ( event.keyCode === 27 ) {
-    console.log("Pressed escape");
+  if ( event.key === 'Escape' ) {
     activeTooltips.forEach((activeTooltip) => {
       hideToolTip (activeTooltip);
     });
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -366,7 +366,7 @@ const setUpAttributes = (tooltipTrigger) => {
 const handleEscape = () => {
   const activeTooltips = selectOrMatches(`.${TOOLTIP_BODY_CLASS}.${SET_CLASS}`);
 
-  if (!activeTooltips) {
+  if (activeTooltips.length === 0) {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


packages/usa-tooltip/src/index.js
```diff
@@ -357,6 +359,24 @@ const setUpAttributes = (tooltipTrigger) => {
   return { tooltipBody, position, tooltipContent, wrapper };
 };
 
+const handleEscape = () => {
+  const activeTooltips = document.querySelectorAll(".is-set");
+  if ( event.keyCode === 27 ) {
```
anselmbradford, line 364: Note that `event.keyCode` is deprecated and should be replaced with `event.key === 'Escape'`

packages/usa-tooltip/src/index.js
```diff
@@ -357,6 +357,15 @@ const setUpAttributes = (tooltipTrigger) => {
   return { tooltipBody, position, tooltipContent, wrapper };
 };
```
amyleadem, line 359: > [!Note]
> Need to add JSDoc comments here

packages/usa-tooltip/src/test/tooltips.spec.js
```diff
@@ -53,6 +80,22 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
 
+    it("tooltip is visible on mouseover", () => {
+      EVENTS.mouseover(tooltipTrigger);
+      assert.strictEqual(tooltipBody.classList.contains("is-set"), true);
+    });
+
+    it("tooltip is hidden on mouseout", () => {
+      EVENTS.mouseout(tooltipTrigger);
+      assert.strictEqual(tooltipBody.classList.contains("is-set"), false);
+    });
```
amyleadem, line 91: > [!note]
> Added some missing tests while I was in there

packages/usa-tooltip/src/test/tooltips.spec.js
```diff
@@ -53,6 +80,22 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
+
+    it("tooltip is hidden on escape keydown", () => {
+      tooltipTrigger.focus();
+      EVENTS.escape(body);
```
amyleadem, line 95: I am struggling to get the escape event to register here. Any ideas?

mejiaj (commented):
Added suggestions. Thanks for creating tests.

packages/usa-tooltip/src/test/tooltips.spec.js
mahoneycm: I've been playing around with this and the issue might have something to do with how we search for `activeTooltips` in the tooltip index.js file.

If I add a console log in lines `376-379`

```js
if (activeTooltips.length === 0) {
  console.log("🟥 no active tooltips");
  return;
}
```
It comes up when running my unit test

```js
  tooltips initialized at document.body
    ✔ trigger is created
    ✔ title attribute on trigger is removed
    ✔ tooltip body is created
    ✔ tooltip is visible on focus
    ✔ tooltip is hidden on blur
    ✔ tooltip is visible on mouseover
    ✔ tooltip is hidden on mouseout
🟥 no active tooltips
    1) tooltip is hidden on escape keydown
    ✔ should not allow for innerHTML of child elements 
```

packages/usa-tooltip/src/test/tooltips.spec.js
amyleadem: Update: I moved these tests to #5919 instead since they felt more appropriate there.

packages/usa-tooltip/src/index.js
```diff
@@ -357,6 +359,29 @@ const setUpAttributes = (tooltipTrigger) => {
...
+ */
+
+const handleEscape = (event) => {
+  if (event.key !== "Escape") {
+    return;
+  }
+
+  const activeTooltips = document.querySelectorAll(
+    `.${TOOLTIP_BODY_CLASS}.${VISIBLE_CLASS}`
+  );
```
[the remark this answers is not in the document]
amyleadem: Good note. I replaced `querySelectorAll` with `selectOrMatches`.

packages/usa-tooltip/src/test/tooltips.spec.js
amyleadem: Update: Replaced `body` with `tooltipTrigger` because we know the unit test knows `tooltipTrigger` exists. Also updated the code in `index.js` to check for the `is-set` class instead of `is-visible` to  determine if there are active tooltips since it looked like the `is-visible` class wasn't being added in the test environment.

amyleadem:
@mejiaj @mahoneycm I updated the code based on your recommendations and our offline discussions. This is ready for re-review. Let me know what you think!

mejiaj (approved):
Looks good. I was able to manually test hover, click, and hitting escape closes the tooltips.

packages/usa-tooltip/src/index.js
mahoneycm, line 1: ## Storybook focus quirk

I was able to confirm that clicking into the iframe storybook preview also allows the escape functionality to work without needing to tab to a trigger button.

I think this is because the second body element within the iframe isn’t capturing the keyboard interactions until the relative "browser focus" is put inside the iframe

mahoneycm (approved):
This looks good to me! 

I did want to flag that it looks like the `escape` event listener is active when there are no active tooltips. Adding a `console.log` within the `handleEscape` shows that the function will trigger on escape.

Is that still a requirement for this functionality? I believe we decided it would be better to add the event listener once rather than every time the triggers are hovered but I wanted to make sure since it was listed in the checklist items. 

## Testing checklist

- [x]  Confirm that active tooltips close when you press `escape`.
- [x]  Confirm no regressions in behavior.
- [x]  Confirm the code meet standards.
- [ ]  Confirm the `escape` event listener is only added when the tooltip is active
    - Escape event listener is triggered anytime escape is pressed
- [x]  Tooltip escape functionality works as expected in sandbox

amyleadem:
@amycole501  and @alex-hull  can you confirm that the tooltip now conforms with the "dismissible" criteria in [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html)?

amycole501 (approved):
@amyleadem yes it works great!! I was able to use the esc key to close the tooltip on all examples.

alex-hull (approved):
Can confirm with Amy C, it should now conform with the dismissible portion of the criteria.

thisisdano (approved):
Thanks all!

PR #5919, opened 2024-05-09
USWDS - Tooltip: Make tooltip content hoverable

### LINKED ISSUE #5918: USWDS - Bug: Tooltip needs "hoverable" functionality
## Summary
[Success Criterion 1.4.13 Content on Hover or Focus](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus) states that for tooltip content must be hoverable:

> Hoverable
> If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;

The tooltip component does not currently meet this criteria.

amyleadem:
# Summary

**Updated the behavior of tooltip to allow hover over tooltip content.** This allows the component to meet the "hoverable" standard outlined in [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html).

> [!Note]
> This issue is slated to be in the same release as #5909, which also edits the tooltip JavaScript. We should confirm there are no conflicts between the two PRs.

## Breaking change

## Related issue

Closes #5918 

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2661)

## Preview link

| Before | After |
|--------|--------|
| [Tooltip](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-tooltip--tooltip) | [Feature preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-tooltip-hoverable/?path=/story/components-tooltip--tooltip) | 

## Problem statement

[Success Criterion 1.4.13 Content on Hover or Focus](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus) states that for tooltip content must be hoverable:

> Hoverable
> If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;

The tooltip component does not currently meet this criteria.

## Solution
Updated the method for detecting how and when the mouse leaves the tooltip. These were the main changes to accomplish this: 
- **Attached the event listener to the `usa-tooltip` wrapper, rather than `usa-tooltip__trigger`.** This added the `usa-tooltip__body` element to the hit target. 
- **Replaced `mouseout` with `mouseleave`.** This allowed us to trigger `hideTooltip` only when the mouse left all elements inside the wrapper, rather than triggering after it left _each_ element. 
- **Removed the `pointer-events: none;` style declaration.** This allowed `mouseleave` to include `usa-tooltip__body` in the hit target. 

## Testing and review
1. Confirm that you move your mouse into the tooltip content without the tooltip closing
1. Confirm that you can move your mouse into the small space between the tooltip and the tooltip trigger without the tooltip closing
1. Confirm that the toolip closes when you leave the `.usa-tooltip` hit boundary
1. Confirm the code meets standards.
1. Confirm the code is compatible with the solution in #5909

> [!Note]
> There is a pre-existing issue with the social media icon tooltips in the [tooltip test story](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-tooltip--test).


packages/usa-tooltip/src/styles/_usa-tooltip.scss
```diff
@@ -49,6 +47,12 @@ $triangle-size: 5px;
     left: 50%;
     margin-left: -$triangle-size;
   }
+
+  &:before {
+    content: "";
+    display: block;
+    position: absolute;
+  }
 }
```
amyleadem, line 56: > [!Note]
> I added these `before` pseudo elements to fill the gap between the tooltip trigger and tooltip. Filling this gap prevents `mouseleave` from triggering when the mouse enters that gap when trying to hover over the tooltip.

packages/usa-tooltip/src/styles/_usa-tooltip.scss
```diff
@@ -28,7 +28,6 @@ $triangle-size: 5px;
   font-size: size("ui", $theme-tooltip-font-size);
   opacity: 0; // Required for recalculating position.
   padding: units(1);
-  pointer-events: none;
```
amyleadem, line 31: > [!Note]
> We needed to remove the `pointer-event` styles here because they were causing the tooltip to be removed from the `usa-tooltip` hit boundary. This caused `mouseleave` to be triggered when hovering over the tooltip.

packages/usa-tooltip/src/index.js
amyleadem, line 1: > [!Note]
> To prevent the tooltip from closing when you move the mouse out of the tooltip trigger, we needed to:
> 1. Target the `usa-tooltip` wrapper instead of the tooltip trigger, and 
> 2. Replace `mouseout` with `mouseleave`. `mouseleave` only triggers when the mouse has exited the element and all its descendents, whereas `mouseout` triggers with each element in the group. See reference below for more details:
>
> "...mouseleave is fired when the pointer has exited the element and all of its descendants, whereas mouseout is fired when the pointer leaves the element or leaves one of the element's descendants (even if the pointer is still within the element)." [^1]

[^1]: https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event

packages/usa-tooltip/src/index.js
```diff
@@ -357,6 +357,12 @@ const setUpAttributes = (tooltipTrigger) => {
   return { tooltipBody, position, tooltipContent, wrapper };
 };
```
amyleadem, line 359: Need to add JSDoc comment

amyleadem:
@mejiaj and @mahoneycm, this still needs JSDoc comments and probably a unit test, but opening this review now so that we can start discussing if there are ways to write this code that better align with standards.

packages/usa-tooltip/src/index.js
```diff
@@ -371,16 +377,15 @@ const tooltip = behavior(
         }
       },
       [TOOLTIP_TRIGGER](e) {
-        const { trigger, body } = getTooltipElements(e.target);
+        const { trigger, body, wrapper } = getTooltipElements(e.target);
 
         showToolTip(body, trigger, trigger.dataset.position);
+        wrapper.addEventListener("mouseleave", closeTooltip);
```
amyleadem, line 383: I was not able to get `mouseleave` to work using the standard behavior method. so instead, I triggered the event listener only when the tooltip is active. Please let me know if you see a better way.

mejiaj: We have to use `mouseout` for it to work with behaviors.

Not opposed to `mouseleave`, but we should add it once on `init()` and remove it on teardown.

---

Wrapper and the `TOOLTIP` (line 370) share the same class, so you _could_ add the show/hide behavior in there instead of relying on trigger.

mejiaj (changes requested):
Thanks for taking this on. 

I hope you don't mind I've edited the PR description. I've simplified the summary for clarification (now its clear we're adding support to the tooltip body). As well as adding a table with preview links to the before/after state for easier testing.

---

This new behavior means we'll have to re-think our current approach to showing/hiding the tooltip. Right now it's listening for any mouseover on the trigger. We should consider either:

1. Toggling visibility on `TOOLTIP`, instead of `TOOLTIP_TRIGGER`.
2. Using `mouseleave`, but ensuring it's only added once and removed on teardown.

packages/usa-tooltip/src/test/test-patterns/test-usa-tooltip-no-wrapper.twig
amyleadem, line 1: > [!Note]
> I added this test because I discovered that my original fixes had the mouse events targeting the wrong tooltips when there were no wrappers around the tooltips. This might not be needed moving forward, but wanted to push it up to confirm the fix is still working. We can remove before merge if this test feels confusing.

mahoneycm (changes requested):
Getting close! A couple small changes I wanted to run by you to make sure we're covering all bases.

packages/usa-tooltip/src/index.js
```diff
@@ -388,6 +388,16 @@ const tooltip = behavior(
...
+
+        const { body, wrapper } = getTooltipElements(tooltipTrigger);
+        wrapper.addEventListener("mouseleave", () => {
+          hideToolTip(body);
+        });
+      });
+    },
+    teardown(root) {
+      selectOrMatches(TOOLTIP, root).forEach((tooltipWrapper) => {
+        tooltipWrapper.removeEventListener("mouseleave", hideToolTip);
```
mahoneycm, line 400: I'm not sure if this teardown is removing the event listener from the right element. It looks like the `forEach` is returning the tooltip trigger based on init.

 Should we rename `tooltipWrapper` to `tooltipTrigger` to match init, then get the element using `getTooltipElements`?

```suggestion
    teardown(root) {
      selectOrMatches(TOOLTIP, root).forEach((tooltipTrigger) => {
        const { wrapper } = getTooltipElements(toltipTrigger)
        wrapper.removeEventListener("mouseleave", hideToolTip);
```

amyleadem: I changed it to `tooltipWrapper` here because (at least in my understanding) `TOOLTIP` has already been transformed into its "wrapper" state now that it has been initialized. So when it looks for `TOOLTIP` here, I believe it is looking for the initialized wrapper version of `usa-tooltip` rather than the original trigger version of `usa-tooltip`. Let me know if that doesn't make sense!

mahoneycm: Ahh yes! Ok that makes sense now considering the lifecycle of the component. Great observation! I'll go ahead and resolve this comment 👍

packages/usa-tooltip/src/test/tooltips.spec.js
```diff
@@ -10,17 +10,45 @@ const tests = [
...
 
+const EVENTS = {
+  escape(el) {
+    const escapeKeyEvent = new KeyboardEvent("keydown", {
+      key: "Escape",
+      bubbles: true,
+    });
+
+    el.dispatchEvent(escapeKeyEvent);
+  },
```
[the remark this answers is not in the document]
amyleadem: I did originally, but realize it is confusing. I removed it in [04124a3](https://github.com/uswds/uswds/pull/5919/commits/04124a3016b1d97d1f553f0bcaddf43edf30a7f2)

packages/usa-tooltip/src/styles/_usa-tooltip.scss
```diff
@@ -66,6 +70,15 @@ $triangle-size: 5px;
   opacity: 1;
 }
 
+.usa-tooltip__body--top {
+  &:before {
+    height: $triangle-size;
+    left: 0;
+    right: 0;
+    top: 100%;
+  }
```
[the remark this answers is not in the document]
amyleadem: I tried this out and it it didn't seem to work for left and right placement. The mixin for this would need to be a bit more involved, so I am torn on the benefit here. I lean towards leaving the code as is. @mejiaj do you have a preference?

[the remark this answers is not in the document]
mejiaj: A mixin would help improve readability. Currently there's no sign of what this pseudoelement is for, but it could be an enhancement. Unless you want to give it another shot.

packages/usa-tooltip/src/styles/_usa-tooltip.scss
mejiaj, line 1: Can we use `::before` and `::after` with double colon instead of single?

packages/usa-tooltip/src/index.js
```diff
@@ -388,6 +388,16 @@ const tooltip = behavior(
     init(root) {
       selectOrMatches(TOOLTIP, root).forEach((tooltipTrigger) => {
         setUpAttributes(tooltipTrigger);
+
+        const { body, wrapper } = getTooltipElements(tooltipTrigger);
+        wrapper.addEventListener("mouseleave", () => {
+          hideToolTip(body);
+        });
```
mejiaj, line 395: Follows the same pattern in teardown line `400`.

```suggestion
        wrapper.addEventListener("mouseleave", () => hideToolTip(body));
```

mejiaj (changes requested):
Minor changes request to pseudoelement format & suggestion for JS. Otherwise LGTM.

packages/usa-tooltip/src/index.js
amyleadem: Updated in [d4ca8c3](https://github.com/uswds/uswds/pull/5919/commits/d4ca8c3f3f85bb9667a667ff84290aa2d7021346)

packages/usa-tooltip/src/styles/_usa-tooltip.scss
amyleadem: Updated in [39534b5](https://github.com/uswds/uswds/pull/5919/commits/39534b5c64d0768f7a92f77249c68df7c3b347ab)

amyleadem:
@amycole501  and @alex-hull  can you confirm that the tooltip now conforms with the "hoverable" criteria in [WCAG 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html)?

amycole501 (approved):
I tested the hoverable action and yes, it is now able to be hovered over with a mouse and then when I move the mouse over the tooltip it does NOT disappear.

packages/usa-tooltip/src/styles/_usa-tooltip.scss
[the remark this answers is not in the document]
amyleadem: I added a comment in  [39534b5](https://github.com/uswds/uswds/pull/5919/commits/39534b5c64d0768f7a92f77249c68df7c3b347ab) to help improve clarity. If you think the comment is insufficient, I can work on the mixin on Monday.

amyleadem:
Sorry @amycole501, I accidentally re-requested your review, which cleared the green approval check 🤦

amycole501 (approved):

packages/usa-tooltip/src/styles/_usa-tooltip.scss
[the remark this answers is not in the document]
amyleadem: Update: @mejiaj created a mixin for this in #5933 (Thanks James!). It was merged into this PR in commit https://github.com/uswds/uswds/pull/5919/commits/babcc2a3e74e22b7793325bd07db512f79efadc1. Going to resolve this comment thread since the task is complete.

mahoneycm (approved):
Great work here! Functionality is working like a charm

## Testing checklist

- [x]  Confirm that you move your mouse into the tooltip content without the tooltip closing
- [x]  Confirm that you can move your mouse into the small space between the tooltip and the tooltip trigger without the tooltip closing
- [x]  Confirm that the toolip closes when you leave the `.usa-tooltip` hit boundary
- [x]  Confirm the code meets standards.
- [x]  Confirm the code is compatible with the solution in [USWDS - Tooltip: Make tooltip dismissible with escape key #5909](https://github.com/uswds/uswds/pull/5909)

mejiaj (approved):
No issues found. Tested in Safari, Firefox, and Chromium.

Thanks for adding the tests.

thisisdano (approved):
Very nice. Both features work together as expected

PR #5920, opened 2024-05-10
USWDS - Tooltip: Add escape to behaviors

mahoneycm:
# Summary

Moves new tooltipEscape function to behaviors map

## Related issue

#5901 

## Related pull requests

Continuation of #5909 

## Preview link

[Tooltip preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-tooltip-escape-behaviors/iframe.html?id=components-tooltip--tooltip&viewMode=story)

## Problem statement

## Solution

## Major changes

## Testing and review

### DIFF
--- a/packages/usa-tooltip/src/index.js
+++ b/packages/usa-tooltip/src/index.js
@@ -3,7 +3,9 @@ const selectOrMatches = require("../../uswds-core/src/js/utils/select-or-matches
 const behavior = require("../../uswds-core/src/js/utils/behavior");
 const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
 const isElementInViewport = require("../../uswds-core/src/js/utils/is-in-viewport");
+const keymap = require("receptor/keymap");
 
+const BODY = "body";
 const TOOLTIP = `.${PREFIX}-tooltip`;
 const TOOLTIP_TRIGGER = `.${PREFIX}-tooltip__trigger`;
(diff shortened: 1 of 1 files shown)


packages/usa-tooltip/src/index.js
```diff
@@ -394,6 +393,9 @@ const tooltip = behavior(
         window.removeEventListener("keydown", handleEscape);
       },
     },
+      keydown: {
+        [BODY]: keymap({ Escape: handleEscape})
+      },
```
amyleadem, line 398: Thanks for making this work with our more standard code patterns, @mahoneycm. 

One thing I noticed with this method is that the page registers `escape` presses even when the tooltip is inactive. What I liked about manually adding and removing the `eventlistener` on mouse over/out or focus in/out is that the window only would register the key presses when the tooltip is active. I feel like this would limit unintended interaction with other components that also rely on the `escape` keypress. Curious what you think!

mahoneycm: That's a great point! To be honest I'm not too familiar with the benefit/loss of having a persistent listener but I understand the principle of only listening for the keypress while the tooltip is active!

I considered having an early return in `handleEscape` if there were no active tooltips but that doesn't change the fact that the listener will be there the entire time.

Happy to move forward with the other solution if we feel like there is additional value there!

mahoneycm:
Closing in favor of #5909

PR #5933, opened 2024-05-20
USWDS - Tooltip: Create a mixin for the tooltip spacer

mejiaj:
# Summary

Created mixin for tooltip spacer to avoid duplicate code.

## Breaking change

## Related issue

#5918

## Related pull requests

Related to #5919.

## Preview link

Preview link: [Tooltip →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-tooltip-hoverable/?path=/story/components-tooltip--tooltip) 

## Problem statement

The CSS for creating the tooltip spacer was repetitive.

## Solution

Created a mixin to manage this in one place.

## Testing and review

Tooltips should create a spacer element between trigger and body that keeps the content visible.

### DIFF
--- a/packages/usa-tooltip/src/styles/_usa-tooltip.scss
+++ b/packages/usa-tooltip/src/styles/_usa-tooltip.scss
@@ -1,5 +1,36 @@
 @use "uswds-core" as *;
 
+/// Create a spacer to increase target area for tooltip triangle.
+///
+/// @param {String} $direction - The direction of the tooltip; can be top, bottom, left, right.
+/// @param {Length} $triangle-size [5px] - The size of the triangle.
+///
+/// @example
+/// @include tooltip-spacer("top");
+///
+/// @output
+/// .usa-tooltip__body--top::before {
+///    top: 100%;
+///    height: 5px;
+///    left: 0;
+///    right: 0;
+///  }
+@mixin tooltip-spacer($direction) {
+  &::before {
+    #{$direction}: 100%;
+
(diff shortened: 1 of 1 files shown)


packages/usa-tooltip/src/styles/_usa-tooltip.scss
```diff
@@ -1,5 +1,36 @@
 @use "uswds-core" as *;
 
+/// Create a spacer to increase target area for tooltip triangle.
+///
+/// @param {String} $direction - The direction of the tooltip; can be top, bottom, left, right.
+/// @param {Length} $triangle-size [5px] - The size of the triangle.
```
mahoneycm, line 6: Is it correct to define `$triangle-size` as a parameter here even though it's not being passed into the mixin?

mejiaj: That's a good point. I don't think it's necessary since triangle size is already customizable.

I've moved the variable above the mixin in 04dcd6be4f to make it clearer that this mixin relies on that.

amyleadem (approved):
Thanks for taking this on @mejiaj! I confirmed that this mixin positions the spacers correctly (See screenshot below. Note: spacers are highlighted in orange only for demonstration purposes). The code looks great and is easy to understand. 

image

mahoneycm (approved):
Working like a charm!

PR #5935, opened 2024-05-21
USWDS - Text Input: Include width variant styles in usa-input package

### LINKED ISSUE #5312: USWDS - Feature: Move usa-input--[width] classes to the usa-input module

We have a need for applying the `.usa-input--[width]` class to a standalone web component. I had hoped that those classes were included with the `usa-input` module but they are instead in `usa-form` nested under the `.usa-form` class.

https://github.com/uswds/uswds/blob/develop/packages/usa-form/src/styles/_usa-form.scss

Being an individual component, I did not think it appropriate to wrap it with `.usa-form` just so that we could access the width classes. It seems like `.usa-form` is intended to be used on a wrapping element around an entire form.

It would be ideal in our use case to have the `usa-input--[width]` available as part of the `usa-input` module. As it is now, we would need to copy those classes into our component library as a mixin which isn't ideal because we'd like to continue to pull them directly from the USWDS library instead.

_No response_

### Additional context

_No response_

### Code of Conduct

aduth:
# Summary

**Add missing styles for Text Input component when using optimized installation.** Width variants for the Text Input component were previously not included when using the `usa-input` Sass package.

## Breaking change

This would only theoretically be a breaking change if someone was importing `usa-form` and not `usa-input`, yet expected to be able to use `usa-input--[width]` classes. This seems exceedingly unlikely, since if someone is using `usa-input--[width]` classes, presumably they're also using `usa-input` classes.

## Related issue

Fixes #5312

## Problem statement

As a developer, I expect that if I include the package listed on a component's documentation (e.g. [Text Input](https://designsystem.digital.gov/components/text-input/)), all of the usage options documented for that component are available, so that I'm not confused by styles not existing despite being documented, or be forced to include more packages than I'd otherwise need for my optimized installation.

## Solution

Moves `usa-input--[width]` styles from `usa-form` to `usa-input` package.


aduth:
> As a developer, I expect that if I include the package listed on a component's documentation (e.g. [Text Input](https://designsystem.digital.gov/components/text-input/)), all of the usage options documented for that component are available, so that I'm not confused by styles not existing despite being documented, or be forced to include more packages than I'd otherwise need for my optimized installation.

As a real-world use-case, we don't include `usa-form` styles in our project, because we find that they're too opinionated (specifically the [unconfigurable `max-width` styling](https://github.com/uswds/uswds/blob/6d1471e5a4ad8e6a164772f47db5edf9586a95c4/packages/usa-form/src/styles/_usa-form.scss#L11-L13)), but we had a need for width-constrained inputs and use `usa-input` styles. We were surprised and confused that the styles weren't being applied as expected, and it took a non-insignificant amount of debugging to figure out what was going on. Eventually we decided to [implement ad-hoc styles](https://github.com/18F/identity-idp/pull/10643) for a width-constrained input, which isn't ideal.

Writing out this use-case also has me thinking that there's room to improve `usa-form` to either make it less opinionated or at least more configurable, so that we would feel more comfortable using those styles.

amyleadem (changes requested):
Thanks for this submission, @aduth. This is generally looking good, but I noticed one issue where these width classes no longer work when used inside a `.usa-form` parent. I've added details in the comment below. 

I checked the following items:
- [x] Confirm that the approach to move `.usa-input--[width]` classes to the `usa-input` package makes sense
- [x] Confirm the `usa-input--[width]` classes work as expected on `usa-input` and `usa-textarea` elements when the `usa-form` package is not present
- [x]  Confirm that this change to `usa-input--[width]` classes does not negatively impact other components when `usa-form` package is not present
    - Note: I tested adding `usa-input—[width]` classes to the form input components in `develop`. Below I have listed the components that did successfully receive the `usa-input--[width]` classes in `develop` and confirmed that they still work in this branch when the `usa-form` package is absent:
        - [x]  Character count
        - [x]  Input mask
        - [x]  Input prefix/suffix
        - [x]  Input
        - [x]  Textarea
- [ ] Confirm no regressions when using `.usa-input--[width]` classes on `.usa-input` and `.usa-textarea` elements placed inside a`.usa-form` parent
    - See details in the comment below
- [x]  Confirm that the sizes for the `usa-input--[width]` classes match the sizes in `usa-form.scss`

Please let me know if you have any questions.

aduth:
That's a good catch @amyleadem ! I'll plan to patch that up.

amyleadem:
Closing this PR in favor of #6232.

Thanks for your work on this @aduth. If you have a moment, we'd appreciate it if you could take a look at PR #6232 to make sure the solution will still work for you.

PR #5970, opened 2024-07-02
USWDS - Pagination: Updated text-decoration value to better show pagination links

### LINKED ISSUE #5939: USWDS - Pagination: Borders around page number have poor color contrast
## Description

On the Pagination Page ( https://designsystem.digital.gov/components/pagination/ ), the borders around the page number have poor color contrast. The light gray ( #d1d1d1 ) on the white background in your example has a color contrast ratio of 1.5:1 or higher. I would think for those with poor vision, low vision, etc., it would be helpful to them to see how large the click or tap area is for the link. I would make the border color be at least #767676, which makes the color contrast ratio 4.5:1.

If you make changes here, you probably need to make the same changes to the code people can download and use.

RachelCorsino:
# Summary

**Added text underline styles to pagination links.** Pagination links now have the same visual indicators as other USWDS text links. This should improve accessibility for the component.

## Breaking change

## Related issue

Closes #5939 

## Related pull requests
[Changelog: Text-Decoration Style Fix for Pagination Component Accessibility #2793](https://github.com/uswds/uswds-site/pull/2793)

## Preview link

[Pagination Component Page](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/rc-update-pagination-border-color/?path=/story/components-pagination--default)

## Problem statement

Pagination links do not provide accessible link styles. Pagination borders around the page number have poor color contrast. The light gray ( #d1d1d1 ) on the white background in the example has a color contrast ratio of 1.5:1. We should provide a clear visual indication that the pagination list items are links.

## Solution

Removed `text-decoration: none` so that the "buttons" are more clearly formatted as links.

## Testing and review

Follow these steps:
1. Visit pagination preview page
2. Ensure pagination "buttons" are showing up as links

- [X] Confirm that this code follows the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [X] Run `git pull origin [base branch]` to pull in the most recent updates from your base and check for merge conflicts. (Often, the base branch is `develop`).
- [X] Run `npm run prettier:sass` to format any Sass updates.

### DIFF
--- a/packages/usa-pagination/src/styles/_usa-pagination.scss
+++ b/packages/usa-pagination/src/styles/_usa-pagination.scss
@@ -119,7 +119,7 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
 
 .usa-pagination__button {
   align-items: center;
-  border-color: color.adjust(color($pagination-current-color), $alpha: -0.1);
+  border-color: color.adjust(color($pagination-current-color), $alpha: -0.8);
   border-radius: radius($theme-pagination-button-border-radius);
   border-style: solid;
   border-width: units($theme-pagination-button-border-width);
@@ -127,7 +127,7 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
   display: inline-flex;
   justify-content: center;
   padding: units($pagination-margin-padding);
-  text-decoration: none;
+  text-decoration: underline;
   width: 100%;
 
   &:hover,
--- a/packages/usa-pagination/src/styles/_usa-pagination.scss
+++ b/packages/usa-pagination/src/styles/_usa-pagination.scss
@@ -98,7 +98,6 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
   align-items: center;
   color: color($pagination-link-token);
   display: inline-flex;
-  text-decoration: none;
 
   &[disabled] {
     opacity: 0.4 !important;
@@ -127,7 +126,6 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
   display: inline-flex;
   justify-content: center;
   padding: units($pagination-margin-padding);
-  text-decoration: underline;
   width: 100%;
 
   &:hover,
--- a/packages/usa-pagination/src/styles/_usa-pagination.scss
+++ b/packages/usa-pagination/src/styles/_usa-pagination.scss
@@ -119,7 +119,7 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
 
 .usa-pagination__button {
   align-items: center;
-  border-color: color.adjust(color($pagination-current-color), $alpha: 0.6);
+  border-color: color.adjust(color($pagination-current-color), $alpha: -0.1);
   border-radius: radius($theme-pagination-button-border-radius);
   border-style: solid;
   border-width: units($theme-pagination-button-border-width);
--- a/packages/usa-pagination/src/styles/_usa-pagination.scss
+++ b/packages/usa-pagination/src/styles/_usa-pagination.scss
@@ -108,7 +108,6 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
   &:focus,
   &:active {
     color: color($pagination-hover-token);
-    text-decoration: underline;
   }
 
   &:visited {
@@ -160,7 +159,6 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
   &:active {
     background-color: color($pagination-current-color);
     color: color($text-color);
-    text-decoration: none;
 
     @media (forced-colors: active) {
       color: buttontext;
--- a/packages/usa-pagination/src/styles/_usa-pagination.scss
+++ b/packages/usa-pagination/src/styles/_usa-pagination.scss
@@ -119,7 +119,7 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
 
 .usa-pagination__button {
   align-items: center;
-  border-color: color.adjust(color($pagination-current-color), $alpha: -0.8);
+  border-color: color.adjust(color($pagination-current-color), $alpha: 0.6);
   border-radius: radius($theme-pagination-button-border-radius);
   border-style: solid;
   border-width: units($theme-pagination-button-border-width);


cathybaptista (commented):
Hi, Rachel

I tested this, and your change does mean a color contrast of 5:1 versus 1:1 so I guess this passes? I'm not an alpha pro.  Just eyeballing it, the change is much darker. I'll approve it if someone with a little more experience with contrast ratios can just verify that the new color contrast works. :)

packages/usa-pagination/src/styles/_usa-pagination.scss
```diff
@@ -119,7 +119,7 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
 
 .usa-pagination__button {
   align-items: center;
-  border-color: color.adjust(color($pagination-current-color), $alpha: -0.8);
+  border-color: color.adjust(color($pagination-current-color), $alpha: 0.6);
```
mahoneycm, line 122: **suggestion (blocking):** Looking into [sass:color adjust](https://sass-lang.com/documentation/modules/color/), this sass function increases or decreases the property of the color by fixed amounts.

By setting a positive value here instead of a negative value, we're increasing the alpha (opacity) by 60%. Since the color that's being passed in is already fully opaque, passing in a positive value does not affect the opacity. The result is a fully opaque border using the `$pagination-current-color`

Instead, I would recommend searching for a _negative_ value that will maintain the original behavior of implementing a slightly transparent border color that will pass the AA non-text contrast minimum value of `3:1`

> [!TIP]
> Once you find a value that works for the default background color, try looking into how `$pagination-current-color` is set and how you might be able to test this feature with different settings!

mahoneycm (changes requested):
You're off to a good start here @RachelCorsino ! Left a change request with some additional context on the color adjust function and additional pitfalls to look out for.

Feel free to ping me if you want any additional help!

packages/usa-pagination/src/styles/_usa-pagination.scss
amyleadem: I wonder if we should move away from opacity shifts here since we have determined that contrast is important here for accessibility. Should we move away from alpha transformations in `color.adjust` and instead use an explicit token here (maybe through [next-token() ](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/functions/color/next-token.scss) or some other method). I'll ask the question in Teams too in case we want to discuss this more there.

mejiaj: Agreed — If LOE is minor we should look into using a theme token.

If a theme token is high LOE, I'm good with this solution. Moving to new functionality, like sass color adjust, could increase scope. We should also start relying less on SASS functions if we can avoid it.

mahoneycm (commented):
Made a suggestion for a slightly more noticeable change though the suggestions still cause the the standard border to be quite a bit darker. I'm wondering if we want to improve the overall logic to maintain something closer to the current styles and offering a fallback instead. 

It might be good to check in with the rest of the team to get a temperature check!

packages/usa-pagination/src/styles/_usa-pagination.scss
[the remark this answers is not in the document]
mejiaj: @mahoneycm, @thisisdano made a recommendation in this [comment](https://github.com/uswds/uswds/issues/5939#issuecomment-2237137797). 

image


This adds immediate clarity without increasing scope. What do you think?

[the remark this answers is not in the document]
mahoneycm: I think that's a great idea! I think you're right about adding immediate clarity to the links. I still think this issue might be a priority though because the UI elements do not meet [WCAG Criteria 1.4.11: Non-text Contrast (Level AA)](Non-text Contrast (Level AA))

This would appear as a failure on our accessibility test pages.

packages/usa-pagination/src/styles/_usa-pagination.scss
```diff
@@ -119,7 +119,7 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
 
 .usa-pagination__button {
   align-items: center;
-  border-color: color.adjust(color($pagination-current-color), $alpha: -0.8);
+  border-color: color.adjust(color($pagination-current-color), $alpha: -0.1);
```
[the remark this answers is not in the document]
mahoneycm: Resolving this conversation since we're taking a different approach

packages/usa-pagination/src/styles/_usa-pagination.scss
```diff
@@ -127,7 +127,7 @@ $pagination-hover-token: list.nth($pagination-link-tokens, 2);
   display: inline-flex;
   justify-content: center;
   padding: units($pagination-margin-padding);
-  text-decoration: none;
+  text-decoration: underline;
   width: 100%;
```
amyleadem, line 132: **issue:** This is my mistake: I advised that we add `text-decoration: underline`, but we can probably just remove `text-decoration` styles completely from the element since it has baseline underline styles. Can you remove the `text-decoration` style declaration here?

packages/usa-pagination/src/styles/_usa-pagination.scss
amyleadem, line 1: **issue:** Can you make the same update to the styles for `.usa-pagination__link` that you did for `.usa-pagination__button`? This should add the text underline to the "Previous" and "Next" links. These should elements have some visual indication for interactivity that the other page links get.

amyleadem (changes requested):
Thanks for making the quick update, @RachelCorsino. I added a couple of suggestions below. 

Also, can you update the PR title and description to reflect the change in solution approach?

mahoneycm (changes requested):
Looks great! This new approach is simple and effective. One request to update the PR description to match this approach and I'll give it the green ✅

**issue (blocking):** The PR description doesn't match this new approach.

**suggestion:** Update PR title, description, and testing instructions to match new text decoration approach.

mejiaj:
@mahoneycm @amyleadem please move to FFR once both have approved.

RachelCorsino:
PR description has been updated to account for `text-decoration` rather than color change.

amyleadem (changes requested):
Thanks @RachelCorsino.

I noticed there are two other `text-decoration` style rules in `usa-pagination.scss` that we should be able to safely remove. Can you check that it is safe to remove those style rules and if so, remove them?

Before approval, we'll also need to attach a changelog PR for this update. (Let me know if you need help creating that).

amyleadem (approved):
Looks good to me!

One last update: Can you update the PR description to be a bit clearer? It might seem nitpicky but this description will be pulled into the release notes and will also be used for future reference as part of the component's history. Maybe something like this?

> **Added text underline styles to pagination links.** Pagination links now have the same visual indicators as other USWDS text links. This should improve accessibility for the component. 

Can you also update the problem statement since it doesn't quite capture the root issue? Maybe something like this:

> **Pagination links do not provide accessible link styles.** Pagination borders around the page number have poor color contrast. The light gray ( #d1d1d1 ) on the white background in the example has a color contrast ratio of 1.5:1. We should provide a clear visual indication that the pagination list items are links."

Feel free to edit these suggestions for clarity.

amyleadem (approved):
LGTM! Thanks @RachelCorsino

mahoneycm (approved):
LGTM! Thanks @RachelCorsino !

thisisdano (approved):
A nice, simple, effective improvement

PR #5999, opened 2024-07-29
USWDS - Forms: Enhance form group error border in desktop widths

### LINKED ISSUE #5998: USWDS - Forms: Form group error left border only viewable on wide widths
# Description

The red border added to `usa-form-group--error` divs has a negative margin of `-2.5` units. This is equal to the border size + its padding so it causes the margin to be pushed off screen unless the user has added additional padding to its parent element.

This error border helps to signify the form's error state and should be maintained in desktop views.

1. Visit form error state storybook preview
2. Resize the window from horizontally
3. In smaller views, the error border will be visible
4. In wider views, the error border will be pushed out of view.

### Related code

https://github.com/uswds/uswds/blob/df175b5916fb5cca5cb380f918fd5e1b1a85018e/src/patterns/components/usa-form-group/_usa-form-group.scss#L1-L11

mahoneycm:
# Summary

**Improved form group error syles.** Form group error border width and relevant spacing have been reduced.

As discussed in [dev sync](https://docs.google.com/document/d/172TU5BVS9kGZQDuPcAhGoZjaTd2ngzcTayq4WRRTBUs/edit#heading=h.3exjgbcdn2d4) (🔒) and this [internal slack thread](https://gsa-tts.slack.com/archives/C050HRGN7/p1722263917352969) (🔒)

## Breaking change

## Related issue

Closes #5998 

## Related pull requests

[Changelog →](https://github.com/uswds/uswds-site/pull/2773)

## Preview link

[Form group error preview →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-border-margin/iframe.html?args=&id=patterns-forms--test-error-form-elements)

[Sandbox demo page →](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/cm-test-5999/pages/prose/)

## Problem statement

In desktop widths, the form group error border uses negative margin and is pushed out of the parent container.

## Solution

After discussion, we opted to keep the original breakpoint to reduce content shift in desktop views, and reduce the size of the border, the padding between the border and the content, as well as the negative margin that pushes it left.

## Major changes

- Form group error styles are now viewable in storybook previews in desktop widths
- Form group error border width is halved
- Padding between border and content is reduced

**Note:** Desktop content does not shift when form group error style are added, mobile content does. This matches the current behavior of develop.

## Testing and review

1. Visit the form group error storybook preview.
2. Resize the window and confirm the left side error border is viewable in all widths.
3. In desktop view, toggle error state control and confirm content does not shift.
4. Confirm visual change is an enhancement and not a regression.

### Screenshots

Screenshots showing the placement of content and border relative to div and margin

| | Before | After |
|--------|--------|--------|
| Desktop | Develop-desktop | Fix-desktop |
| Mobile | Develop-mobile | Fix-mobile | 
| Demo page (Content has `2rem` of margin from `grid-layout`) | Develop-demo-desktop | Fix-demo-desktop |

### DIFF
--- a/packages/usa-form-group/src/styles/_usa-form-group.scss
+++ b/packages/usa-form-group/src/styles/_usa-form-group.scss
@@ -11,7 +11,13 @@
 
 // Block input elements
 .usa-form-group--error {
-  @include u-border-left(0.5, "error-dark");
-  padding-left: units(2);
   position: relative;
+
+  &::before {
+    @include u-border-left(0.5, "error-dark");
+    content: "";
+    height: 100%;
+    left: units("neg-105");
+    position: absolute;
+  }
 }
--- a/packages/usa-form-group/src/styles/_usa-form-group.scss
+++ b/packages/usa-form-group/src/styles/_usa-form-group.scss
@@ -1,10 +1,9 @@
 @use "uswds-core" as *;
 
-$form-error-border-width: 2px; 
+$form-error-border-width: 2px;
 $form-error-border-padding: units(1);
-$form-error-negative-margin: -( 
-  $form-error-border-width + rem-to-px($form-error-border-padding)
-);
+$form-error-negative-margin: -($form-error-border-width +
+      rem-to-px($form-error-border-padding));
 
 .usa-form-group {
   margin-top: units(3);
--- a/packages/usa-form-group/src/styles/_usa-form-group.scss
+++ b/packages/usa-form-group/src/styles/_usa-form-group.scss
@@ -11,13 +11,11 @@
 
 // Block input elements
 .usa-form-group--error {
+  @include u-border-left(2px, "error-dark");
+  padding-left: units(1);
   position: relative;
 
-  &::before {
-    @include u-border-left(0.5, "error-dark");
-    content: "";
-    height: 100%;
-    left: units("neg-105");
-    position: absolute;
+  @include at-media("desktop") {
+    margin-left: units("neg-2px");
   }
 }
--- a/packages/usa-form-group/src/styles/_usa-form-group.scss
+++ b/packages/usa-form-group/src/styles/_usa-form-group.scss
@@ -1,5 +1,11 @@
 @use "uswds-core" as *;
 
+$form-error-border-width: 2px; 
+$form-error-border-padding: units(1);
+$form-error-negative-margin: -( 
+  $form-error-border-width + rem-to-px($form-error-border-padding)
+);
+
 .usa-form-group {
   margin-top: units(3);
 
@@ -11,11 +17,11 @@
 
 // Block input elements
 .usa-form-group--error {
-  @include u-border-left(2px, "error-dark");
-  padding-left: units(1);
+  @include u-border-left($form-error-border-width, "error-dark");
+  padding-left: $form-error-border-padding;
   position: relative;
 
   @include at-media("desktop") {
-    margin-left: -10px;
+    margin-left: $form-error-negative-margin;
   }
 }
--- a/packages/usa-form-group/src/styles/_usa-form-group.scss
+++ b/packages/usa-form-group/src/styles/_usa-form-group.scss
@@ -14,8 +14,4 @@
   @include u-border-left(0.5, "error-dark");
   padding-left: units(2);
   position: relative;
-
-  @include at-media("desktop") {
-    margin-left: units(-2.5);
-  }
 }
--- a/packages/usa-form-group/src/styles/_usa-form-group.scss
+++ b/packages/usa-form-group/src/styles/_usa-form-group.scss
@@ -16,6 +16,6 @@
   position: relative;
 
   @include at-media("desktop") {
-    margin-left: units(-1);
+    margin-left: -10px;
   }
 }
--- a/packages/usa-form-group/src/styles/_usa-form-group.scss
+++ b/packages/usa-form-group/src/styles/_usa-form-group.scss
@@ -16,6 +16,6 @@
   position: relative;
 
   @include at-media("desktop") {
-    margin-left: units("neg-2px");
+    margin-left: units(-1);
   }
 }


mahoneycm:
@cathybaptista Thank you for flagging!

This stems from the `usa-form` max-width styles. Once the viewport is beyond the `mobile-lg` token width, the form gets a max width of our `mobile` token (which by default is 20rem). Since the max width is lower than the breakpoint that enables it, you see the snap back to the `mobile` width.

```scss
.usa-form {
  ...
  @include at-media("mobile-lg") {
    max-width: units("mobile");
  }
}
```

This behavior is outside of the scope of this work and can be viewed without the error states enabled in any form that uses the `usa-form` class (without the `usa-form--large` class), such as the [Reset Password](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/patterns-forms--reset-password) pattern on the `develop` branch!

Because of this, we shouldn't have to worry about it as being part of this work

packages/usa-form-group/src/styles/_usa-form-group.scss
```diff
@@ -11,11 +11,13 @@
 
 // Block input elements
 .usa-form-group--error {
-  @include u-border-left(0.5, "error-dark");
-  padding-left: units(2);
```
mejiaj, line 15: **issue (non-blocking)**: The lack of padding is noticeable on small screens.

| Before | After |
| :----- | :---- |
| image | image | 

Side-by-side for comparison before/after.
image

packages/usa-form-group/src/styles/_usa-form-group.scss
```diff
@@ -11,11 +11,13 @@
...
   position: relative;
 
-  @include at-media("desktop") {
-    margin-left: units(-2.5);
+  &::before {
+    @include u-border-left(0.5, "error-dark");
+    content: "";
+    height: 100%;
+    left: units("neg-105");
```
mejiaj, line 20: **question:** This brings the border much closer to the content, was that intentional?

mejiaj (commented):
No issues found in Safari 17.6, Chromium 128, or Firefox 130.

Just a minor question and non-blocking issue. I can approve PR once resolved.

packages/usa-form-group/src/styles/_usa-form-group.scss
mahoneycm: Yeah, removing the `units(-2.5)` of margin prevents the border from being pushed out of the parent div. This is what was causing the border to disappear from our preview in desktop widths.

packages/usa-form-group/src/styles/_usa-form-group.scss
mahoneycm: I agree that this is a small amount of padding... the benefit here is that there is no content shift as a result of the form entering the error state. This was a cause for concern with Character Count in #5908, where the error state is added during user input.

We should consider the pros/cons of the reduction in content shift vs. the lack the padding between the error border and the input.

mahoneycm: For reference, this is the related [internal slack thread](https://gsa-tts.slack.com/archives/C050HRGN7/p1722263917352969) (🔒) where we discussed this approach

mejiaj (approved):
Thanks for those answers, @mahoneycm.

amyleadem (changes requested):
Thanks for your work on this @mahoneycm. 

I have a few concerns about the design of the left bar here: 
1. The spacing between the form elements and the error bar is very tight. It doesn't feel consistent with the other spacing in our components. 
1. The bar exists in the negative margin, outside of the expected content region. This feels a bit risky to me. If users have a color shift at the edge of the content, or have split their layout at all, it might cause some unintended negative consequences. 

To show what it might look like in the real world, I created a quick [demo of the error states in a page layout in uswds-sandbox](https://federalist-d5e7c07c-6ffa-4b8e-935f-49a7df24a505.sites.pages.cloud.gov/preview/uswds/uswds-sandbox/al-test-5999/pages/prose/).

**Possible solutions:**
1. I wonder if we can consider allowing the error state to shift the content, like it currently does at narrower window widths. In #5908, I actually found the content shift helpful because it drew my attention more to the error ([Preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-char-count-error-state/iframe.html?args=&id=components-form-inputs-character-count--character-count&viewMode=story)). We would have to investigate if there are any accessibility or usability concerns with a content shift like this, but my immediate instinct is that content shift might not necessarily be a bad thing.
2. If we decide we must prevent any content shift and are comfortable living in the negative margins, maybe we can use `$theme-site-margins-width` to help inform the spacing. For example, if the site margins are at least >= 4 units, then move the bar -2 or -2.5 units. If the site margins are <4 units, move the bar to -1.5 units (or something). This might be fairly limited in thinking though &mdash; it might not address forms nested inside of components, etc. Alternatively, maybe we can create a setting that allows users to control the position of the error bar? 
3. Consider using icon or other visual indicator instead of the vertical bar to visually indicate an error state (this is probably better as a follow-on task)

Curious what you all think.

cc: @mejiaj, @jaclinec, @thisisdano, @annepetersen

jaclinec:
I think we should consider using a different visual indicator (i.e. icon) instead of the bar. To me, it stands out better than the vertical bar and doesn't have the content shift issue. It's also an expected and familiar convention used for errors in forms. 

I worry that the vertical bar is not noticeable, and see your point about the content shift possibly bringing needed attention to it. My initial instinct is that a content shift might not be a big deal usability- or accessibility-wise, but it's hard to know without testing or more research.

There's a really good [article on form error messages](https://www.smashingmagazine.com/2022/08/error-messages-ux-design/) and it does mention layout shifts can be "noisy" but doesn't say it's a deal-breaker, FWIW.

mejiaj:
> 1. The spacing between the form elements and the error bar is very tight. It doesn't feel consistent with the other spacing in our components.
> 2. The bar exists in the negative margin, outside of the expected content region. This feels a bit risky to me. If users have a color shift at the edge of the content, or have split their layout at all, it might cause some unintended negative consequences.

No issues from me, but I agree it looks tight. Increasing the spacing will cause it to break out even more or be even closer to the label.

Changing the error style is a possible alternative, but higher LOE. We'd need to spike to find alternative error styles.


> **Possible solutions:**
> 
> 1. I wonder if we can consider allowing the error state to shift the content, like it currently does at narrower window widths. In [USWDS - Character count: Enhance visual cue when limit is exceeded #5908](https://github.com/uswds/uswds/pull/5908), I actually found the content shift helpful because it drew my attention more to the error ([Preview](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-char-count-error-state/iframe.html?args=&id=components-form-inputs-character-count--character-count&viewMode=story)). We would have to investigate if there are any accessibility or usability concerns with a content shift like this, but my immediate instinct is that content shift might not necessarily be a bad thing.

I mentioned in our meeting about layout shift, but according to this Smashing Magazine article, Core Web Vital's don't count shifts caused by user interaction.[^1]. 

@amyleadem are you suggesting we leave the form elements as-is if we decide not proceed with removing the layout shift?

[^1]: [Shifts caused by user interaction are not counted. ](https://www.smashingmagazine.com/2021/06/how-to-fix-cumulative-layout-shift-issues/#:~:text=Shifts%20caused%20by%20user%20interaction%20are%20not%20counted.)


> 2. If we decide we must prevent any content shift and are comfortable living in the negative margins, maybe we can use `$theme-site-margins-width` to help inform the spacing. For example, if the site margins are at least >= 4 units, then move the bar -2 or -2.5 units. If the site margins are <4 units, move the bar to -1.5 units (or something). This might be fairly limited in thinking though — it might not address forms nested inside of components, etc. Alternatively, maybe we can create a setting that allows users to control the position of the error bar?

Allowing users to control the position of the error bar seems like a good idea. I'd avoid `$theme-site-margins-width` because that setting has caused issues with other components.

> 3. Consider using icon or other visual indicator instead of the vertical bar to visually indicate an error state (this is probably better as a follow-on task)

This would require a spike for research and re-work that would increase the LOE. Not saying we shouldn't do it, but we should make these decisions **much** earlier to avoid scrapping work.

---

### My recommendation

1. Give users the option to control border spacing
2. Go back to previous method
3. Re-do error styles [^2]

[^2]: ❗This would be a roadmap item

amyleadem:
I agree that choosing new error style (Option 3) would be a large LOE that would require more research, discussion, and prototyping than this issue/release would allow. 

In my opinion, the cleanest option is to just extend the current mobile styles (that have a content shift) to the desktop state (Option 1). But this might be informed too heavily by personal visual preference. 

@jaclinec, have we done any testing of error states in the "mobile" view? Curious if we have heard any complaints about the current content shift in error states there.

mahoneycm:
> 1. The spacing between the form elements and the error bar is very tight. It doesn't feel consistent with the other spacing in our components.
>
> 2. The bar exists in the negative margin, outside of the expected content region. This feels a bit risky to me. If users have a color shift at the edge of the content, or have split their layout at all, it might cause some unintended negative consequences.

I agree that it's very tight and not ideal. The original styles work as long as enough margin is added to the container. The original styles push the bar _further_ towards / out of the edge of the content.

|  | Old error styles | New error styles |
|--------|--------|--------|
| Padding: 2rem | image | image |
| Padding: 1rem | image | image |

It does feel like the safest option might be to remove the negative margins/placement and allow the content to shift with the border.

Would there be any benefit to adding a slight delay to error styles being added? Maybe like a 500ms delay after the user stops typing?

jaclinec:
> I agree that choosing new error style (Option 3) would be a large LOE that would require more research, discussion, and prototyping than this issue/release would allow.
> 
> In my opinion, the cleanest option is to just extend the current mobile styles (that have a content shift) to the desktop state (Option 1). But this might be informed too heavily by personal visual preference.
> 
> @jaclinec, have we done any testing of error states in the "mobile" view? Curious if we have heard any complaints about the current content shift in error states there.

We tested some error handling in the zebra batch testing, but it was error handling from the browser so it looked different:

image

packages/usa-form-group/src/styles/_usa-form-group.scss
```diff
@@ -11,11 +11,11 @@
...
 .usa-form-group--error {
-  @include u-border-left(0.5, "error-dark");
-  padding-left: units(2);
+  @include u-border-left(2px, "error-dark");
+  padding-left: units(1);
   position: relative;
 
   @include at-media("desktop") {
-    margin-left: units(-2.5);
+    margin-left: -10px;
```
mahoneycm, line 19: Matching the existing behaviors, this number is equal to the negation of border width + padding. Here I'm using a magic number but we could make the styles smarter by assigning variables and calculating the `margin-left`.

```scss
$form-error-border-width: 2px;
$form-error-border-padding: rem-to-px(units(1)); // Alternatively, we could just set this to 8px
$form-error-negative-margin: -(
  $form-error-border-width + $form-error-border-padding
);

...

.usa-form-group--error {
  @include u-border-left($form-error-border-width, "error-dark");
  padding-left: $form-error-border-padding;
  position: relative;

  @include at-media("desktop") {
    margin-left: $form-error-negative-margin;
  }
}

mejiaj (approved):
I was able to use the form error story to confirm these styles.

image

amyleadem (approved):
Confirming that this matches the styles we discussed. Noting that I find the styles to still be a bit tight, but it feels necessary given the discussion we've had, and I don't want to block this PR for my personal preference. 

## Possible follow-on tasks
1. Elevate `$form-error-negative-margin` and `$form-error-border-width` to theme settings so that users can customize error styles to fit their project needs.
2. Research alternative error state styles that replace the left border with an icon or other design

image

amyleadem:
@thisisdano @heymatthenry @annepetersen 

This PR is approved from a technical standpoint, but we need your decision about if we should proceed with this change. We discussed in previous meetings that we might not want to release this until we have some data or feedback to back up the need for it. Curious what you think!

mahoneycm:
Closing in favor of #6204

mahoneycm:
Closing in favor of #6204

PR #6013, opened 2024-08-07
USWDS - Date Picker: Updated toggleCalendar function to account for aria-disabled attribute + readonly

### LINKED ISSUE #5982: USWDS - Date Picker:  Prevent calendar form opening with `aria-disabled`
## Description

The current date picker behaviors could be improved for a better experience and match accessibility requirements.

On the toggle `toggleCalendar()` function, we check if the input has a `disabled` attribute. If so, it returns early and prevents the calendar from opening. If we add a check for the `aria-disabled` attribute, we can prevent users from selecting a date while allowing the button to remain discoverable, which is required for accessibility.

This would make it easier for users to follow recommended disabled guidance by alleviating the responsibility of the user to prevent the click behaviors.

> Using "aria-disabled=true" instead of "disabled" indicates the presence of a disabled element. Use JavaScript to programmatically disable the element to prevent clicks.

## Related code

https://github.com/uswds/uswds/blob/369cf3fdf9abf721a357e864a0f6adbf667ba862/packages/usa-date-picker/src/index.js#L1297-L1298

```diff
const toggleCalendar = (el) => {
-  if (el.disabled) return;
+  if (el.disabled || el.hasAttribute("aria-disabled")) return;
```

## Additional information

[Notes from CS/UX sync](https://docs.google.com/document/d/1IAs5HfPXmD3zWRVB5H_ia2HHXMb5yt-gUqJiUectGWQ/edit#heading=h.vhcvw8ojdcme) (🔒)

RachelCorsino:
# Summary

**Updated toggleCalendar function to account for `aria-disabled attribute + readonly`**. The new code added prevents users from selecting a date while allowing the button to remain discoverable, which is required for accessibility. It also adds the `readonly` which effectively replicates disabled behaviors without the accessibility concerns.

## Breaking change

## Related issue

Closes #5982 

## Related PR
[Changelog PR](https://github.com/uswds/uswds-site/pull/2794)

## Preview link

Preview link:
[Date Picker](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/rc-prevent-calendar-form-opening-aria-disabled/?path=/story/components-form-inputs-date-picker--range-date)

## Problem statement

The date picker does not fully comply with accessibility requirements. Users who rely on assistive technologies may experience frustration or confusion while attempting to interact with the date picker, and users who follow recommended disabled guidance may find it difficult to do so.

## Solution

This PR adds a check for the `aria-disabled` attribute to the toggleCalendar() function, ensuring that the date picker remains discoverable while preventing users from selecting a date when the element is disabled. It also adds the `readonly` attribute when `aria-disabled` is used to help the date picker be discoverable by screenreaders. This change aligns with accessibility requirements and simplifies the process for users to follow recommended disabled guidance.

## Testing and review

1. Under 'Date Picker - Range Date' in Storybook, change "Disabled State" to aria-disabled. Refresh page to make change.
2. Ensure calendar form does not open when calendar icon is clicked
3. Use a screen reader to test that the toggle button is announced as disabled.

- [X] Confirm that this code follows the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://pages.18f.gov/accessibility/checklist/).
- [X] Run `npm run prettier:sass` to format any Sass updates.
- [X] Run `npm test` and confirm that all tests pass.
- [X] Run your code through [HTML_CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) and make sure it’s error free.


alex-hull (commented):
Not yet approving based on observation: Now, this could be a JAWS functionality, as it does not happen in NVDA. When I tab into the calendar icon, while it does tell me it is unavailable, instructions are still announced telling the user to "Press space to activate, use the arrow keys to select". I feel like this could be a bit confusing to the user if they are originally told this is unavailable.

amycole501:
NVDA and Narrator don't announce those extra instructions but I can confirm that in JAWS 2024 and Chrome you do hear all those instructions read aloud.

I'm wary of also having a focusable element that is inoperable BUT can't find specific WCAG criteria that sates it's a failure. Since no one can operate the element it's not technically failing any criteria that we can tell.

packages/usa-date-picker/src/index.js
```diff
@@ -1295,6 +1295,10 @@ const selectDate = (calendarDateEl) => {
  * @param {HTMLButtonElement} el An element within the date picker component
  */
 const toggleCalendar = (el) => {
+  if (el.hasAttribute("aria-disabled")) {
+    el.setAttribute("readonly", true)
+    return;
+  }
   if (el.disabled) return;
```
mahoneycm, line 1302: ## toggleCalendar function

**issue (blocking)** With the curent approach, the `readonly` attribute is only being added when a user clicks the calendar button.

**suggestion:** We should only prevent calendar functionality here. We can move the `readonly` attributre to the `ariaDisable()` function on line lines [721-726](https://github.com/uswds/uswds/blob/9c3967195abe62643302d0cc91b79e3beba1381f/packages/usa-date-picker/src/index.js#L721-L726)

```suggestion
  if (el.disabled || el.hasAttribute("aria-disabled")) return;
```

packages/usa-date-picker/src/index.js
mahoneycm, line 1: **suggestion**: Since `readonly` is a boolean value, we should pass an empty string to `setAttribute()` instead of `true`

```diff
- el.setAttribute("readonly", true)
+ el.setAttribute("readonly", "")
```

mahoneycm (changes requested):
Looks like the calendar functionality is being blocked as expected, but `readonly` isn't being appended on initialization. Left a couple change requests to address!

mahoneycm:
> NVDA and Narrator don't announce those extra instructions but I can confirm that in JAWS 2024 and Chrome you do hear all those instructions read aloud.
> 
> I'm wary of also having a focusable element that is inoperable BUT can't find specific WCAG criteria that sates it's a failure. Since no one can operate the element it's not technically failing any criteria that we can tell.

@amycole501 I believe this would be the intended behavior for `aria-disabled` buttons. We want them to be discoverable via keyboard but have functionality prevented via javascript.

packages/usa-date-picker/src/index.js
mejiaj: Using `true` seems more explicit. MDN has some related guidance on this related to [`hidden`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden#javascript). Validates setting explicitly.

packages/usa-date-picker/src/index.js
RachelCorsino: Resolved in [172c512](https://github.com/uswds/uswds/pull/6013/commits/172c512057f6bf100a7a31171bce0c6878ea56c7). Let me know if this is what was recommended here or if I misunderstood!

packages/usa-date-picker/src/index.js
RachelCorsino: Resolved in [172c512](https://github.com/uswds/uswds/pull/6013/commits/172c512057f6bf100a7a31171bce0c6878ea56c7). Let me know if this is what you were thinking!

packages/usa-date-picker/src/index.js
```diff
@@ -721,6 +721,7 @@ const disable = (el) => {
 const ariaDisable = (el) => {
   const { externalInputEl, toggleBtnEl } = getDatePickerContext(el);
 
+  el.setAttribute("readonly", "")
```
mahoneycm, line 724: **issue (blocking):** This is incorrectly placing the `readonly` attribute on the `usa-date-picker` div element instead of the input element.

Can you update this to append the attribute to the correct element and confirm that the text box is preventing input?

packages/usa-date-picker/src/index.js
mahoneycm, line 1: **issue (blocking):** CircleCI is reporting a [formatting error](https://app.circleci.com/pipelines/github/uswds/uswds/6101/workflows/a1f72435-65be-454a-a86b-ab83ca2e96a9/jobs/10564). 

**suggestion:** Run prettier script to format files after you complete the requested changes.

mahoneycm:
**chore:** Could you also create a Site change log entry for this item? Refer to the [changelog README](https://github.com/uswds/uswds-site/blob/b98f40a00bf52182f72be4de6dd725e244ebb246/_data/changelogs/_CHANGELOG-README.md#L4) and reach out if you have any questions!

packages/usa-date-picker/src/index.js
mejiaj: @mahoneycm we had conflicting guidance here, but we should agree as a team on this style and document it.

For now, no additional changes necessary.

mejiaj (commented):
Thanks @RachelCorsino.

### Testing and review

I was able to test and confirm the following:
- [x] Input is unclickable in `aria-disabled`
- [x] On aria disabled: button shows as `dim` and unclickable.
- [x] On aria disabled: I can still enter numbers in the date.
      

On the last item, is that the intended behavior? You can't click calendar to select date, but you can still enter it manually?

Otherwise LGTM and I can approve once we have clarity on question above and Charlie's changes have been addressed. Thanks again

packages/usa-date-picker/src/index.js
mahoneycm: @mejiaj I based my recommendation off of the MDN guidance for [boolean attributes](https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML)

> If the attribute is present, it can have one of the following values:
>
> - no value at all, e.g. attribute
> - the empty string, e.g. attribute=""
> - attribute's name itself, with no leading or trailing whitespace, e.g. attribute="attribute"
>
> Note: The strings "true" and "false" are invalid values. To set the attribute to false, the attribute should be omitted altogether. Though modern browsers treat any string value as true, you should not rely on that behavior.



I believe this is also a difference between the `hidden` attribute and other attributes.

In our example, adding `extenalInputEl.hidden = true` will appropriately add the `hidden` attribute to the element without a string value. There is no `=""`.

If we do the same with readonly: `externalInputEl.readonly = true`, the attribute is not added to the element.

When using `setAttribute("readonly", "")`, just the boolean attribute is added: `readonly`

packages/usa-date-picker/src/index.js
```diff
@@ -721,6 +721,7 @@ const disable = (el) => {
 const ariaDisable = (el) => {
```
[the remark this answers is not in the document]
RachelCorsino: Resolved in [14f5cb1](https://github.com/uswds/uswds/pull/6013/commits/14f5cb112eaa1dec968244878608f653557cd734)

packages/usa-date-picker/src/index.js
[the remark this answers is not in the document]
RachelCorsino: Resolved in [14f5cb1](https://github.com/uswds/uswds/pull/6013/commits/14f5cb112eaa1dec968244878608f653557cd734)

RachelCorsino: Resolved in [14f5cb1](https://github.com/uswds/uswds/pull/6013/commits/14f5cb112eaa1dec968244878608f653557cd734)

packages/usa-date-picker/src/index.js
RachelCorsino: Resolved in [14f5cb1](https://github.com/uswds/uswds/pull/6013/commits/14f5cb112eaa1dec968244878608f653557cd734)

RachelCorsino:
> Thanks @RachelCorsino.
> 
> ### Testing and review
> I was able to test and confirm the following:
> 
> * [x]  Input is unclickable in `aria-disabled`
> * [x]  On aria disabled: button shows as `dim` and unclickable.
> * [x]  On aria disabled: I can still enter numbers in the date.
>   
> 
> On the last item, is that the intended behavior? You can't click calendar to select date, but you can still enter it manually?
> 
> Otherwise LGTM and I can approve once we have clarity on question above and Charlie's changes have been addressed. Thanks again

I think that was a mistake on my end with one of the functions. After fixing, the input element does not allow for input.

packages/usa-date-picker/src/index.js
```diff
@@ -714,14 +714,14 @@ const disable = (el) => {
...
- * @param {HTMLElement} el An element within the date picker component
+ * @param {HTMLElement} el - The date picker element
  */
 const ariaDisable = (el) => {
   const { externalInputEl, toggleBtnEl } = getDatePickerContext(el);
 
-  el.setAttribute("readonly", "")
+  externalInputEl.setAttribute("readonly", "")
   toggleBtnEl.setAttribute("aria-disabled", true);
   externalInputEl.setAttribute("aria-disabled", true);
```
mahoneycm, line 726: **quibble (non-blocking):** Super minor thing but for consistency, can we group `externalInputEl` changes?

```suggestion
  toggleBtnEl.setAttribute("aria-disabled", true);
  externalInputEl.setAttribute("aria-disabled", true);
  externalInputEl.setAttribute("readonly", "")
```
_Moved line 724 under 726_

packages/usa-date-picker/src/index.js
```diff
@@ -734,6 +734,8 @@ const ariaDisable = (el) => {
 const enable = (el) => {
   const { externalInputEl, toggleBtnEl } = getDatePickerContext(el);
 
+  el.removeAttribute("readonly");
+  externalInputEl.removeAttribute("aria-disabled");
   toggleBtnEl.disabled = false;
   externalInputEl.disabled = false;
```
mahoneycm, line 740: **issue:** Similar to the `ariaDisabled` function, we'll want to update `el` to `externalInputEl`.

**issue:** We removing `disabled` from the `toggleBtnEl` but not `aria-disabled`. We should make sure to remove this as well.

I'm also grouping element changes in this suggestion:

```suggestion
  toggleBtnEl.disabled = false;
  toggleBtnEl.removeAttribute("aria-disabled");
  
  externalInputEl.disabled = false;
  externalInputEl.removeAttribute("aria-disabled");
  externalInputEl.removeAttribute("readonly");
```

packages/usa-date-picker/src/index.js
mahoneycm, line 1: **issue:** I'm seeing a CI build error that should be resolved by pulling in the latest from `develop`

```node
git fetch
git pull origin develop
```

mahoneycm (changes requested):
Just a few fine grain details then we should be good to go! Thanks for your persistence @RachelCorsino

packages/usa-date-picker/src/index.js
RachelCorsino: Resolved in [18e578c](https://github.com/uswds/uswds/pull/6013/commits/18e578c8c71e20ea38cf57edc3e0fdb130e76c0f)

packages/usa-date-picker/src/index.js
RachelCorsino: Resolved in [18e578c](https://github.com/uswds/uswds/pull/6013/commits/18e578c8c71e20ea38cf57edc3e0fdb130e76c0f)

mejiaj (approved):
Thanks @RachelCorsino. Confirming that the `aria-disabled` button AND input are no longer editable.

amycole501 (approved):
Sounds good now in JAWS; works as expected.

thisisdano (approved):
Thank you!

Issue #5910, opened 2024-05-03
USWDS - Memorable date: Input options don't meet some users expectations


jaclinec:
# Summary
Users, especially those with visual impairments, expect to be able to search by typing either the number or the name for the month in memorable date. The dropdown for the 'month' input in memorable date only allows for typing the number of the month, but usability testing showed that it'd be beneficial to allow for both name or number searching.

# Observations
This mostly applied to participants who are visually impaired (2 blind, 2 low vision) in our testing, although we did have one 'typically sighted' participant who commented that we should allow users to search with either name or number.

Those with visual impairments who typed in the combo box to search for and select a month were split on what they typed: half of them typed the number (for example, "01" for January) and half typed the name of the month (e.g. "D" for December). The combo box does not allow searching my month name which was a friction point for those users. See [video clip](https://drive.google.com/file/d/1g_mBG7PWpdLPuPQ0oscff7VWP-sXeTNg/view?usp=sharing) :lock:

# Affected user groups

- Screen reader users
- Visually impaired
- All users

# Research method
Usability testing with 8 participants with disabilities: 
- 5 visually impaired (3 blind, 2 low vision/legally blind (1 with nystagmus))
- 1 motor impairment
- 1 ADHD
- 1 anxiety and depression

Screen reader software participants used: JAWS, NVDA, and VoiceOver

[Link to public findings report forthcoming...here is the internal [Zebra findings report](https://docs.google.com/document/d/1vbGSh4UNZiseyzdGXa1pJwU_3M4y7HaU3QPkXOa7gsI/edit?usp=sharing)] :lock:

# Next steps

- [ ] Update combo box search functionality that allows users to search by either the month name or number. This solution could be affected/combined by the work we're doing in #5647 

Note: This issue is related to #5911

mejiaj:
USA Memorable date shown [here](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-form-inputs-memorable-date--memorable-date&p=all) uses USA Select, not USA Combo box.

## Todo

- [ ] See if we can make USA memorable date more responsive to keyboard input (_modifying markup_).
- [ ] Confirm if USA Combo box is a viable replacement
  - [ ] If it is, confirm if we need to enhance combo box

annepetersen:
[...]

PR #6028, opened 2024-08-19
USWDS - Memorable Date: Remove numbers from the dropdown list of months

cathybaptista:
# Summary

**Removed the leading numbers from the month options in the memorable date component.** Recent usability testing indicated that having both numbers and names to represent months was confusing for screen reader users.

## Breaking change

## Related issue

Closes https://github.com/uswds/uswds/issues/5910

## Related pull requests

Changelog uswds/uswds-site#2787

## Preview link

Preview link: [Memorable Date preview
](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cb-memorable-date-input-options/?path=/story/components-form-inputs-memorable-date--memorable-date)

## Problem statement

The desired state for the memorable date component is that there is only one option for the values in the months dropdown (either numbers or month names) so that users can search by typing in the month name. The current state is that both the number and the month name are included in the option dropdown, which means users cannot search by the month name. 

## Solution

The solution is to remove the numbers from the memorable date component option dropdown. Another approach was to make both options (numbers AND month names) available, but the risk of this approach did not outweigh the rewards. The team decided that including month names is the best way to resolve this accessibility issue. 

## Testing and review

1. Opened federalist link and inspected source.
2. Verify that the id="date_of_birth_month" options do not include a number.
3. Opened the component in production.
4. Verify that the id="date_of_birth_month" options includes a number.
5. Verify that if you tab to the memorable date month field and hit the first letter of each month on the keyboard the correct month appears.
6. Verify that the correct code appears on uswds-site for this component when the "component code" button is selected. This code should not include numbers.

### DIFF
--- a/packages/usa-memorable-date/src/usa-memorable-date.twig
+++ b/packages/usa-memorable-date/src/usa-memorable-date.twig
@@ -9,18 +9,18 @@
         {% elseif disabled_state == 'aria-disabled' %} aria-disabled="true" {%- endif %}
       >
         <option value>- Select -</option>
-        <option value="1">01 - January</option>
-        <option value="2">02 - February</option>
-        <option value="3">03 - March</option>
-        <option value="4">04 - April</option>
-        <option value="5">05 - May</option>
-        <option value="6">06 - June</option>
-        <option value="7">07 - July</option>
-        <option value="8">08 - August</option>
-        <option value="9">09 - September</option>
-        <option value="10">10 - October</option>
-        <option value="11">11 - November</option>
-        <option value="12">12 - December</option>
+        <option value="1">January</option>
+        <option value="2">February</option>
+        <option value="3">March</option>
+        <option value="4">April</option>
+        <option value="5">May</option>
+        <option value="6">June</option>
+        <option value="7">July</option>
+        <option value="8">August</option>
+        <option value="9">September</option>
+        <option value="10">October</option>
+        <option value="11">November</option>
+        <option value="12">December</option>
       </select>
     </div>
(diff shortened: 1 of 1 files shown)


mejiaj (approved):
Confirming I can navigate using keyboard more easily. Thanks Cathy!

[uswds-memorable-date-6028.webm](https://github.com/user-attachments/assets/c13ea62d-c670-4d9f-8c5a-8dcb4467a223)

mahoneycm (approved):
Searching by name is working for me! I think this will clear up a lot of the confusion.

Thanks @cathybaptista !

amyleadem (changes requested):
Thanks for making this update @cathybaptista. The code update looks good!

Quick request before approval: Can you update the summary in the PR description to be a bit more succinct? We try to pull that PR summary into the release notes and the current one is a bit too long to be useful in that context. If desired, we can move the additional information that is currently in the "summary" section down to the "solution" section. 

I'm thinking something like this should work (Feel free to edit):

**Removed the leading numbers from the month options in the memorable date component.** Recent usability testing indicated that having both numbers and names to represent months was confusing for screen reader users.

mejiaj:
Moving the original issue back to `In progress` to capture requested changes @amyleadem mentioned.

amyleadem (approved):
Looks good to me! Thanks @cathybaptista 

Moved the issue to FFR

thisisdano (approved):

PR #6029, opened 2024-08-19
USWDS - Error state preview: Add developer preview alert.

### LINKED ISSUE #5992: USWDS - Feature Request: Add "Developer preview" alert to Test Form Error story
# Description

Add "developer preview" alert to the Test Form Error story. This will be used to reference our components current error state styles. The idea is to "code out in the open" and be more transparent while we improve our guidance and implementation.

This was approach inspired by the Aria-Practice Group's [tooltip preview](https://codepen.io/Moiety/pen/LaPvWy)

mahoneycm:
# Summary

Added developer preview alert message to error state form preview.

## Breaking change

## Related issue

Closes #5992

## Related pull requests

*No changelog. Not a user facing change at this time.*

## Preview link

[Preview link →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-preview-alert/?path=/story/patterns-forms--test-error-form-elements)

## Problem statement

Form error state preview code is a test story and should not be referred to production ready code.

## Solution

Add “developer preview” alert to test pattern page.

## Testing and review

1. Visit [test error form element page →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-preview-alert/?path=/story/patterns-forms--test-error-form-elements)
2. Confirm alert is displaying as intended.
3. Approve copy for alert message.
4. Inspect alert import and implementation.
5. Confirm code styles are up to USWDS standard.

### DIFF
--- a/packages/usa-form/src/test/test-pattern/test-usa-form.twig
+++ b/packages/usa-form/src/test/test-pattern/test-usa-form.twig
@@ -1,3 +1,8 @@
+{% if error_state %}
+{% include '../../../../usa-alert/src/usa-alert.twig' with alert %}
+{% endif %}
+
+
 {# *
 * While we could add the usa-form-group--error class to <form class="usa-form"> elements, we should 
 * avoid doing this and only add it to usa-form-group elements.
--- a/packages/usa-form/src/test/test-pattern/test-usa-form.twig
+++ b/packages/usa-form/src/test/test-pattern/test-usa-form.twig
@@ -2,7 +2,7 @@
   {% include "@components/usa-alert/src/usa-alert.twig" with {
     modifier: "usa-alert--warning",
     title: "Heads up!",
-    text: "This is a developer preview. The patterns presented on this page may not be suitable production level code.",
+    text: "This is a developer preview. The patterns presented on this page may not be suitable for production or reflect our most recent research-informed recommendations.",
   } %}
 {% endif %}
 
--- a/packages/usa-form/src/test/test-pattern/test-usa-form.twig
+++ b/packages/usa-form/src/test/test-pattern/test-usa-form.twig
@@ -1,5 +1,5 @@
 {% if error_state %}
-  {% include '../../../../usa-alert/src/usa-alert.twig' with alert %}
+  {% include "@components/usa-alert/src/usa-alert.twig" with alert %}
 {% endif %}
 
 {# *
--- a/packages/usa-form/src/test/test-pattern/test-usa-form.twig
+++ b/packages/usa-form/src/test/test-pattern/test-usa-form.twig
@@ -1,5 +1,9 @@
 {% if error_state %}
-  {% include "@components/usa-alert/src/usa-alert.twig" with alert %}
+  {% include "@components/usa-alert/src/usa-alert.twig" with {
+    modifier: "usa-alert--warning",
+    title: "Heads up!",
+    text: "This is a developer preview. The patterns presented on this page may not be suitable production level code.",
+  } %}
 {% endif %}
 
 {# *
--- a/packages/usa-form/src/test/test-pattern/test-usa-form.twig
+++ b/packages/usa-form/src/test/test-pattern/test-usa-form.twig
@@ -1,7 +1,6 @@
 {% if error_state %}
-{% include '../../../../usa-alert/src/usa-alert.twig' with alert %}
+  {% include '../../../../usa-alert/src/usa-alert.twig' with alert %}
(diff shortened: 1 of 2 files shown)


packages/usa-form/src/test/test-pattern/test-usa-form.twig
```diff
@@ -1,3 +1,7 @@
+{% if error_state %}
+  {% include '../../../../usa-alert/src/usa-alert.twig' with alert %}
```
mejiaj, line 2: **suggestion:** There's a path shorthand available.

```suggestion
  {% include "@components/usa-alert/src/usa-alert.twig" with alert %}
```

packages/usa-form/src/usa-form.stories.js
```diff
@@ -10,6 +10,12 @@ import EsContent from "../../templates/usa-sign-in/usa-sign-in~lang-es.json";
 import MultipleContent from "../../templates/usa-sign-in/usa-sign-in--multiple/usa-sign-in--multiple.json";
 import EsMultipleContent from "../../templates/usa-sign-in/usa-sign-in--multiple/usa-sign-in--multiple~lang-es.json";
 
+const InfoContent = {
+  modifier: "usa-alert--warning",
+  title: "Heads up!",
+  text: "This is a developer preview. The patterns presented on this page may not be suitable production level code.",
+};
+
```
mejiaj, line 18: **polish:** This isn't part of the form story. You could move this to where the template is being called. Example below.

https://github.com/uswds/uswds/blob/42907b9ef1a8dfa859730e91d03482f2283b0392/packages/templates/usa-password-reset-form.twig#L20-L30

mejiaj (commented):
Looks good @mahoneycm, added minor suggestions.

packages/usa-form/src/usa-form.stories.js
```diff
@@ -10,6 +10,12 @@ import EsContent from "../../templates/usa-sign-in/usa-sign-in~lang-es.json";
 import MultipleContent from "../../templates/usa-sign-in/usa-sign-in--multiple/usa-sign-in--multiple.json";
 import EsMultipleContent from "../../templates/usa-sign-in/usa-sign-in--multiple/usa-sign-in--multiple~lang-es.json";
 
+const InfoContent = {
+  modifier: "usa-alert--warning",
+  title: "Heads up!",
+  text: "This is a developer preview. The patterns presented on this page may not be suitable production level code.",
```
amyleadem, line 16: **suggestion:** Add some small copy tweaks for clarity. 
```suggestion
  text: "This is a developer preview. The patterns presented on this page may not be suitable for production.",
```

Alternatively, wondering if we can give more detail about "developer preview" means. Maybe something like this? 
```suggestion
  text: "This is a test story that shows the current state of USWDS error styles. We are currently researching best practices for error states, so the patterns presented on this page may not be suitable for production.
```

mahoneycm: Thanks for these! I struggled to find the right wording so opted for a less is more approach.

One of the big inconsistencies with these implementations is the inconsistent use of things like `usa-form`. I think I hesitated to say that it was the "current state" of the error styles because this implimentation is peculiar but at the end of the day that is what we're using this form for!

@annepetersen do you have a preference for the copy here?

packages/usa-form/src/usa-form.stories.js
mahoneycm: WOW this is an awesome change! Being able to isolate all the alert data right within the template is a great improvement. Thanks for flagging this!

Resolved in [b65075a](https://github.com/uswds/uswds/pull/6029/commits/b65075a91e6030998612cd38b4c5c42bfc1c666d)

packages/usa-form/src/usa-form.stories.js
annepetersen: "Suitable for production" is better, I think. 

But we could even append:
"suitable for production or reflect our most recent research-informed recommendations."

mahoneycm: Got it! I think that accomplishes the goal @amyleadem's second suggestion that added the note about ongoing research.

Added in [d079149](https://github.com/uswds/uswds/pull/6029/commits/d079149d68d305413fbc448a9784f3ce75b6c0c5)

annepetersen (approved):
LGTM!

PR #6037, opened 2024-08-27
USWDS - Header: Search input appears last in menu for mobile header.

### LINKED ISSUE #3526: Mobile navigation tab order does not follow visual order
## Description

When tabbing through elements in the mobile menu navigation drawer, the tab order may not align to an expected order based on what is visually present on the page. The search field is shown visually toward the top of the drawer, but is last in the tab order. This can cause confusion for users who navigate using keyboard tabbing, or iOS VoiceOver / Android TalkBack gesture navigation.

1. Navigate to [Basic Header demo](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/release-2.7.1/components/preview/header--basic.html)
2. Shrink the page, or zoom in to the point where the mobile navigation styling is activated
3. <kbd>Tab</kbd> to and activate the menu button using <kbd>Space</kbd>
4. Observe that focus is placed on the "Close" button in the navigation drawer
5. Press <kbd>Tab</kbd>

**Expected behavior:** Focus is placed in the search field, as it is visually located immediately following the current active element
**Actual behavior:** Focus is placed in the navigation menu links. Continuing to press <kbd>Tab</kbd>, you will observe that the search field is last in the tab order

## Additional information

Screen recording:

uswds-menu-tab-order

Relevant success criteria [2.4.3 Focus Order](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html):

>For example, a screen reader user interacts with the programmatically determined reading order, while a sighted keyboard user interacts with the visual presentation of the Web page. Care should be taken so that the focus order makes sense to both of these sets of users and does not appear to either of them to jump around randomly.

Relevant code, which places the navigation _visually_ last, even if in markup it occurs before the search field:

https://github.com/uswds/uswds/blob/9ca144f77120b18e6f22544fe66f7e3f727b4e4e/src/stylesheets/components/_navigation.scss#L73

Relevant ["Accessibility concerns" guidance for `order` CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/order#Accessibility_concerns):

>Using the `order` property will create a disconnect between the visual presentation of content and DOM order. This will adversely affect users experiencing low vision navigating with the aid of assistive technology such as a screen reader. If the visual (css) order is important, then screen reader users will not have access to the correct reading order.

Note: This issue does not occur on the main [USWDS website](https://designsystem.digital.gov/), only because the `.usa-nav__primary` on which the `order` styling is applied is wrapped by a `.usa-nav__inner`.

Possible remediation options:

- Search should be visually displayed the same as it is in markup order (i.e. remove `order`, allow search field to show at the bottom of the drawer)
- Search should be moved before the navigation links in the markup. Note that this could have adverse effects on the intended desktop styling.
- Search field should be programmatically moved to occur in the DOM prior to navigation links at smaller viewport sizes. Note that it's unclear if assistive technologies is accommodating to these programmatic changes.

### LINKED ISSUE #3920: Reordering display navigation may be disorienting
As a user, I expect that the order of content in tab navigation and accessibility tree matches the visual presentation, so that I'm not disoriented by focus shifting unpredictably while navigating content (via <kbd>Tab</kbd>, for example).

As a developer, I expect that when I place content in markup after `.usa-nav__primary`, it will be shown visually after the navigation content.

An example of this is in the primary navigation mobile styling:

https://github.com/uswds/uswds/blob/3b166be8d20f6888913a3d96707df7ce07bc63bc/src/stylesheets/components/_navigation.scss#L64-L73

You can verify this by tabbing through the mobile navigation menu in the [component demo](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.app.cloud.gov/preview/uswds/uswds/release-2.10.0/components/preview/header--basic.html), where you will find that:

- Display order: Close button, search field, search button, menu links
- Tab order: Close button, menu links, search field, search button

Interestingly, this problem does not affect the [USWDS main website](https://designsystem.digital.gov/), since `.usa-primary__nav` is rendered within a (non-flex) `.usa-nav__inner` container.

Related guidance:

- [C27: Making the DOM order match the visual order](https://www.w3.org/TR/WCAG20-TECHS/C27.html)
  - >The objective of this technique is to ensure that the order of content in the source code is the same as the visual presentation of the content. [...] Each order may be meaningful in itself but may cause confusion for assistive technology users.
- [G59: Placing the interactive elements in an order that follows sequences and relationships within the content
](https://www.w3.org/TR/WCAG20-TECHS/G59.html)
   - >When the order of the HTML source matches the visual order of the Web page, tabbing through the content follows the visual layout of the content. When the source order does not match the visual order, the tab order through the content must reflect the logical relationships in the content that are displayed visually.

Related resources:

- https://developer.mozilla.org/en-US/docs/Web/CSS/order#Accessibility_concerns
- https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Ordering_Flex_Items#The_order_property_and_accessibility
- https://tink.uk/flexbox-the-keyboard-navigation-disconnect/
- https://adrianroselli.com/2015/10/html-source-order-vs-css-display-order.html
- https://www.youtube.com/watch?v=8MAvH6vYbDo
- https://alastairc.uk/2017/06/the-responsive-order-conflict/

Possible improvements:

- Update guidance to recommend two copies of the search form or menu items for mobile and desktop, where respectively the search field is shown prior to menu items (mobile) and after menu items (desktop).
- Dynamically update DOM in JavaScript to reorder search field, optionally in combination with [`MediaQueryList#addListener`](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener) to update after viewport size changes.

### LINKED ISSUE #5971: USWDS - Header: Tab order does not match visual order in mobile view
## Summary
The tab order in the mobile header menu does not currently match the visual order. In standard variants of the header, the search bar is visually at the top of the mobile menu but appears last in the tab order. This issue means that the component does not meet [WCAG criterion 2.4.3 (Focus order)](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-focus-order).

Note that this does issue not happen in the extended variant of the component. In the extended variant, the search bar visually appears at the bottom of the mobile menu, which matches the tab order.

cathybaptista:
# Summary
**Removed the CSS order property from mobile header navigation.** Now, the visual order of the component matches the tab order. 

## Details
This change visually moves the search bar to appear last in the list of menu items in the mobile header. This change results in the header being compliant with [WCAG criterion 2.4.3 (Focus order)](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-focus-order) in mobile view.

## Breaking change

This change causes the search bar to appear at the bottom of the menu. If you would like to visually keep the search bar at the top of the menu, you will need to reorder your markup in the mobile view.

## Related issue

Closes #5971, closes #3526, closes #3920.

## Related pull requests

Changelog uswds/uswds-site#2798

## Preview link

Preview link: [Header Component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cb-remove-order-from-usa-nav/?path=/story/components-header--default)

## Problem statement

Web pages should be able to be navigated in an order that preserves meaning and operability. Currently, the tab order in the mobile header menu does not match the visual order. Therefore, the component does not meet [WCAG criterion 2.4.3 (Focus order)](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-focus-order).

## Solution

By removing the `order: 2; style` from `.usa-nav__primary`, the search bar appears at the bottom of the menu in the mobile header, maintaining tab order and preserving meaning. This approach was chosen based on a [conversation in dev sync on 8/6/24](https://docs.google.com/document/d/172TU5BVS9kGZQDuPcAhGoZjaTd2ngzcTayq4WRRTBUs/edit#heading=h.rcadppdhx7dt). This will impact users that wish to visually keep the search bar at the top of the menu; these users will need to reorder their markup in the mobile view.

## Testing and review

1. In Storybook, select the header component.
2. Resize the browser to minimum width so that the menu is collapsed.
3. Select the "menu" button and validate that the search input field appears after other sample links in the menu.
4. Confirm the tab order is appropriate and consistent for each of the header variants.

I am looking for confirmation that the tab order makes sense and that it is consistent with [WCAG criterion 2.4.3 (Focus order)](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-focus-order),


packages/usa-nav/src/styles/_usa-nav.scss
```diff
@@ -130,8 +130,6 @@ $expand-less-icon: map-merge(
   @include at-media-max($theme-header-min-width) {
     @include nav-list("nav");
     margin-top: units(3);
-    order: 2;
-
     a {
       @include u-padding-y(1.5);
     }
```
mahoneycm, line 135: **polish:** We should maintain a space between these styles and the declaration

```suggestion

    a {
      @include u-padding-y(1.5);
    }
```

mahoneycm (changes requested):
Thanks @cathybaptista! The headers are looking consistent across the board.

One minor change to add an empty line before the anchor declaration

packages/usa-nav/src/styles/_usa-nav.scss
mahoneycm: Thanks!

mahoneycm (approved):
Lgtm!

- [x] Header search placement is consistent across variants
- [x] Tab order matches expectation

> [!IMPORTANT]
> We'll want to mark this change our release notes. If we offer the suggestion to reapply order via CSS, we should flag that it goes against accessibility guidelines.

amyleadem (changes requested):
Thanks for this work @cathybaptista! I confirmed that the tab order now matches the visual order. 

I noticed that the vertical margins between mobile nav elements are a bit different across variants. Can you update the margins in the Default variant above the search bar and primary navigation to align with the margins in the Extended megamenu variant? I've highlighted the differences in the screenshot below:

image

amyleadem (changes requested):
Thanks @cathybaptista. 

I noticed that there is still a difference in the spacing between the search bar and link list. The basic variant has a margin-top of .5rem while the extended has a margin-top of 1rem. 

Instead of the changes you made to `usa-header.scss`, I'm wondering if we can use the `margin-top` rule from `.usa-nav__secondary .usa-search` ([line 342 of usa-nav.scss](https://github.com/uswds/uswds/blob/3a6c114dbf13e876199790d3e96f65d58b86298b/packages/usa-nav/src/styles/_usa-nav.scss#L342)) and make it so that it applies more generically (potentially [line 113 of the same file](https://github.com/uswds/uswds/blob/3a6c114dbf13e876199790d3e96f65d58b86298b/packages/usa-nav/src/styles/_usa-nav.scss#L113)). Curious what you think!

amyleadem (approved):
Looks good to me! Thanks @cathybaptista 

@mahoneycm I am clearing your review since some changes have been made since you last saw it. Can you take another look when you get a moment?

packages/usa-nav/src/styles/_usa-nav.scss
```diff
@@ -111,6 +111,7 @@ $expand-less-icon: map-merge(
   }
 
   .usa-search {
+    margin-top: units(2);
```
mahoneycm, line 114: **Issue (blocking):** It looks like this change causes an unintentional visual regression in the full header display for default and megamenu variants. The top margin pushes down the search bar and it is no longer vertically centered to match the nav buttons.

| | Develop | Working branch |
|--------|--------|--------|
| Default | image | image |
| Megamenu | image | image | 

**Suggestion:** This may be resolved by removing the margin in the media query below (I haven't confirmed).

```diff
margin-top: units(2);
+ // Additional empty space before media query to scan easier
 @include at-media($theme-header-min-width) {
   margin-left: units(2);
+  margin-top: 0;
}

**Review by mahoneycm — CHANGES_REQUESTED** [review]
I noticed a regression in desktop views! Would you mind looking into this?

File: packages/usa-nav/src/styles/_usa-nav.scss (hunk 2 above)
Comment (mahoneycm) [reply packages/usa-nav/src/styles/_usa-nav.scss:114]: lgtm! Resolved in [adb5ba9](https://github.com/uswds/uswds/pull/6037/commits/adb5ba91bd7b3e7008a72f8d8bc43c77f1e0e0fc)

File: packages/usa-nav/src/styles/_usa-nav.scss [hunk 3]
```diff
@@ -111,8 +111,10 @@ $expand-less-icon: map-merge(
...
```
mahoneycm, line 114: **polish:** Let's add an empty line before the media-query to make this a bit more scannable 🙂 

```suggestion
    margin-top: units(2);
    
```

aduth:
Does this fix #3920 and/or #3526?

mahoneycm:
@aduth it looks like it!

mahoneycm (approved):
lgtm!

packages/usa-header/src/styles/_usa-header.scss
```diff
@@ -28,6 +28,7 @@ $z-index-overlay: 400;
 
   // The search <form>
   .usa-search {
+    margin-top: units(1);
```
mejiaj, line 31: **question**: Is this still necessary if we're also setting and overriding in `_usa-nav.scss`?

packages/usa-header/src/styles/_usa-header.scss
```diff
@@ -150,7 +151,7 @@ $z-index-overlay: 400;
     .usa-nav {
       @include u-flex("row", "align-center", "justify-end");
       display: flex;
-      padding: 0 0 units(0.5) units(1);
+      padding: 0 0 units(1) units(1);
```
mejiaj, line 154: **thought**: The clarity of this is nice. We can shorthand, but this approach seems better.

mejiaj (commented):
@cathybaptista nice job. 

Functionally, everything looks and works well. Tested on both mobile and desktop breakpoints. No issues found with tabbing.  One minor note to confirm on the header margin — I can approve once confirmed.

I've also updated the PR description to link the other issues that this feature closes.

```diff
- Closes #5971
+ Closes #5971, closes #3526, closes #3920.
```

Thanks for flagging that @aduth.

packages/usa-header/src/styles/_usa-header.scss
mahoneycm: Looks like the value used is the value found in `_usa-nav.scss`. We can probably remove here.

image

mejiaj (approved):
I was able to successfully test and confirm that the search input appears last.

I compared the feature branch with what's currently on develop. Before/after links below.

| Before | After |
| :----- | :---- |
| [Develop](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-header--extended)  |    [Feature branch](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cb-remove-order-from-usa-nav/?path=/story/components-header--extended)     |

thisisdano (approved):
This one is probably technically a breaking change because of the UX and design implications — but it might be worth noting that I don't believe teams will have to make any changes _to their code_ to make this work, so change does not introduce a breaking change on the code side

thisisdano:
I approve this PR, but we may want to double-check the changelog entry and release notes.

amyleadem:
@thisisdano Added a [suggestion to the changelog](https://github.com/uswds/uswds-site/pull/2798/files#r1786416549) in an attempt to clarify the user action for the breaking change here. Once we figure out a solution in the changelog, we can include the changelog in the release notes and PR description.

PR #6058, opened 2024-09-11
USWDS: Add instructions to provide a reproducible example to the bug issue template

### LINKED ISSUE #6040: USWDS - Issue template: Add prompt for reproducible example
## Summary
Add a note to the bug issue template to include a link to a repo, reproducible code example, or sandbox for testing.

amyleadem:
# Summary

Added instructions to provide a reproducible example to the bug issue template

## Breaking change

## Related issue

Closes #6040 

## Related pull requests
N/A 

## Preview

**Before:** 

image

**After:**

image

## Testing and review

- Confirm that the updated instructions are appropriate and make sense
- Confirm no spelling or grammatical errors

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -13,8 +13,8 @@ body:
   - type: textarea
     id: reproduce
     attributes:
-      label: Instructions for reproducing the bug
-      description: Provide a link to a public repo or code preview that demonstrates this issue. Alternatively, describe how to reproduce this issue.
+      label: Steps to reproduce the bug
+      description: Describe how to reproduce this issue. If the issue is found only in your project, provide a link to a public repo or code preview that demonstrates this issue.
       placeholder: |
         1. Go to '...'
         2. Click on '....'
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -13,8 +13,8 @@ body:
   - type: textarea
     id: reproduce
     attributes:
-      label: Steps to reproduce the bug
-      description: Describe how to reproduce this issue.
+      label: Instructions for reproducing the bug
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -14,7 +14,7 @@ body:
     id: reproduce
     attributes:
       label: Steps to reproduce the bug
-      description: Describe how to reproduce this issue.
+      description: Describe how to reproduce this issue. If the issue is found only in your project, provide a link to a public repo or code preview that demonstrates this issue.
```
mahoneycm, line 17: **Question:** What do you think about making the demo its own, non-required, field? I think it'd bring additional attention to the ask and could see a benefit of separating it from the testing instructions so users could provide the demo, and then specifically outline the steps to reproduce.

mahoneycm (approved):
This looks good to me but also left a question regarding separating the code demo request from the "steps to reproduce" section. Let me know what you think!

.github/ISSUE_TEMPLATE/bug_report.yaml
amyleadem: My instinct is to leave it as is in a single field to see if it helps before adding additional fields. That said, happy to defer to the group if we decide we want an additional field. @mejiaj, do you have a preference?

heymatthenry:
**thought (non-blocking):** It would be great if we had a page somewhere that explained a little more about how to make a good reproducible example, and why it's important. The issue template probably isn't the place for that, but it we could link to it from here if we had that guidance somewhere else.

The R community is really good about this. I love the way [this page](https://www.tidyverse.org/help/#reprex) provides guidance on constructing a good example. But it also explains the benefits of going through the process of setting up the minimal test case not just for the team reviewing the bug, but for the submitters themselves.

heymatthenry (approved):
Thanks for adding this!

thisisdano (approved):

PR #6074, opened 2024-09-20
USWDS - Combo box search results

### LINKED ISSUE #5647: Combo box search results do not match user expectations
# Summary
Users expect first letter navigation when they type into a combo box for selecting states or countries. In other words, if they type "T," they expect to see results of states starting with the letter T, like Tennessee, Texas, etc. The current way the combo box functions is to show all results _containing_ the letter "t", so for example Connecticut may show up in results, confusing users.

# Observations
4 out of 5 participants in usability testing were confused by the search results. They were tasked with selecting "Texas" from U.S. states combo box in a form prototype. Those 4 users typed "T" and were confused that they did not get results for only states starting with "T," and often would just scroll down the whole list to get to Texas. One users commented that that would be especially annoying if you were navigating a long list of 200+ countries.

[Video clips](https://drive.google.com/file/d/1h7TnVkejKvQGhPamrcZ5Z6rZxRrNl5A7/view?usp=sharing) :lock:

# Affected user groups

- All users
- Screen reader users
- Screen magnification users

# Research method
Usability testing with 5 visually impaired participants. See [findings report ](https://github.com/uswds/uswds/wiki/Usability-research-findings-for-Q3-2023)for details.

# Recommendations for next steps

Currently there are no clear solution recommendations. The team discussed whether the unexpected behaviors could be avoided by updating the combo box results to be sorted by relevance rather than alphabetically. Another possibility is adding screen reader instructions to enter a search term. Is it possible or recommended to implement first-letter navigation?

# Next steps

Decide on a solution to mitigate the friction users experience with unexpected search results and test it again in a new round of usability testing.

kodyjackson:
# Summary
Implement First Letter Navigation for Combo Box Filtering

## Breaking change
No breaking change

## Related issue

Closes #5647 

## Problem statement
Users interacting with the combo box for selecting states or countries expect intuitive behavior where typing the first letter of an option narrows down the results to only those that begin with that letter. Currently, the combo box shows all results containing the typed letter, leading to confusion, as options like "Connecticut" appear when users type "T." This does not align with user expectations and can hinder the overall usability of the component.

## Solution
The solution implements first letter navigation in the combo box's filtering logic. By modifying the generateDynamicRegExp function, we ensure that only the options starting with the typed letter are displayed, providing a more accurate and user-friendly experience.

## Major changes

1. Modification of `generateDynamicRegExp` Function:

    -  Updated the regex generation to anchor matches to the start of the string, ensuring that only options beginning with the input character(s) are returned.
    - Adjusted the logic to prioritize first-letter matches over partial matches.

2. Testing Enhancements:

    - Added new test cases to validate the updated behavior, ensuring that typing a letter displays only options starting with that letter.
    - Ensured backward compatibility by retaining existing functionalities while focusing on first-letter navigation.

### DIFF
--- a/packages/usa-time-picker/src/test/time-picker-regex.spec.js
+++ b/packages/usa-time-picker/src/test/time-picker-regex.spec.js
@@ -1,11 +1,10 @@
 const assert = require("assert");
 const { FILTER_DATASET } = require("../index");
 
-
 const generateDynamicRegExp = (filter, query = "", extras = {}) => {
   const escapeRegExp = (text) =>
     text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&");
-  
+
   let find = filter.replace(/{{(.*?)}}/g, (m, $1) => {
     const key = $1.trim();
     const queryFilter = extras[key];
@@ -19,16 +18,14 @@ const generateDynamicRegExp = (filter, query = "", extras = {}) => {
 
       return "";
     }
-      
+
     return escapeRegExp(query);
   });
 
-    find = `^(?:${find})$`;
+  find = `^(?:${find})$`;
 
   return new RegExp(find, "i");
 };
-
-
 
 describe("time picker regex", () => {
   const { filter, ...dataset } = FILTER_DATASET;
--- a/packages/usa-time-picker/src/test/time-picker-regex.spec.js
+++ b/packages/usa-time-picker/src/test/time-picker-regex.spec.js
@@ -1,6 +1,34 @@
 const assert = require("assert");
-const { generateDynamicRegExp } = require("../../../usa-combo-box/src/index");
 const { FILTER_DATASET } = require("../index");
+
+
+const generateDynamicRegExp = (filter, query = "", extras = {}) => {
+  const escapeRegExp = (text) =>
+    text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&");
+  
+  let find = filter.replace(/{{(.*?)}}/g, (m, $1) => {
+    const key = $1.trim();
+    const queryFilter = extras[key];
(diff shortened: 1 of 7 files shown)


packages/usa-time-picker/src/test/time-picker-regex.spec.js
```diff
@@ -1,7 +1,35 @@
 const assert = require("assert");
-const { generateDynamicRegExp } = require("../../../usa-combo-box/src/index");
```
mejiaj, line 2: @kodyjackson the component still relies on combo box. Is there a particular reason why we removed this?

mejiaj (commented):
@kodyjackson thanks for creating this. Left a question on time picker and can you double-check tests and `prettier:check` passes?

packages/usa-time-picker/src/test/time-picker-regex.spec.js
kodyjackson: The time-picker relied on an earlier version of `generateDynamicRegExp`, which matched strings by checking if any character from the list was present. The updated version, however, creates a regular expression that matches the starting characters of the list elements with the search query.

jaclinec:
Considering findings and recommendations from the [combo box landscape analysis](https://docs.google.com/document/d/1JfsnjWjIfbrKVB7JN9OsRcJXg306b7JX8xTTHf_MD_I/edit?usp=sharing) :lock: done earlier this year and conversations in CS UX sync from last April, it looks like we decided to try implementing a combo box solution similar to[ Elastic UI's](https://eui.elastic.co/#/forms/combo-box#sorting-matches) where first letter navigation is prioritized, but that it also returns results with characters contained in the search terms. Kind of a "best of both worlds." 

Elastic UI example:
Image

mejiaj:
Closing in favor of #6122.

PR #6048, opened 2024-09-05
USWDS - Utilities: Custom breakpoints

mahoneycm:
# Summary

*Continuation of #5985. Thank you for your contribution @jamigibbs.*

**Added new `$theme-utility-breakpoints-custom` setting.** This setting generates responsive variants of USWDS utility classes at custom breakpoints.

Being able to add a custom breakpoint to the utilities is useful for integrating USWDS with existing frameworks or other design requirements. Custom breakpoints will generate utilities alongside the default system breakpoints.

## Sample implementation
To generate a custom breakpoint for your USWDS utilities, add custom breakpoint values to a [Sass map](https://sass-lang.com/documentation/values/maps/) for `$theme-utility-breakpoints-custom` in your [USWDS settings configuration](https://designsystem.digital.gov/documentation/settings/#configuring-custom-uswds-settings). Here is an example implementation:

```scss
@use "uswds-core" with (
  $theme-utility-breakpoints-custom: (
    "new-breakpoint": 800px
  )
);
```

> [!Important]
> Custom breakpoints must be defined with `px` values. 

> [!Tip]
> Every breakpoint added to the system adds about 50kb to the compiled CSS. Be careful to only turn on the breakpoints you need for your project. You can turn off default USWDS breakpoints with the `$theme-utility-breakpoints` setting. More details about `$theme-utility-breakpoints` can be found  on the [Settings page](https://designsystem.digital.gov/documentation/settings/#utiltiies-settings). 

## Breaking change

This is not a breaking change.

## Related issue

Closes #5984

## Related pull requests

Continuation of #5985

[Site documentation update →](https://github.com/uswds/uswds-site/pull/2837)

## Preview link

[Preview link →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-custom-breakpoints/)

## Problem statement

It is currently not possible for users to implement custom breakpoints and use them alongside our utility classes

## Solution

Create new scss map for users to define their own breakpoints and build out utility classes with.

## Major changes

- Created `$theme-utility-breakpoints-custom` map for users to name and define their own breakpoints
- Created `px-to-user-em()` sass function
- Collect the breakpoints in system properties
- Update `_utility-builder.scss` to take custom breakpoints
- Update `at-media` mixin to create utility classes with custom breakpoints
- At this time, breakpoints can only be set in `px` units.
- Custom breakpoints follow the same `em` conversation methods as standard USWDS breakpoints
[...]

Issue #5984, opened 2024-07-19
USWDS - Feature: Add theme setting for custom breakpoints


jamigibbs:
### Is your feature request related to a problem? Please describe.

When the USWDS system is integrated into an existing site, there are sometimes additional breakpoints already integrated that do not exist in the USWDS system. This means that the USWDS utilities are incomplete because the system doesn't allow for adding additional breakpoints.

### Describe the solution you'd like

I would like to be able to use the USWDS system breakpoints as well as optionally add a custom breakpoint for generating utility classes.

### Additional context

The VA.gov system uses a breakpoint called `medium-screen` which is 768px (or 48rem). This breakpoint does not exist in the USWDS system but we would like to continue to use it along with the other breakpoints USWDS offers. This breakpoint is being used in [167+ files in VA.gov](https://github.com/search?q=repo%3Adepartment-of-veterans-affairs%2Fvets-website%20medium-screen&type=code).

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Feature+Request%22) for duplicate feature requests.

mejiaj:
Moving issue back to `In progress` while we quickly look into supporting layout grid classes based on custom breakpoint.

CC @amyleadem

PR #6083, opened 2024-09-26
USWDS - Layout grid utility: Custom breakpoints 

amyleadem:
# Summary

A potential addition to PR #6048

Added support for generating custom breakpoints for layout grid utility classes.

## Breaking change

## Related issue

Related to #5984

## Related pull requests

Potential addition to  #6048

## Preview link
[Storybook](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-custom-breakpoints/?path=/story/design-tokens-fonts--fonts)

## Problem statement

In PR #6048, we added support to create custom breakpoints with the theme setting `$theme-utility-breakpoints-custom`. However, this setting did not generate custom breakpoints for layout grid. 

## Solution

This PR generates custom responsive variants for `grid-col`, `grid-gap`, `grid-offset`, and `grid-container` classes.
    
> [!note]
> The media queries for grid-container have some pre-existing quirks that should probably be broken out into a separate issue. Here is an example:
> ```
> @media all and (min-width: 30em) and (min-width: 64em){
>   .mobile-lg\:grid-container-card{
>     padding-left:2rem;
>     padding-right:2rem;
>   }
> }
> ```
> These media queries have multiple `min-width` declarations, which are redundant and at times contradictory. This PR does not attempt to address this pre-existing issue.

### Performance impact
Including custom breakpoints for layout grid classes adds 8kb to the compiled CSS file. Here are the details on CSS size:

| Description | CSS size |
|:--------|:--------|
| With default breakpoints | 524kb |
| With 1 new custom breakpoint _without_ layout grid (`cm-custom-breakpoints` branch) | 572kb |
| With 1 new custom breakpoint _with_ layout grid (this branch) | 580kb | 

## Testing and review
- Checkout the [`al-test-custom-breakpoints` branch in uswds-sandbox](https://github.com/uswds/uswds-sandbox/tree/al-test-custom-breakpoints). 
- Use the following code in [`uswds-theme.scss`](https://github.com/uswds/uswds-sandbox/blob/al-test-custom-breakpoints/src/_styles/_uswds-theme.scss):
  ```scss
  @use "uswds-core" with (
    $theme-utility-breakpoints-custom: (
      "new-breakpoint": 800px
    ),
    $theme-utility-breakpoints: (
      widescreen: true,
    )
  );
  ```
- Run `npx gulp buildSass` and navigate to  `_site/assets/css/styles.css`
- In styles.css, search for:
    - `.widescreen\:grid-container` and `.new-breakpoint\:grid-container`. Confirm that the style rules match for the two rules, with the exception of a different media query width. 
    - `.widescreen\:grid-gap` and `.new-breakpoint\:grid-gap`. Confirm that the style rules match for the two rules, with the exception of a different media query width. 
    - `.widescreen\:grid-col` and `.new-breakpoint\:grid-col`. Confirm that the style rules match for the two rules, with the exception of a different media query width. 
    - `.widescreen\:grid-offset` and `.new-breakpoint\:grid-offset`. Confirm that the style rules match for the two rules, with the exception of a different media query width.

### DIFF
--- a/packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
+++ b/packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
@@ -6,6 +6,9 @@
 
 $namespace-grid: ns("grid");
 
+$custom-breakpoints: map-deep-get($system-properties, breakpoints, extended);
+$all-breakpoints: map-collect($system-breakpoints, $custom-breakpoints);
+
 // basic container
 .#{$namespace-grid}container {
   $props: append-important($grid-global, $theme-grid-container-max-width);
@@ -56,7 +59,7 @@ $namespace-grid: ns("grid");
   }
 
   // responsive column gaps
-  @each $mq-key, $mq-value in $system-breakpoints {
+  @each $mq-key, $mq-value in $all-breakpoints {
     @if map.get($theme-utility-breakpoints-complete, $mq-key) {
       @include at-media($mq-key) {
         @each $gap-key,
@@ -103,7 +106,7 @@ $namespace-grid: ns("grid");
 }
 
 // responsive columns
-@each $mq-key, $mq-value in $system-breakpoints {
+@each $mq-key, $mq-value in $all-breakpoints {
   @if map.get($theme-utility-breakpoints-complete, $mq-key) {
     @include at-media($mq-key) {
       .#{$mq-key}#{$separator}#{$namespace-grid}col {
@@ -142,7 +145,7 @@ $namespace-grid: ns("grid");
 }
 
 // responsive offsets
-@each $mq-key, $mq-value in $system-breakpoints {
+@each $mq-key, $mq-value in $all-breakpoints {
   @if map.get($theme-utility-breakpoints-complete, $mq-key) {
     @each $width-key, $width-value in $system-layout-grid-widths {
       @include at-media($mq-key) {
--- a/packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
+++ b/packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
@@ -16,7 +16,7 @@ $all-breakpoints: map-collect($system-breakpoints, $custom-breakpoints);
 }
 
 // container with custom widths
-@each $width-key, $width-value in $system-breakpoints {
+@each $width-key, $width-value in $all-breakpoints {
   .#{$namespace-grid}container-#{$width-key} {
     $props: append-important($grid-global, $width-key);
     @include grid-container($props);
@@ -24,7 +24,7 @@ $all-breakpoints: map-collect($system-breakpoints, $custom-breakpoints);
 }
 
 // responsive containers...
-@each $mq-key, $mq-value in $system-breakpoints {
+@each $mq-key, $mq-value in $all-breakpoints {
   @if map.get($theme-utility-breakpoints-complete, $mq-key) {
     @include at-media($mq-key) {
       .#{$mq-key}#{$separator}#{$namespace-grid}container {
@@ -33,7 +33,7 @@ $all-breakpoints: map-collect($system-breakpoints, $custom-breakpoints);
       }
 
       // ...with custom widths
-      @each $width-key, $width-value in $system-breakpoints {
+      @each $width-key, $width-value in $all-breakpoints {
         .#{$mq-key}#{$separator}#{$namespace-grid}container-#{$width-key} {
           $props: append-important($grid-global, $width-key);
           @include grid-container($props);
--- a/packages/uswds-core/src/styles/mixins/layout-grid/_grid-container.scss
+++ b/packages/uswds-core/src/styles/mixins/layout-grid/_grid-container.scss
@@ -1,9 +1,10 @@
 @use "sass:list";
+@use "sass:map";
 @use "../../settings" as *;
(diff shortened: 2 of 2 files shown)


packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
```diff
@@ -6,6 +6,9 @@
 
 $namespace-grid: ns("grid");
 
+$custom-breakpoints: map-deep-get($system-properties, breakpoints, extended);
+$all-breakpoints: map-collect($system-breakpoints, $custom-breakpoints);
```
amyleadem, line 10: >[!note]
> Interestingly, layout grid and icon list size use [$system-breakpoints in breakpoints.scss](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/tokens/units/breakpoints.scss) and the other utilities build from [`$system-properties.breakpoints` in _properties.scss](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/_properties.scss#L157). Both seem to collect the same values:

```scss
/*breakpoints.scss*/
$system-breakpoints: general.map-collect(
  map.get(spacing.$system-spacing, "large"),
  map.get(spacing.$system-spacing, "larger"),
  map.get(spacing.$system-spacing, "largest")
);
```

```scss
/*_properties.scss*/
  breakpoints: (
    standard:
      map-collect(
        map.get($system-spacing, "large"),
        map.get($system-spacing, "larger"),
        map.get($system-spacing, "largest")
      ),
    extended: (),
  ),
```

Flagging in case we want to standardize this and potentially replace `$system-breakpoints` with `$system-properties, breakpoints, standard` here. To keep things scoped, I am leaving `$system-breakpoints` as is for now.

packages/uswds-core/src/styles/mixins/layout-grid/_grid-container.scss
```diff
@@ -13,7 +14,11 @@
     $props: list.nth($props, 1);
     $margin-x: append-important($props, auto);
     @include u-margin-x($margin-x);
-    @include u-maxw($props);
+    @if map.has-key($custom-breakpoints, $props) {
+      max-width: map.get($custom-breakpoints, $props);
+    } @else {
+      @include u-maxw($props);
+    }
```
amyleadem, line 21: @mejiaj not sure if nested conditional statements are ideal here. Let me know if you have suggestions to clean this up!

packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
mejiaj: @amyleadem good find. Do you have an initial recommendation?

packages/uswds-core/src/styles/mixins/layout-grid/_grid-container.scss
mejiaj: No issues. We can revisit if we run into anything.

packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
amyleadem: We could remove the reference to `$system-breakpoints` and instead use the [same technique from _utility-builder.scss](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/mixins/_utility-builder.scss#L356) to grab `$our-breakpoints` from `$system-properties`. Initial testing on this branch shows that this change creates the correct number of references in the CSS, but it would need further testing to confirm that the generated CSS is correct. What do you think - should we pursue now or explore as a possible follow-up?


```suggestion
$our-breakpoints: map-deep-get($system-properties, breakpoints, standard);
$all-breakpoints: map-collect($our-breakpoints, $custom-breakpoints);
```

mejiaj (approved):
@amyleadem a nice addition, good work.

I was able to confirm new layout utils are generated when I add a custom breakpoint.

**Input**
```scss
// _settings-utilities.scss
$theme-utility-breakpoints-custom: (
  "jumbo": 2000px,
)
```

**Output**
```css
// uswds.css
/* … */
@media all and (min-width: 125em){
  .grid-row.jumbo\:grid-gap-0{
    margin-left:0;
    margin-right:0;
  }
  .grid-row.jumbo\:grid-gap-0 > *{
    padding-left:0;
    padding-right:0;
  }
```

packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
mejiaj: @amyleadem inclined to say follow-up so as to not delay the release.

RachelCorsino (approved):
Confirming the following:
- `.widescreen\:grid-container` and `.new-breakpoint\:grid-container` match after updating custom breakpoint
- `.widescreen\:grid-gap` and `.new-breakpoint\:grid-gap` match after updating custom breakpoint
- `.widescreen\:grid-col` and `.new-breakpoint\:grid-col` match after updating custom breakpoint
- `.widescreen\:grid-offset` and `.new-breakpoint\:grid-offset` match after updating custom breakpoint

Thus, everything looks good to me!

amyleadem:
Since we have two approvals on this, I'm going to merge it into the `cm-custom-breakpoints` branch.

PR #6093, opened 2024-10-04
USWDS - Footer: Show overflow (for focus rings)

### LINKED ISSUE #6092: USWDS - Bug: Focus ring in footer links is cut off
### Describe the bug

The focus ring around footer links is cut off.

Image

1. Create a footer using code like so:
```html
<!DOCTYPE html>
<html>
<head>
  <link href="https://cdn.jsdelivr.net/npm/@uswds/uswds@3.9.0/dist/css/uswds.min.css" rel="stylesheet">
</head>
<body class="display-flex flex-column" style="min-height: 100vh;">
  <main class="flex-1"></main>
  <footer class="usa-footer usa-footer--medium">
    <div class="usa-footer__primary-section">
      <nav class="usa-footer__nav">
        <ul class="grid-row grid-gap">
          <li class="desktop:grid-col-auto usa-footer__primary-content mobile-lg:grid-col-4">
            <a class="usa-footer__primary-link" href="#">Test</a>
          </li>
          <li class="desktop:grid-col-auto usa-footer__primary-content mobile-lg:grid-col-4">
            <a class="usa-footer__primary-link" href="#">Test</a>
          </li>
        </ul>
      </nav>
    </div>
  </footer>
</body>
</html>
```
2. Observe that the top half of the focus ring is cut off

### Expected Behavior

The focus ring should show fully around the links in the footer.

Image

### Related code

_No response_

### Screenshots

Current behavior:

Image

Expected behavior:

Image

### System setup

- USWDS version: 3.9.0
- browser: all major browsers

### Additional context

_No response_

### Code of Conduct

6TELOIV:
# Summary
Remove `overflow: hidden` from `usa-footer` to let focus ring show.

## Breaking change
None

## Related issue
Closes #6092

## Related pull requests
None

## Preview link
None

## Problem statement
Focus rings are visually cut off at the top in the footer.

## Solution
Removing `overflow: hidden` allows focus rings to show at the edge of the footer.

## Testing and review
Should visually review the various recommended `usa-footer` layouts to make sure this doesn't introduce any overflow bugs.


amyleadem (changes requested):
Thanks for submitting this PR @6TELOIV. The focus outline now appears on all sides of the link elements. 

I noticed that there is a small visual regression in the mobile view, where the list element borders are now visible outside the footer container. I've attached a screenshot that demonstrates this:

image

This is likely not typically visible in project environments where the footer spans from edge to edge, but it is something that should be resolved. I do not see any other visual regressions after removing `overflow: hidden`. 

One possible fix for this is to move the `border-top` that is currently set on `.usa-footer__primary-content` to `.usa-footer__primary-link`.

Please let us know if you have any questions.

6TELOIV:
I think the solution to that regression is to move `border-top: 1px solid #adadad;` to the `.usa-footer__primary-link` and `.usa-footer__primary-link--button` classes. Currently they're on the grid item, but should probably be on the child instead (based on the design).

Additionally, this fixes a different visual inconsistency in the Slim footer. notice the difference in length of the border on the primary links vs the contact links.

| Version | Image |
|-|-|
| Current | image |
| Fixed | image |

packages/usa-footer/src/styles/_usa-footer.scss
```diff
@@ -161,14 +162,6 @@ $-chevron-expand-more: map-merge(
...
 }
 
-.usa-footer__primary-content {
-  border-top: 1px solid color("base-light");
-
-  @include at-media("mobile-lg") {
-    border: none;
-  }
-}
-
```
mahoneycm, line 171: **Issue (blocking):** While this fixed default and slim variants, we've lost the border on the big footer accordion buttons:

| Before | After |
|--------|--------|
| image | image | 

**Suggestion:** Since the markup varies between variants we might just want to store this border as a variable and reapply it for the `.usa-footer__primary-link-button` class:

```scss
$primary-content-border: 1px solid color("base-light")

.usa-footer__primary-link {
   border-top: $primary-content-border;
    ...
    &-button {
    border: 0;
    border-top: $primary-content-border;
    ...
    };
};
```

If there's a cleaner way to accomplish this I'm open to other suggestions as well

mahoneycm (changes requested):
Thanks for keeping up with this @6TELOIV ! The default and slim variants are looking great!

Left one suggestion to resolve a big footer variant regression.

packages/usa-footer/src/styles/_usa-footer.scss
6TELOIV: alternatively, wouldn't elevating the border reset to the link itself, and then removing the border reset from the button work?
```scss
.usa-footer__primary-link {
  border: 0;
  border-top: 1px solid color("base-light");
  /* ... */
  &--button {
    /* remove border: 0 */
  }
}```

mahoneycm: Good idea! I hadn't considered that. If that works without regression then we can take that route!

6TELOIV: From my testing, this seems to fix the issue, and IMHO it is a little cleaner, as it keeps the border definition in one place for all footer primary links, regardless of if they're truly links or instead collapsible buttons. Let me know your thoughts, and I can submit the change @mahoneycm

6TELOIV: Sounds good, I think we commented at the same time. Will make the change.

amyleadem:
Thanks for the update @6TELOIV. I found one more minor visual regression in the slim footer (see screenshot below). 

@mahoneycm, @cathybaptista, @mejiaj 
I opened up PR #6237 to explore a solution that removes the mobile x-padding from `.usa-footer__primary-container`. This could be considered scope creep, so I partitioned it out for now. However it solves two problems at once: 
1. Fixes the (minor) visual regression from this PR 
1. Fixes the pre-existing issue in the slim footer variant that caused the top border of the list items to not extend the full width of the component. 

If we like this solution, we can either incorporate that change in this branch or just merge in that PR. Let me know what you think. 

Regression in slim footer variant:
image

6TELOIV:
@amyleadem sorry, I haven't done reviews of my own before, usually I'm requesting others to review! I think #6237 makes sense and just merging that and closing this one seems good (correct me if I'm misunderstanding anything here!).

PR #6094, opened 2024-10-04
Add notifications for 3.9.0

thisisdano:
This adds the breaking changes and markup changes from `3.9.0` into the Sass notifications.

```
3.9.0
Release notes: github.com/uswds/uswds/releases/tag/v3.9.0

- Accessibility, Display: [usa-header] Removed the CSS order property from the mobile
  view in standard variants of the header component. Now, the visual order of the
  component matches the tab order.
- Accessibility, Markup: [usa-footer] Added the autocomplete="email" attribute to
  the big footer variant and the "Create an account" template. This attribute allows
  the components to meet the standards outlined in WCAG 1.3.5. Teams should update their
  markup if they use an email field in their big footer.
- Accessibility, Markup: [usa-identifier] Updated the USA.gov link in Spanish versions
  of the identifier. The link text now reads "Visite USAGov en Español" and the link
  url is now "https://www.usa.gov/es/". Teams should update this text if they use the
  Spanish-language identifier.
- Accessibility, Markup: [usa-memorable-date] Removed numeric representation of
  months in the memorable date component. Recent usability testing indicated that having
  both numbers and names to represent months was confusing for screen reader users.
  Teams should update their memorable date component to remove the leading numbers.
- Display: [usa-alert, usa-site-alert] Fixed a bug that caused
  $theme-site-margins-width to unexpectedly adjust the alignment inside the alert and
  site alert components. Alignment on the alert and site alert components will likely
  shift from this change. Confirm that your implementation of the component aligns
  as expected.
- Display: [usa-button] Updated the width of unstyled buttons at narrow screen widths.
  Now, unstyled buttons receive a width of auto to better match USWDS link styles.
  Users should confirm that the variant visually displays as expected in their projects.
- Display: [usa-card] Fixed a bug that caused the component to ignore the
  $theme-card-font-family setting. Confirm that your implementation of the card
  component displays with the expected font family.
```

Note that they get a bit hidden at compile time because of all the Sass deprecation errors. We'll want to address those errors asap, not only to fix the code, but because it adds noise on top of potentially useful information.


amyleadem (approved):
LGTM! 
- Confirmed all breaking and markup changes are accounted for
- Confirmed the descriptions are accurate
- Confirmed the notifications display as expected in the terminal

PR #6146, opened 2024-10-23
USWDS - Step indicator: Remove aria-label from wrapper

### LINKED ISSUE #6113: USWDS - Bug: Accessibility Aria Failure on Step Indicator
### Describe the bug

According to https://www.siteimprove.com/ `aria-label` is a prohibited for use in a `<div>`

All uses of the USWDS Step Indicator are failing accessibility test because of this.

1. Use the code from USWDS Step Indicator: https://designsystem.digital.gov/components/step-indicator/

Image

### Expected Behavior

Code should pass accessibility check for all conformance levels

### Related code

<div class="usa-step-indicator" aria-label="progress">

### Screenshots

Image

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

amyleadem:
# Summary

**Removed the aria-label from the wrapper of the step indicator component.** This resolves an automated testing error related to having an invalid attribute on a `div` element. 

## Markup change
To remove automated testing errors, users must update the markup to remove the `aria-label` on the `usa-step-indicator` element:
```diff 
- <div class="usa-step-indicator" aria-label="progress">
+ <div class="usa-step-indicator">

```

## Breaking change

## Related issue

Closes #6113 

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2892)

## Preview link

[Step indicator component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-step-indicator-label/?path=/story/components-step-indicator--default)

## Problem statement

The step indicator component fails SiteImprove accessibility tests because it has an invalid `aria-label` attribute on a `<div>` element.

## Solution

This PR removes the `aria-label` from the `.usa-step-indicator` div element. This was done for the following reasons:
- The `aria-label` attribute is not valid on a `<div>` element. In some cases this means the label is not read out, and in others, it causes accessibility scan failures.
- This `aria-label` was determined to be an unnecessary addition. The accessibility team reported that user interviews suggested the label did not provide necessary context.
    - Note: JAWS does not currently read out the aria-label, but VoiceOver does

## Testing and review

1. If available, confirm that this issue is now resolved in SiteImprove scans
2. Confirm that the `aria-label` has been successfully removed
3. Confirm that removing the `aria-label` is appropriate
4. Confirm no regressions

### DIFF
--- a/packages/usa-step-indicator/src/usa-step-indicator.twig
+++ b/packages/usa-step-indicator/src/usa-step-indicator.twig
@@ -1,4 +1,4 @@
-<div class="usa-step-indicator {{ modifier }}" aria-label="progress">
+<div class="usa-step-indicator {{ modifier }}">
   <ol class="usa-step-indicator__segments"{% if noLabels %} aria-hidden="true"{% endif %}>
     {% for step in steps %}
       <li class="usa-step-indicator__segment{% if step.status == "complete" %} usa-step-indicator__segment--complete{% endif %}{% if step.status == "current" %} usa-step-indicator__segment--current{% endif %}"{% if step.status == "current" and showLabels %} aria-current="true"{% endif %}>


amyleadem:
@amycole501 and/or @alex-hull, can you confirm that this fix resolves the SiteImprove error described in the related issue?

amycole501 (approved):
Removing the label is the cleanest and most straightforward fix. Thank you!

packages/usa-step-indicator/src/usa-step-indicator.twig
```diff
@@ -1,4 +1,4 @@
-<div class="usa-step-indicator {{ modifier }}" aria-label="progress">
+<div class="usa-step-indicator {{ modifier }}">
```
mahoneycm, line 1: **Note (non-blocking):** I noticed that my screen reader no longer focuses the entire time picker "group".

Before:

1. Screen reader focuses entire component group
2. Screen reader then focuses step indicator bar and reads through steps
3. Screen reader then reads out header element

After:

1. Screen reader focuses step indicator bar and reads through steps
2. Screen reader then focuses header element

This doesn't seem to be an issue but wanted to flag since the behavior changes slightly

mahoneycm (approved):
Looks good to me! Left one comment about a small change in screen reader behavior.

### Testing checklist

- [x]  If available, confirm that this issue is now resolved in SiteImprove scans
- [x]  Confirm that the `aria-label` has been successfully removed
- [x]  Confirm that removing the `aria-label` is appropriate
- [x]  Confirm no regressions

packages/usa-step-indicator/src/usa-step-indicator.twig
amyleadem: Thanks for flagging!

@amycole501 and @alex-hull, would this change in behavior count as an accessibility regression? Wanting to check in with you both before moving forward with this.

alex-hull: I agree with Charlie. I don't think this will be an issue. It appears that the flow is not that much different between the before and after interaction, so that should not be too big of a change. I also think as long as the user hears what current step they are on, this should not be an issue. @amyleadem

alex-hull (approved):
Address changes to focus on whole component. Approve on my end.

thisisdano (approved):

PR #6085, opened 2024-09-27
USWDS - Breadcrumb: Add keyboard navigation for current page's link

kodyjackson:
# Summary

This pull request enhances the accessibility of the breadcrumb component by ensuring that keyboard and screen reader users can interact with the current page link. This improvement aims to create a more inclusive navigation experience for all users.

## Breaking change

This is not a breaking change.

## Related issue

Closes #6054 

## Problem statement

The breadcrumb component was not fully interactive for keyboard users and screen reader users. Specifically, the last link, which represents the current page, was not reachable or identifiable through keyboard navigation, making it difficult for these users to understand their current location within the site.

## Solution

To resolve this issue, we made the following changes:

  1. Changed the Last Breadcrumb Item: The last breadcrumb link was modified to use an anchor (<a>) tag instead of a span. This allows it to be focusable and interactable via keyboard navigation.
  
  2. Added Accessibility Attributes: We included aria-current="page" to the last breadcrumb item, ensuring screen readers announce it as the current page.
   
  3. Styled the Link: We introduced a new class (.usa-breadcrumb__link--no-style) that removes any underlines and prevents the link from changing color when clicked, maintaining visual consistency.

These adjustments ensure that the breadcrumb component is fully accessible, allowing all users to navigate and understand their position within the site effectively.

PR #6148, opened 2024-10-24
USWDS - Breadcrumb: Make current page keyboard focusable

### LINKED ISSUE #6054: USWDS - Breadcrumb: Enhance breadcrumb and make current page keyboard focusable
### Describe the bug

The breadcrumb component isn't fully interactive with a keyboard. Keyboard users and screen reader users aren't able to hear/interact with the current page's link.

1. Go to https://designsystem.digital.gov/components/breadcrumb/
2. Try using your keyboard to tab into the current page in the example code
3. Try listening to the current page in a screen reader

### Expected Behavior

Go to https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/ to see how the current page link is handled. It is both keyboard focusable and a screen reader is able to tell which page they're on without having to listen to the whole list. The current page is in the focus order.

### Screenshots

_No response_

### System setup

Chrome, PC< JAWS

### Additional context

_No response_

### Code of Conduct

mejiaj:
# Summary

**Breadcrumb current page is now focusable.** Users can now tab to focus to breadcrumb's current page.

### Markup changes

```diff
<!-- usa-breadcrumb.html -->
- <li class="usa-breadcrumb__list-item usa-current" aria-current="page">
-  <span>Economically disadvantaged women-owned small business federal contracting program</span>
+ <li class="usa-breadcrumb__list-item usa-current">
+   <a href="" class="usa-breadcrumb__link" aria-current="page">
       Economically disadvantaged women-owned small business federal contracting program
+   </a> 
</li>
```

Users will need to update to the following markup for the current breadcrumb item:
```html
<!-- project-breadcrumb.html -->
<li class="usa-breadcrumb__list-item usa-current">
  <a href="" class="usa-breadcrumb__link" aria-current="page">
    {{ PROJECT_CURRENT_PAGE_TITLE }}
  </a>
</li>
```

## Breaking change

This is not a breaking change. The [original a11y test](https://designsystem.digital.gov/components/breadcrumb/accessibility-tests/) passes. No failures if users aren't able to update right away.

## Related issue

Closes #6054.

## Related pull requests

Changelog PR: uswds/uswds-site#2918

## Preview link

image

| Before | After |
| :----- | :---- |
| [Develop] |  [Feature]       |

[Develop]: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-breadcrumb--default&p=all
[Feature]: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-breadcrumb-current-focusable/?path=/story/components-breadcrumb--default

## Problem statement

Users can't tab to focus on the current breadcrumb link.

## Solution

Follows Also follows recommended markup from W3 [Breadcrumb pattern](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/examples/breadcrumb/).

## Major changes

1. **Updated breadcrumb current page markup.** The current breadcrumb page is now a link, instead of a span.
2. **Fixed mixin `add-link-styles`.** This mixin used color util function, but the output was incorrect due to missing import. Added SASSDoc comments and unit tests.
3.  **Current page link appears as normal text.** Instead of using USA link primary colors, this uses the theme text color to make it appear like normal text.
4. **Removed unstyled button colors.** Redundant because we're already setting link styles on `usa-breadcrumb__link`. It's also outputting _a lot_ more CSS than we actually need.

## Testing and review

1. Visit Breadcrumb [default] and [wrap] variant.
2. Confirm current page is keyboard focusable
3. Confirm that current page link looks like normal text.

[default]: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-breadcrumb-current-focusable/?path=/story/components-breadcrumb--default
[wrap]: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jm-breadcrumb-current-focusable/?path=/story/components-breadcrumb--wrap

### DIFF
--- a/packages/usa-breadcrumb/src/styles/_usa-breadcrumb.scss
+++ b/packages/usa-breadcrumb/src/styles/_usa-breadcrumb.scss
@@ -185,6 +185,14 @@ $breadcrumb-back-icon-aspect: (
   }
 }
 
+.usa-breadcrumb__link[aria-current="page"] {
+  @include add-link-styles(
+    $theme-text-color,
+    $hover: $theme-text-color,
+    $active: $theme-text-color
+  );
+}
+
 // ---------------------------------
 // Variations
 // ---------------------------------
--- a/packages/uswds-core/src/styles/mixins/helpers/add-link-styles.scss
+++ b/packages/uswds-core/src/styles/mixins/helpers/add-link-styles.scss
@@ -1,6 +1,35 @@
 @use "sass:meta";
 @use "sass:list";
+@use "../../functions/utilities/color" as *;
 
+/// Adds link default and state (visited, hover, active) styles.
+///
+///
+/// @param {List} $color - A list of colors for default, hover, and active.
+/// @param {String} $hover [null] - The link's hover color.
+/// @param {String} $active [null] - The link's active color.
+///
+/// @example
+/// @include add-link-styles(
+///   $theme-text-color,
+///   $hover: "primary",
+///   $active: "primary-darker"
+/// );
+///
+/// @output
+/// &:link{
+///   color:#1b1b1b;
+/// }
+/// &:visited{
+///   color:#1b1b1b;
+/// }
+/// &:hover{
+///   color:#005ea2;
+/// }
+/// &:active{
+///   color:#162e51;
+/// }
+///
 @mixin add-link-styles($color, $hover: null, $active: null) {
   @if meta.type-of($color) == "list" {
     $active: list.nth($color, 3);
--- a/packages/usa-breadcrumb/src/usa-breadcrumb.twig
+++ b/packages/usa-breadcrumb/src/usa-breadcrumb.twig
@@ -15,8 +15,10 @@
         <span>Contracting assistance programs</span>
       </a>
     </li>
-    <li class="usa-breadcrumb__list-item usa-current" aria-current="page">
-      <span>Economically disadvantaged women-owned small business federal contracting program</span>
+    <li class="usa-breadcrumb__list-item usa-current">
+      <a href="" class="usa-breadcrumb__link" aria-current="page">
+        Economically disadvantaged women-owned small business federal contracting program
+      </a>
     </li>
   </ol>
 </nav>
--- a/packages/uswds-core/src/test/tests.scss
+++ b/packages/uswds-core/src/test/tests.scss
@@ -685,4 +685,50 @@ $this-function: "is-color-dark()";
   }
 }
 
+@include describe("[mixin] add-link-styles()") {
+  @include it("Adds link styles from a SASS list of color tokens") {
+    @include assert {
+      @include output {
+        $link-colors: ("primary", "primary-vivid", "primary-darker");
+        @include add-link-styles($link-colors);
+      }
+      @include expect {
+        &:link {
+          color: #005ea2;
+        }
+        &:visited {
+          color: #005ea2;
+        }
+        &:hover {
+          color: #0050d8;
+        }
+        &:active {
+          color: #162e51;
+        }
+      }
+    }
+  }
+  @include it("Adds link styles from color tokens") {
+    @include assert {
+      @include output {
+        @include add-link-styles("ink", "primary", "primary-darker");
+      }
+      @include expect {
+        &:link {
+          color: #1b1b1b;
+        }
+        &:visited {
+          color: #1b1b1b;
+        }
+        &:hover {
+          color: #005ea2;
+        }
+        &:active {
+          color: #162e51;
+        }
+      }
+    }
+  }
+}
+
 // Debug


packages/uswds-core/src/styles/mixins/helpers/add-link-styles.scss
```diff
@@ -1,6 +1,35 @@
 @use "sass:meta";
 @use "sass:list";
+@use "../../functions/utilities/color" as *;
```
mejiaj, line 3: > [!TIP]
> Color tokens weren't working as expected because of missing util. Imported missing util, added docs, and created a unit test to ensure this works as expected.

amyleadem (changes requested):
Looks good! Just one small question about the empty href in the twig file and a request to move the markup change documentation higher up in the PR description. 

- [x]  Confirm that the updated markup structure matches the W3C recommendations and makes sense
- [x]  Confirm the visual presentation matches W3C recommended pattern (No text underline, default text color)
    - Note: W3C uses bolded text on the current link but that is ok here. It is not outlined as a requirement in the w3c docs.
- [x]  Confirm that hover, focus, and active states of the current page link appear as expected (does not change style across states)
- [ ]  Confirm the PR description has appropriate summary and instructions
    - [ ]  Examples make sense, but I'd recommend the section gets moved higher up in the PR description, potentially right below the summary. I have been adding a “Markup change” section right next to “Breaking change” for this in my recent PRs ([example](https://github.com/uswds/uswds/pull/6146)).I think it would be good to create a consistent place for users to find markup changes, and I like it being in the top part of the description for easy understanding.
- [x]  Confirm there is no break if the markup is not updated
- [x]  Confirm no visual or behavioral regressions in breadcrumb
- [x]  Confirm no visual regressions as a result of the fix in `add-link-styles` mixin
    - Note: This new reference to `@include add-link-styles` is the only use of the mixin in the system. As a result, there are no related regressions. Current page link appears as expected in all states.
- [x]  Confirm Sass unit test  for add-link-styles works as expected
- [x]  Confirm documentation for `add-link-styles` makes sense
- [x]  Confirm looks good in mobile
- [x]  Confirm current page status is clear in VoiceOver
- [x]  Confirm unit tests pass

packages/usa-breadcrumb/src/styles/_usa-breadcrumb.scss
```diff
@@ -185,6 +185,14 @@ $breadcrumb-back-icon-aspect: (
...
 }
 
+.usa-breadcrumb__link[aria-current="page"] {
+  @include add-link-styles(
+    $theme-text-color,
+    $hover: $theme-text-color,
+    $active: $theme-text-color
+  );
+}
+
```
mahoneycm, line 195: **Question:** Wouldn't this be considered an a11y failure since there is no visual indication that it's a link?

From my understanding, [this is an example of a failure to comply with criterion 1.4.1](https://www.w3.org/TR/WCAG20-TECHS/F73.html). WCAG recommends [two visual indicators](https://www.w3.org/TR/WCAG20-TECHS/G182.html) that it is different from normal body text.

CC: A11y team @alex-hull @amyleadem

mahoneycm (commented):
Looking good! Leaving a question about an accessibility concern with the current link styles.

packages/usa-breadcrumb/src/styles/_usa-breadcrumb.scss
amycole501: It is focusable, however, and is read aloud in screen readers as "current".

mahoneycm: @amyleadem the thing is with this example is that I think the bolded text counts as a "visual indicator other than color"

It's similar to [this example](https://arc.net/l/quote/cxhitkjm) provided by W3C

> A news site lists links to the articles appearing on its site. Additional information such as the section the article appears in, the time the article was posted, a related location or an indication that it is accompanied by live video appears in some cases. The links to the articles are in a different color than the additional information but **the links are not underlined, and each link is presented in a larger font** than the rest of the information so that users who have problems distinguishing between colors can identify the links more easily.

mahoneycm: I'm uncertain if this is _only_ required if a different color is used, and I'm having trouble finding a criterion that denotes a link must have visual indicators. I feel like we've identified this as an accessibility issue in the past 🤔

mejiaj:
Moved to draft because we're going to take a closer look at UX and A11Y concerns.

opensource-joe:
The need here is still real (#6054 open, nothing else fixing it), and closing #6085 in favor of this one made sense. Two things worth flagging. First, this no longer applies cleanly to develop, the breadcrumb stylesheet picked up a wrapping/zoom fix for #6689 in the same region this touches, so at least that hunk needs a rebase. Second, on the a11y thread that's been open since October: I read the actual diff rather than just the conversation, and the new .usa-breadcrumb__link[aria-current="page"] rule only sets color, matching the surrounding text exactly in every state. So it's not a question of whether bold-without-underline is enough, there's currently no visual indicator at all distinguishing the current-page link from plain text. Might be worth deciding that explicitly (bold, a subtle underline, something) before this comes out of draft.

ryparker:
@coderabbitai full review

PR #6171, opened 2024-11-01
USWDS - Core: Fix test task to include all JavaScript test files

aduth:
# Summary

Fixes automated tests to include all files.

## Breaking change

## Related pull requests

Split from #5793

## Preview link

N/A

## Problem statement

As a developer, I expect that failing tests will be included and trigger a failing build in continuous integration, so that I have confidence that my code is working as intended.

## Solution

Include all test files in Gulp test task.

Previously, test files included in `uswds-*` packages were not included.

## Testing and review

Include a failing test assertion in a spec file for a `uswds-*` package.

Example:

```diff
diff --git a/packages/uswds-core/src/js/utils/test/select.spec.js b/packages/uswds-core/src/js/utils/test/select.spec.js
index b628239cf..9d63723ef 100755
--- a/packages/uswds-core/src/js/utils/test/select.spec.js
+++ b/packages/uswds-core/src/js/utils/test/select.spec.js
@@ -19,6 +19,7 @@ describe("select", () => {
   });
 
   it("returns an empty array if given a non-string selector", () => {
+    assert(false);
     assertArrayWithLength(select(undefined), 0);
   });
```

Running `npx gulp unitTests` should result in a non-zero exit code (a failure). On `develop`, this currently passes.


mahoneycm (commented):
Lgtm! The five JS unit tests within `uswds-core` are now appropriately running with the rest of the unit tests.

**question:** I noticed that there are two unit tests in the [src/test](https://github.com/uswds/uswds/tree/b5a332a49169db5ccebc209059283f05a721d7f1/src/test) directory that are not included in these tests. Should these be included as well?

aduth:
> **question:** I noticed that there are two unit tests in the [src/test](https://github.com/uswds/uswds/tree/b5a332a49169db5ccebc209059283f05a721d7f1/src/test) directory that are not included in these tests. Should these be included as well?

It does look like those two aren't being referenced anywhere currently, so I agree they should be included somehow.

But I'm less clear _where_ they should be included, since tests are currently broken down between `unitTests` and `sassTests`, and those two files seem to be more on the side of compilation / Sass tests.

A couple ideas:

- We could include them in the existing `sassTests` task. (Based on their content, they feel more relevant to group with the existing Sass tests)
- We could dissolve the groupings and have a singular test task, which may also allow for simpler file path patterns (`**/test/*.spec.js`)

I also noticed that [we only pass the Mocha configuration for `unitTests`](https://github.com/uswds/uswds/blob/b5a332a49169db5ccebc209059283f05a721d7f1/tasks/test.js#L12) and [not `sassTests`](https://github.com/uswds/uswds/blob/b5a332a49169db5ccebc209059283f05a721d7f1/tasks/test.js#L17). The Mocha configuration includes [DOM global registration](https://github.com/uswds/uswds/blob/b5a332a49169db5ccebc209059283f05a721d7f1/packages/uswds-core/src/js/utils/test/.mocharc.json#L3) which might imply that the separation is partly to do with whether a global DOM is needed for the tests, but there's other configuration that isn't directly related to DOM and isn't being applied for the Sass tests (`exit` and `timeout`). Depending on how we choose to proceed, I'd think it could make sense for all tests to use the same Mocha configuration?

mahoneycm:
@aduth Thank you for outlining that. For the time being, could you add the compile tests to the sassTests like your first suggestion? We're thinking we'll keep them separate for now.

I appreciate the help!

aduth:
@mahoneycm I started working on this and it turned into quite a rabbit-hole, since the current tests do not pass as-is.

You can check this by running them directly:

```
npm exec mocha src/test/include.spec.js src/test/build.spec.js
```

There's a number of preexisting issues:

- The tests include timeouts which run after very long delays, without anything telling Mocha to wait for their results, perhaps giving false positives on test results
- [Test utility paths](https://github.com/uswds/uswds/blob/b5a332a49169db5ccebc209059283f05a721d7f1/packages/uswds-core/src/js/utils/test/util.js#L5-L7) are outdated and seem to reflect the pre-`packages` folder structure
- The tests reference Gulp tasks which no longer exist (`sass`, `copy-dist-sass`)
- The tests are meant to verify that `node_modules/@uswds/uswds/dist` is supported as a load path for Sass, but this is _actually broken_ in the published package (the [relative path referenced in `dist/scss/stylesheets/uswds.scss`](https://unpkg.com/@uswds/uswds@3.11.0/dist/scss/stylesheets/uswds.scss) does not exist)

I've fixed most of these in my local copy of this branch, except for the issues with `dist/scss`, which are harder to solve.

Perhaps the `dist/scss` folder should be removed altogether? It doesn't seem to work, [the documentation only mentions `pacakges`](https://designsystem.digital.gov/documentation/migration/#3-update-your-sass-compiler-settings-and-recompile-css-2), and there's an inline comment referencing the fact that copying `scss` may not be needed:

https://github.com/uswds/uswds/blob/b5a332a49169db5ccebc209059283f05a721d7f1/tasks/copy.js#L16

mahoneycm:
@aduth Thank you for taking a swing at this and the breakdown of the issues. My gut reaction is that this will take a bit of discussion from the team and should likely not block you from getting the rest of this work through.

I've created #6294 for us to take another look and track the issue. 

How do you feel about moving forward with this PR as it is in its current state?

aduth:
@mahoneycm I'm fine to move forward with this as-is! I think it's a good call to deal with those broken tests as a separate task.

mahoneycm (approved):
Thanks @aduth

mejiaj (approved):
LGTM, thank you!

We can make additional improvements in the future, like creating separate tasks for component tests, core utils, SASS unit tests, and then a script that runs them all.

That way we have the option of running specific tests or all during active development.

mahoneycm (approved):

thisisdano (approved):

PR #6173, opened 2024-11-01
USWDS - Core: Replace receptor library "ignore" with Element#contains

aduth:
# Summary

Reduced the JavaScript bundle size. By optimizing the implementation of core component code, overall JavaScript sizes are reduced.

## Breaking change

## Related pull requests

## Preview link

N/A

## Problem statement

See #5793

## Solution

Replace use of `receptor/ignore` with equivalent [`Node#contains`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) condition.

The intended purpose is the same.

[`receptor.ignore`](https://github.com/shawnbot/receptor?tab=readme-ov-file#ignore):

>Returns a delegated function that only calls the callback if the event's target isn't contained by the provided element. This is useful for creating event handlers that only fire if the user interacts with something outside of a given UI element.

[`Node#contains`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains)

>The contains() method of the [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node) interface returns a boolean value indicating whether a node is a descendant of a given node, that is the node itself, one of its direct children ([childNodes](https://developer.mozilla.org/en-US/docs/Web/API/Node/childNodes)), one of the children's direct children, and so on.

## Testing and review

To be totally honest, I'm not sure what this behavior is meant to control. References to `js-search-button` CSS class do not appear in any relevant Storybook stories or [documentation](https://search.usa.gov/search?utf8=%E2%9C%93&affiliate=uswds&query=js-search-button&commit=).

### DIFF
--- a/packages/usa-search/src/index.js
+++ b/packages/usa-search/src/index.js
@@ -1,4 +1,3 @@
-const ignore = require("receptor/ignore");
 const behavior = require("../../uswds-core/src/js/utils/behavior");
 const select = require("../../uswds-core/src/js/utils/select");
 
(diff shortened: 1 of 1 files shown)


mahoneycm:
Looking into it, it looks like this might be a relic of the old components structure? Here's a link to the [components dir](https://github.com/uswds/uswds/tree/release-2.14.0/src/js/components) from the release-2.14.0 branch.

I also can't find where this JS file is being used at all. This looks like a suitable replacement but I'm curious if we should just remove from this version of USWDS since it's not being used by the design system.

@amyleadem @mejiaj Any thoughts? Should we upkeep despite this or is it just tech debt?

dct2023:
@mahoneycm Adding you as a reviewer for this to get us unblocked on #6291

PR #6174, opened 2024-11-01
USWDS - Core: Replace receptor library "keymap" with custom implementation

aduth:
# Summary

**Reduced the JavaScript bundle size.** By optimizing the implementation of core component code, overall JavaScript sizes are reduced.

## Breaking change

## Related pull requests

Includes additional inline code comments based on review feedback from that pull request.

## Preview link

N/A

## Problem statement

See #5793

## Solution

Replace use of `receptor/keymap` with equivalent custom implementation.

## Testing and review

Verify unit tests pass:

```
npx gulp unitTests
```

For user acceptance testing, Modal is a good component to test for regressions, since its focus trap includes both simple key mapping (<kbd>Tab</kbd>) as well as modifier combinations (<kbd>Shift+Tab</kbd>).

1. Go to http://localhost:6006/?path=/story/components-modal--default
2. Click "Open default modal"
3. Press <kbd>Tab</kbd> or <kbd>Shift+Tab</kbd>
4. Observe that modal focus wraps as expected in focus trap


packages/uswds-core/src/js/utils/test/keymap.spec.js
mahoneycm, line 1: **Praise:** Thank you for including these unit tests!

packages/uswds-core/src/js/utils/test/keymap.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    const onTab = sinon.stub();
+    const handler = keymap({ Tab: onTab });
+    document.body.addEventListener("keydown", handler);
+    const event = new KeyboardEvent("keydown", { key: "Tab" });
+    document.body.dispatchEvent(event);
+
+    assert(onTab.calledOnceWithExactly(event));
+  });
+
+  it("does not call handler if combo does not include modifier from keypress", () => {
```
mahoneycm, line 26: **Suggestion:**

```suggestion
  it("does not call event if keypress includes modifier that the handler does not", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    const onTab = sinon.stub();
+    const handler = keymap({ Tab: onTab });
+    document.body.addEventListener("keydown", handler);
+    const event = new KeyboardEvent("keydown", { key: "Tab", shiftKey: true });
+    document.body.dispatchEvent(event);
+
+    assert(onTab.notCalled);
+  });
+
+  it("calls handler if all modifiers in combo are pressed", () => {
```
mahoneycm, line 36: **suggestion:**

```suggestion
  it("calls event when all modifiers and key in handler are pressed", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    const onTab = sinon.stub();
+    const handler = keymap({ Tab: onTab });
+    document.body.addEventListener("keydown", handler);
+    const event = new KeyboardEvent("keydown", { key: "Escape" });
+    document.body.dispatchEvent(event);
+
+    assert(onTab.notCalled);
+  });
+
+  it("calls handler if combo does not include modifier and neither does keypress", () => {
```
mahoneycm, line 16: **polish:** This test description is a bit confusing. 

I read this as: "It calls the event handler if the combination of keys does not include a modifier or the pressed key". Which can't be the case because we are pressing the `tab` key in this test.

If I'm understanding correctly, maybe something like:

```suggestion
  it("calls event if keypress matches keymap handler", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
```diff
@@ -0,0 +1,65 @@
+const sinon = require("sinon");
+const assert = require("assert");
+const keymap = require("../keymap");
+
+describe("keymap", () => {
+  it("does not call handler if keypress is not for the mapped key", () => {
```
mahoneycm, line 6: **suggestion:**

```suggestion
  it("does not call event if keypress does not match handler", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    const onTab = sinon.stub();
+    const handler = keymap({ "Shift+Tab": onTab });
+    document.body.addEventListener("keydown", handler);
+    const event = new KeyboardEvent("keydown", { key: "Tab", shiftKey: true });
+    document.body.dispatchEvent(event);
+
+    assert(onTab.calledOnceWithExactly(event));
+  });
+
+  it("does not call handler if some but not all modifiers in combo are pressed", () => {
```
mahoneycm, line 46: **Suggestion:**

```suggestion
  it("does not call event if some but not all modifiers in handler are pressed", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    const onTab = sinon.stub();
+    const handler = keymap({ "Shift+Ctrl+Tab": onTab });
+    document.body.addEventListener("keydown", handler);
+    const event = new KeyboardEvent("keydown", { key: "Tab", shiftKey: true });
+    document.body.dispatchEvent(event);
+
+    assert(onTab.notCalled);
+  });
+
+  it("does not call handler if event has modifiers but combo does not", () => {
```
mahoneycm, line 56: **suggestion:** For clarity:

```suggestion
  it("does not call event if keypress contains modifiers handler does not", () => {
```

mahoneycm (changes requested):
This seems to be working great @aduth ! The simple replacement in each of the component JS files is a good sign.

I left a few comments for added clarity. Some of the unit test descriptions were confusing to me. I added some suggestions but let me know if I'm off base. Hopefully we can improve their readability a bit so users can understand easily.

Open to other suggestions as well!

packages/uswds-core/src/js/utils/keymap.js
```diff
@@ -0,0 +1,37 @@
...
+ * handlers.
+ *
+ * @param {KeymapConfig} map
+ * @return {KeyboardEventHandler}
+ */
+module.exports = (map) => (event) =>
+  Object.keys(map).forEach((combo) => {
+    // Each key combination can have one or more modifier, where each modifier is prefixed with the
+    // modifier name and "+".
+    const parts = combo.split("+");
```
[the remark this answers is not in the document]
aduth: That's a good catch. I think what you're seeing is the `keymap` handler for tooltip, which is handled on any key press within the body element:

https://github.com/uswds/uswds/blob/95b83cdc637a859e40f38cbc249c7b9e24414ed0/packages/usa-tooltip/src/index.js#L402-L404

packages/uswds-core/src/js/utils/keymap.js
```diff
@@ -0,0 +1,37 @@
...
+
+    // The keyboard key name should come as the last part of a "+" delimited key. Popping the item
+    // will result in `parts` only containing the modifiers to be checked.
+    const key = parts.pop();
+
+    const isModifierMatch = ["Shift", "Alt", "Ctrl", "Meta"]
+      .filter((mod) => event.getModifierState(mod) || parts.includes(mod))
+      .every((mod) => event.getModifierState(mod) && parts.includes(mod));
```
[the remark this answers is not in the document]
aduth: > **quibble:** It might be worth added a comment to address what filter and every is doing here.

I agree! Let me know if 4167c08a2 helps clarify.

packages/uswds-core/src/js/utils/test/keymap.spec.js
aduth: Similar note to above.

Maybe middle-ground ... ?

```suggestion
  it("calls handler when all modifiers and key in combo are pressed", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
aduth: Yeah, the purpose of this is to verify the handler being called with an unqualified (no modifier) keypress, i.e. only if both the keypress event has no modifier and the expected key combination has no modifier.

Maybe something like ... ?

```suggestion
  it("calls handler if keypress matches keymap handler with no modifiers", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
aduth: For the rest of these comments, I think it stems back to my earlier comment https://github.com/uswds/uswds/pull/6174#discussion_r1911157811 in how I had intended to use the terms "handler" and "combo"/"key". Open to suggestions on clearer naming!

mejiaj:
@mahoneycm re-requesting review to confirm your comments have been addressed.

mahoneycm:
@aduth I left one more suggestion for the direction of the test descriptions! Apologies for not getting back to you sooner.

Thank you for your continued support and input 🙏

aduth:
@mahoneycm Thanks for the re-review. I applied the latest suggestions from the review threads. Let me know if there are others you'd still like to see updated from the unresolved threads.

dct2023:
@mlloydbixal Adding you as a reviewer for this to get us unblocked on #6291

packages/uswds-core/src/js/utils/test/keymap.spec.js
mahoneycm: **Suggestion:** Match new language used in other tests for consistency:

```suggestion
  it("calls handler if when no modifiers are expected or pressed", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
mahoneycm: **Suggestion:** Updated for new convention

```suggestion
  it("does not call handler if incorrect key is pressed", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
mahoneycm: **Suggestion:** Updated for new convention

```suggestion
  it("does not call handler if keypress does not include all expected modifiers", () => {
```

packages/uswds-core/src/js/utils/test/keymap.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    const onTab = sinon.stub();
+    const handler = keymap({ Tab: onTab });
+    document.body.addEventListener("keydown", handler);
+    const event = new KeyboardEvent("keydown", { key: "Tab", shiftKey: true });
+    document.body.dispatchEvent(event);
+
+    assert(onTab.notCalled);
+  });
```
mahoneycm, line 64: **Issue:** Looks like this is actually a duplicate test to the one found on lines 26-34. 

https://github.com/uswds/uswds/blob/0b721726bc745bccfb09f7da3f86088eef081b7e/packages/uswds-core/src/js/utils/test/keymap.spec.js#L26-L34

We should be safe to remove:

```suggestion

```

mahoneycm (changes requested):
@aduth Ok last batch of suggestions to make the existing tests match the new convention and to remove a duplicate test for unexpected modifiers in keypress!

After this we should be good to go

packages/uswds-core/src/js/utils/test/keymap.spec.js
aduth: > **Issue:** Looks like this is actually a duplicate test to the one found on lines 26-34.

Good catch! Trying to think back to what I might have been trying to account for in this test, I adapted this in a0ab5a1 to check for a scenario slightly different than the other modifier tests.

- ✅ All modifiers are pressed (line 36)
- ✅ Some but not all modifiers are pressed (line 45)
- ✅ **New:** Different modifier entirely is pressed (line 56)

aduth:
> @aduth Ok last batch of suggestions to make the existing tests match the new convention and to remove a duplicate test for unexpected modifiers in keypress!

Thanks for the review. Pushed updates in d0840da and a0ab5a1 based on your feedback.

mahoneycm (approved):
LGTM! Thanks so much @aduth 👍

heymatthenry (approved):

heymatthenry (approved):

Issue #6118, opened 2024-10-15
USWDS - Bug: Icon list needs alt text for specific list types


amycole501:
### Describe the bug

The first example of the [icon list ](https://designsystem.digital.gov/components/icon-list/) shows green checks indicating "yes" and a red x indicating "no" or "don't".

<img width="150" alt="Image" src="https://github.com/user-attachments/assets/ac793c3d-940e-419d-bc05-b77965a73553"/>

The icons on that list are marked as decorative.

We need to add alt text to those icons as they indicate meaning whereas the other icon list examples don't (i.e. those icons are decorative).

### Steps to reproduce the bug

1. Go to https://designsystem.digital.gov/components/icon-list/
2. Listen to the icon lists using JAWS, VO, or NVDA
3. There are no indicators of a "yes" or "no" that reflect the differences in the icons and the meaning the icons imply.

### Expected Behavior

For any icons that convey meaning such as a "yes" or "no" or "do" or "don't" we should add alt text for those indicators.

It may behoove us to also add instructions to that effect on the accessibility section on the page. "Ensure that for any icons that convey meaning, you add relevant alt text"

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [x] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

mlloydbixal:
Next steps is to follow the guidance already outlined in the accessibility section of the icons list component page (just on the list of yes/no icons that are non-decorative"

"If you wish to expose icons to screen readers:
- Provide descriptive text for each icon.
- Remove the aria-hidden="true" attribute and add a aria-labelledby attribute with a value that matches the id of a title element added inside the svg."

And then update the first bullet point to read "Provide descriptive text for each icon that conveys meaning" to include some of what Amy suggested! cc: @amycole501

mejiaj:
@mlloydbixal @amycole501 did we look into alternative solutions and why this approach was chosen?

mlloydbixal:
@mejiaj  I was referencing our own accessibility guidance section for this component and info I found here: https://accessibilityinsights.io/info-examples/web/svg-img-alt/ 

cc @amycole501

amycole501:
A bit more context:
[...]

PR #6194, opened 2024-11-08
USWDS - usa-icon-list: Add appropriate alt text to non-decorative icons

mlloydbixal:
# Summary
**Updates title text and aria-label of non-descriptive icons in icons list.** The icons that convey meaning in the icon list (indicating "yes" or "no") have appropriate title text and aria-labelledby attribute.

## Breaking change

## Related issue

Closes [#6118](https://github.com/uswds/uswds/issues/6118)

## Problem statement
One of the lists with icons have icons that are marked as decorative when they are conveying meaning (yes/no or do/don't). The ideal state is they have appropriate "alt text" to indicate meaning. This is only an issue on the list with green/red icons.

## Solution
By adding title tags inside the SVG icons with a unique id that matches the aria-labelledby attribute on the SVG element, screen readers gain accessibility to a description of what they mean.

Some additional info on this strategy and others that were also options: https://accessibilityinsights.io/info-examples/web/svg-img-alt/

## Testing and review
1. Go to https://designsystem.digital.gov/components/icon-list/
2. Using JAWS, VO, or NVDA listen to the icon description.
3. Confirm description is read for the non-decorative icons only.
4. Can also inspect the element to confirm these labels are available for non-decorative icons only.

## Preview
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/ml-icon-list-alt-txt/?path=/story/components-icon-list--default

### DIFF
--- a/packages/usa-icon-list/src/content/usa-icon-list.json
+++ b/packages/usa-icon-list/src/content/usa-icon-list.json
@@ -4,7 +4,7 @@
       "icon": {
         "name": "check_circle",
         "color": "green",
-        "alt_text": "Yes"
+        "title": "Yes"
       },
       "content": "Wash your hands for 20 seconds with soap"
     },
@@ -12,7 +12,7 @@
       "icon": {
         "name": "check_circle",
         "color": "green",
-        "alt_text": "Yes"
+        "title": "Yes"
       },
       "content": "Stay six feet away from others"
     },
@@ -20,7 +20,7 @@
       "icon": {
         "name": "cancel",
         "color": "red",
-        "alt_text": "No"
+        "title": "No"
       },
       "content": "Avoid large gatherings"
     }
--- a/packages/usa-icon-list/src/content/usa-icon-list.json
+++ b/packages/usa-icon-list/src/content/usa-icon-list.json
@@ -3,21 +3,24 @@
     {
       "icon": {
         "name": "check_circle",
-        "color": "green"
+        "color": "green",
+        "alt_text": "Yes"
       },
       "content": "Wash your hands for 20 seconds with soap"
     },
     {
       "icon": {
         "name": "check_circle",
-        "color": "green"
+        "color": "green",
+        "alt_text": "Yes"
       },
       "content": "Stay six feet away from others"
     },
     {
       "icon": {
         "name": "cancel",
-        "color": "red"
+        "color": "red",
+        "alt_text": "No"
       },
       "content": "Avoid large gatherings"
     }
--- a/packages/usa-icon-list/src/usa-icon-list.twig
+++ b/packages/usa-icon-list/src/usa-icon-list.twig
@@ -8,7 +8,7 @@
   {% for item in items %}
   <li class="usa-icon-list__item">
     <div class="usa-icon-list__icon {% if item.icon.color %} text-{{ item.icon.color }}{% endif %}">
-      <svg class="usa-icon{% if item.icon.utilities %} {{ item.icon.utilities }}{% endif %}" aria-hidden="true" role="img">
+      <svg class="usa-icon{% if item.icon.utilities %} {{ item.icon.utilities }}{% endif %}" aria-hidden="true" role="img" {% if item.icon.alt_text %} aria-label={{ item.icon.alt_text }}{% endif %} >
         <use xlink:href="{{ img_path }}/sprite.svg#{{ item.icon.name }}"></use>
       </svg>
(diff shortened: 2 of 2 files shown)


packages/usa-icon-list/src/usa-icon-list.twig
```diff
@@ -5,16 +5,18 @@
...
 <ul class="{{ list_classes | join(' ') | trim }}">
-  {% for item in items %}
-  <li class="usa-icon-list__item">
-    <div class="usa-icon-list__icon {% if item.icon.color %} text-{{ item.icon.color }}{% endif %}">
-      <svg class="usa-icon{% if item.icon.utilities %} {{ item.icon.utilities }}{% endif %}" aria-hidden="true" role="img">
+  {% for key, item in items %}
+    <li class="usa-icon-list__item">
+      <div class="usa-icon-list__icon {% if item.icon.color %} text-{{ item.icon.color }}{% endif %}">
+        <svg class="usa-icon{% if item.icon.utilities %} {{ item.icon.utilities }}{% endif %}"{% if item.icon.title %} aria-labelledby="usa-icon__title_{{ key }}"{% endif %} aria-hidden="{% if item.icon.title %}false{% else %}true{% endif %}" role="img">
+        {% if item.icon.title %}<title id="usa-icon__title_{{ key }}">{{ item.icon.title }}</title>{% endif %}
```
mahoneycm, line 12: **Question:** Reading the provided recommendations, was there a reason we didn't take the `aria-label` approach in the "better" category?

In their example, they're using `aria-labelledby` to point to a visible paragraph element and didn't create a `title` element. I'm curious if `aria-labelledby` is the approach we would want to take if there is on screen content that would describe the icon.

My hunch is we can use `aria-label` to describe the icon without the need to create a `title` element on the page, resulting in less markup.

mahoneycm: Ahh ok thanks for clarifying! I'll mark this as resolved 👍

packages/usa-icon-list/src/content/usa-icon-list.json
```diff
@@ -3,21 +3,24 @@
     {
       "icon": {
         "name": "check_circle",
-        "color": "green"
+        "color": "green",
+        "title": "Yes"
```
mahoneycm, line 7: I know there was some discussion on what was considered "meaningful" text and how these read to the screen reader. Would it be better to use more descriptive text pertaining to the icon rather than the "meaning" ? 

Maybe we can use "green check / red x" instead of yes/no to help prevent confusion in this example? Not sure if it's the best path forward but figured I'd bring it up for discussion

amyleadem (changes requested):
Thanks for your work on this @mlloydbixal.

I am a bit concerned about moving forward with this approach because it feels like [this concern](https://github.com/uswds/uswds/issues/6118#issuecomment-2477564929) from the issue is still not addressed. 

> Maybe it's worth changing the example, since these should be built to work without the icons (and I would say that technically these do — since the list makes just as much sense with the icons as without them). In fact it would be worse if the screen reader read something like "DON'T Avoid large gatherings"

I am finding the readout confusing because in VO it reads "No, image, avoid large gatherings". The meaning is much more straightforward if we do not include the icon in the readout. (One could argue that the visual icon might also be confusing, but that's probably another conversation)

We do offer [guidance](https://designsystem.digital.gov/components/icon-list/#accessibility-icon-list) to include the `title` attribute when the icon has meaning. I am wondering - is it necessary to illustrate it in the component preview? 

I see a couple of options for this:
1. Create a component example that has clearly useful "meaningful" icons
2. Opt to not provide an example and rely solely on guidance for users to add the `title` attribute.

annepetersen:
[Discussion in dev sync](https://docs.google.com/document/d/172TU5BVS9kGZQDuPcAhGoZjaTd2ngzcTayq4WRRTBUs/edit?tab=t.0#heading=h.ze2i8xu3oauw) has led to not making this change, but instead having the Accessibility team do some more research. 

How/should we provide guidance about meaningful icons?

This should probably be converted to an issue and have assignees/labels added — would be helpful if someone else could pick that up.

annepetersen (changes requested):
(see my comment above; this should become an issue)

annepetersen:
Next steps over on the [previously opened issue here](https://github.com/uswds/uswds/issues/6118)

PR #6195, opened 2024-11-08
USWDS - Combo Box: Prevent list HTML from rebuilding every time the list is displayed [WIP]

### LINKED ISSUE #6193: USWDS - Enhancement: Combo Box list HTML is rebuilt every time the list is displayed
### Describe the bug

Every time the combo box list is opened, the JS rebuilds every `li` element within the unordered list whether there has been a change to the results or not.

1. Go to Combo Box
2. Inspect HTML - At the start you see an empty `ul` element
3. Open the combo box drop down
4. `li` elements are appended to the DOM
5. While viewing the li elements in the devTools window, close and reopen the combo box
6. `li` elements flash because they are removed and re-appended to the DOM

### Expected Behavior

`li` elements should only be removed and added to the DOM when there is a change to the list or required attributes.

### Related code

Code that builds the options HTML:
https://github.com/uswds/uswds/blob/f0684a3812d85d26c3868c9368ec8bf1689a0d52/packages/usa-combo-box/src/index.js#L409-L439

Code that determines when to append the HTML:
https://github.com/uswds/uswds/blob/f0684a3812d85d26c3868c9368ec8bf1689a0d52/packages/usa-combo-box/src/index.js#L447-L455

### Screenshots

Recording of this behavior:

https://github.com/user-attachments/assets/5a5e1478-e393-4d4b-92c1-eda1be63f92d

### System setup

- USWDS Version: `3.9.0`
- Operating System: MacOS Sonoma 14.6.1
- Browser: Arc (chromium)

### Additional context

_No response_

### Code of Conduct

mahoneycm:
# Summary

> [!NOTE]
> This is a work in progress!

Performance enhancement to prevent combo box list HTML from appending to the DOM every time list is opened. Now, the DOM will only be modified when there are changes to the options list.

## Breaking change

This is _not_ a breaking change

## Related issue

Closes #6193 

## Related pull requests

_TK_

## Preview link

Combo box →
Time picker →

## Problem statement

Every time the combo box list is opened, the JS rebuilds every `li` element within the unordered list whether there has been a change to the results or not.

## Solution

Use a persistent `optionsCache` to compare the previous options array to the new options array. If options array matches cache, do not append a new list.

## Major changes

**TK**

## Testing and review

1. Go to Combo Box
2. Inspect HTML - At the start you see an empty `ul` element
3. Open the combo box drop down
4. `li` elements are appended to the DOM
5. While viewing the li elements in the devTools window, close and reopen the combo box
6. `li` elements are not affected
7. Use combo box input to filter results
8. Confirm combo box works as expected
9. Repeat steps for date picker

### DIFF
--- a/packages/usa-combo-box/src/index.js
+++ b/packages/usa-combo-box/src/index.js
@@ -35,6 +35,8 @@ const DEFAULT_FILTER = ".*{{query}}.*";
 
 const noop = () => {};
 
+let optionsCache = [];
+
 /**
  * set the value of the element and dispatch a change event
  *
@@ -478,6 +480,10 @@ const displayList = (el) => {
     }
   });
 
+  const optionsMatchCache = options.toString() === optionsCache.toString();
+
+  optionsCache = options;
+
   const numOptions = options.length;
   const optionHtml = options.map((option, index) => {
     const optionId = `${listOptionBaseId}${index}`;
@@ -517,14 +523,16 @@ const displayList = (el) => {
 
   listEl.hidden = false;
 
-  if (numOptions) {
-    listEl.innerHTML = "";
-    optionHtml.forEach((item) =>
-      listEl.insertAdjacentElement("beforeend", item),
-    );
-  } else {
-    listEl.innerHTML = "";
-    listEl.insertAdjacentElement("beforeend", noResults);
+  if (!optionsMatchCache) {
+    if (numOptions) {
+      listEl.innerHTML = "";
+      optionHtml.forEach((item) =>
+        listEl.insertAdjacentElement("beforeend", item),
+      );
+    } else {
+      listEl.innerHTML = "";
+      listEl.insertAdjacentElement("beforeend", noResults);
+    }
   }
 
   inputEl.setAttribute("aria-expanded", "true");


opensource-joe:
The idea here is right, but the cache has a real bug: optionsCache is one module-level variable shared across every combo box on the page, not scoped per instance. Two combo boxes with matching option sets will break each other, the second one to open finds a false cache match against the first and never populates its own list. There's a second issue in the same spot: since optionsCache starts as [], a combo box whose very first render has zero results will also false-match and skip inserting the "no results" message. Neither is about the ten months since this opened, they're bugs in the caching approach itself and would show up as soon as this got tested on a page with more than one combo box.

ryparker:
I'm closing this PR because the proposed cache can show stale results. The list still rebuilds on each filter pass at `packages/usa-combo-box/src/index.js:520` and `:525`, so the behavior you identified remains.

The cache key uses `options.toString()` on an array of `HTMLOptionElement` nodes. Each node becomes `[object HTMLOptionElement]`, so different option sets with the same number of items produce the same key. We need a key based on the option values, or a dirty flag updated when the list changes, before relying on the cache.

This is worth following up on. If you'd like to revise the approach, comment here or reopen the PR and I'll review it with you. Thanks for digging into it.

PR #6200, opened 2024-11-13
USWDS - File input: Add role="alert" to error message [TEST]

mahoneycm:
# Summary

Alternative approach for #6168.

This approach adds `role="alert"` attribute to the file input error message. While testing, I heard the alert exactly one time the first time I tested this approach. After refreshing, it does not call out the alert again

## Preview link

[File input: Specific →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-file-input-sr-only-error/iframe.html?args=&id=components-form-inputs-file-input--specific&viewMode=story)


ethangardner:
This was an alternate solution to a branch that has subsequently been merged in #6168. Closing this PR.

PR #5882, opened 2024-04-23
Cm create error components

mahoneycm:
# Summary

This work creates "error" state previews for each form element component. Keeping this work in case of future need but holding for now.

PR #6204, opened 2024-11-14
USWDS - Storybook: Add margin to form error test preview

### LINKED ISSUE #5998: USWDS - Forms: Form group error left border only viewable on wide widths

mahoneycm:
# Summary

Improved the display of form error states in desktop views by adding margin to the form element. Now, error borders are viewable at all widths in the preview.

This is a lighthanded alternative approach from #5999 

## Breaking change

This is _not_ a breaking change

## Related issue

Closes #5998

## Related pull requests

#5999 

## Preview link

[Form error test story →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin/?path=/story/patterns-forms--test-error-form-elements)

## Problem statement

Error states use a negative margin in desktop widths. When previewed in storybook, they are generally on their own and thus, do not have enough margin to preview error states.

## Solution

Add margin to error state preview using utility class to properly preview error states.

### Alternative solutions considered

This approach only affects the larger developer preview for error states. We are currently not offering individual error state previews per component but if we do, we should find another approach so that utility classes do not make their way into component code previews.

## Major changes

- Error state borders are now viewable in test story at desktop widths.

## Testing and review

1. Visit [Test error form elements](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin/?path=/story/patterns-forms--test-error-form-elements) storybook preview
2. When `error_state` is `true`, confirm you can see left border error


packages/usa-form/src/test/test-pattern/test-usa-form.twig
```diff
@@ -13,383 +13,1177 @@
 * Setting the error on the form element would add the error class to the entire form and not just
 * the form question that holds the error.
 *#}
-<form class="usa-form">
-  <div class="usa-character-count">
-    <div class="usa-form-group {% if error_state == true %}usa-form-group--error{% endif %}">
-      <label class="usa-label {% if error_state == true %}usa-label--error{% endif %}" for="with-hint-input">Character count</label>
-      <span id="with-hint-input-hint" class="usa-hint">This is an input with a character counter.</span>
+<div class="margin-205">
```
mejiaj, line 16: **issue**: This template is re-used for the disabled states story.    We should only apply code to the templates/stories that need it.

Consider removing this wrapper and using a Storybook decorator[^1] to avoid confusion, unintended side effects, and make it clear that we only want this spacing for the error state.

```js
// usa-form.stories.js
TestErrorFormElements.decorators = [
  (Story) =>
    `<div class="padding-205">
      ${Story()}
    </div>`,
];

```

[^1]: https://storybook.js.org/docs/writing-stories/decorators

packages/usa-form/src/test/test-pattern/test-usa-form.twig
mejiaj, line 1: **thought**: Concerns about maintaining this large hardcoded template moving forward. What would be the LOE to use component includes for the form elements?

Not only would it make this test cleaner and easier to maintain. It will also improve the individual form component templates by making them more robust.

packages/usa-form/src/test/test-pattern/test-usa-form.twig
```diff
@@ -13,383 +13,1177 @@
...
+      </div>
+      <span id="with-hint-input-info" class="usa-character-count__message">You can enter up to 25 characters</span>
+    </div>
+  </form>
+
+
+  {# *
+  * usa-legend and usa-label output the same base styles. Because of this, we can add the
+  * usa-label--error class without any additional changes or unexpected style conflicts.
+  *#}
```
mejiaj, line 40: **issue**: Let's not go too far from the standard twig comments. If we need the better comments highlighting, let's only apply it to the lines that need it.

```suggestion
{#
  * While we could add the usa-form-group--error class to <form class="usa-form">
  * elements, we should avoid doing this and only add it to usa-form-group elements.
  * Setting the error on the form element would add the error class to the entire 
  * form and not jus the form question that holds the error.
#}
```

Can you update the other comments too?

mejiaj (changes requested):
@mahoneycm thanks for this work. Re-scoping it to this issue makes the review much easier!

One thing I noticed is that the templates are rendering what seems like duplicates multiple times, is this intentional? Screenshot attached.

CleanShot 2024-11-18 at 11 51 38@2x

packages/usa-form/src/test/test-pattern/test-usa-form.twig
mahoneycm: Looking into this I think we'll need to finish #5882 which brings error states to into the individual stories. After that, we should be able to reference them pretty easily in this test story.

I believe we held off on this approach before while we research our current error states.

packages/usa-form/src/test/test-pattern/test-usa-form.twig
mahoneycm: Holding off on this since we're no longer touching the twig file. We can easily accomplish this when we refactor this test template to include other stories rather than adding the direct markup

packages/usa-form/src/test/test-pattern/test-usa-form.twig
mahoneycm: This is awesome! Thanks for sharing

Resolved in [eb71284](https://github.com/uswds/uswds/pull/6204/commits/eb712848f9add16164119cfba648b67a2779985e)

mahoneycm:
@mejiaj  Ready for re-review!

packages/usa-form/src/usa-form.stories.js
```diff
@@ -67,3 +67,9 @@ TestErrorFormElements.argTypes = {
     table: { disable: false },
   },
 };
+TestErrorFormElements.decorators = [
+  (Story) =>
+    `<div class="padding-205">
```
amyleadem, line 72: **suggestion:** Can we target just the horizontal padding here? 
```suggestion
    `<div class="padding-x-205">
```

amyleadem (changes requested):
Thanks for this @mahoneycm. 

I had a question about scope for this change. Is it possible to extend this solution to the components that have built in error states or error stories? I think it just includes text input, file input, and character count (but it is worth double checking).

I was also curious if we could target the horizontal padding only.

mlloydbixal:
@mahoneycm not sure if it would be out of scope for this task but since I've got RTL styling always top of mine I took a look and noticed this border does not mirror correctly when the language direction is switched. Should be a quick easy lift if it's within scope! You can test by `dir="rtl"` to `<html>`

packages/usa-form/src/usa-form.stories.js
mahoneycm: Resolved in [59a32c8](https://github.com/uswds/uswds/pull/6204/commits/59a32c8769a400cf8196ae065d417b10207f09ab)

packages/usa-form/src/usa-form.stories.js
```diff
@@ -67,3 +67,9 @@ TestErrorFormElements.argTypes = {
     table: { disable: false },
   },
 };
+TestErrorFormElements.decorators = [
+  (Story) =>
+    `<div class="padding-x-205">
```
amyleadem, line 72: **suggestion (non-blocking):** Sorry I missed this earlier, but since we are honing in on the error state story, thinking it would be good to have the change apply only when the [negative margin on the error states kicks in](https://github.com/uswds/uswds/blob/develop/packages/usa-form-group/src/styles/_usa-form-group.scss#L18). 
```suggestion
    `<div class="desktop:padding-x-205">
```

amyleadem (changes requested):
One more small change request. Not a blocker, but would make mobile testing a bit easier for this story.

packages/usa-form/src/usa-form.stories.js
mahoneycm: @amyleadem  I lean against making this change because I believe it would give us false confidence in our desktop preview. I don't think the difference between desktop and mobile display would be as clear since the difference in spacing would disappear when switching views.

Now, this is only a test page and it may be ok for _this_ story, so I'm open to other opinions but I think we should allow the difference in mobile and desktop to show.

mahoneycm: I converted a previous #5884 task to an issue: #6240. I believe this work can be included in the scope of adding error stories to the rest of the components given the solution might be scoped to only the components that have error states.

Alternatively, I can create it's own issue to add margin to all components/current error components. Lmk what you think!

amyleadem (approved):
LGTM! Thanks @mahoneycm. I like the simple solution here and look forward to continuing the work in #6240

thisisdano (approved):
Simple and effective

Issue #5967, opened 2024-07-01
USWDS - Feature: Set minimum heading count for in-page navigation


stphnwlkr:
### Is your feature request related to a problem? Please describe.

There are often times where having one or two headings on a page makes the in-page navigation feature of limited values, especially with only a small amount of content on the page.

### Describe the solution you'd like

Ideally, we would have a data attribute **data-minimum-heading-count** that could be set so that the in-page navigation doesn't render if there are less than or equal to the value in the attribute. This would build on the feature request https://github.com/uswds/uswds/issues/5610

### Describe alternatives you've considered

I am currently working with several WordPress table of contents plugins that have this feature, and this would be a great alternative for ensuring consistency in the coding.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Feature+Request%22) for duplicate feature requests.

PR #6205, opened 2024-11-14
USWDS - In-Page Navigation: Add data-mininum-heading-count attribute to in-page navigation

mlloydbixal:
# Summary
**Hide in-page navigation when there are fewer than 2 headings .**. The data-minimum-heading-count setting determines when the in-page navigation should be displayed or hidden if the minimum number of headings is not reached.

Previously, in-page navigation was shown even when there were no headings, resulting in an empty component.

## Breaking change

## Issue
Closes [#5967](https://github.com/uswds/uswds/issues/5967)

## Related Pull Requests
USWDS-Site: Add changelog for in-page navigation [#3000](https://github.com/uswds/uswds-site/pull/3000)

## Problem statement
In scenarios where there is limited content on the page, the in-page navigation does not provide much value.  Currently the in-page navigation is available to the user even if only 1 header was present. This issue can be solved by adding a data attribute to only show the in-page navigation element when the defined minimum number of headers is reached.

## Solution
Provide an attribute called data-minimum-heading-count where the number value sets the minimum amount of headings needed on the page to show the in-page navigation.

## Preview link
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/ml-inpage-nav-heading-count/?path=/story/components-in-page-navigation--default

## Major changes
-Customizable data attribute `data-minimum-heading-count` added to in-page navigation element
-New test pattern added to display usability
-JS functionality to not render any children in the navigational element when the minimum amount of headers is not met (default or custom setting)

## Testing and review
1. Navigate to the in-page-navigation component
2. Change storybook control for `data-minimum-heading-count` to be higher than the number of headings available
3. Confirm the navigational element is hidden with no children if the minimum heading count has not been reached
4. Change storybook control for `data-minimum-heading-count` to be lower than the number of headings available
5. Confirm it is visible with the correct children if the minimum heading count has been reached
6. Check all controls and visuals for regressions
7. Ensure all tests pass


mlloydbixal:
@mejiaj I fixed the regression and switched the css rule to use empty(). I think this will work better so the element will not be hidden in the case that a user adds other content into the `usa-in-page-nav` element.

mahoneycm (changes requested):
The feature is working pretty well! Excited to get this out. I left a couple questions and suggestions for discussion.

Chore: The PR description could use a bit of a quick polish. Could you add the related issue, a changelog PR, and testing steps? This will help us streamline testing and get this merged quicker!

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
```diff
@@ -175,6 +180,40 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
+    describe("Minimum heading count tests", () => {
+      beforeEach(() => {
+        body.innerHTML = TEMPLATE;
+        theNav = document.querySelector(THE_NAV);
+      });
+
+      afterEach(() => {
+        body.innerHTML = "";
+        window.location.hash = "";
+      });
```
mejiaj, line 192: **issue**:  The fact that we reset in `beforeEach()` and `afterEach()` makes me lean towards separating this test into its own file.

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
```diff
@@ -175,6 +180,40 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
 
+    describe("Minimum heading count tests", () => {
+      beforeEach(() => {
+        body.innerHTML = TEMPLATE;
+        theNav = document.querySelector(THE_NAV);
+      });
+
+      afterEach(() => {
+        body.innerHTML = "";
+        window.location.hash = "";
```
mejiaj, line 191: **issue**: We're not modifying hash, so this isn't needed here.

```suggestion

```

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
```diff
@@ -175,6 +180,40 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
+      });
+
+      it("does not render the in-page navigation when minimum heading count is too high", () => {
```
mejiaj, line 194: **question**: What's the max number of headings?

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
```diff
@@ -175,6 +180,40 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
+        behavior.off(containerSelector(body));
+        behavior.on(containerSelector(body));
+        assert.strictEqual(
+          theNav.hasChildNodes(),
+          false,
+          "In-page navigation should not have child nodes when the count is too high",
+        );
+      });
+
+      it("does render the in-page navigation when minimum heading count is low", () => {
```
mejiaj, line 205: **question**: Confirming, this hides the in-page navigation when the minimum heading count is low right?

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
```diff
@@ -9,6 +9,7 @@
     data-scroll-offset="0"
     data-root-margin="0px 0px 0px 0px"
     data-threshold="1"
+    data-minimum-heading-count="3"
```
mejiaj, line 12: **issue**: Is this needed if the default is `3`?

```suggestion

```

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
mlloydbixal: I am rewording this to say "does not render the in-page navigation when the default minimum heading count of 3 is reached" since template.html does not override the default value.

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
mlloydbixal: I will reword this to be "does render the in-page navigation when default minimum heading count of 3 has not been reached"

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
mlloydbixal: I removed this in favor of making it a storybook controlled value per @mahoneycm's suggestion!

packages/usa-in-page-navigation/src/index.js
```diff
@@ -14,6 +14,7 @@ const IN_PAGE_NAV_TITLE_HEADING_LEVEL = "h4";
 const IN_PAGE_NAV_SCROLL_OFFSET = 0;
 const IN_PAGE_NAV_ROOT_MARGIN = "0px 0px 0px 0px";
 const IN_PAGE_NAV_THRESHOLD = "1";
+const IN_PAGE_NAV_MINIMUM_HEADING_COUNT = 3;
```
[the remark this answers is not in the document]
mlloydbixal: Hm now I'm thinking of a scenario where even if there are only 2 headers maybe the content is longer form in which the navigation might be helpful to flag that there's more content below?  Although if that's a minor edge case the user can just utilize the new data attribute.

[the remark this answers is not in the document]
mlloydbixal: Updated!

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
mahoneycm: Looking at this test, this part is testing to make sure the in-page nav is hidden when the number of headings is lower than what is defined as the minimum heading count. I don't believe there is any "max number" of headings that could be generated at this time.

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
mahoneycm: **question:** Would we want to consider making the default minimum a system setting? This would enable users to update _all_ in page nav components on their site in one place, and use the data attribute for specific instances. 🤔

packages/usa-in-page-navigation/src/test/in-page-navigation-minimum-heading.html
amyleadem, line 1: **issue:** To match convention, the file name for this should include `template.html` (`in-page-navigation-minimum-heading.html` --> `in-page-navigation-minimum-heading.template.html`).

amyleadem (changes requested):
Thanks for this work @mlloydbixal. 

I had a few of top-level notes about this PR before diving deeper into my review: 
1. **Unexpected behavior (for me, at least):** When the minimum header count is higher than the number of headers on the page, the `usa-in-page-nav` aside remains in the DOM with empty content. Is it possible (and reasonable) to remove the aside in this scenario? This might require some team discussion.
1. **Needs Storybook controls:** It would be helpful for testing if we had both a Storybook control and test story for this attribute. It looks like you have the start to this in the twig file with your addition for `minimum_heading_count`, but I don't see that being pulled into `usa-in-page-navigation.stories.js`. Let me know if you have any questions about this. 
1. **The default value feels too opinionated:** I added a comment below to add my thoughts to the existing conversation. IMO, changing the default value to 1 is the least opinionated default we can have, and would also prevent the need to set the attribute value on all of our test templates. 

Curious to hear thoughts!

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
mejiaj: The issue itself says _don't_ display if there are one or two headings.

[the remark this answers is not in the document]
amyleadem: The issue does say that. However, I think we need to reconsider setting 3 as the hard-coded default, since it runs a much higher risk of becoming a breaking change. 

If we set the hard-coded default value to 3, a lot of current implementations of the component will disappear from the page unless the user takes action. I would rather users choose to turn off the components according to their site's content strategy, rather than being forced to turn it on. 

IMO, simply adding this attribute will resolve the issue by opening the door for the user to choose that they want to have a minimum of three headings. I don't think we need to hard code it.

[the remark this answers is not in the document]
mejiaj: ### Decision

After internal discussions we decided a minimum of two headings was a good starting point.

- Two headings fits a real user need (the USWDS website's [z-index page](https://designsystem.digital.gov/design-tokens/z-index/)) and is mentioned in the original issue.
- Single headings are usually at the top, so wouldn't make sense to have a minimum of zero or one.
- This is a dynamically generated enhancement that **doesn't** impact critical content.
- It would save users from additional custom code code to hide the empty menu.

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
mejiaj: I see. The navigation still gets created, but is visually hidden because the minimum heading count is too low.

mlloydbixal:
> Thanks for this work @mlloydbixal.
> 
> I had a few of top-level notes about this PR before diving deeper into my review:
> 
> 1. **Unexpected behavior (for me, at least):** When the minimum header count is higher than the number of headers on the page, the `usa-in-page-nav` aside remains in the DOM with empty content. Is it possible (and reasonable) to remove the aside in this scenario? This might require some team discussion.
> 2. **Needs Storybook controls:** It would be helpful for testing if we had both a Storybook control and test story for this attribute. It looks like you have the start to this in the twig file with your addition for `minimum_heading_count`, but I don't see that being pulled into `usa-in-page-navigation.stories.js`. Let me know if you have any questions about this.
> 3. **The default value feels too opinionated:** I added a comment below to add my thoughts to the existing conversation. IMO, changing the default value to 1 is the least opinionated default we can have, and would also prevent the need to set the attribute value on all of our test templates.
> 
> Curious to hear thoughts!

I updated the default to 2 as discussed. I also added a story with controls to manage the number of headers on page and the minimum number of headers required to show the in-page-navigation. I agree with. @mejiaj that it seems heavy handed to remove the <aside> element from the DOM since it is placed by the user. Keeping the `display:hidden` rule instead allows for easy override. 

This is ready for re-review! Thanks for all the input!

packages/usa-in-page-navigation/src/test/in-page-navigation-minimum-heading.html
mlloydbixal: Fixed in [fdbd228](https://github.com/uswds/uswds/pull/6205/commits/fdbd2281ac74231198d9b1521a0ce2f90427c250)!

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
mlloydbixal: Tests were updated in [fdbd228](https://github.com/uswds/uswds/pull/6205/commits/fdbd2281ac74231198d9b1521a0ce2f90427c250)

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
mlloydbixal: Tests were updated in [fdbd228](https://github.com/uswds/uswds/pull/6205/commits/fdbd2281ac74231198d9b1521a0ce2f90427c250)

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
mlloydbixal: Fixed in [d74c91c](https://github.com/uswds/uswds/pull/6205/commits/d74c91c5c2b9f8ac81e764697ed696ee419dc923)

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
mlloydbixal: This was fixed in [1d137b2](https://github.com/uswds/uswds/pull/6205/commits/1d137b2ac71a01132b6e9a5fedb92b6ac936895f) and further updated since then

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
```diff
@@ -9,6 +9,7 @@
     data-scroll-offset="0"
     data-root-margin="0px 0px 0px 0px"
     data-threshold="1"
+  {% if minimum_heading_count %} data-minimum-heading-count={{minimum_heading_count}} {% endif %}
```
mahoneycm, line 12: Looks like the html templates aren't building because the attribute value isn't being quoted!

```suggestion
  {% if minimum_heading_count %} data-minimum-heading-count="{{minimum_heading_count}}" {% endif %}
```

mlloydbixal: Resolved in [63c9b39](https://github.com/uswds/uswds/pull/6205/commits/63c9b39b3381b157426380d92a4cac2c37ab9172) 🥳

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
```diff
@@ -95,7 +95,12 @@ tests.forEach(({ name, selector: containerSelector }) => {
 
     afterEach(() => {
       sinon.resetHistory();
-      behavior.off(containerSelector(body));
+      if (theNav) {
+        const container = containerSelector(body);
+        if (container) {
+          behavior.off(container);
+        }
+      }
```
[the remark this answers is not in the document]
mlloydbixal: Resolved in [1d137b2](https://github.com/uswds/uswds/pull/6205/commits/1d137b2ac71a01132b6e9a5fedb92b6ac936895f)

packages/usa-in-page-navigation/src/test/in-page-navigation.spec.js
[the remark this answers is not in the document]
mlloydbixal: Resolved in [fade30a](https://github.com/uswds/uswds/pull/6205/commits/fade30a9a561aeaa04f63627539a72d903817bee) but further updates have been made in [fdbd228](https://github.com/uswds/uswds/pull/6205/commits/fdbd2281ac74231198d9b1521a0ce2f90427c250) since we lowered the default

packages/usa-in-page-navigation/src/index.js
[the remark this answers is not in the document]
mlloydbixal: This update was made in [fdbd228](https://github.com/uswds/uswds/pull/6205/commits/fdbd2281ac74231198d9b1521a0ce2f90427c250)

packages/usa-in-page-navigation/src/usa-in-page-navigation.twig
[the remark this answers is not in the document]
mlloydbixal: This line was changed in favor of using a storybook control

mlloydbixal:
@amyleadem @mahoneycm @mejiaj ready for re-review 🥳

packages/usa-in-page-navigation/src/usa-in-page-navigation.stories.js
```diff
@@ -42,3 +44,19 @@ TestCustomHeaderSelector.argTypes = {
...
+    options: ["1", "2", "3", "4"],
+    control: { type: "select" },
+  },
+  minimumHeaderCount: {
+    defaultValue: "1",
+    name: "Minimum number of headers required to show navigation",
+    options: ["1", "2", "3", "4"],
+    control: { type: "select" },
+  },
+};
```
mahoneycm, line 62: **issue:** Thank you for creating these test controls! For some reason I'm unable to use them. 
- Changing the control causes the template to reload without the JS
- Refreshing the page resets the control to it’s defaults

I've seen this JS behavior in other stories but the others keep the set value after reloading. Can you take a look at how the other in-page navigation stories are set up and see if we can get the test story to display correctly?

packages/usa-in-page-navigation/src/usa-in-page-navigation.stories.js
mahoneycm, line 1: **Issue:** I also noticed that "Test Custom Header Selector" test is no longer working. I'm curious if this is just an issue with the twig file or stories.js file that's causing the regression.

| Develop | Working branch |
|--------|--------|
| [Before](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-in-page-navigation--test-custom-header-selector) | [After](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/ml-inpage-nav-heading-count/?path=/story/components-in-page-navigation--test-custom-header-selector) |

mahoneycm (changes requested):
Functionality is looking good! The test stories aren't working as expected and making it a bit harder to test. Would you mind taking a look?

packages/usa-in-page-navigation/src/usa-in-page-navigation.stories.js
mlloydbixal: Fixed in [6869b3b](https://github.com/uswds/uswds/pull/6205/commits/6869b3b59da46f3fb87969b8152a5053c87af932)

packages/usa-in-page-navigation/src/usa-in-page-navigation.stories.js
mlloydbixal: Fixed in [18eeec9](https://github.com/uswds/uswds/pull/6205/commits/18eeec9d17b4b8ee251a10d595b869716c319efe)

packages/usa-in-page-navigation/src/index.js
```diff
@@ -14,6 +14,7 @@ const IN_PAGE_NAV_TITLE_HEADING_LEVEL = "h4";
 const IN_PAGE_NAV_SCROLL_OFFSET = 0;
 const IN_PAGE_NAV_ROOT_MARGIN = "0px 0px 0px 0px";
 const IN_PAGE_NAV_THRESHOLD = "1";
+const IN_PAGE_NAV_MINIMUM_HEADING_COUNT = 2;
```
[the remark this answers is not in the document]
mlloydbixal: I do like that - I'll create it as a separate issue for future enhancement so we can get a team discussion going without blocking this.

packages/usa-in-page-navigation/src/test/test-patterns/test-minimum-header.twig
```diff
@@ -0,0 +1,22 @@
...
+    <p>Section content</p>
+    {% endif %}
+    {% if headerLevels >= 3 %}
+    <h4>Heading level 4</h4>
+    <p>Section content</p>
+    {% endif %}
+    {% if headerLevels >= 4 %}
+    <h5>Heading level 5</h5>
+    <p>Section content</p>
+    {% endif %}
```
mahoneycm, line 20: **Issue (blocking):** The current tests don't work for headings above 2 levels

image


**Suggestion** We should limit these heading levels to level 2 and 3

By default, the in-page nav selection only uses `h2` & `h3` unless unless specified otherwise using the [data-heading-elements](https://arc.net/l/quote/uivmjhkb) data-attribute.

Limiting to only `h2` and `h3` will showcase default behavior of the component a bit better! 👍

mlloydbixal: Fixed in [ff2ce12](https://github.com/uswds/uswds/pull/6205/commits/ff2ce12aa8d92f083f068efd61641156419dec87) thanks for pointing that out :)

mahoneycm (commented):
Looking good! One remaining question for a conditional statement in one of the tests but other than that things are working great!

packages/usa-in-page-navigation/src/test/in-page-navigation-minimum-heading-default.spec.js
```diff
@@ -0,0 +1,47 @@
...
+      body.innerHTML = "";
+    });
+
+    it("does not render the in-page navigation when default heading count of 2 has not been reached", () => {
+      behavior.off(containerSelector(body));
+      behavior.on(containerSelector(body));
+      assert.strictEqual(
+        theNav.hasChildNodes(),
+        false,
+        "In-page navigation should not have child nodes when the default minimum heading count of 2 has been reached",
```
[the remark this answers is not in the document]
mlloydbixal: Fixed in [3c145f9](https://github.com/uswds/uswds/pull/6205/commits/3c145f9170b7f38f88c2c70ffeaddbf8c5f6bc56)

thisisdano (approved):
Good communication to get this one right

PR #6230, opened 2024-11-27
USWDS - Storybook: Add padding to form elements with error states

mahoneycm:
# Summary

Adds horizontal padding to stories that feature an error state component in order to display form group error border if present.

## Breaking change

This is _not_ a breaking change.

## Related pull requests

Additional work to potentially be included in #6204

## Preview links

[Character Count (default) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-character-count--character-count)
[Character Count (Test no label) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-character-count--test-no-label)
[Character Count (Test no form group) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-character-count--test-no-form-group)

[File input (error) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-file-input--error)
[File input (Specific) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-file-input--specific)

> [!NOTE]
> You can trigger an error state in file input specific but it does _not_ have the form group error styles. Therefore, the error border will not display

[Input Prefix →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-input-prefix-or-suffix--prefix)

> [!NOTE]
> Input prefix also does not have the form group error styles, therefore, will not have the error border.

[Text input →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-text-input--input)
[Text input (Error) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-text-input--error)
[Text input (State Showcase) →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-error-story-margin-addl/?path=/story/components-form-inputs-text-input--error)

## Problem statement

Form group error borders are pushed off screen and are hidden.

## Solution

Add horizontal margin to storybook previews featuring error states

## Testing and review

1. Visit each of the error state previews above
2. Confirm we want to add the margin decorator to each of the examples.
    * I've made note that some of the error states do not have the error border due to form group classes missing. I opted to keep these in to make it clear that it was _not_ present in these examples rather than hidden off-screen
3. Confirm there are no other examples we want to add margin to


packages/usa-character-count/src/usa-character-count.stories.js
amyleadem, line 1: **suggestion:** Move this decorator to the story level so that you only need to add it once to the file. [^1]

Alternatively, I wonder if there is any value (or harm) to setting this at the global level in preview.js. That would prevent us from needing to remember this decorator for every component. 

```diff
export default {
  title: "Components/Form Inputs/Character Count",
+  decorators: [
+    (Story) =>
+      `<div class="padding-x-205">
+        ${Story()}
+      </div>`,
+  ],
  argTypes: {
    disabled_state: {
      name: "Disabled state",
      control: { type: "radio" },
      options: ["none", "disabled", "aria-disabled"],
    },
  },
};
```
[^1]: https://storybook.js.org/docs/writing-stories/decorators

amyleadem (changes requested):
Thanks for this @mahoneycm. I like the idea of adding the padding to these components. I don't think this should be a blocker for #6204, but I think it would be a good addition if we can make it. I added a suggestion below to elevate the location of the decorators to add them at either the global or component level. Let me know if you have any questions.

packages/usa-character-count/src/usa-character-count.stories.js
mahoneycm: I was wondering if we could take a similar approach to updating some base styles similar to how we do in the [web components repo](https://github.com/uswds/web-components/blob/develop/storybook/index.css). I feel like globally adding a bit of margin so that the component previews are all uniform might be the better idea 🧐

mejiaj: I don't see any harm in increasing padding, but could also see it as a follow-up enhancement.

opensource-joe:
This isn't tied to an open issue and #6204, the PR it was supplementary to, already merged in December 2024, so nothing is blocked on it. All three reviewers landed on the same point though: the padding is worth having, but a per-story decorator is the wrong place for it. Amy suggested component-level or global preview.js, and Charlie separately pointed at the same pattern already used in the web-components repo's Storybook config. That's a small rework, not a rejection, just never got picked back up.

ryparker:
I'm closing this during the backlog cleanup because we don't have capacity to finish it now. The reviews supported adding padding, with the remaining suggestion to move the decorator to the component or global level instead of repeating it in each story. #6204 has already merged, so this follow-up isn't blocking that change.

If you'd like to pick this back up, comment here or reopen the PR. I'll help with the rebase and review. Thanks for the work, and sorry we left the follow-up waiting.

PR #6232, opened 2024-11-29
USWDS - Input: Move input width classes out of usa-form package

### LINKED ISSUE #5312: USWDS - Feature: Move usa-input--[width] classes to the usa-input module

amyleadem:
# Summary
**Moved `.usa-input--[width]` and `.usa-input-group--[width]` classes out of the `usa-form` package.** 
These classes are now generated in the `usa-input` and `usa-input-prefix-suffix` packages and can be used without the `.usa-form` parent element.

> [!Note]
> This is an alternative solution to PR #5935 (Thanks @aduth!)

## Breaking change

## Related issue

Closes #5312

## Related pull requests

[Changelog PR](https://github.com/uswds/uswds-site/pull/2997)

## Preview link

- [Input component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-input-width/?path=/story/components-form-inputs-text-input--input)
- [Input prefix/suffix component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-input-width/?path=/story/components-form-inputs-input-prefix-or-suffix--prefix)
- [Mailing address pattern](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/al-input-width/?path=/story/patterns-forms--mailing-address)

## Problem statement

As a developer, I expect that if I include the package listed on a component's documentation (e.g. [Text Input](https://designsystem.digital.gov/components/text-input/)), all of the usage options documented for that component are available, so that I'm not confused by styles not existing despite being documented, or be forced to include more packages than I'd otherwise need for my optimized installation.

## Solution
- Moved `.usa-input--[width]` and `.usa-input-group--[width]` class defintions into the `usa-input` and `usa-input-prefix-suffix` package, respectively.
    - Created a new `$system-input-widths` map in the `uswds-core` package so that the values can be defined in a single location
    - Used the `$system-input-widths` map to generate `usa-input--[width]` and `usa-input-group--[width]` classes
- Removed the `max-width: none` declaration from `.usa-form .usa-input` and `.usa-form .usa-textarea` to prevent specificity overrides

## Testing and review
- Confirm the `.usa-input--[width]` and `.usa-input-group--[width]` classes match the CSS (class names and values) in `develop`
- Confirm `usa-input--[width]` and `usa-input-group--[width]` classes work as expected both _with_ AND _without_ a `.usa-form` parent on the following elements:
    - `usa-input`
    - `usa-textarea`
    - `usa-input-group`
    - Character count input (currently accepts the `usa-input--[width]` classes in `develop`)
    - Input mask input (currently accepts the `usa-input--[width]` classes in `develop`)
- Confirm no regressions as a result of reducing the specificity of `max-width: none` on `.usa-form` children:
    - Set `$theme-input-max-width` to a value less than mobile (like "card" or 15)
    - Open the local build of [mailing address pattern story](http://localhost:6006/?path=/story/patterns-forms--mailing-address) 
    - Confirm that all elements in the form, including the zip code input, are the same width as they are in [develop](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/patterns-forms--mailing-address)
- Confirm that creating `$system-input-widths` in `uswds-core` makes sense and lives in the appropriate location


packages/uswds-core/src/styles/tokens/units/input-widths.scss
amyleadem, line 1: **question:** Is this the best location for this?

packages/usa-form/src/styles/_usa-form.scss
```diff
@@ -17,41 +17,12 @@
     text-decoration: none;
   }
 
-  .usa-input,
   .usa-range,
-  .usa-select,
-  .usa-textarea {
+  .usa-select {
```
amyleadem, line 21: > [!note]
> Removed these elements because the specificity of this rule caused an [override](https://github.com/uswds/uswds/pull/5935#discussion_r1853057805) of the `usa-input--[width]` classes when they were used inside `.usa-form`. We should test that there are no regressions as the result of this change.

mahoneycm: **question:** Should we also remove `usa-range` and `usa-select` for uniformity?

packages/usa-input-prefix-suffix/src/styles/_usa-input-prefix-suffix.scss
```diff
@@ -58,6 +58,13 @@ $icon-offset: $icon-width + ($input-padding * 2);
   }
 }
 
+// Create input group max width utility classes
+@each $name, $width in $system-input-widths {
+  .usa-input-group--#{$name} {
+    max-width: #{$width};
+  }
+}
+
```
mahoneycm, line 67: **thought:** These classes feel a bit redundant.

I understand why we might not want to drop support for them but it feels like we could set these styles with `usa-input--` classes or even better, the `maxw-` utility classes. 

Since we're addressing them now, I wanted to bring this up in case we wanted to explore other options for how users would customize these inputs. That said, we might be better for simply supporting them because users might already be using these classes. Let me know what you think!

mahoneycm (commented):
This is all looking and working pretty well @amyleadem ! I left a couple thoughts in comments but nothing that is necessarily a blocker.

packages/usa-input-prefix-suffix/src/styles/_usa-input-prefix-suffix.scss
amyleadem: I originally thought the same thing, but ended up landing on keeping them for the following reasons:
1. It felt more more intuitive to add `usa-input-group--[width]` to `usa-input-group` elements and `usa-input--[width]` classes to `usa-input` elements. It felt weird to mix them. 
2. The `maxw` classes are not ideal for this since they are defined with `rem` rather than `ex`. The `ex` unit makes more sense for inputs, and having a set of standard widths for inputs makes sense to me. 
3. And most importantly, preserving the classes as they are keeps this from being a breaking change. 
 
How we set input widths is something worth exploring in the Elements version of the component, though. Good to flag it for future discussion.

packages/usa-form/src/styles/_usa-form.scss
amyleadem: > **thought:** This might make it harder to customize entire forms.
> 
> Currently, setting `max-width: none` here allows the inputs to match the max width of the form element. Removing this means the form and inputs have separate max-widths. I'm not sure if this should be a blocker or not but I'm curious how this might affect downstream users.

This is a fair note. This update would cause a regression in:
- `.usa-form` child inputs if `$theme-input-max-width` is set to a width narrower than "mobile"
- `.usa-form--large` child inputs if `$theme-input-max-width` is set to a width narrower than "mobile-lg"

**Preview of regression:**
image

This is because the max-width for `usa-form` in desktop widths is set to ["mobile"/20rem](https://github.com/uswds/uswds/blob/develop/packages/usa-form/src/styles/_usa-form.scss#L12) and `usa-form--large` is ["mobile-lg"/ 30rem](https://github.com/uswds/uswds/blob/develop/packages/usa-form/src/styles/_usa-form.scss#L71). By default, `$theme-input-max-width` is set to `mobile-lg`, so it is either the same width or wider than the `.usa-form` elements. But if the input width is set to less than that, we start to see regressions. 

For this reason, I'm thinking we restore this `max-width: none` rule in `usa-form` and instead explore alternatives, either reducing specificity in this rule or increasing specificity in the `usa-input` and `usa-input-prefix-suffix` packages. I'll come back with solutions.

packages/usa-form/src/styles/_usa-form.scss
```diff
@@ -17,41 +17,11 @@
     text-decoration: none;
   }
 
-  .usa-input,
-  .usa-range,
-  .usa-select,
-  .usa-textarea {
+  :where(.usa-input, .usa-textarea, .usa-select, .usa-range) {
```
amyleadem, line 20: > [!note]
> Reduced specificity with `where()` here because this rule was causing .usa-form to override input width classes because it was more specific.

packages/usa-form/src/styles/_usa-form.scss
amyleadem: Update: proposing using `:where()` here instead of removing the rules entirely. This will preserve the the `usa-form` max-width on child input elements in the default state, but will also allow a custom max width `usa-input--[width]` and `usa-input-group--[width]` classes. 

Curious what you think.

mahoneycm (dismissed):
LGTM! No perceived regressions from `develop`

## Testing checklist

- [x] Confirm the `.usa-input--[width]` and `.usa-input-group--[width]` classes match the CSS (class names and values) in develop.
- [x] Confirm `usa-input--[width]` and `usa-input-group--[width]` classes work as expected both with AND without a `.usa-form` parent on the following elements:- [ ]  Confirm no regressions as a result of removing `max-width: none` from `.usa-form . usa-input` and `.usa-form . usa-input`
- [x]  Confirm that creating `$system-input-widths` in `uswds-core` makes sense

mejiaj (approved):
Nice work, the new features are a nice enhancement that avoid breaking changes. Also, shoutout to @mahoneycm for the thorough review.

I tested:
- [x] Updating `$theme-input-max-width` and ensure there aren't conflicts.
- [x] Different sizes for input group for input prefix/suffix.
- [x] Different sizes for input widths
- [x] Compared against `develop`

packages/usa-form/src/styles/_usa-form.scss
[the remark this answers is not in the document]
amyleadem: Good idea. Added a comment in [eb51128](https://github.com/uswds/uswds/pull/6232/commits/eb511281f04453e7961817aefdfe134ca65e4493)

packages/usa-input-prefix-suffix/src/styles/_usa-input-prefix-suffix.scss
```diff
@@ -58,6 +58,13 @@ $icon-offset: $icon-width + ($input-padding * 2);
   }
 }
 
+// Create input group max width utility classes
+@each $name, $width in $system-input-widths {
+  .usa-input-group--#{$name} {
+    max-width: #{$width};
+  }
+}
```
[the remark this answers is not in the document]
amyleadem: Added some generic "utility classes" controls to the text input and input prefix/suffix stories in [6d3483d](https://github.com/uswds/uswds/pull/6232/commits/6d3483d0c5bfb19814797067e1c602c63e2a954e). It is just a basic text control so we can add whatever utility classes we need. Let me know if you had something else in mind.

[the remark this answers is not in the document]
mejiaj: @amyleadem thanks, that works. 

I didn't mean to add additional work to this PR and added to create #6236.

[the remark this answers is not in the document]
amyleadem: Awesome. In #6236, we can consider if we want to improve the controls at all by creating a select with a list of preconfigured width classes, etc.

mlloydbixal (approved):
Confirmed all the testing and review steps and confirmed previous requested changes look good. Approving!

packages/usa-form/src/styles/_usa-form.scss
```diff
@@ -17,41 +17,13 @@
     text-decoration: none;
   }
 
-  .usa-input,
-  .usa-range,
-  .usa-select,
-  .usa-textarea {
-    // max width not needed since .usa-form has one of its own
+  // Clear max-width in children since .usa-form has one of its own.
+  // Used :where() to reduce specificity and
```
aduth, line 21: Does this create an implicit requirement that the `usa-form` styles have to be loaded before the `usa-input` styles, since they're still equal specificity? (e.g. if a project has an [optimized build](https://designsystem.digital.gov/documentation/migration/#6-optimize-your-installation:~:text=Load%20the%20your%20project%E2%80%99s%20necessary%20component%20packages%20in%20your%20Sass%20entry%20point.%20If%20you%20find%20a%20hit%20for%20the%20class%20name%20in%20your%20codebase%2C%20include%20the%20relevant%20package%20in%20your%20Sass%20entry%20point.))

mahoneycm: Good catch @aduth 

After some quick testing in sandbox, it does look like loading `usa-form` after `usa-input` _will override_ the utility class with the `max-width: none` set here.  

Since the specificity is lower it also allows for other utility classes to override the max width easier.

Experimented with moving the following into various orders:

```sss
@forward "usa-form";
@forward "usa-input";
@forward "usa-select";
@forward "uswds-utilities";
```

|  Forward order | Develop | Working branch |
|--------|--------|--------|
| form, input, select, utilities | image | image | 
| input, form, select, utilties | image | image | 
| select, form, input, utilities | image | image | 
| utilities, form, input, select | image | image | 
| utilities, input, select, form | image | image | 

Findings:

- In `develop`
  - The high specificity within `usa-form` made things consistent no matter what order the packages were forwarded
  - Max width utility could not penetrate styles
- With this change:
  - Forward order does matter
  - Form being imported first seems to be the best option for versatility
  - Forwarding form after input _or_ select packages means the form max-width will take priority
  - Forwarding form after `uswds-utilities` classes means the form max-width will take priority
  - Forwarding form last means the form max-width will take priority over all of the above

I'm going to look at this again on Monday along side things like `uswds-form-controls` which imports all of the following:

```
usa-character-count, usa-checkbox, usa-combo-box, usa-date-picker, usa-error-message, usa-fieldset, usa-file-input, usa-form, usa-form-group, usa-hint, usa-input, usa-input-prefix-suffix, usa-label, usa-legend, usa-memorable-date, usa-radio, usa-range, usa-select, usa-textarea, usa-time-picker
```

amyleadem: Good notes. 

To improve resiliency and address this issue, I've updated the Sass to generate both `.usa-form .usa-input--[width]` and `.usa-input--[width]` in [d9e1740](https://github.com/uswds/uswds/pull/6232/commits/d9e1740129c63bdf7b70b9d37af939b40fab606d) (same for usa-input-group). I also removed the `where()` from `usa-form` in [d6d2eb4](https://github.com/uswds/uswds/pull/6232/commits/d6d2eb40cf437ffa01ea527f9f6057de3bda688a) because it was no longer needed. 

amyleadem:
@mlloydbixal @mahoneycm (cc: @aduth)

I've updated the Sass to generate explicit `.usa-form .usa-input--[width]` classes (see [related comment](https://github.com/uswds/uswds/pull/6232#discussion_r1879040086) for details). Can you re-review and make sure it still works as expected?

packages/usa-form/src/styles/_usa-form.scss
mahoneycm: Continuing my testing, it looks like any time a component that extends `%block-input-` placeholders from `_form.scss` is forwaded _before_ `usa-form`, it causes the `max-width: none` definition to take priority.

This appears to be true even for something like `usa-checkbox`:

**Importing usa-form first:**

```scss
@forward "usa-form";
@forward "usa-checkbox";
@forward "usa-input";
```

Results in:

image

**Importing some other stylesheet with `%block-input-` placeholder first:**

```scss
@forward "usa-checkbox";
@forward "usa-form";
@forward "usa-input";
```

image

I'm a bit puzzled by this tbh but can spend more time diving in if we consider this a blocker for release.

**note:** As you can see in these screenshots, the form max-width classes have gone back to taking priority over our utility classes. This is the current behavior on `develop` (see table screenshots in [this comment](https://github.com/uswds/uswds/pull/6232#discussion_r1874142139))

amyleadem: Update: I restored `where()` in `usa-form` to reduce specificity on those max-width rules. It looks like it solves the issue. Let me know if it gives you any trouble, @mahoneycm.

mahoneycm: This is looking great! From my testing, the package forward/import order no longer affects which `max-width` takes priority. This branch now matches behaviors found in `develop`. 

Great work @amyleadem !

mahoneycm (approved):
This is working great! Great job covering all test cases and maintaining the current behaviors.

✅ ✅ ✅ 


## Testing checklist

- [x]  Confirm the `.usa-input--[width]` and `.usa-input-group--[width]` classes match the CSS (class names and values) in `develop`        
- [x]  Confirm `usa-input--[width]` and `usa-input-group--[width]` classes work as expected both *with* AND *without* a `.usa-form` parent on the following elements:
    - [x]  `usa-input`
    - [x]  `usa-textarea`
    - [x]  `usa-input-group`
    - [x]  Character count input (currently accepts the `usa-input--[width]` classes in `develop`)
    - [x]  Input mask input (currently accepts the `usa-input--[width]` classes in `develop`)
- [x]  Confirm no regressions as a result of removing `max-width: none` from `.usa-form . usa-input` and `.usa-form . usa-input`
    - Looks good. The max width from `usa-form` manages the max width without the need to set it individually on nested elements.
    - [x]  Should we remove range and select?
- [x]  Confirm that creating `$system-input-widths` in `uswds-core` makes sense

thisisdano (approved):
Nice work, this one got a bit complicated!

PR #6268, opened 2024-12-20
USWDS - Core: Add transition utility with reduced motion preference handling

### LINKED ISSUE #5256: USWDS - Feature: Add support for prefers-reduced-motion to disable transitions

Some users who experience distraction or nausea from animated content may specify system preferences to express their desire for reduced motion to eliminate inessential animation. The U.S. Web Design System should ideally respect this preference to disable transitions which are inessential (e.g. mobile menu flyout animation).

When the [`@media (prefers-reduced-motion)` media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) is present for a user, [the USWDS `$project-easing` variable](https://github.com/uswds/uswds/blob/develop/packages/uswds-core/src/styles/variables/project-easing.scss) should effectively be set to `0`.

_No response_

### Additional context

- [`prefers-reduced-motion` MDN documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)
- [WCAG Technique C39: Using the CSS reduce-motion query to prevent motion](https://www.w3.org/WAI/WCAG21/Techniques/css/C39)
- [web.dev article: _prefers-reduced-motion: Sometimes less movement is more_](https://web.dev/prefers-reduced-motion/)

### Code of Conduct

aduth:
# Summary

**Animated transitions now respect the system's reduced motion preference**. A new transition utility handles default behavior for easing and disabling inessential animation.

## Breaking change

## Related issue

Closes #5256

## Preview link

- [Tooltip](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/aduth-reduced-motion/?path=/story/components-tooltip--tooltip)
- [Skipnav](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/aduth-reduced-motion/?path=/story/components-skipnav--skipnav)
- [Modal](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/aduth-reduced-motion/?path=/story/components-modal--default)

In local Storybook, can be previewed at:

- http://localhost:6006/?path=/story/components-tooltip--tooltip
- http://localhost:6006/?path=/story/components-skipnav--skipnav
- http://localhost:6006/?path=/story/components-modal--default

## Problem statement

As a user who may be distracted or experience harm such as nausea due to animated content, I expect that the U.S. Web Design System will respect my preference to experience reduced motion where possible.

Related:

- [WCAG Success Criterion 2.3.3: Animation from Interactions (Level AAA)](https://www.w3.org/WAI/WCAG21/Understanding/animation-from-interactions)

## Solution

Adds a new utility `u-transition` with smart defaults to handle:

- Default easing
- Optional `$essential` argument (default `false`) to control whether transition is disabled when reduced motion is preferred

Related:

- [WCAG Technique C39: Using the CSS reduce-motion query to prevent motion](https://www.w3.org/WAI/WCAG21/Techniques/css/C39)

## Testing and review

See "Preview links" above.

Unless system preferences are configured to prefer reduced motion, there should be no changes in the appearance of animated transition effects.

Instructions for configuring reduced motion preference:

- macOS: https://support.apple.com/guide/mac-help/stop-or-reduce-onscreen-motion-mchlc03f57a1/mac
- Other: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#user_preferences

----


aduth:
This addresses all `transition` styles in stylesheets. I discovered there's also one use of `animation` which should probably similarly be disabled when reduced motion is preferred.

https://github.com/uswds/uswds/blob/0cf7c7e24d6c95b6eefc760cae7c063fac8555e9/packages/usa-nav/src/styles/_usa-nav.scss#L103

A couple ideas:

- We could create a `u-animation` mixin very similar to what's introduced here with `u-transition`
- Alternatively, we could refactor the approach to either inline the media-query, or introduce a mixin which wraps `@content` with `@media not(prefers-reduced-motion)`

mejiaj:
I've created a local branch to get the previews and added them to the description.

- [Tooltip](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/aduth-reduced-motion/?path=/story/components-tooltip--tooltip)
- [Skipnav](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/aduth-reduced-motion/?path=/story/components-skipnav--skipnav)
- [Modal](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/aduth-reduced-motion/?path=/story/components-modal--default)

CC @uswds/a11y

packages/uswds-core/src/styles/mixins/utilities/_transition.scss
```diff
@@ -0,0 +1,20 @@
+@use "sass:meta";
+@use "sass:string";
+@use "../../variables/project-easing" as *;
+
+// Outputs transition with project default easing and media query to disable inessential animation
+// when reduced motion is preferred.
+
```
mejiaj, line 7: SASSDoc style comments for mixins and functions would be really helpful.

```suggestion
/// Transitions with default easing and option to disable inessential
/// animations when reduced motion is preferred.
///
/// @param {String} $prop - Property to transition (e.g., 'top', 'opacity', 'all').
/// @param {String | Variable} $easing [$project-easing] - Easing function.
/// @param {Boolean} $essential [false] - Toggle essential transitions.
///
/// Example usage:
/// @include u-transition('top');
/// @include u-transition('opacity', 'ease-in-out');
/// @include u-transition('all', cubic-bezier(0.17, 0.67, 0.83, 0.67), true);
```

mejiaj (commented):
@aduth thanks for submitting this! I've tested with both `Reduce motion` settings and everything looks good. Minor suggestion for SASSDoc mixin.

_I personally, prefer reducing motion._

---

I also want to make sure I'm understanding the second point in this [comment](https://github.com/uswds/uswds/pull/6268#issuecomment-2607527677) correctly.

>- Alternatively, we could refactor the approach to either inline the media-query, or introduce a mixin which wraps @content with @media not(prefers-reduced-motion)

Do you mean inline `prefers-reduced-motion` **strictly** for this single animation? If so, it seems fair to add it inline because:

1. We're not using many animations and it seems unnecessary to create a mixin for this
2. The animation itself is **only** found in this file
3. We already have several other media queries in `usa-nav`, so wouldn't look out of place

aduth:
> Do you mean inline `prefers-reduced-motion` **strictly** for this single animation? If so, it seems fair to add it inline because:

Yep, that's what I meant. I agree that -- at least in the short term -- adding the media query to the existing style is the most direct option. Maybe an extra mixin could be added in the future.

Updated in 6ca2a6f.

packages/uswds-core/src/styles/mixins/utilities/_transition.scss
aduth: > SASSDoc style comments for mixins and functions would be really helpful.

Looks good 👍 Added in fdac9048d.

packages/uswds-core/src/styles/mixins/utilities/_index.scss
```diff
@@ -42,6 +42,7 @@
 @forward "text-indent";
 @forward "text";
 @forward "top";
+@forward "transition";
```
mahoneycm, line 45: **thought:** We should consider adding guidance to our [Utilities pages](https://designsystem.digital.gov/utilities/) to document this new utility for users.

packages/uswds-core/src/styles/mixins/utilities/_transition.scss
```diff
@@ -0,0 +1,28 @@
...
+///
+/// @param {String} $prop - Property to transition (e.g., 'top', 'opacity', 'all').
+/// @param {String | Variable} $easing [$project-easing] - Easing function.
+/// @param {Boolean} $essential [false] - Toggle essential transitions.
+///
+/// @example
+///   @include u-transition('top');
+///   @include u-transition('opacity', 'ease-in-out');
+///   @include u-transition('all', cubic-bezier(0.17, 0.67, 0.83, 0.67), true);
+@mixin u-transition($prop, $easing: $project-easing, $essential: false) {
```
mahoneycm, line 16: **Praise:** `$essential` is a nice touch! Matches w3c guidance and allows this utility to be very resilient. Thanks!

mahoneycm, line 16: **Thought:** Would there be any benefit to defining the `$prop` default to `all`? 

Users could then add `@include u-transition` with no props to add project standard transitions to all transitions.

```suggestion
@mixin u-transition($prop: "all", $easing: $project-easing, $essential: false) {
```

mahoneycm (approved):
This is great! This will benefit users and developers alike. Thanks @aduth 

Left a couple comments for additional considerations but nothing blocking

packages/uswds-core/src/styles/mixins/utilities/_index.scss
aduth: Just to clarify, this would be a separate pull request in the `uswds-site` repository, correct?

packages/uswds-core/src/styles/mixins/utilities/_transition.scss
aduth: > **Thought:** Would there be any benefit to defining the `$prop` default to `all`?

I'd be open to it. I do like the idea of a fully-defaulted version of a mixin. On the other hand, I personally can't think of a single instance that I've used `transition` and not had a specific property in mind that I was trying to animate, vs. explicitly wanting the transition for everything. And I'd worry that setting that as a default behavior could create the potential for unintended animations.

mejiaj: Using `all` could have performance drawbacks if used in many places. 

The current solution seems like a good place to start.

mejiaj (approved):
Thanks for the quick updates!

packages/uswds-core/src/styles/mixins/utilities/_index.scss
mahoneycm: @aduth correct! Just wanted to flag for internal discussion.

@mejiaj right, just noting we have utility pages that capture the rest of the available utilities, so it's worth discussing if we want to create guidance for this utility in the future

packages/uswds-core/src/styles/mixins/utilities/_transition.scss
mahoneycm: Sounds good! I'll resolve this thread

heymatthenry (approved):
Thank you, @aduth 🫡

PR #6269, opened 2024-12-20
USWDS: Add Prettier configuration file

aduth:
# Summary

Added a Prettier configuration file to the project to ensure consistent defaults across developer systems.

## Breaking change

## Problem statement

As a contributor, USWDS expects me to run Prettier on all files, but this can have unpredictable results if I have a Prettier configuration file in a parent directory outside the USWDS project, since USWDS doesn't define its own configuration.

A developer may add a Prettier configuration to a common parent directory from their code projects in order to have some sensible default behavior absent a project's own configuration when used in combination with a Prettier editor integration (e.g. VS Code Prettier extension with "Format on save").

## Solution

Adds a configuration file so that Prettier will consistently use default settings.

## Testing and review

Add a Prettier configuration file in a parent directory from the USWDS project with configuration values different from the default, and verify that lints pass:

```
echo "printWidth: 100" > ../.prettierrc.yml
npm exec gulp lintSass
```

On `develop`, the above would fail, because Prettier will traverse up from the USWDS directory to try to find a configuration file.

---


amyleadem:
> [!note]
> We do something similar to this in the web-components directory in [.prettierrc](https://github.com/uswds/web-components/blob/develop/.prettierrc)

.prettierrc.json
amyleadem, line 1: **question (non-blocking):** Does the team have a preference about making this `.prettierrc` vs `.prettierrc.json`?

amyleadem (approved):
Looks good to me! Thanks @aduth for the contribution. 

I was able to successfully run prettier and get the expected results.

.prettierrc.json
mahoneycm: Personally, I like `.pretierrc` because it's cleaner _but_ we have a `.mocharc.json` and `.eslintrc.yml` so it might be better to keep the file name for consistency. It also directly defines that this file is written in `json` and not YAML

mahoneycm (approved):

mahoneycm:
@aduth thank you for the clear testing instructions! I was able to compare to `develop` and this is a welcomed change!

.prettierrc.json
anselmbradford: Since this is an empty config, would it make sense to instead add a `prettier` key to the `package.json` (which has higher precedence, see --> https://prettier.io/docs/en/configuration)? This would also be consistent with existing precedent, e.g. the mocha config (https://github.com/uswds/uswds/blob/develop/package.json#L51).

aduth: > Since this is an empty config, would it make sense to instead add a `prettier` key to the `package.json` (which has higher precedence, see --> https://prettier.io/docs/en/configuration)? This would also be consistent with existing precedent, e.g. the mocha config (https://github.com/uswds/uswds/blob/develop/package.json#L51).

I don't personally have any strong preference, but FWIW I noticed that the Mocha configuration value referenced there doesn't actually work, partly because the file doesn't exist, but also because Mocha doesn't support referencing config files like that.

thisisdano (approved):

PR #6289, opened 2025-01-07
Add explicit background color to footer links

anmazz:
# Summary
(Sorry I've been trying to set up signed commits, history has been a bit wonky as I try to fix it).
Add explicit base-lightest background color to the footer primary and secondary links. In Chromium browsers, these links looked fine due to the default button color being #EFEFEF.

Chrome footer:
button usa-footer_

and the Chrome default button:
Hello

## Breaking change

## Preview link

Preview link:

## Problem statement
Inconsistent big footer button colors.

## Solution
Normalizes footer links across browsers with explicit background color.

## Testing and review

Check all browsers in mobile mode <480px wide for big footer colors. I verified in Safari, Chrome, and Firefox locally.

### DIFF
--- a/packages/usa-footer/src/styles/_usa-footer.scss
+++ b/packages/usa-footer/src/styles/_usa-footer.scss
@@ -84,6 +84,7 @@ $-chevron-expand-more: map-merge(
   border: 0;
   border-top: 1px solid color("base-light");
   display: block;
+  background-color: color("base-lightest");
 
   @include at-media("mobile-lg") {
     @include u-padding-x(0);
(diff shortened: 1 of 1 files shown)


opensource-joe:
This applies cleanly against develop. I checked the color choice against the rest of the file: `.usa-footer__primary-section`, the parent of both link classes here, already uses `background-color: color("base-lightest")`, so this isn't a new value, it's just making the links match a background their container already declares. That also explains the bug: an opaque default button background paints over the inherited color in some browsers but not others. On the signed commits worry in the description, the one commit here is verified fine.

ryparker:
@coderabbitai full review

PR #6290, opened 2025-01-07
USWDS - Exclude links with usa-button class from text color change

### LINKED ISSUE #6258: USWDS - Bug: dark background link color overrides link button color
### Describe the bug

The styles for links within a `.usa-dark-background` container override the styles for an `<a>` element with the `.usa-button` class. If your button variant has a dark text color, it will be overridden with the light link color. This is not a problem for `<button>` elements; it's only a problem for links with a button class. 

1. Create a button variant that uses a light background and dark text. 
2. Apply that variant to a link element (not a button)
3. Put that link within a `.usa-dark-background` container
4. See how the dark button text is overridden with the style for light-colored links within dark backgrounds 

### Expected Behavior

The color of text within button variants should not depend on whether or not that button is a `button` element or an `a` element. A button variant with dark text should remain dark within a dark container, regardless of being a button or a link. 

### Related code

In the screenshot below, I used the code sample from https://designsystem.digital.gov/components/button/ but replaces the `button` elements with links. 

### Screenshots

Image

### System setup

- "@uswds/uswds": "^3.7.1",
- Macbook
- Sonoma 14.5
- Chrome Version 131.0.6778.140 (Official Build) (x86_64)

### Additional context

_No response_

### Code of Conduct

anmazz:
# Summary
Adds `:not(.usa-button)` to links contained in dark backgrounds so that the light colored link text does not override button styling.

## Breaking change

## Related issue

Closes [#6258](https://github.com/uswds/uswds/issues/6258)

## Problem statement
Links styled with usa-button on dark backgrounds prior to this change take on the dark background link color.
This is a link styled as a button

## Solution
There might be a better solution here, especially if the dark background impacts the link text color in other areas that might need to be excluded. I have excluded any links with classes `usa-button` from having their color overridden.

With exlcusion:
This is a link styled as a button

## Testing and review
Added `<a class="usa-button ...">` to the test files.

### DIFF
--- a/packages/usa-dark-background/src/styles/_usa-dark-background.scss
+++ b/packages/usa-dark-background/src/styles/_usa-dark-background.scss
@@ -10,7 +10,7 @@ $background-context: "Background";
     color: color($theme-text-reverse-color);
   }
 
-  a {
+  a:not(.usa-button) {
     @include set-link-from-bg(
       "base-darker",
       $theme-link-reverse-color,
--- a/packages/usa-link/src/usa-link.twig
+++ b/packages/usa-link/src/usa-link.twig
@@ -10,7 +10,16 @@
   This is a link that opens in a <em>new</em> tab and goes to an <a class="usa-link usa-link--external" rel="noreferrer" target="_blank"  href="https://i.giphy.com/media/WPzQF6ruiIIVzHNlwX/source.gif">external website</a>.
 </p>
 
+<div class="padding-1">
+  <a class="usa-button">This is a link styled as a button</a>
+  <a class="usa-button usa-button--accent-cool">This is a link styled as an accent button</a>
+</div>
+
+
 <div class="usa-dark-background padding-1 display-inline-block">
   <p>This is <a class="usa-link" href="javascript:void(0);">a text link on a dark background</a>.</p>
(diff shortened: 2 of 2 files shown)


opensource-joe:
This applies cleanly against develop, and #6258 is still open with nothing else fixing it. `:not(.usa-button)` is a good, narrow fix: plain CSS, no new class needed. The question you raised yourself in the description is the one worth answering: are there other components rendered as anchors, like tags, pagination, or breadcrumbs, that would hit the same dark-background override if nested there? Might be worth deciding before merge whether this needs to generalize beyond usa-button, or if that's fine as a follow-up.

ryparker:
@coderabbitai full review

ryparker:
@coderabbitai full review

ryparker:
At `ff6677c8`, excluding button-styled links also changes the priority of ordinary link styles. The [new selector](https://github.com/uswds/uswds/blob/ff6677c87f2bc6f2cab6e6d6c2ac6a10a3c9ccbf/packages/usa-dark-background/src/styles/_usa-dark-background.scss#L13) changes `.usa-dark-background a` from `(0,1,1)` specificity to `.usa-dark-background a:not(.usa-button)` at `(0,2,1)`. A later `.usa-dark-background .custom-link { color: #fff; }` rule previously won and now loses. That's the reason for the breaking-change label.

Could we use `a:where(:not(.usa-button))` inside `.usa-dark-background`? It applies the same exclusion without increasing specificity. Please check ordinary and button-styled links, their state colors, and a later custom override. The PR needs to stay on hold until those existing overrides work.

Issue #6117, opened 2024-10-15
USWDS - Bug: Range slider focus improvements


amycole501:
### Describe the bug

We found two potential issues in the range slider audit:

1. The current version of the range slider receives keyboard focus around the entire component. The only interactive element is the button which also receives focus. The dual focused elements doesn't seem intuitive as the slider bar isn't the interactive part.

2. Also, the focus indicator on the button is quite small. We recommend increasing it at the very least to match closer to what W3C offers in their examples.

### Steps to reproduce the bug

1. Go to https://designsystem.digital.gov/components/range-slider/
2. Keyboard tab into the range slider
3. Note that the entire element receives focus. 
4. Note also how small and nearly imperceptible the focus indicator is on the button. It's a small blue outline and not offset or larger. The touch target may not meet the enhanced size.

### Expected Behavior

See an example of other range sliders where focus is only on the button:
https://carbondesignsystem.com/components/slider/usage/#feedback
https://www.w3.org/WAI/ARIA/apg/patterns/slider/examples/slider-color-viewer/  <<the entire component receives focus but the button's focus is much clearer. The other slider examples from W3 only show the button receiving focus.

### System setup

PC, Chrome

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [x] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

amyleadem:
Next steps: Evaluate the landscape to understand what is the common focus state for range sliders

alex-hull:
Will add a more high level summary in another comment, but link to the [findings document](https://docs.google.com/document/d/1hF5uEp8eO4UBQtBkksX_G5FASu462-6_qgA62sCgYGc/edit?usp=sharing) is available

amycole501:
Following up on this, we received some great feedback from a user who is visually impaired/low vision. We asked them a few questions to test the range slider focus area and these were the responses:

Here is my take on the range slider.  I did not attempt to use any keyboard shortcuts with the slider, though I did use Tab and Shift + Tab to see how the focus moved.
 
1.	**Are you able to tell where the focus should be on the slider button? Does it have enough focus?**
[...]

PR #6302, opened 2025-01-13
USWDS - Range Slider Missing Cues

### LINKED ISSUE #6116: USWDS - Bug: Range slider missing visible cues for value
### Describe the bug

The range slider has audible cues that indicate "22 out of 100" for example, but that type of granular detail isn't clear when viewing the range slider. The location of the button doesn't indicate the current value nor does the entire slider bar indicate the capacity (0-100 for example).

2. View the range slider
3. Operate the range slider using JAWS, VO, or NVDA
4. Hear the precise values announced.
5. Note that there are no visible indications that reflect those exact numbers.

### Expected Behavior

We want to give the most accessible example of the component for users to follow. By allowing SR users to hear the precision value of the amount, we are not providing an equal experience for sighted users who instead get a rougher, "ballpark" range experience. Sighted users also are unable to tell the full amount (in the case of our example, it goes from 0-100).

### Related code

_No response_

### Screenshots

_No response_

### System setup

JAWS 2024, PC, Chrome

### Additional context

_No response_

### Code of Conduct

cathybaptista:
# Summary

Currently, the range slider has audible cues that aren't available for sighted users. This update adds a value to the slider so that the current value is visible.

## Breaking change

## Related issue

Closes #6116

## Related pull requests

[Range Slider Changelog](https://github.com/uswds/uswds/pull/6302)

## Preview link

https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cb-range-slider-missing-cues/?path=/story/components-form-inputs-range--range

## Problem statement

We want to give the most accessible example of the component for users to follow. Currently, only screen reader users hear the slider's value. This results in an inconsistent experience for users.

Currently, the fix is to provide a text box that tracks the range's value. (https://app.mural.co/t/gsa6/m/gsa6/1731686667058/827f9384efe66e042c0d44463a82922ebf8fc674)

Other additions and future improvements are included in #6117 

See examples that do include the numerals:
https://carbondesignsystem.com/components/slider/usage/
https://www.lightningdesignsystem.com/components/slider/

## Solution

Add a current value field to the range slider. This field indicates the current value of the slider. This plus min/max values on the slider (the user can get there using the range slider) means sighted users and SR users get the same information.

This solution uses javascript to add and update a "current value" field. Because there was only one element before and now there are two in a side-by-side format, an additional div aligns the range and field.

## Major changes

**This change requires new unit tests**

## Testing and review

Run the code in storybook. Test that:
1. The range slider and all of its variants function as expected without error.
2. Add some beginning, ending and interim values to the slider to make sure it still works as expected.
3. Check for regressions.


alex-hull (dismissed):
Sounds good with JAWS and looks good zoomed in at 200%!

amyleadem (changes requested):
@cathybaptista Thanks for the work here! The element updated the visual display of the value as expected. Cool to see it!

I added some comments below to make more readable and resilient. Please let me know if you have any questions.

packages/usa-range/src/index.js
mejiaj, line 1: ### Dynamically added elements

Delete the previous volume example. Open up your browser devtools, go to console, and enter the following to create a range slider.

<details>
<summary>Dynamically create range</summary>

```js
// Open console and enter the following JS.

// Create a document fragment and elements.
const fragment = document.createDocumentFragment();

const label = document.createElement('label');
label.setAttribute('for', 'volume');
label.classList.add('usa-label');
label.textContent = 'Volume';

const input = document.createElement('input');
input.classList.add('usa-range');
input.type = 'range';
input.id = 'volume';
input.name = 'volume';
input.min = '0';
input.max = '11';

// Append the label and input to the fragment.
fragment.appendChild(label);
fragment.appendChild(input);

// Append fragment to the `.usa-form` selector.
const usaForm = document.querySelector('.usa-form');
usaForm.appendChild(fragment);
```
</details>

Notice that a span for value isn't created. We should check if a span already exists and create one if it doesn't.

capture-Arc-2025-02-10@2x

packages/usa-range/src/styles/_usa-range.scss
mejiaj, line 1: ### Layout shift

When the numbers change the layout seems to shift. For example, going from single to double digit and double to triple.

capture-Arc-2025-02-10

mejiaj (changes requested):
Visually, it looks good @cathybaptista, adding some comments on functional issues I found when testing.

packages/usa-range/src/styles/_usa-range.scss
mahoneycm: After the latest update, I'm still seeing a layout shift once the span reaches 4 digits

[Kapture 2025-02-14 at 10.56.23.webm](https://github.com/user-attachments/assets/4b9827f8-8e11-4366-853d-487b3e4ce043)

We should try to refactor the styles so that the span length will never affect our range slider.

Maybe compare styles to the [Carbon Design System](https://carbondesignsystem.com/components/slider/usage/) slider you listed as an example.

Are there flex properties we can utilize to ensure the range doesn't shrink? Or some text property that would make the span extend to the right rather than the left?

packages/usa-range/src/index.js
```diff
@@ -5,7 +5,31 @@ const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
...
+ * Dynamically create a span field with the range slider's initial value.
+ *
+ * @comment This is the visual representation of the value of the range slider
+ * and appears in an adjacent span. This runs on init.
+ */
+const createEnhancedRangeSlider = (el) => {
+  const rangeSliderEl = el;
+  const rangeSliderVal = document.createElement("span");
+  rangeSliderVal.className = VALUE_WRAPPER;
+  rangeSliderVal.textContent = rangeSliderEl.value;
```
mahoneycm, line 23: **Enhancement:** Utilize `data-text-unit`

We use `data-text-unit` to add a unit suffix to the range slider screen reader callouts.

To keep information consistent, we should consider adding this optional suffix to the visual span.

packages/usa-range/src/index.js
```diff
@@ -5,7 +5,31 @@ const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
...
 
+const RANGE_WRAPPER = `${RANGE_CLASSNAME}__wrapper`;
+const VALUE_WRAPPER = `${RANGE_CLASSNAME}__value`;
+/**
+ * Dynamically create a span field with the range slider's initial value.
+ *
+ * @comment This is the visual representation of the value of the range slider
+ * and appears in an adjacent span. This runs on init.
+ */
+const createEnhancedRangeSlider = (el) => {
```
mahoneycm, line 19: **Suggestion (blocking):** Reduce redundant information for screen readers

When using VoiceOver, I hear the expected callouts for the range slider.

When moving to the span, I hear the value, and then clickable.

ex:

> 20, clickable

We might want to consider ignoring this span since screen readers have their own callouts thanks to the dynamic `aria-valuetext` that is created by the `updateCallout` function.

Pinging @alex-hull  to make sure this approach sounds correct from an accessibility perspective!

packages/usa-range/src/test/template.html
mahoneycm, line 1: **Suggestion:** Multiple sliders test

Since we are dynamically generating the span, and since we saw issues with it appending to the wrong range slider during early work, we should consider adding a test with two spans and confirm they are added and updated correctly.

Can we update this template and tests to include two sliders?

This will also help reduce redundancy between this template and the `sr-callout.spec.html`

packages/usa-range/src/test/range.spec.js
mahoneycm, line 1: **Praise:** Great work on your first batch of unit tests! These look and work great!

packages/usa-range/src/usa-range.twig
```diff
@@ -12,5 +12,5 @@
     {% if text_preposition %}data-text-preposition="{{ text_preposition }}"{%- endif %}
     {% if disabled_state == 'disabled' %} disabled {%- endif %}
     {% if disabled_state == 'aria-disabled' %} aria-disabled="true" {%- endif %}
-  >
+    >
```
mahoneycm, line 15: **Polish:** Looks like an accidental space was committed here!

```suggestion
  >
```

packages/usa-range/src/styles/_usa-range.scss
mahoneycm: Looks great!

mahoneycm (changes requested):
Just two minor styling fixes! Let me know if you have any questions

amycole501 (approved):
Works in NVDA. I no longer have JAWS but I'll accept Alex's previous review using JAWS as proof that this works in both screen readers on Windows.

heymatthenry:
Recognizing that there are small refinements that can be done here, but merging this in as-is and will open another issue for those enhancements.

PR #6306, opened 2025-01-14
USWDS - Skipnav: Fix visible wrapping into viewport at high zoom

### LINKED ISSUE #6260: USWDS - Bug: skipnav text overlays content at high zoom levels
### Describe the bug

At high zoom levels, the "Skip to main content" skipnav component's text becomes visible when it should not be, and overlays content. I have attached a screenshot showing this on the USWDS site itself.

Most users will not browse the web this way, but there's a certain cruel irony in this is most likely to affect users with reduced vision and it's an accessibility component that's getting in the way.

This came up as an issue on Login.gov but in debugging I realized it applies to USWDS itself.

1. Browse a USWDS site in Chrome on an iPhone (this likely applies more broadly)
2. Using the three dots at the bottom of the Chrome UI, select "Zoom text..."
3. Zoom in multiple levels

### Expected Behavior

The skipnav content does not become visible unless triggered, regardless of zoom level.

### Related code

_No response_

### Screenshots

Image
^ Chrome on iPhone (iOS 18.2 on an iPhone 14)

Image
^ same Chrome on iPhone setup, but on login.gov where the issue was first reported

Image
^ a less realistic use case, but Chrome on a Mac (visiting local login.gov dev site)

### System setup

- USWDS version: 3.9.0 locally, but also designsystem.digital.gov itself
- Device: iPhone 14
- OS: iOS 18.2
- Browser: Chrome (131.0.6778.154)

I cannot reproduce it in Safari on iOS, but do not have a variety of devices to test on.

I was only able to reproduce this on a desktop by making the browser window very small and zooming in a ton; it made it easier for me to poke at for debugging, but I suspect it is far less likely to be encountered in the wild this way.

### Additional context

I was experimentally able to fix this by changing the `skipnav` component's `top` attribute in CSS from `-3.8rem` to `-38rem`, but this is likely an inelegant fix. (I am not a designer.)

### Code of Conduct

aduth:
# Summary

**Fixed an issue where inactive skipnav links would overlap page content at very high zoom levels.**

## Breaking change

## Related issue

Closes #6260

## Preview link

In local Storybook:

- Default: http://localhost:6006/?path=/story/components-skipnav--default
- Focus: http://localhost:6006/?path=/story/components-skipnav--focus

## Problem statement

Previously, if the user zoomed in enough to cause the skipnav link content to wrap onto multiple lines, the content would overlap the page.

## Solution

Changes from hard-coding an offset around an assumed height of the skipnav to aligning the bottom of the skipnav element with the top of the page. This addresses the issue because the problem stems from the element height changing when the text wraps onto multiple lines.

Codepen visualization: https://codepen.io/aduth/pen/mybWbea

## Testing and review

Verify zoom never causes link to appear:

1. Go to http://localhost:6006/?path=/story/components-skipnav--default
2. In the Storybook preview controls, click the zoom icon button icon repeatedly
3. Verify that the text content of the inactive skipnav never appears

Before|After
---|---
before|after

Verify that focused skipnav appears as expected when tabbed:

1. Go to http://localhost:6006/iframe.html?args=&id=components-skipnav--default&viewMode=story
2. Tab into page content
3. Verify that the skipnav animates into view


opensource-joe:
The direction here is still right and #6260 is still open with no competing fix. One heads up: this no longer applies cleanly to `develop`. The SCSS hunk expects a `transition: $project-easing;` line that isn't there anymore, it looks like it got replaced by a `@include u-transition("top")` mixin call at some point after this PR opened. The other files (JSON split, stories) still apply as-is; it's just the stylesheet that needs a rebase.

ryparker:
@coderabbitai full review

PR #6308, opened 2025-01-15
USWDS - utils: replace `resolve-id-refs` dependency

heymatthenry:
# Summary

This change removes the `resolve-id-refs` package and re-implements the functionality it was providing for the password reset form.

## Breaking change

## Problem statement

`resolve-id-refs` is triggering a [CodeQL vulnerability alert](https://github.com/uswds/uswds-tutorial/pull/67/checks?check_run_id=35615138469) in the compiled USWDS javascript. This PR just removes that dependency and re-implements the same functionality.

## Testing and review

I tested this change by running the unit test suite and by viewing the password reset form in Storybook to verify that there were no regressions.

| `resolve-id-refs`  |     0.1.0     |     --      |

### DIFF
--- a/packages/uswds-core/src/js/utils/toggle-form-input.js
+++ b/packages/uswds-core/src/js/utils/toggle-form-input.js
@@ -1,4 +1,3 @@
-const resolveIdRefs = require("resolve-id-refs");
 const toggleFieldMask = require("./toggle-field-mask");
 
 const CONTROLS = "aria-controls";
@@ -13,6 +12,21 @@ const HIDE_ATTR = "data-hide-text";
  */
 const getHideText = (showText) =>
   showText.replace(/\bShow\b/i, (show) => `${show[0] === "S" ? "H" : "h"}ide`);
+
+const resolveIdRefs = (id) => {
+  const ids = id.trim()
+    .split(/\s+/);
+
+  return ids
+    .map((i) => {
+      const el = document.getElementById(i);
+      if (!el) {
+        const err = `no element with id: ${i}`;
+        throw new Error(err);
+      }
+      return el;
+    })
+}
 
 /**
  * Component that decorates an HTML element with the ability to toggle the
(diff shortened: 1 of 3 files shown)


packages/uswds-core/src/js/utils/toggle-form-input.js
```diff
@@ -14,6 +13,19 @@ const HIDE_ATTR = "data-hide-text";
...
+
+  return ids.map((i) => {
+    const el = document.getElementById(i);
+    if (!el) {
+      const err = `no element with id: ${i}`;
+      throw new Error(err);
+    }
+    return el;
+  });
+};
```
aduth, line 27: I know it matches current behavior of `resolve-id-refs`, but I'd be curious if it's a good idea to interrupt code execution with a thrown error on a missing element, vs. silently ignoring (or `console.log`-ing) the missing element.

e.g.

```suggestion
const resolveIdRefs = (idRefs) =>
  idRefs
    .trim()
    .split(/\s+/)
    .map((id) => document.getElementById(id))
    .filter(Boolean);
```

heymatthenry: Thank you for looking! FWIW, I don't disagree at all. That said (and I recognize I'm probably being overly conservative!), for something we're trying to get in ASAP I'd prefer to leave the current behavior for now. This feature also doesn't have adequate test coverage. So what I'd prefer to do is to make an issue for your suggested improvements and write some tests such that next time something comes up here we can make changes more confidently.

aduth: Sounds reasonable to me 👍

mejiaj (approved):
Thanks @heymatthenry! Confirming this matches the current behavior. Due to urgency, I agree with follow-up work to improve this work with test cases to match.

thisisdano (approved):
Toggle works as expected and uses a better method for handling the string. Nice to see our dependencies down to one.

PR #6309, opened 2025-01-15
USWDS - Core: Fix broken project-wide Mocha configuration reference

aduth:
# Summary

Fixed project test configuration for contributors running single-file tests.

## Breaking change

## Related issue

Related discussion: https://github.com/uswds/uswds/pull/6269#discussion_r1904435115

## Preview link

(No visual changes)

## Problem statement

As a contributor iterating on JavaScript changes, I want to be able to run single test files in isolation so that there's a short feedback loop between code changes and test runs, so that I don't have to run the entire test suite to verify changes to a single component.

This was seemingly intended to be supported with the existing `mocha` entry in `package.json`, but (a) the file referenced in this directive no longer exists and (b) while Mocha does support a `mocha` key in `package.json`, it only supports individual configuration values and not a reference to a configuration file.

## Solution

Removes `mocha` from `package.json` and re-exports the updated project-wide configuration file path from the root directory `.mocharc.js`.

An alternative solution could move `.mocharc.json` from `packages/uswds-core/src/js/utils/test` to the root directory, though there's some precedent with `.eslintrc.yml` existing in that same directory.

## Testing and review

Run the Mocha command directly with one or more file paths:

```
npm exec mocha packages/uswds-core/src/js/utils/test/select.spec.js
```

On `develop`, this command will fail because the test depends on a DOM document initialized through `jsdom-global/register` in the project-wide `.mocharc.json`. On this branch, the tests should pass.

Before|After
---|---
image|image


ethangardner (approved):
Nice work as usual. Thanks for the contribution. This was a welcome improvement and will be useful for future work.

PR #6313, opened 2025-01-16
fix(core): preserve password toggles with missing targets

### LINKED ISSUE #6310: USWDS: Utils improvements after removing `resolve-id-refs`
This is follow-on work from #6308. In that PR we replaced an old dependency with our own implementation of the same functionality. However, there are a couple aspects of that work that were less than ideal. First, a comment on the PR (correctly) identified one shortcoming with the original library code that also applies to our own implementation:

              Granting that this matches current behavior of `resolve-id-refs`, but I'd be curious if, rather than interrupting code execution with a thrown error on a missing element, it might be better to silently ignore (or `console.log`) the missing element.

e.g.

```suggestion
const resolveIdRefs = (idRefs) =>
  idRefs
    .trim()
    .split(/\s+/)
    .map((id) => document.getElementById(id))
    .filter(Boolean);
```

_Originally posted by @aduth in https://github.com/uswds/uswds/pull/6308#discussion_r1916973736_ 

This was done for the sake of expedience and to make sure that the new code did more-or-less the same thing as the original dependency code. Also, because there is no test coverage for the specific functionality in the password reset form that uses this code, I wanted to change as little as possible. To complete this issue, a dev should:

- [ ] Handle errors more gracefully than the current implementation; and
- [ ] Improve test coverage such that we can work more confidently in this corner of the codebase

aduth:
## Summary

Simplify `resolveIdRefs` and preserve password-toggle behavior when referenced fields are missing. A missing ID no longer prevents valid targets from toggling. If no targets resolve, the control keeps its existing label, pressed state, and focus instead of claiming that it revealed a password. An absent `aria-controls` is handled without throwing.

Closes #6310. This follows the resolver introduced in #6308.

## Breaking change

This is not a breaking change. Valid single-target and multiple-target markup keeps its existing behavior. The intentional behavior change concerns invalid references: unresolved IDs are skipped and an empty result leaves the control unchanged. No documented contract requiring callers to catch an exception for invalid `aria-controls` was identified.

## Testing and review

Verified at `82861584d1e449301d330f4268145d1d0072b10e`:

- Eleven focused Mocha tests pass. Five new regression cases fail against the previous implementation and pass with this repair.
- Resolver fixtures are cleaned between tests, and assertions check element identity.
- Chromium, Firefox, and WebKit probes cover valid single/multiple targets, a missing middle target, all targets missing, empty references, absent `aria-controls`, repeated keyboard activation, and retained focus.
- Paired base/repair results match for valid markup. With no resolved targets, the label, pressed attribute, input types, and focus remain unchanged.
- Formatting and whitespace checks pass. Current build, signature, CodeQL, and security checks pass. Before/after screenshots are retained in the local validation report.

To review locally, open the sign-in form story, enter a password, and activate Show password twice by keyboard. Verify masking, the control label and pressed state, and retained focus. Repeat with a missing reference alongside a valid field, then with no resolvable targets.

WebKit execution does not establish released Safari or screen-reader speech. Before merge, refresh the branch base and rerun final verification. The current `PR title` workflow checks an older trusted base that lacks `.github/scripts/check-pr-title.test.mjs`, so it fails before validating the otherwise conforming title. This PR remains open.


opensource-joe:
The fix applies cleanly against `develop`, unchanged since #6308. Replacing the throw with `.filter(Boolean)` is a real behavior improvement, not just style: today one bad ID in `aria-controls` throws and breaks the whole password toggle, including the elements that were referenced correctly; this only drops the missing one. Worth knowing as a tradeoff: a typo'd ID now fails silently instead of loudly, though breaking the whole component is worse. Tests cover both the happy path and the missing-ID case, and no other open PR touches this function.

ryparker:
@coderabbitai full review

ryparker:
Fixed the missing-target edge cases in `82861584`. Previously, the button could say “Hide my typing” even when none of its referenced fields existed, and a missing `aria-controls` attribute threw an error.

The button now stays unchanged when there are no valid targets. Valid single and multiple targets still toggle as before, and a missing reference between two valid ones no longer prevents them from working. The resolver tests also clean up their fixtures and check that the returned elements are the right ones.

All 11 focused Mocha tests pass. The five new regression cases fail with the previous implementation and pass with this fix. Chromium, Firefox, and WebKit checks cover valid, partly missing, empty, and absent target lists, repeated keyboard activation, and retained focus. The before/after comparisons preserve valid behavior; invalid lists leave the label, pressed state, input types, and focus unchanged without errors. Formatting checks pass, and screenshots are saved with the local validation report.

This doesn't require changes to existing valid markup. The PR is still open for review. Released Safari and screen-reader speech were not tested.

PR #6320, opened 2025-01-23
USWDS: Add notifications for 3.12.0 

amyleadem:
# Summary

Added Sass notifications for 3.12.0

## Related issue

N/A

## Related pull requests
N/A

## Preview
image

## Testing and review
Run `npx gulp buildSass` and view the release notifications
- Confirm the relevant items from 3.12.0 are included in the notifications
- Confirm the notifications are formatted as expected


mejiaj (approved):
LGTM

capture -Warp-2025-01-23@2x

packages/uswds-core/src/styles/_notifications.scss
```diff
@@ -206,6 +206,15 @@ $uswds-notifications:
...
 + "\A   to allow the full focus outline to show."
++ "\A "
++ "\A --------------------------------------------------------------------"
++ "\A 3.12.0"
++ "\A Release notes: github.com/uswds/uswds/releases/tag/v3.12.0"
++ "\A "
++ "\A - Display: [usa-in-page-navigation]"
++ "\A   Fixed a bug that prevented in-page navigation from scrolling to nested headings."
++ "\A   ✏️ Teams that use `data-scroll-offset` should confirm this change"
++ "\A     does not cause regressions in scroll behaviors."
```
mejiaj, line 217: **issue (non-blocking)**: @amyleadem the other in-page navigation change also has an edit note and could be useful.

```suggestion
+ "\A     does not cause regressions in scroll behaviors."
+ "\A - Display: [usa-in-page-navigation]"
+ "\A   Added the `data-minimum-heading-count` property to the in-page navigation component."
+ "\A   ✏️ Teams should customize the value of this property based on their content needs."
```

amyleadem: I was torn on including this one because I didn't see other references to `data-` properties in the file. I support including it because teams might need to take action on it, but going to hold off on committing this in case anyone else has thoughts.

mahoneycm, line 217: **Idea:** I noticed we've included Sass notifications in the past. Since the Sass depcrecations are getting well known to devs, it might be a good addition?

```suggestion
+ "\A     does not cause regressions in scroll behaviors."
+ "\A - Sass: Resolved Sass deprecation warnings related to the color function. 
+ "\A    This change ensures compatibility with Dart Sass 2.0.0 and eliminates the use of deprecated color functions.
```

mahoneycm (commented):
One non-blocking suggestion, to accompany James! Otherwise lgtm.

packages/uswds-core/src/styles/_notifications.scss
thisisdano: I think my perspective here when it comes to notifications, is that we focus on what might potentially be breaking or what needs manual intervention. In this case, it's not that devs should use this new feature, but that they should confirm that it does not cause regressions on pages with only a single header

thisisdano:
```suggestion
+ "\A     does not cause regressions in scroll behaviors."
+ "\A - Display: [usa-in-page-navigation]"
+ "\A   Added the `data-minimum-heading-count` property to the in-page navigation component."
+ "\A   ✏️ Teams should check to make sure the default minimum of 2 does not result"
+ "\A   in regressions."
Commit suggestion
Add suggestion to batch
```

Something like that?

thisisdano: I don't think we need this one here

heymatthenry (approved):
LGTM, and incorporates feedback from the team.

juliaelman (approved):

PR #6385, opened 2025-02-07
USWDS - Character Count: Fix label selector

### LINKED ISSUE #6384: USWDS - Bug: Character Count component's selector query for label's `for` attribute does not include quotes
### Describe the bug

When you register the character count component at a node via `characterCount.on(node)` it queries for a label by its `for` attribute using the `id` of the input.  If that `id` contains certain characters such as a colon `:`, it will fail in its query for the label's `for` attribute:

```
SyntaxError: Failed to execute 'querySelector' on 'Document': 'label[for=character-count-:r0:]' is not a valid selector.
```

It should be placing the `for` value in quotes, e.g. `node.querySelector("label[for='character-count-:r0:']")`.  This will enable the character count component to work for any valid HTML id.

I do plan to open a PR with the fix!

Assuming you are in a project that is set up for the USWDS, in your HTML body use the example from https://designsystem.digital.gov/components/character-count/ and use a valid ID on the `input` that would break a selector query.  In this case it is `character-count-:r0:`:

```html
<script>
  const container = document.getElementById("container");

  // This should throw an error
  characterCount.on(container);
</script>

<form class="usa-form">
  <div id="container" class="usa-character-count" data-maxlength="25">
    <div class="usa-form-group">
      <label class="usa-label" for="character-count-:r0:">Text Input</label>
      <span class="usa-hint" id="character-count-:r0:-hint">This is an input with a character counter.</span>
      <input
        class="usa-input usa-input--lg usa-character-count__field"
        id="character-count-:r0:"
        aria-describedby="character-count-:r0:-hint character-count-:r0:-info"
        name="example"
      />
    </div>
    <span
      id="character-count-:r0:-info"
      class="usa-character-count__message usa-sr-only"
    >
      You may enter up to 25 characters
    </span>
  </div>
</form>
```

### Expected Behavior

You should be able to use any valid HTML ID for your input.

### Related code

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

ealexhaywood:
# Summary

Fix label selector query for character count component.

## Breaking change

## Related issue

Closes #6384 

## Problem statement

The character count component can fail if its input element ID contains characters that would make a selector query invalid.  

## Solution

Placing the ID in quotes in the selector query fixes this.


jerseycheese:
Tested this against current `develop`. The branch here is very behind, but the one-line change cherry-picks clean and the target line is unchanged upstream, so it still applies as-is.

Reproduced the bug first: built a Character Count whose input uses a React `useId()`-style id (`:r0:`), and `CharacterCount.on()` throws exactly what the issue reports:

```
SyntaxError: 'label[for=character-count-:r0:]' is not a valid selector
```

With this PR's change (`label[for="${inputID}"]`) the same setup initializes cleanly. So the fix does what it says.

Worth flagging why this isn't an edge case: `:r0:` is literally what React 18+'s `useId()` emits, and pairing `useId()` across the input and its `<label for>` is the existing pattern. Any React/Next consumer wiring Character Count up that way hits it.

Two small things, neither a blocker:

- Quoting handles every id that's valid HTML, which is the actual bug here. If you wanted to be bulletproof against an id containing a literal `"` or `\`, `CSS.escape(inputID)` is the belt-and-suspenders version, but that's beyond what's reported.

ethangardner (approved):
LGTM, and thanks for the contribution. Sorry for the looooooong delay in getting this reviewed and merged.

PR #6387, opened 2025-02-07
USWDS - Character Count: Remove status message elements when `.off()` is called  

### LINKED ISSUE #6386: USWDS - Bug: Character Count component does not clean up its status messages when `.off()` is called
### Describe the bug

The character count component adds two status elements on `init()`, one for display and one for the screen reader.  It does not remove these elements from the DOM when `.off()` is called.

Assuming you are in a project that is set up for the USWDS, in your HTML body use the example from https://designsystem.digital.gov/components/character-count/ 

```html
<script>
  const container = document.getElementById("container");

  // This creates the status message elements (one for visual, one for sr)
  characterCount.on(container);

  // This does not clean up the status messages
  characterCount.off(container);
</script>

<form class="usa-form">
  <div id="container" class="usa-character-count" data-maxlength="25">
    <div class="usa-form-group">
      <label class="usa-label" for="character-count">Text Input</label>
      <span class="usa-hint" id="character-count-hint">This is an input with a character counter.</span>
      <input
        class="usa-input usa-input--lg usa-character-count__field"
        id="character-count:"
        aria-describedby="character-count-hint character-count-info"
        name="example"
      />
    </div>
    <span
      id="character-count-info"
      class="usa-character-count__message usa-sr-only"
    >
      You may enter up to 25 characters
    </span>
  </div>
</form>
```

### Expected Behavior

I would expect calling `.off()` would remove any elements it added to the DOM that were added in `on()` or `init()`.  This causes an issue in [React's strict mode](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-ref-callbacks-in-development) where ref callbacks are ran twice causing too many status message elements to be rendered (and aren't cleaned up when `.off()` is called).

### Related code

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

ealexhaywood:
# Summary

This PR adds a `teardown()` method to remove the status message elements that are added to the DOM when `.on()` is called for the character count module. 

## Breaking change

## Related issue

Closes #6386 

## Problem statement

The character count module adds two status elements (one for visual, one for screen reader) to the DOM via `createStatusMessages()` on `init()`.  However, it does not remove these elements when `.off()` is called.  

This prevents this module from being used properly in [React's strict mode](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-ref-callbacks-in-development) because rendering and callback refs run twice.  Since `.on()` gets called twice, and `.off()` (in a cleanup function) does not remove the status elements, you end up having 4 total status elements in the DOM (instead of the intended 2 elements).

## Solution

This PR adds a `teardown()` method to the character count module that matches on the `CHARACTER_COUNT` class at the specified `root`, and queries for the elements there were added by `createStatusMessages()` and removes them.

Let me know if you all think there is a better way!

### DIFF
--- a/packages/usa-character-count/src/index.js
+++ b/packages/usa-character-count/src/index.js
@@ -1,4 +1,5 @@
 const select = require("../../uswds-core/src/js/utils/select");
+const selectOrMatches = require("../../uswds-core/src/js/utils/select-or-matches");
 const behavior = require("../../uswds-core/src/js/utils/behavior");
 const debounce = require("../../uswds-core/src/js/utils/debounce");
 const { prefix: PREFIX } = require("../../uswds-core/src/js/config");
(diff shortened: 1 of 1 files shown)


opensource-joe:
This fixes a real bug well diagnosed, #6386 is open and the StrictMode double-invoke explanation is exactly right. I checked that `teardown` is an actual lifecycle hook `off()` calls in `behavior.js`, not just a plausible-sounding method name, so this is using the framework correctly. One detail worth noting since it's easy to get wrong: `getElementsByClassName` returns a live collection, and this correctly spreads it into an array before removing nodes, which avoids the classic bug of skipping elements when you mutate a live collection while iterating it.

ryparker:
@coderabbitai full review

PR #6388, opened 2025-02-10
USWDS - [date-picker]: Fix formatting on narrow parent containers

### LINKED ISSUE #6069: USWDS - Bug: Date picker year button overlapping with days below
### Describe the bug

When open the date picker year button appears to wrap and overlaps with days below

1. Have a 2 grid layout 3-9 with grid-gap-4
2. Add a date picker in the left column
3. Open date picker

### Expected Behavior

Date picker remains inline no matter what

Workaround `usa-date-picker__calendar__cell usa-date-picker__calendar__month-labe` added `display: flex`

### Related code

<div class="padding-2">
  <div class="grid-row">
    <div class="grid-col-3">
      <div class="usa-form-group">
        <label class="usa-label" id="appointment-date-label" for="appointment-date"
          >Appointment date</label
        >
        <div class="usa-hint" id="appointment-date-hint">mm/dd/yyyy</div>
        <div class="usa-date-picker">
          <input
            class="usa-input"
            id="appointment-date"
            name="appointment-date"
            aria-labelledby="appointment-date-label"
            aria-describedby="appointment-date-hint"
          />
        </div>
      </div>
    </div>
    <div class="grid-col-9">.grid-col-9</div>
  </div>

### Screenshots

image

### System setup

- USWDS version 3.8.1
- Browser Chrome

### Additional context

See slack thread https://gsa-tts.slack.com/archives/C3F14AHSQ/p1726526329713889

### Code of Conduct

mandylloyd:
# Summary
**Enhanced date picker responsiveness** Replaced media query-based styling with JavaScript-driven approach to adapt month/year labels to parent container width to prevent overlap issues.

## Breaking change

## Related issue

Closes #6069

## Related pull requests
None

## Preview link

[Preview link](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/ml-date-picker/iframe.html?args=&id=components-form-inputs-date-picker--default&viewMode=story)

## Problem statement
The date picker component was experiencing a layout issue where the year button would overlap with the days below when opened, causing a poor user experience.

## Solution
By using JavaScript to check the parent container width size, the appropriate styles are applied to resolve the layout issue, rather than relying on CSS media queries.

## Major changes

- Added a function `checkWidth()` to check parent container width size
- Applied existing styles to resolve layout issue
- Removed use of media query

## Testing and review
[Sandbox for reproducing exact 2 column layout bug](https://github.com/uswds/uswds-sandbox/tree/ml-date-picker)

1. Navigate to the date picker component in storybook
2. Confirm month/year does not overlap any any screen size
3. Open and link sandbox
4. Confirm month/year does not overlap in either column


packages/usa-date-picker/src/index.js
```diff
@@ -1332,6 +1334,35 @@ const selectDate = (calendarDateEl) => {
...
+ * @param {HTMLElement} el An element within the calendar component
+ */
+function checkWidth(el) {
+  const parent = el.querySelector(`.${CALENDAR_ROW_CLASS}`);
+  const child =
+    el.querySelector(`.${CALENDAR_MONTH_LABEL_CLASS}`) ||
+    el.querySelector(`.${CALENDAR_MONTH_LABEL_CLASS_NARROW}`);
+
+  // Add narrow class styling when parent width is less than 300
+  if (parent.offsetWidth < 300) {
```
mlloydbixal, line 1349: I picked 300 based on the bug example that was provided and felt this was the smallest size the styles should be switched at before it started looking bad but I'm open to other opinions here

packages/usa-date-picker/src/test/date-picker.spec.js
```diff
@@ -17,6 +17,36 @@ const tests = [
...
+  resizeListeners.forEach((listener) => {
+    originalRemoveEventListener.call(window, "resize", listener);
+  });
+  resizeListeners = [];
+  window.addEventListener = originalAddEventListener;
+  window.removeEventListener = originalRemoveEventListener;
+});
+
+// --- End Global override ---
+
```
mlloydbixal, line 49: This global clean up method was added so the resize event listener doesn't stick around and get called again on subsequent unit tests that use resize events, which was previously causing failures. 

I only added it to this file because it's the last one to run that causes the issue of all the date-picker unit tests. If this file were to be removed for some reason this logic would need to be moved to one of the other files. I was unsure if I should include it in every date-picker test file because there were quite a few and it's technically only needed at the end. LMK if anyone feels differently!

amyleadem (commented):
@mlloydbixal 

Thanks for working on this. I like the idea to check for the parent container width rather than relying on viewport width. However, I am curious if there is a non-JS approach to this issue. Could [CSS container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries) possibly work here? 

Update: More discussion on this can be found in this [Slack thread](https://gsa-tts.slack.com/archives/C050HRGN7/p1740053855608299) (🔒).

opensource-joe:
#6069 is still open and this is the only PR addressing it. There's a real bug in `checkWidth()` though: it adds a new `resize` listener to `window` every time the calendar opens and never removes the old one, and since `debounce()` returns a new function reference each call, there's no way to target the previous listener for removal even after the fact. Open and close the date picker repeatedly and listeners just accumulate. I think this is what the test file's global `addEventListener` patching is working around, stripping resize listeners after each test because leftover ones from prior tests were interfering, rather than something wrong with the tests themselves. Also still open: Amy's question about CSS container queries instead of JS, which would sidestep this whole category of bug since there'd be no listener to manage at all.

ryparker:
@coderabbitai full review

ryparker:
The responsive fix at `9a377326` removes a class that custom styles or scripts may depend on. Below 300px, [`checkWidth` replaces `.usa-date-picker__calendar__month-label`](https://github.com/uswds/uswds/blob/9a377326a9daf12c79124eff576c0b1b66c081a5/packages/usa-date-picker/src/index.js#L1351) with the narrow class. Previously, the base class stayed present at every width, including when zoom or a small column made the calendar narrow. That's the reason for the breaking-change label.

Please keep the base class and toggle an additional narrow modifier. The narrow styles can override what they need without removing the existing class or its baseline styles. Check transitions in both directions and an existing custom selector before lifting the hold.

Issue #6295, opened 2025-01-08
USWDS - Bug: In-page navigation breaks for headings inside a USWDS alert component


mgwalker:
### Describe the bug

The alert component creates a positioned container element (`.usa-alert__body`) that creates a new offset frame. The in-page navigation component then inserts an anchor inside the alert component. When clicking the link in the in-page navigation, it uses the anchor's `offsetTop`, which is the vertical distance between the selected element *and its nearest offset parent*. In this case, it's nearest offset parent is the alert container, not the page body itself.

### Steps to reproduce the bug

1. Browse to [https://handbook.tts.gsa.gov/tools/git-signing](https://handbook.tts.gsa.gov/tools/git-signing)
2. Click `About the cryptographic algorithm` in the in-page nav
3. Note that the page scrolls to the top rather than to the linked heading

### Expected Behavior

The browser should scroll to the selected heading, perhaps using [`getBoundingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) instead of `offsetTop` since the former computes absolute values (relative to the page, not any other elements in the tree).

### Related code

https://github.com/uswds/uswds/blob/main/packages/usa-in-page-navigation/src/index.js#L202

### System setup

- USWDS version: 3.11.0
- All browsers

### Additional context

Possibly related to:
- #5255

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [x] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

mejiaj:
Need to confirm if fixed by uswds/uswds#5878.

mejiaj:
@cathybaptista scheduling for confirmation please ping me or @mahoneycm if you can't get to this on 01/24.

dct2023:
Confirmed this is done.

PR #5878, opened 2024-04-22
USWDS - In-Page Navigation: Fix scrolling to nested headers

jhancock532:
# Summary

**Fixed issue with in-page navigation being unable to scroll to nested headings.** The in-page navigation can now smooth scroll to headings within the summary box and card components.

> [!important]
> There is currently a build error related to prettier. This should be resolved by PR #6319.

## Breaking change

This is not a breaking change.

## Related issue

Closes #5255

## Related pull requests

_N/A_

## Preview link

[Preview link](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/jh-fix-in-page-navigation-scroll-bug/?path=/story/components-in-page-navigation--default)

## Problem statement

The in-page navigation should navigate the user to the section of the page with the indicated heading when the heading link is clicked. 

Currently, clicking on links that refer to nested headings, within the summary box or card components (or any other custom, non-standard USWDS component) causes the page to reset its scroll. 

This poor user experience hinders users from accessing content they are seeking quickly and may mislead users to thinking desired content doesn't exist on the page. Erroneous behaviour reduces users trust in the site and USWDS branded sites as a whole. 

## Solution

The existing code uses the `offsetTop` value of the element to identify how much to scroll down the page. This value is relative to the parent container. To correctly scroll to the location of the element, the `offsetTop` must be added to the `parentOffset` of all the elements parent components. This MR adds a function to calculate the elements true offset in this manner.

An alternative solution would be to refactor the code to use `scrollIntoView` instead. This would involve refactoring the approach we use to test this component, which is being supported using a mocked offsetTop value and watching the `window.scroll` function with `sinon`. 

## Major changes

No major changes to component implementation / testing approach made.

## Testing and review

A new Storybook story in the in-page navigation component has been created for testing. If running storybook locally, go to http://localhost:6006/iframe.html?args=&id=components-in-page-navigation--test-nested-headers&viewMode=story and click on each link in the in-page navigation. The page should correctly scroll to each heading in turn.

Any feedback on the solution would be helpful. I would like to add a test case to cover this fix, but I'm not sure how I'd go about mocking offsetParent only for certain elements. It also feels like if I was to mock this for specific contexts, the test would be verifying a very specific scenario instead of being representative. Using a testing framework such as Playwright instead of Jest would be helpful here.

Issue #5255, opened 2023-04-27
USWDS - Bug: In-Page Navigation scrolling does not work with Summary Box


mattcoulon-ia:
### Describe the bug

When an h2 or h3 tag is used inside of a Summary Box component, it will appear in the In-page navigation component. Clicking on its corresponding link in the navigation does not scroll to its anchor, but rather scrolls to the top of the page. 

### Steps to reproduce the bug

1. Set up a page to use the In-page Navigation component.
2. Add sections of content with h2 and/or h3 tags (the more vertical space this takes up the easier it will be to see the issue).
3. Add a Summary Box component to the page.
4. Make the Summary Box header an h2
5. Click on the link in the In-page Navigation that is supposed to anchor to the Summary Box
6. Instead of scrolling to the Summary Box, the page will scroll to the top of `main`

### Expected Behavior

The In-page Navigation component should scroll to a Summary Box anchor when that Summary Box is part of the nav. 

### Related code

The issue can be seen by visiting [https://designsystem.digital.gov/components/summary-box/](https://designsystem.digital.gov/components/summary-box/) and clicking on 'Key information' under **On this page**.

### System setup

- USWDS version 3.4.1
- Device: Windows PC
- Operating System: Windows 10 Enterprise
- Browser and version: Chrome Version 112.0.5615.138 (Official Build) (64-bit)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

richard8:
I'm experiencing this issue too — not only with summary boxes, but with headers inside other components as well. 

The page navigation and anchor links are generated, and navigation links are highlighted as expected when scrolling through the page. But clicking on a navigation link will scroll to the top of the page rather than to the corresponding anchor.

jhancock532:
I've experienced this issue recently with the card component. You can see a live demo of the issue in action by visiting the [card component documentation page](https://designsystem.digital.gov/components/card/), then clicking on one of the card headings.

![in-page-navigation-bug](https://github.com/uswds/uswds/assets/18164832/d0705ac5-01d8-4e2d-bef2-79d06ad690e8)

[...]

PR #6397, opened 2025-02-11
Get heading/TOC offset without walking up the DOM tree

IHIutch:
# Summary

Rather than recursively walking up the DOM tree which could be quite deep, `el.getBoundingClientRect().top + window.scrollY` provides the position of the element on the page immediately.

## Related issue

- https://github.com/uswds/uswds/issues/5255
- https://github.com/uswds/uswds/issues/6295

## Related pull requests

Updates the work on this PR: https://github.com/uswds/uswds/pull/5878

## Preview link

[Preview link](https://uswds-git-develop-jonathan-hutchisons-projects.vercel.app/?path=/story/components-in-page-navigation--test-nested-headers)

## Problem statement

The current implementation walks up the DOM tree which could be extremely deep and expensive to do

## Solution

Uses a more optimal method to detect the vertical position of an element on a page

## Major changes

n/a

## Testing and review

Updates tests to match the fact that JSDOM cannot get `window.scrollY` or `el.getBoundingClientRect().top`. By using `getBoundingClientRect()` we cannot simply add up the `offsetTop` values of elements to verify the scroll position.

### DIFF
--- a/packages/usa-in-page-navigation/src/index.js
+++ b/packages/usa-in-page-navigation/src/index.js
@@ -189,24 +189,6 @@ const getSectionId = (value) => {
 };
 
 /**
- * Calculates the total offset of an element from the top of the page.
- *
- * @param {HTMLElement} el A heading element to calculate the offset for.
- * @returns {number} The total element offset from the top of its parent.
- */
-const getTotalElementOffset = (el) => {
-  const calculateOffset = (currentEl) => {
-    if (!currentEl.offsetParent) {
-      return currentEl.offsetTop;
-    }
-
-    return currentEl.offsetTop + calculateOffset(currentEl.offsetParent);
-  };
-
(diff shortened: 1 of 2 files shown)


opensource-joe:
Both linked issues are already closed by #5878, so this isn't fixing a live bug, it's a follow-up on that PR's own recursive offset function. The replacement is a real improvement though: getBoundingClientRect() is the standard way to do this and avoids correctness gaps the recursive walk has with transforms and fixed-position ancestors, not just faster. One thing worth a look: three tests went from checking an exact scroll position to just checking that scroll was called at all, because JSDOM can't compute real geometry. Understandable given the constraint, but it does mean those tests no longer catch a regression in the offset math itself.

ryparker:
@coderabbitai full review

PR #6411, opened 2025-02-14
USWDS: Combo Box used in Time Picker Component Ignores Hint Text

### LINKED ISSUE #6315: USWDS - Bug: Time picker ignores hint text
### Describe the bug

On initialization, the time picker ignores the hint text attributes.

1. Go to [time picker](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/iframe.html?p=all&id=components-form-inputs-time-picker--time-picker&viewMode=story) component
2. Ensure is initialized properly, _i.e. should be a combo box_
3. Inspect hint text and view ID `appointment-time-hint`
4. Search for aria attributes that have the ID
5. Verify that there are none

### Expected Behavior

As a user, I expect hint text to be linked when the component is initialized. This **should** be a requirement to all dynamically generated components.

### Related code

_No response_

### Screenshots

Image

### System setup

- USWDS version: 3.11.0
- MacOS M1 Laptop
- Sequoia
- Chromium 132

### Additional context

Discovered in prototype work uswds/uswds-sandbox#168.

### Code of Conduct

cathybaptista:
# Summary

On initialization, the combo box used by the time-picker component was ignoring its hint text attributes.

## Breaking change

## Related issue

Closes #6315 

[Changelog Entries Combobox and Time Picker](https://github.com/uswds/uswds-site/pull/3117)

## Preview link

Preview link:
[Time Picker using Combo Box with aria-describedby
](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cb-time-picker-ignores-hint-text/?path=/story/components-form-inputs-time-picker--time-picker-default-value)

## Problem statement

The combo box used in time picker was not linking any aria-describedby attributes for the "appointment-time-hint" field. As a result, the combo box hint text was not being read by screen readers.

## Solution

The aria-describedby text was not being included when the combo box was initialized as part of the time-picker component. It is now initialized and set as an attribute when the input field that is part of the combo box element is created.

Combo boxes use the arria describedby as:

aria-describedby="fruit-hint" (A single value with a single "-".)

## Testing and review

Go to [time picker](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cb-time-picker-ignores-hint-text/?path=/story/components-form-inputs-time-picker--time-picker-default-value) component
Ensure is initialized properly, i.e. should be a combo box
Inspect hint text and view ID appointment-time-hint
Search for aria attributes that have the ID
Verify that the appointment-time-hint text is found.


amyleadem (commented):
Interested to talk through this one because I am curious about the approach.

In the [time picker twig](https://github.com/uswds/uswds/blob/develop/packages/usa-time-picker/src/usa-time-picker.twig#L10), we explicitly define `aria-describedby` on the `<input>` element. This leads me to believe that our expectation has been that users are responsible for manually defining the `aria-describedby` in their HTML. From there, I also assume that it is our responsibility to make sure that attribute is carried over in the dynamically-generated version of the component.

This PR seems to take a different approach. It looks like it is looking for a hint element with a specific name space and then defines `aria-describedby` with it. This approach feels a bit opinionated and fragile to me since the assumed naming structure feels rigid and it also sets the value of `aria-describedby` whether the hint is present or not. 

Curious if it is more appropriate to just look for the user-defined `aria-describedby` then transfer it to the dynamic component. I am coming at this with some assumptions, so let's get on a call and talk through this! 

image

opensource-joe:
#6315 is still open, still needed. Looks like the design question from amyleadem's review got addressed, the final version just adds aria-describedby to the existing attribute transfer list alongside required, aria-label, and aria-labelledby, rather than guessing at a hint element by naming convention. That was exactly what the review asked for. Just needs someone to confirm that resolves it and take another look, it's been sitting since that fix with no follow-up.

ryparker:
@coderabbitai full review

Issue #6415, opened 2025-02-14
USWDS - [Enhancement] Unit Tests: Update build.spec.js to build entire dist package


mahoneycm:
# Description

Our `build.spec.js` unit test is currently only running `compileSass` and testing for the output CSS.

## Expected behavior

Build would test for all expected dist files when running `buildUSWDS`

dct2023:
@mejiaj Thoughts on if we need this for the current sprint? Weight? Complexity?

ryparker:
I'm closing this tracker and keeping the complete-distribution tests in [#6924](https://github.com/uswds/uswds/issues/6924#issuecomment-5730233156).

The requirements there cover expected output files, Sass compilation by a site using USWDS, failure exit codes, and version and source information in the artifact we publish. Those tests still need to be implemented; #6417 didn't complete the work.

Please flag any missing requirement in #6924.

PR #6417, opened 2025-02-18
USWDS - Unit Tests: Fix USWDS build unit test

### LINKED ISSUE #6413: USWDS - Unit Tests: Fix build test
# Description

Our unit test to check build functionality is not functional

## Expected behavior

We should be able to run gulp commands to accurately check gulp build functions.

mahoneycm:
# Summary

Fix the existing `build.spec.js` test which tests USWDS css output. 

## Breaking change

## Related issue

Closes #6413

## Preview link

[Preview link →](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/cm-fix-build-test/)

## Problem statement

The current unit test fails when being ran and does not have a script to easily run it.

## Solution

- Update test to run `compileSass` instead of `sass` which no longer exists
- Update test to properly await gulp compilation before testing
- Create `buildTests` test script in `test.js`
    - This test is being included and exported in our gulpfile
    - I also created a new script in our `package.json` to run this test

## Major changes

- Test can now be run using `npm run test:build`
- Test properly calls for sass compilation and checks output after dist is built
- Mocha timeout increased by 2 seconds to give compilation more time to complete

> [!NOTE]
This build script is currently only testing output `css`. Since this is all it was checking before I wanted to scope the work to fixing the existing functionality.
I’ve created #6415 to consider adding additional tests to check for other dist files.
> 

## Testing and review

1. Checkout this branch locally
2. Delete the `/dist` directory
3. Run `npm run test:build`
4. Confirm `/dist` is rebuilt with css dir:
    - `uswds.css`
    - `uswds.min.css`
    - `uswds.min.css.map`
5. Confirm tests run without failure

### DIFF
--- a/packages/uswds-core/src/js/utils/test/.mocharc.json
+++ b/packages/uswds-core/src/js/utils/test/.mocharc.json
@@ -3,5 +3,5 @@
     "jsdom-global/register"
   ],
   "exit": true,
-  "timeout": 4000
+  "timeout": 6000
 }
--- a/src/test/build.spec.js
+++ b/src/test/build.spec.js
@@ -4,8 +4,8 @@ const path = require("path");
 const pkg = require("../../package.json");
 const { runGulp, distCssPath } = require("../../packages/uswds-core/src/js/utils/test/util");
 
-before(() => {
-  setTimeout(() => runGulp("sass"), 20000);
+before(async () => {
+  await runGulp("compileSass");
 });
 
 describe("build output", () => {
--- a/tasks/test.js
+++ b/tasks/test.js
@@ -22,4 +22,8 @@ module.exports = {
   sassTests() {
     return src("packages/uswds-core/src/test/sass.spec.js").pipe(mocha());
   },
+
+  buildTests() {
+    return src("src/test/build.spec.js").pipe(mocha(mochaConfig));
+  },
 };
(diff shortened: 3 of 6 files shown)


packages/uswds-core/src/js/utils/test/.mocharc.json
```diff
@@ -3,5 +3,5 @@
     "jsdom-global/register"
   ],
   "exit": true,
-  "timeout": 4000
+  "timeout": 6000
```
mahoneycm, line 6: **note:** Timeout increased to give the compilation process more time to complete. Alternatively, we can set this timeout specifically for the `before()` hook in `build.spec.js`

tasks/test.js
```diff
@@ -22,4 +22,8 @@ module.exports = {
   sassTests() {
     return src("packages/uswds-core/src/test/sass.spec.js").pipe(mocha());
   },
+
+  buildTests() {
+    return src("src/test/build.spec.js").pipe(mocha(mochaConfig));
+  },
```
mahoneycm, line 28: **note:** I opted to create a separate test function here so since this test is unrelated to component code. Including it in our other tests would slow them down due to the compilation process.

src/test/build.spec.js
mahoneycm, line 1: **question:** This test currently checks for `uswds.css.` and `uswds.min.css`. Should we add a new test for `uswds.min.css.map` since it's output by default?

ethangardner:
Given the approach in the file currently that runs a gulp build with a longer timeout using mocha, I rescoped this to a new issue here https://github.com/uswds/uswds/issues/6877. We already do a full build elsewhere in CI, so we can get a similar level of coverage without the extra overhead of adding a standalone build test that runs an additional sass build. I'll close this PR and issue #6413 due to this rescoping.

PR #6425, opened 2025-02-26
USWDS - [site-alert]: sync variants with alert

nektro:
# Summary

Adds the "error", "warning", "success" style variants to `site-alert`. These put it in sync with the available `alert` styles.

## Breaking change

## Problem statement

I attempted to use the `.usa-site-alert--error` class in my code and was met with an uncolored gray box with no icon. upon checking https://unpkg.com/browse/@uswds/uswds@3.11.0/dist/css/uswds.css and https://unpkg.com/browse/@uswds/uswds@3.11.0/packages/usa-site-alert/src/styles/_usa-site-alert.scss I realized it was because no such class existed.

## Solution

https://designsystem.digital.gov/components/alert/ and https://designsystem.digital.gov/components/site-alert/ being more in sync will cause less developer confusion and users will already be familiar with the styles.

### DIFF
--- a/packages/usa-site-alert/src/styles/_usa-site-alert.scss
+++ b/packages/usa-site-alert/src/styles/_usa-site-alert.scss
@@ -3,6 +3,9 @@
 
 // Alert variables ---------- //
 $site-alert-icons: (
+  success: "check_circle",
+  warning: "warning",
+  error: "error",
   info: "info",
   emergency: "error",
 );


sknep (approved):
~thanks so much!~

Edit to add:
I almost missed something key -- the `site-` part of your example isn't part of the class name that the Alert component uses. 
I'm going to look into your suggestion further, but I wanted to point this out in the meantime. The BEM-style classes are addictive, so you may try:
```html
<div class="usa-alert usa-alert--error">
...</div>
```

annepetersen (changes requested):
Oh interesting! Hmm. 

I appreciate your intention regarding consistency here, but icons being limited in Site alert was an intended design decision when that component was created. It does help differentiate `Site alert` from `Alert`, for one thing.

From [the component page, Site alert](https://designsystem.digital.gov/components/site-alert/#about-the-site-alert-component) is intended for "critical, time-sensitive warnings or directions across every page so that users see it whenever they visit the site. Using the same site alert treatment across agencies helps create a consistent and predictable way for users to find urgent information across all government websites." Limiting the variants of Site alert also helps with that consistency.

During the work on this (in 2021), we decided against dictating more levels of site alert, which would be indicated by having these additional three icons available — or more other icons, for that matter. Among other reasons, some of the community feedback we received at the time included several very specific warnings against expanding these options further.  

Being constrained and specific about alert level for alerts that appear site-wide also help prevent alert blindness for users (when people start ignoring something that's consistently there all the time).

So, long and short, we very much appreciate this contribution, but it's not one we intend to integrate right now.

Issue #6426, opened 2025-02-26
USWDS - Feature: Dynamic tooltip changes


sarahwylie:
### Is your feature request related to a problem? Please describe.

The tooltip can be created dynamically, but it currently cannot be updated once created. If the state of a button changes, for example, the tooltip text on that button cannot update to reflect the new state change. Instead, multiple tooltips are created and continue to propagate. 

### Describe the solution you'd like

In the `setUpAttributes()` function, wrap a conditional around the creation of the `span` element so additional tooltips are not built on top of the original.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [x] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Feature+Request%22) for duplicate feature requests.

venkatajanapareddy:
I've created a [CodeSandbox](https://codesandbox.io/p/sandbox/dynamic-tooltip-issue-fhqwfl) that reproduces this issue.

I was able to reproduce this problem in Google Chrome, Firefox, and Safari on a MacBook. Additionally, I noticed an extra issue where newly created tooltips persist in the UI even after the mouse hover ends. Please follow the steps to reproduce mentioned in the CodeSandbox link to see this in action. This behavior can lead to screen readers announcing multiple, conflicting tooltips, and it causes the DOM to become unnecessarily cluttered.

PR #6431, opened 2025-03-04
USWDS - Tooltip: Allow for dynamic changes

sarahwylie:
# Summary

Updated tooltip generation so devs can change tooltip text dynamically.

## Breaking change

This is not a breaking change

## Related issue

Resolves #[6426](https://github.com/uswds/uswds/issues/6426)

## Preview link

Preview link:
https://github.com/user-attachments/assets/9d3cc8ab-a7e3-4072-9442-85056d946494

## Problem statement

I applied the USWDS tooltip to a button component whose state changes based on the surrounding elements. When the state of the button changes, the tooltip text no longer conveys the appropriate meaning. When trying to dynamically update the text, a new `<span>` element was created each time the state of the text was updated, which quickly spiraled the element out of control.

## Solution

This code checks whether a tooltip has already been initiated on the target element before implementing the next steps. If the trigger has not yet been applied to the element, then the tooltip will proceed through the full setup steps. If the trigger already exists, then the code will check if the title has been changed, if the position has been changed, and if any additional classes have been added to the element, then makes only the necessary updates. 

## Testing and review

I made this change as a patch for a project, and I wanted to contribute the work back in case it is useful for others. 

## Dependency updates
N/A


packages/usa-tooltip/src/index.js
```diff
@@ -313,50 +313,87 @@ const hideToolTip = (tooltipBody) => {
...
-  const additionalClasses = tooltipTrigger.getAttribute("data-classes");
-  let position = tooltipTrigger.getAttribute("data-position");
-
-  // Apply default position if not set as attribute
-  if (!position) {
-    position = "top";
-    tooltipTrigger.setAttribute("data-position", position);
-  }
+  // If the tooltip has already been instantiated, don't create another instance
+  const tooltipExists = () => {
```
venkatajanapareddy: Thanks for addressing this tooltip issue. I like how you've separated the logic between creating and updating tooltips.

A few suggestions to make the code more maintainable:
1. Consider moving the `tooltipExists` and `newTooltip` functions outside of `setUpAttributes` to improve readability
2. The function name `tooltipExists` is a bit confusing since it actually creates a new tooltip - perhaps `createTooltip` would be more accurate?
3. Similarly, `newTooltip` could be renamed to `updateTooltip` to better reflect its purpose?

packages/usa-tooltip/src/index.js
```diff
@@ -313,50 +313,87 @@ const hideToolTip = (tooltipBody) => {
...
+    return { tooltipBody, position, tooltipContent, wrapper };
+  };
+  const newTooltip = () => {
+    const tooltipContent = tooltipTrigger.getAttribute("title");
+    const additionalClasses = tooltipTrigger.getAttribute("data-classes");
+    let position = tooltipTrigger.getAttribute("data-position");
+    // Identify the existing span element that wraps the trigger
+    const wrapper = tooltipTrigger.parentNode;
+    // Identify the existing span element that contains the tooltip
+    const tooltipBody = tooltipTrigger.nextElementSibling;
```
venkatajanapareddy: Using `nextElementSibling` to find the tooltip body is potentially fragile if the DOM structure changes. Consider using `wrapper.querySelector(`.${TOOLTIP_BODY_CLASS}`)` instead, which would be more resilient to structural changes.

packages/usa-tooltip/src/index.js
```diff
@@ -313,50 +313,87 @@ const hideToolTip = (tooltipBody) => {
...
+      const classesArray = additionalClasses.split(" ");
+      classesArray.forEach((classname) => wrapper.classList.add(classname));
+    }
+    // place the text in the tooltip
+    tooltipBody.textContent = tooltipContent;
+
+    return { tooltipBody, position, tooltipContent, wrapper };
+  };
+
+  if (!tooltipTrigger.classList.contains(TOOLTIP_TRIGGER_CLASS)) {
```
venkatajanapareddy: The current approach checks if the trigger has the TOOLTIP_TRIGGER_CLASS class, which works for the initial implementation but may have edge cases.

When a tooltip is dynamically updated, the key check is:
```javascript
if (!tooltipTrigger.classList.contains(TOOLTIP_TRIGGER_CLASS)) {
  tooltipExists(); // Create new tooltip
} else {
  newTooltip(); // Update existing tooltip 
}
```

This relies on the TOOLTIP_TRIGGER_CLASS already being present to determine if it's an update scenario. A more reliable approach would be to check the actual DOM structure - The proper tooltip structure consists of a parent wrapper (.usa-tooltip) containing both the trigger element and the tooltip body:
```javascript
const isTooltipSetup = tooltipTrigger.parentNode && 
                       tooltipTrigger.parentNode.classList.contains(TOOLTIP_CLASS) &&
                       tooltipTrigger.parentNode.querySelector(`.${TOOLTIP_BODY_CLASS}`);

if (!isTooltipSetup) {
  // Create new tooltip
} else {
  // Update existing tooltip
}
```
This would check if this trigger already properly wrapped in a tooltip structure.

venkatajanapareddy:
Hey @sarahwylie , Thank you once again for working on this fix. I was playing around with the solution using the comments I suggested above by extracting tooltip creation and update logic into separate functions for better maintainability and also created a storybook test.

You can find my implementation in the `fix/dynamic-tooltip-fix` branch of my fork.

If you would like to check it out:
1. Clone my fork: `git clone https://github.com/venkatajanapareddy/uswds.git`
2. Checkout the branch: `git checkout fix/dynamic-tooltip-fix`
3. Run Storybook to see the test, I created a new file under tooltip storybook named test-usa-tooltip-dynamic.twig.

Feel free to incorporate any parts that work well for your implementation. Great work on the fix.

packages/usa-tooltip/src/index.js
sarahwylie: Excellent call - thanks!

packages/usa-tooltip/src/index.js
sarahwylie: Thank you - I didn't have the time to get quite this complex while I was working on this, so I appreciate your thorough review!

packages/usa-tooltip/src/index.js
sarahwylie: My apologies for the delayed response, but you are absolutely right about the function names! It's one of those things that made sense during development for me, but when I step back it's definitely confusing. 

I'm going to push back a little on the readability for moving each of the functions out of `setUpAttributes`, since the other two functions set up the tooltip attributes. I cannot, however, think of a better name for the `setUpAttributes` function that wouldn't be confusing elsewhere, so consider this note my pushback.

sarahwylie:
This is ready for review @venkatajanapareddy

venkatajanapareddy:
> This is ready for review @venkatajanapareddy

@sarahwylie  Great work implementing the feedback! The changes look excellent:

  - Function naming is much clearer (`createTooltip`/`updateTooltip`)
  - Robust DOM structure checking with proper `querySelector` usage
  - Clean separation of create vs update logic
  - Prevents tooltip element proliferation as intended

  The implementation properly addresses the dynamic tooltip requirements while maintaining good code practices.

venkatajanapareddy (commented):
LGTM. 👍

opensource-joe:
#6426 is still open and real. I traced how this hooks into the rest of the module, setup() is already a public API, and the create/update split correctly reuses the existing show/position logic rather than duplicating it, so the design holds up. Good community review here too, though no USWDS team member has weighed in yet. One gap: no test coverage was added for the new update path, that's the actual behavior being fixed so it'd be good to see at least one test pinning it down before this merges.

ryparker:
@coderabbitai full review

ryparker:
The update behavior looks useful, but `tooltip.setup(trigger)` no longer returns its result at `d745d5ff`. That's why this has the breaking-change label. Normal initialization still works because `init` gets the elements separately.

Previously, callers could read `tooltip.setup(trigger).tooltipBody`. Both inner helpers still return the expected object, but the [outer function discards those returns](https://github.com/uswds/uswds/blob/d745d5ffebaf3f861dfa51b0bf40fcf4886133d7/packages/usa-tooltip/src/index.js#L407), so the [exported setup method](https://github.com/uswds/uswds/blob/d745d5ffebaf3f861dfa51b0bf40fcf4886133d7/packages/usa-tooltip/src/index.js#L471) returns `undefined`.

Please return the chosen helper's result from the outer function. Tests should check the returned body, wrapper, position, and content on both initial setup and update, and confirm that repeated setup preserves the existing content when no new title is supplied. That would keep existing calls working alongside the new update behavior.

PR #6450, opened 2025-03-25
USWDS - Date Picker: prevent focus out out-of-order issue in safari

### LINKED ISSUE #6449: USWDS - Bug: Clicking Date Picker calendar button to close calendar doesn't work
### Describe the bug

The Date Picker in Safari doesn't work properly when trying to close the calendar via the calendar button. 

1. Go to https://designsystem.digital.gov/components/date-picker/ in a Mac or other iOS device Safari browser
2. Open the calendar date selector using the button
3. Try to close the button using the same button
4. The calendar will flash closed and reopen
5. Compare this same functionality in Chromium-based browsers or Firefox

### Expected Behavior

Clicking on the calendar button a second time should close the calendar date picker.

### Related code

_No response_

### Screenshots

_No response_

### System setup

USWDS 3.12
Mac
Sequoia 15.3.1
Safari 18.3

### Additional context

It appears that in Safari, the sequence of events being fired is
- Focusout
- Click (on calendar button)
The focusout will close the calendar, and the click will reopen it.

In Chrome or Firefox, the order is reversed
- Click
- Then focusout
This ordering makes it so that the focusout will make the calendar stay closed. 

### Code of Conduct

anmazz:
# Summary

Fixes Safari issues where clicking the calendar button a second time would not close calendar.

## Breaking change

## Related issue

Closes [#6449 ](https://github.com/uswds/uswds/issues/6449)

## Problem statement
The Date Picker in Safari doesn't work properly when trying to close the calendar via the calendar button.

## Solution
Sets a "preventFocusOut" flag during the mousedown event to capture the event prior to focusout or click.
The order is as follows in Safari from my testing:
- User clicks button
- mousedown fired (set preventFocusOut to true)
- focusout fired (since preventFocusout is set to true, ignore the calendar toggle)
- click (reset preventFocusOut)

Adding the mousedown event listener allows for the ability to intercept the click on the button before focusout is fired. 
In Chromium browsers/Firefox, the order of events is click and then focusout, which is why those browsers don't have issues. The click toggles the calendar off, and the focusout will try to hide the calendar again, which is fine since it's already closed.

## Testing and review

The date picker should be tested with all browsers that clicking the calendar button once opens the calendar, and clicking it again closes it. Clicking outside of the element should close the calendar as well. 

It didn't seem as though the calendar button was keyboard accessible, so I'm wondering if there is anything to be done in that regard.


opensource-joe:
#6449 describes a real Safari-specific bug and the mousedown-flag approach makes sense for the event-ordering quirk. But this no longer applies to develop, the focusout handler it patches was removed, looks like as part of #6750 which merged yesterday. Calendar dismissal now goes through a pointerdown listener that skips clicks inside the widget entirely, so the button's own click handler is the only thing that toggles it now. That might mean this exact race is already gone, but I can't test Safari from here to confirm. Worth checking against current develop before rebasing, in case this is already fixed a different way.

ryparker:
@coderabbitai full review

PR #6460, opened 2025-04-02
USWDS: Port web component banner to main USWDS codebase

heymatthenry:
## Summary

This PR adds the USWDS Elements `:pour-one-out:` banner component to the main USWDS codebase.

## Detailed description

With USWDS Elements work on indefinite hiatus, we will add web component variants of existing components to the main USWDS codebase as we develop them. The `usa-banner` component is the first such variant, and so requires some additional configuration and tooling to be added to the project. To that end, in addition to adding the web component banner itself, this PR adds Vite and related dependencies to build that code. For now, Vite runs separately from Gulp/Compile, and only touches the web component code. *Eventually this is likely to change*. Relatedly, the banner code is written in ECMAscript module syntax. This is also a break from existing practice, but this is also a change that is likely to roll out more broadly throughout the codebase over time. Because the addition of Vite and the partial adoption of ESmodules apply only to new code, we do not consider them to be breaking changes.

As for as the build configuration, the initial `usa-banner` web component in `dist/components` is a "batteries-included" version that is bundled with all of its dependencies. It is intended to be self-contained, and support plug-and-play usage. Users who need to customize the component or to optimize their own builds, should use the source code in a way that aligns with their own project-specific needs and tooling. (More detailed documentation and usage information will follow in other PRs on the `uswds-site` repo.)

The component code itself has already been vetted, so for review purposes, there shouldn't be a need to dive too deeply into the component itself or its styles.

**Update 5/19**: Successfully tested this alongside the relevant branch of USWDS Compile (uswds/uswds-compile#152) on the documentation site and verified that they work together.

~Still to do:~

- [x] Fix imports for assets imported into CSS
- [x] Re-implement custom props
- [x] Fix layout inside open state
- [x] Fix eslint config to support classes (this might mean just replacing the config) (done in #6463)
- [x] Support single-file (CDN-compatible) version of component

## Dependency updates

| [Updated dependency example] |     [1.0.0]      |   [1.0.1]   |
| @rollup/plugin-commonjs | -- | 28.0.3 |
 | @spiriit/vite-plugin-svg-spritemap | -- | 4.0.0  |
| eslint-plugin-lit | -- | 2.0.0 |
| lit     |        --        |   3.2.1   |
| vite |  -- | 6.2.2 |
| vite-plugin-svg-sprite | -- |  0.6.2 |
| sass |     1.8.4     |     --      |


heymatthenry:
This isn't building right when installing to the docs site. Setting this back to draft while I fix that. `:hold-the-phone:`

packages/usa-banner/src/styles/_usa-banner.component.css
```diff
@@ -0,0 +1,91 @@
+:host {
+  --icon-lock: url("../img/lock.svg");
```
sknep, line 2: So this would expect external resources to exist in the parent/image directory? It seems a little opinionated, no?

I only ask because I remember you saying something about base64 encoding the images. Is there a way to set an asset path prefix with css custom props?

packages/usa-banner/src/styles/_usa-banner.component.css
```diff
@@ -0,0 +1,91 @@
...
+  --theme-banner-link-color-hover: var(--theme-link-hover-color, #1a4480);
+  /* Missing theme-banner-max-width because we can't pass custom property to media queries. */
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.usa-banner {
+  background-color: var(--theme-banner-background-color);
```
sknep, line 20: this part makes me really excited for a css custom props-themable USWDS future

packages/usa-banner/src/usa-banner.component.js
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  constructor() {
+    super();
+    this.lang = "en";
+    this.isOpen = false;
+    this.tld = "gov";
+
+    this.data = {
+      en: {
```
sknep, line 66: I really appreciate the integrated multilingual support using established language codes. Is this approach documented anywhere? I hope it continues!

packages/usa-banner/src/usa-banner.component.js
```diff
@@ -0,0 +1,250 @@
...
+            "Comparta información sensible sólo en sitios web oficiales y seguros.",
+        },
+      },
+    };
+  }
+
+  // Get English or Spanish strings. Default to English if an unknown `lang` is passed.
+  // Ex: <usa-banner lang="zy"></usa-banner>
+  get _bannerText() {
+    const content = this.data[this.lang] || this.data.en;
```
sknep, line 116: would it be appropriate to fall back to the document-provided `lang` attr?
`document.documentElement.lang` https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/lang

packages/usa-banner/src/usa-banner.component.js
```diff
@@ -0,0 +1,250 @@
...
+    `;
+  }
+
+  static lockIcon() {
+    return html`
+      <span
+        class="usa-banner__icon-lock"
+        role="img"
+        aria-label="Locked padlock icon"
+        part="lock-icon"
```
sknep, line 158: I'm not familiar with the `part` attribute -- does it serve a specific purpose here?

packages/usa-banner/src/usa-banner.component.js
```diff
@@ -0,0 +1,250 @@
...
+        </p>
+      </div>
+    `;
+  }
+
+  static styles = [unsafeCSS(bannerStyles), unsafeCSS(bannerComponentStyles)];
+
+  render() {
+    const classes = { "usa-banner__header--expanded": this.isOpen };
+    // ? Is there a better way to fallback to a default value is passed value doesn't match?
```
sknep:
```suggestion
    // ? Is there a better way to fallback to a default value if passed value doesn't match?
```

sknep: for what it's worth, I think your ternary operator below is appropriate for now.

packages/usa-banner/src/usa-banner.component.js
heymatthenry: `part` exposes a [CSS Shadow Part](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_shadow_parts)—basically a way of poking a hole in the shadow DOM that styles can get through.

sknep:
Is this the place where I can review this? I can't tell if this is the WC version or not:

https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/mh/port-banner/?path=/story/components-banner--default

heymatthenry:
@sknep The web component isn't in storybook yet. Elements has/had its own Storybook, and I haven't ported this stuff over yet. The docs do use the web component where appropriate, so you can see it working in the [WIP docs](https://federalist-ead78f8d-8948-417c-a957-c21ec5617a57.sites.pages.cloud.gov/preview/uswds/uswds-site/mh-add-web-component-banner-docs/components/banner/#banner-web-component-2)

packages/usa-banner/src/styles/_usa-banner.component.css
heymatthenry: This one needs to be set a little differently from the ones we inline. And it just shares the path with the non-web-component banner. But at the end of the day, setting the path in a custom prop makes it part of the public API for the component, so it's easy to change if someone wants to change it.

packages/usa-banner/src/usa-banner.component.js
heymatthenry: This is something we were/are trying out so it's not documented because the underlying implementation could change

packages/usa-banner/src/usa-banner.component.js
heymatthenry: This is a good idea! However, unless/until the design system has guidance about setting that attribute I wouldn't necessarily want to do that here. (To be clear, everyone should always set the `lang` attribute if for no other reason than it's important for performance and accessibility—i just don't necessarily want to bak assumptions about best practices into a component without making adoption of those practices an explicit part of the design system.)

heymatthenry:
@sknep I added a couple other small changes (new props mostly) but otherwise this seems good to me if it seems good to you.

sknep (approved):

PR #6463, opened 2025-04-08
USWDS: Update eslint config to modern syntax

heymatthenry:
# Summary

This PR ports the old YAML-based eslint config to the modern flat config syntax. There are no user-facing aspects of this change.

## Breaking change

## Problem statement

Updating this configuration will enable us to use more modern code and libraries without eslint causing builds to fail. This turned out to be trickier than one might hope, because the airbnb ruleset we use doesn't support the new configuration style.

## Dependency changes

| `eslint-plugin-airbnb-base`    |        --        |   0.0.1-security   |


annepetersen (approved):
LGTM

PR #6475, opened 2025-05-21
fix(skipnav): preserve target focus outlines

### LINKED ISSUE #6474: USWDS - Bug: Visual Focus Indicator Disappears from Target Element After SkipNav Activation
### Describe the bug

When a keyboard user activates the skip link provided by the `usa-skipnav` component (typically labeled “Skip to main content”), the keyboard focus correctly moves to the designated main content area. However, no visible focus indicator (such as a border, outline, or focus ring) appears on that target element. This creates a poor experience for keyboard users, who are left unsure where their focus is located until they press `Tab` again and see focus move.

**Impact on Accessibility:**

* **Violates WCAG 2.4.7 Focus Visible (Level AA):** This Success Criterion requires that "any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible." Actively causing the focus indicator to disappear on the target of a keyboard navigation mechanism fails this.
* **User Disorientation and Reduced Usability:** Lack of a visible focus indicator can disorient users who navigate by keyboard, particularly those with low vision, attention limitations, or certain cognitive disabilities who depend on these visual cues.

1. Go to a webpage that correctly implements the USWDS `usa-skipnav` component.
2. Using only the keyboard, press the `Tab` key. The "Skip to main content" link should appear at the top of the page and receive a visible focus outline.
3. With the "Skip to main content" link focused, press `Enter`.
4. **Observe the main content area of the page**:
    - Keyboard focus will have moved to this area (e.g., the `<main>` element).
    - **Issue**: The main content area will *not* have a visible keyboard focus indicator (like the standard blue USWDS outline or the browser’s default outline).
5. Press `Tab` again:
    - The next focusable element *within* or *after* the main content area will correctly show a visible focus indicator.
    - This confirms the issue is specific to the **immediate target of the skip link**, not general focus behavior.

https://github.com/user-attachments/assets/851c7c58-e98c-47e5-a5a7-0ba01daa323d

### Expected Behavior

When a keyboard user activates the `usa-skipnav` link, and their focus is programmatically moved to the target main content area, the target element should **visually indicate it has received focus**—consistent with [WCAG 2.4.7 Focus Visible (Level AA)](https://www.w3.org/WAI/WCAG21/quickref/#focus-visible).

To see a correct implementation in the wild:
- Go to [GitHub homepage](https://github.com/) **(must be logged in)** and press `Tab` to activate the skip link.

You can see the main content area shows a visible outline or other visual focus indicator when it receives keyboard focus.

### Related code

_No response_

### Screenshots

_No response_

### System setup

- USWDS version: 3.12.0
- Device: Macbook
- Operating system: MacOS
- Browser: Reproduced in both Google Chrome & Firefox

### Additional context

_No response_

### Code of Conduct

venkatajanapareddy:
## Summary

Preserve visible focus on skipnav and return-to-top targets by removing the forced inline `outline: 0`. Focus movement and temporary tabindex cleanup remain unchanged. Authored CSS and inline outlines survive activation and blur.

Closes #6474.

## Breaking change

This is not a breaking change. The fix restores the target's existing focus styling without removing a public API or changing the focus lifecycle.

## Visual evidence

The contributor's original recordings show the focus indicator [before the fix](https://github.com/user-attachments/assets/63405f9d-01bf-4383-bc65-1945c67610e9) and [after the fix](https://github.com/user-attachments/assets/4e9ff6e5-81e7-4949-b548-6974fc77455e). Fresh matched screenshots are also retained in the local validation report.

## Testing and review

Verified at `62cff6322a111a566d277d0c82f612b8151b5bb9`:

- Five skipnav unit tests pass. The two new tests fail against develop's old implementation and pass with this change.
- New regression tests cover both skip links and return-to-top links, including authored outlines, focus, blur, and tabindex cleanup. The follow-up commit changes tests only.
- Six matched Chromium keyboard cases cover default, custom CSS, and authored inline outlines. The base suppresses each to 0px; this PR retains the 4px default outline, 4px custom CSS outline, and 5px dotted inline outline.
- Enter activation, Tab into content, blur cleanup, and return-to-top focus pass. Prettier and whitespace checks pass. The focused test file passes current develop's ESLint configuration; the shared dependency installation cannot load this older branch's `@eslint/eslintrc` dependency.
- Current build, signature, and security checks pass.

Before merge, run VoiceOver with Safari or NVDA with Firefox. Confirm skip-link activation, reading position, visible focus, and navigation after leaving the target. Browser keyboard checks do not establish screen-reader results.

Refresh the branch base before final merge verification. The current `PR title` workflow checks an older trusted base that lacks `.github/scripts/check-pr-title.test.mjs`, so it fails before validating the otherwise conforming title. This PR remains open.


opensource-joe:
#6474 is a real gap, keyboard users get no visible confirmation of where focus landed after using a skip link. The fix matches the bug exactly, one line removed, nothing else touched. I checked the file's history for why that suppression might have been added in the first place and didn't find anything tying it to a specific reason, so removing it doesn't look like it's undoing an intentional fix for something else. Looks ready.

ryparker:
@coderabbitai full review

ryparker:
Added two regression tests in signed commit `62cff6322a111a566d277d0c82f612b8151b5bb9`; production code is unchanged. They check that authored outlines survive activation and blur for skip links and return-to-top links, while focus and temporary `tabindex` cleanup keep working. All five skipnav tests pass. Running them against `develop` produces the expected two outline failures and three existing passes.

In Chromium 147, keyboard checks covered default styling, custom CSS, and inline outlines. `develop` reduced each outline to 0px; this PR keeps the 4px default, 4px custom CSS, and 5px inline dotted outlines. Enter activation, Tab into content, blur cleanup, and return-to-top focus also pass. Production JavaScript matches the contributor version already tested.

Prettier and whitespace checks pass. The shared dependency install is missing this older branch's `@eslint/eslintrc`, so I checked the focused spec with `develop`'s ESLint configuration.

We still need VoiceOver with Safari or NVDA with Firefox checks for activation, reading position, focus visibility, and navigation after leaving the target. The Chromium keyboard checks don't establish screen-reader behavior.

PR #6502, opened 2025-07-11
USWDS - Readme: Update README.md to state 2.x is no longer maintained.

chriscct7:
# Summary

Clarified that version 2.x of USWDS is no longer maintained

## Problem statement

The current site incorrectly states that USWDS is still maintained. 
If it is in fact still maintained, a seperate PR would be required due to the "supported until" of the same sentence being incorrectly in the past.


annepetersen (approved):
Helpful catch. Thanks!

Issue #6216, opened 2024-11-19
USWDS - Sass 3.0.0: `@import` and global built-in functions deprecation


mahoneycm:
## Description

Updating `sass` package to `>=1.80.0` brings in a new sass deprecation warning. These warnings happens many times during compilation.

**Differences from other warnings**

Unlike other deprecations, this deprecation will not take place in Sass `2.0.0` but rather **`3.0.0`**. Currently scheduled ≈ two years out.

<details>
<summary>SASS Announcement</summary>

> Sass @import rules and global built-in function calls now emit deprecation warnings. While Dart Sass 2.0.0 will be released soon with various smaller breaking changes, we don’t expect to remove Sass @import rules or global built-in functions until Dart Sass 3.0.0, which will be released no sooner than two years after Dart Sass 1.80.0.
>
>Eventually, all @import rules will be treated as [plain CSS @imports](https://sass-lang.com/documentation/at-rules/import/#plain-css-imports), likely after an intermediate period where anything that used to be a Sass @import throws an error. [^1]

Source: [Breaking Change: `@import` and global built-in functions](https://sass-lang.com/documentation/breaking-changes/import/)

</details>

[^1]: [Transition period](https://arc.net/l/quote/htzbmdad)

## Compilation warnings

```js
Deprecation Warning: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import
```

## Suggested change

- [ ] Replace `map-has-key` → `map.has-key`
- [ ] Replace `length()` → `list.length`; _we'll need to confirm this is correct in the context of our code_
- [ ] Replace `unquote` with `string.unquote`
- [ ] Replace `map-merge` → `map.merge`; _Flagged 20 times_

Sass is offering an [automatic migration tool](https://sass-lang.com/documentation/breaking-changes/import/#automatic-migration) to ease the transition to the new modular approach.

Additionally, the migration tool allows us to migrate away from global built-in functions separately from migrating `@import` rules if we want to take an incremental approach towards migration.

> [!NOTE]
> Since this deprecation won't take effect until Sass `3.0.0` we could decide to include these changes with the immediate changes needed for `2.0.0` or silence the warning for downstream users in compile and resolve this at a later time.

mejiaj:
I've made the following changes:
- Updated issue to remove duplicate warning
- Updated metadata
  - Reduced weight because this change is happening much later.
  - Reduced complexity because the errors are low LOE

[...]

PR #6519, opened 2025-08-21
USWDS - SASS: Address map deprecation warnings

mejiaj:
# Summary

This work replaces deprecated SASS `map-merge` and `map-has-key`. It avoids issues when SASS updates and gets rid of 23 warnings (469 [`develop`] → 446 [this branch]).

## Breaking change

## Related issue

Part of #6216.

## Solution

Uses new map functions based on [SASS map guidance](https://sass-lang.com/documentation/modules/map/).

## Major changes

Updated map methods and imported `sass:map` with `@use "sass:map";` where it was missing.

## Testing and review

- [ ] There shouldn't be any errors on compile
- [ ] Map warnings should be gone on compile
- [ ] There shouldn't be any visual regressions in: Accordion, Banner, Combo box, Date picker (and range), Footer, Icon list, Navigation, Add Aspect, and Layout (at-media).


jonathanbobel (approved):
LGTM. Ran fine locally.

PR #6592, opened 2026-03-26
USWDS - Keymap: Guard against non-keyboard events from datalist selections

### LINKED ISSUE #6567: USWDS - Bug: Console error when using `list` and `<datalist>` on text inputs
### Describe the bug

Might be an edge case because it doesn't seem to be an issue with a USWDS component, but when we have a text input that uses a datalist, we see a console error

The easiest way to reproduce the error is going to the USWDS and in the Inspector view, updating the code for the text input. 

```html
<input list="ice-cream-flavors" class="usa-input" id="input-type-text" name="input-type-text">
<datalist id="ice-cream-flavors">
  <option value="Chocolate"></option>
  <option value="Coconut"></option>
  <option value="Mint"></option>
  <option value="Strawberry"></option>
  <option value="Vanilla"></option>
</datalist>
```

1. Add all the `<datalist>` HTML
2. Add `list="ice-cream-flavors"` to the `<input>`
3. Click on the input
4. Click on an option to make a selection
5. There should be a console error

### Expected Behavior

There should not be any console errors

### Related code

_No response_

### Screenshots

Image

### System setup

- Browser: Chrome

### Additional context

_No response_

### Code of Conduct

daresTheDevil:
When you add a native `<datalist>` to a `usa-input` and select an option in Chrome, you get a `TypeError` in the console:

```
TypeError: a.getModifierState is not a function
    at keymap.js:34
    at behavior.js:84
```

Chrome dispatches what looks like a keyboard event through the `keydown` delegation chain when you pick from a datalist, but it's not actually a `KeyboardEvent`. No `event.key`, no `getModifierState`. The `keymap` utility assumes both exist, so it blows up.

Fix is a one-line early return in `keymap.js`. If `event.key` is undefined, bail. Non-keyboard events skip the key matching instead of throwing. No change in behavior for actual keyboard events.

Verified the fix manually. Built the patched bundle, loaded a test page with `usa-input` + `<datalist>`, selected an option, no error. Same page throws with the unpatched bundle. Added a unit test covering this case. 743/743 passing.

Credit to @jaredcunha for finding this and providing the repro steps.

Closes #6567


ethangardner (changes requested):
Thanks for the contribution. There was a similar change from #6567 that also addressed the `datalist` scenario in the library code. If you can accept the version from the `develop` branch with the change I approved and merged, that should resolve the conflict in `keymap.js`. 

I think the test you wrote for it is great, and I'd like to get that part merged once the conflict is resolved.

ethangardner (approved):

annepetersen (approved):

PR #6593, opened 2026-03-27
USWDS - Date Picker: Add aria-current to today's date button

### LINKED ISSUE #6546: USWDS - Bug: Date Picker Widget does not use Aria-Current to indicate current date
### Describe the bug

The current date or month is not programmatically indicated, making it difficult for users relying on assistive technologies to identify the current date or month within a calendar or date picker.
The current date is currently only visually indicated with the usa-date-picker__calendar__date--today class. Users that use a screen reader will not be informed to identify the current date or month, leading to confusion and inefficiency in navigating date-related content.

1. Navigate to https://designsystem.digital.gov/components/date-picker/
2. Select the date-widget icon and open up the calendar widget
3. Right click the highlighted date and select inspect from dropdown
4. The  inspect menu should appear with the html focused on the highlighted element. 
5. The presence of aria-current is missing despite a class of the element of the element being usa-date-picker__calendar__date--today

### Expected Behavior

aria-current is used to convey that the initial focused date is the current date
Example: <td aria-current="date">15</td> 

### Related code

_No response_

### Screenshots

Image

### System setup

- USWDS 3.12
- Dell
- Windows 11
- Edge Version 142.0.3595.94

### Additional context

_No response_

### Code of Conduct

daresTheDevil:
# Summary

**Improved screen reader identification of the current date.** Today's date button in the calendar widget now includes `aria-current="date"` so assistive technologies can programmatically identify it.

## Breaking change

## Related issue

Closes #6546

## Related pull requests

None.

## Problem statement

The date picker calendar visually highlights today's date with the `usa-date-picker__calendar__date--today` CSS class, but there's no programmatic indicator for screen readers. Users relying on assistive technologies can't tell which date is today when navigating the calendar. This is a gap in WCAG 1.3.1 (Info and Relationships).

## Solution

Added `aria-current="date"` to the today button during calendar rendering. The attribute is only set on the single button matching today's date. No other dates are affected.

## Major changes

- Extract `isToday` boolean from existing `isSameDay(dateToRender, todaysDate)` check
- Set `aria-current="date"` on the button element when `isToday` is true

## Testing and review

- All 273 date picker tests pass (4 new)
- New test confirms today's button has `aria-current="date"`
- New test confirms non-today buttons do not have `aria-current`
- Run `npm test` to verify


jerseycheese (commented):
Verified the bug independently on the live design system docs (https://designsystem.digital.gov/components/date-picker/): today's button carries the `--today` class but no `aria-current` anywhere in the rendered calendar. The patch closes that gap exactly. LGTM.

## Good Calls
- Reusing the existing `isSameDay(dateToRender, todaysDate)` check by extracting `isToday` keeps the change to a few lines and avoids a second date comparison.
- `aria-current="date"` is the right token value per ARIA 1.2 (vs. `"true"` or omitting the value).
- Tests cover both directions (positive on today, negative on the rest), which is the right shape for an attribute that should appear exactly once.

ethangardner (approved):
LGTM - thanks for the contribution!

PR #6594, opened 2026-03-27
fix: prevent console error when using datalist with text inputs

### LINKED ISSUE #6567: USWDS - Bug: Console error when using `list` and `<datalist>` on text inputs

vijaygovindaraja:
## Summary
Fixes console error when native HTML `<datalist>` elements are used with USWDS text inputs.

### 1. `validate-input.js` — null check ordering bug
The `validatorCheckbox` element was used (line 30) before being checked for null (line 32):

```javascript
// Before — crashes if validatorCheckbox is null
const checked = validatorPattern.test(el.value);
validatorCheckbox.classList.toggle(CHECKED_CLASS, checked); // TypeError!
if (!validatorCheckbox) { throw ... }

// After — check existence first
if (!validatorCheckbox) { throw ... }
const checked = validatorPattern.test(el.value);
validatorCheckbox.classList.toggle(CHECKED_CLASS, checked);
```

### 2. `behavior.js` — guard against missing `closest()` in event delegation
Added a check that `event.target` supports the `closest()` method before calling it. Some elements (datalist options, certain SVG elements) may not support `closest()` in all browsers, causing a TypeError during event bubbling.

```javascript
// Before
const target = event.target && event.target.closest(selector);

// After
const target =
  event.target &&
  typeof event.target.closest === "function" &&
  event.target.closest(selector);
```

## Test plan
- [ ] Add a `<datalist>` to a `usa-input` text field
- [ ] Click the input and select a datalist option
- [ ] Verify no console errors appear
- [ ] Verify validation components still work correctly (with `data-validation-element`)
- [ ] Verify character count components still work correctly
- [ ] Run existing test suite

Fixes #6567


vijaygovindaraja:
updated based on @daresTheDevil's analysis in #6567.

the root cause was in keymap.js — when you pick a datalist option, Chrome fires an InputEvent, not a KeyboardEvent. the keymap handler was calling getModifierState() and accessing event.key on it, which blows up because InputEvent doesn't have those.

changes in this update:
- **added**: keymap.js guard — early return when the event isn't a KeyboardEvent (checks for event.key and getModifierState before proceeding)
- **removed**: behavior.js closest() guard — daresTheDevil confirmed event.target is always the <input>, not a datalist <option>, so closest() is always available. that guard was unnecessary.
- **kept**: validate-input.js null check fix — separate bug but worth fixing (validatorCheckbox used before null check)

@daresTheDevil appreciate the thorough testing and comparison — that event logging approach nailed it. the InputEvent vs KeyboardEvent distinction was the key insight.

packages/uswds-core/src/js/utils/keymap.js
```diff
@@ -17,7 +17,16 @@
  * @param {KeymapConfig} map
  * @return {KeyboardEventHandler}
  */
-module.exports = (map) => (event) =>
+module.exports = (map) => (event) => {
+  // Bail out if this is not a KeyboardEvent (e.g. InputEvent from datalist
+  // selection). Only KeyboardEvents have `key` and `getModifierState`.
+  if (
```
ethangardner, line 23: There's one small addition I think would be good here. Would you mind adding an additional check for `!(event instanceof KeyboardEvent)` to the other two conditions so we're left with 

```js
!(event instanceof KeyboardEvent) ||
typeof event.key === "undefined" ||
typeof event.getModifierState !== "function"
```

ethangardner (changes requested):
Thank you so much for this contribution! I appreciated your notes about the diagnosis and how to reproduce the issue, as well as the comments you added to the file. I have one small change request, but this looks good otherwise.

ethangardner:
I was able to confirm the behavior and validate the fix. There is one small change in review.

vijaygovindaraja:
@ethangardner done, added the instanceof check.

ethangardner (approved):
Thank you!

PR #6595, opened 2026-03-29
USWDS - Character Count: Defer aria-live to prevent iOS VoiceOver announcement on page load

### LINKED ISSUE #6541: USWDS - Bug: Character Count announced on component render on iOS VoiceOver
### Describe the bug

VoiceOver announces the [character count component](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-form-inputs-character-count--character-count) text content without user interaction when the component is initialized on render.

1. On an iOS device, go to https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-form-inputs-character-count--character-count
2. Turn on VoiceOver and reload the page so the component is initialized
3. Note that the character count message for both the input and textarea are read on render.

### Expected Behavior

I am expecting the screen reader to only read the character count message when the user inputs text into the `<input>` element.

### Related code

This bug appears to be the result of the logic in the [`createStatusMessages` function](https://github.com/uswds/uswds/blob/9999d08fe6c8e7f0aaa3f24b86ba37b281c25b54/packages/usa-character-count/src/index.js#L74) in the component's `index.js` file and how browsers treat the dynamic creation of elements with `aria-live="polite"`. I spun up the repo locally and was able to resolve the bug by wrapping the setting of that attribute in `setTimeout()` with a 100ms delay.

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

daresTheDevil:
# Summary

Deferred the `aria-live` attribute on character count status messages so iOS VoiceOver doesn't announce them on page load.

## Breaking change

## Related issue

Closes #6541

## Related pull requests

None.

## Problem statement

When a page with character count inputs loads, iOS VoiceOver immediately announces the status messages ("25 characters allowed", "50 characters allowed") before the user has done anything. Browsers treat the initial content of a newly-inserted `aria-live` region as a live update, so VoiceOver reads it out loud on mount.

This affects every iOS VoiceOver user who lands on a page with character count components. It's confusing and noisy -- you hear counts for inputs you haven't touched yet.

## Solution

Defer setting `aria-live="polite"` by 100ms after the status message element is inserted into the DOM. The element gets appended first, then the live region attribute is added after the delay. By the time `aria-live` is present, the browser no longer treats the existing content as a new update.

This is the same approach @mfairchild365 suggested in the issue. Simple, no side effects. Live updates still fire normally during typing since `aria-live` is in place well before any user interaction.

## Major changes

- `packages/usa-character-count/src/index.js`: Wrapped `setAttribute("aria-live", "polite")` in a `setTimeout` with 100ms delay inside `createStatusMessages`.

## Testing and review

- Verified on iPhone 16 Pro Max, iOS 26.4, Safari with VoiceOver. No announcement on page load. Live updates still work during typing.
- Added two tests: one confirming `aria-live` is not set synchronously, one confirming it's present after the delay.
- All existing character count tests pass (44/44).


ethangardner (commented):
@chandracarney - Code-wise this looks fine to me, but I'd like to defer to you on the screen reader behavior.

chandracarney:
Thanks for this contribution @daresTheDevil. We've been sorting out a lot of logistics, including testing. I'm going to tag in @jonathanbobel as a reviewer, then hopefully can get this merged soon!

jonathanbobel (approved):
This is an acceptable solution to this issue. 🚀

jerseycheese:
Confirmed on iPhone 16 Pro, iOS 26.5 (Safari + VoiceOver): on page load VoiceOver announces the status message ("You can enter up to 25 characters") on its own, before any interaction, so the bug reproduces.

Worth flagging for anyone else confirming: it's load-timing-sensitive. It reproduces on a real page that initializes on `DOMContentLoaded` (and on the live docs site), but not in the Storybook story, where init runs inside a requestAnimationFrame and the live region lands at a different point relative to VoiceOver's page snapshot. So confirm against a real page load, not the story.

This PR resolves it on my device.. silent on load, and still announces the count while typing.

chandracarney:
Thanks for the review and confirmation, @jonathanbobel and @jerseycheese! 🚀

PR #1207, opened 2016-05-27
Screener/CircleCI integration

el-mapache:
## Screener Integration

This PR affects the test builds of the repository. During a build, it connects the CI server to Screener's (https://screener.io) cloud server and checks for visual regressions in the page's layout. These tests are defined via Screener's interface and must be manually configured. Some basic checks have been put in place for purposes of testing these changes, they can be found in the `uswds` project under the name `styleguide-ci`.

## Description

* Amends circle.yml file to install screener's tunnel and shell scripts

## Additional information

The changes made to the circle.yml file are pretty basic, there are probably optimizations that can be made. Currently, running these regression tests appears to add about 5 minutes to the build; this represents a significant increase in running time.

Issue #805, opened 2015-10-07
Duplicative selectors in _buttons.scss?


rhewitt22:
It appears that you've duplicated three attribute selectors in the [buttons partial](https://github.com/18F/web-design-standards/blob/890640d73f598ccdc9bade37b073ff0130f3c230/assets/_scss/elements/_buttons.scss#L12-L17).

```css
.usa-button,
.usa-button-primary,
.usa-button:visited,
.usa-button-primary:visited,
button,
[type="button"],  <=
[type="submit"],  <=
[type="reset"],   <=
[type="button"],  <=
[type="submit"],  <=
[type="reset"],   <=
[type="image"] {
  ...
}
```

msecret:
Good catch!

PR #6597, opened 2026-03-31
USWDS - Build: Replace postcss, autoprefixer with lightningcss

IHIutch:
##   Breaking change

This is not intended to be a breaking change, though may require additional manual testing to confirm.

## Problem Statement

The current Sass compilation pipeline uses multiple gulp plugins (`gulp-sass`, `gulp-postcss`, `autoprefixer`, `postcss-discard-comments`, `postcss-csso`), causing redundant streams over the full CSS output.

Using `lightningcss` handles this in a single pass is more performant and delivers almost the exact output as before.

## Solution

**1. Replace PostCSS with LightningCSS** — The entire PostCSS streaming pipeline (`gulp-postcss`,`autoprefixer`, `postcss-discard-comments`, `postcss-csso`, `gulp-sourcemaps`, `gulp-replace`, and `gulp-rename`) is replaced with direct calls to `sass-embedded` and `lightningcss.transform()`. LightningCSS handles autoprefixing and minification in a single native pass. The `compileSass()` function is rewritten as a simple `async` function that compiles Sass, transforms the output, and writes files directly. 

This improves the compilation speed by ~0.7s.

## Major changes

- `tasks/sass.js`: Rewrote `compileSass()` from a Gulp stream pipeline to an async function using `sass-embedded` + `lightningcss` directly
- `.browserslistrc`: Deleted, config moved to `package.json`
- Added `lightningcss` and `browserslist` as dev dependencies

### CSS output differences

The minified output is functionally equivalent (~35 bytes difference). The expanded output differs primarily in formatting and vendor prefix selection.

#### Size & line count

| | Baseline (PostCSS) | New (LightningCSS) | Delta |
|---|---|---|---|
| **Expanded size** | 651,972 bytes | 631,641 bytes | **-20,331 bytes (-3.1%)** |
| **Expanded lines** | 30,077 | 30,968 | +891 lines |
| **Minified size** | 521,664 bytes | 521,699 bytes | **+35 bytes (~identical)** |

#### Vendor prefixes dropped

LightningCSS considers these unnecessary for the current browserslist targets (`> 2%, last 2 versions, not dead`):

| Prefix | Baseline occurrences | New occurrences |
|---|---|---|
| `-moz-column-*` (count, fill, gap, break-inside) | 7 | **0** |
| `-moz-user-select` | 4 | **0** |
| `-moz-appearance` | 8 | **1** |
| `-o-object-fit` | 9 | **0** |

## Testing and review

Benchmarked with `hyperfine --warmup 1 --runs 5 'npx gulp compileSass'` on macOS ARM64:

| Step | Mean | vs Baseline |
|------|------|-------------|
| Baseline (develop) | 4.523s ± 0.127s | — |
| + LightningCSS | 3.802s ± 0.088s | **-16%** |

To verify correctness:
1. Run `npx gulp compileSass` and confirm `dist/css/uswds.css`, `dist/css/uswds.min.css`, and `dist/css/uswds.min.css.map` are generated
2. Diff the expanded CSS output against a baseline build to check for functional regressions (minor whitespace or comment differences are expected from the tooling change)
4. Run `npx gulp buildSass` to confirm the lint + compile pipeline still works end-to-end
5. Spot-check that vendor-prefixed properties are still present in the output
6. Verify sourcemaps resolve correctly in browser devtools

## Dependency changes

| Dependency name | Previous version | New version |
| --- | :-: | :-: |
| lightningcss | -- | ^1.31.1 |
| browserslist | -- | ^4.28.1 |


ethangardner:
Thank you for suggesting this change. I did an analysis of the changes to the built file, and moving this direction with the tooling might make sense to do in the future. However, as it stands today, adopting Lightning CSS introduces some significant changes to how CSS properties are ordered in the built output. Shipping this safely would require a thorough audit and a major version bump to USWDS, which is more than our team has the bandwidth to take on at the moment.

To give some context on the complexity involved, here are a few known issues related to CSS reordering in Lightning CSS:

- [Add an option to prevent re-ordering of rules - Issue #231](https://github.com/parcel-bundler/lightningcss/issues/231) 
- [Lightning CSS will shuffle CSS properties order - Issue #805](https://github.com/parcel-bundler/lightningcss/issues/805) 
- [Property reordering during minification causes certain settings to be ignored - Issue #1207](https://github.com/parcel-bundler/lightningcss/issues/1207)
- [Properties order not maintained - Issue #572](https://github.com/parcel-bundler/lightningcss/issues/572)
 
Once these upstream concerns are more fully resolved and we have capacity for a major release, we'd be happy to revisit this.

IHIutch:
Hey @ethangardner, thanks for the reply. 

I should've done more homework here, so I apologize for wasting your time. Let me put a more thorough write-up together to document what's actually changing. Hopefully, that will make future contributions easier and maybe function as something of a roadmap. 

Appreciate your time

annepetersen:
Appreciate all that, @IHIutch! I'll close this one until you have that documentation in hand, then.

IHIutch:
I recognize that its going to be impossible to be a 100% confident with this change and it makes sense that it would be introduced in a major. Though, I might propose an alternative approach: Many new projects these days are using Vite (React, Vue, Astro, etc), which uses `sass-embedded` and `lightningcss` under the hood (we're do this on a CMS site, [Blue Button](https://github.com/CMSgov/bluebutton-site-static/blob/master/astro.config.mjs#L40-L49)). To maintain parity with these modern implementations and start to gain confidence with this approach, it would be great to see this adopted as an experimental feature. Something like, `exports.compileSass__expirementalLigntingcss = compileWithLightningcss;`

This would let greenfield projects opt into this approach and highlight any issues. This way, if/when LightningCSS is adopted, there's already confidence that it works as expected.

My 2¢.

---

For documentation purposes, I put together some pretty comprehensive checks and analysis. I'm going to summarize the changes here; it will be a little lengthy, but I wanted it to be comprehensive since the output isn't a part of the repo. Buckle up!

## CSS output review: `develop` vs this branch

There are no obvious regressions in the output except the monospace font issue, which is documented below.

**Shipped file size**: 522,211 bytes → 522,235 bytes. No components dropped, no components added, no selectors reordered in a cascade-breaking way.

Two primary things to address:

1. `monospace, monospace` gets normalized, undoing the [`monospace, monospace` hack](https://github.com/uswds/uswds/blob/4b71e0c0a10180f5954105bff4e803b1de219b0e/packages/uswds-elements/lib/_normalize.scss#L55) from Normalize.css. 
2. **Safari 16.4 browser support** (see "Risks" below).

### What changed in the output

| | develop (csso) | this PR (lightningcss) |
|---|---:|---:|
| `uswds.min.css` | 522,211 B | 522,235 B |
| `@media` rules | 339 | 339 |
| `@font-face` rules | 26 | 26 |
| components present | 93 | 93 |
| components byte-identical | — | 7 |
| components with non-trivial diff | — | 86 |

The 86 components that "differ" do so almost entirely from superficial changes. Here's every added/removed line classified across all 86 component diffs, sorted by count:

| Lines | Category |
|---:|---|
| 8,830 | Declaration reordering, selector rewrap, formatting |
|   482 | Media query: `(min-width: N)` → range `(width >= N)` |
|   461 | Pseudo-element: `::after` → `:after` (and `::before`) |
|   430 | Color/value normalization (hex, `rgba()` → `#RRGGBBAA`, named → hex) |
|   177 | Vendor prefix drops and `-webkit-mask` value normalization |
|   121 | `url(foo)` → `url("foo")` quoting |
|    76 | `center center` → `50%` |
|    36 | `@font-face` multi-word `font-family` unquoting |
|    24 | `0px`/`0rem` → `0` |

The 8,830 "reordering" dwarfs everything else. The remaining ~1,800 lines are the semantic changes noted above, and all of them are discussed in the risks section below. No component's computed values change.

Reproduce with `node vrt/static-diff/scripts/classify-diff.mjs vrt/static-diff/artifacts/min-diffs`.

### Risks

**1. Media Queries Level 4 range syntax:** lightningcss rewrites 241 of the 339 `@media` rules from `@media (min-width: 64em)` to `@media (width >= 64em)`. csso doesn't do this, so `develop` currently ships the legacy form.

Browser support for the range syntax:

| Engine | Supported since |
|---|---|
| Chrome/Edge | 104 (Aug 2022) |
| Firefox | 102 (Jun 2022) |
| Safari / iOS Safari | **16.4 (Mar 2023)** |

Below those versions, the `@media` block is discarded and the responsive rules inside it silently fail. The current `browserslist` in `package.json` (`> 2%, last 2 versions, not dead`) already excludes anything older, so the rewrite is follows. There's a couple ways to address this, if necessary:

1. Update `browserslist` to name an older floor (e.g. `ios_saf >= 14`).
2. Pass `exclude: Features.MediaQueries` to `lightningcss.transform` in `tasks/sass.js`.
3. Accept the rewrite and document Safari 16.4+ as the minimum.

**2. Vendor prefixes dropped:** lightningcss regenerates prefixes from `browserslist` rather than using autoprefixer's table. It removes 22 lines of `-moz-appearance`, `-moz-column-*`, `-moz-user-select`, `-webkit-appearance: listbox/menulist/none`, and `-o-object-fit`. Each targets Firefox < 60 or pre-Blink WebKit — well below `browserslist`. No currently-needed prefix was dropped. Low risk.

**3. Monospace font shrinkage:** Every `<code>`, `<pre>`, `<kbd>`, `<samp>` element renders at **13px / 14.95px line-height** on this branch vs **16px / 18.4px line-height** on develop. Spot-checked with Playwright chromium measuring the first `<code>` element on `components-add-aspect--add-aspect`:

```
develop:      fontSize=16px,  lineHeight=18.4px,  width=115.22px
lightningcss: fontSize=13px,  lineHeight=14.95px, width=93.63px
```

lightningcss collapses `font-family: monospace, monospace` → `font-family: monospace` as a perceived redundancy. Options to address:

1. **Replace the hack:** Replace with an explicit font stack (e.g. `font-family: ui-monospace, SFMono-Regular, Consolas, "Liberation Mono", monospace`). This is the cleanest fix and is forward-compatible with any future optimizer.
2. **Preserve the duplication:** Use `font-family: "monospace", monospace` (one quoted, one bare), so lightningcss preserves the duplication.

**4. Banner comment:** The `/*! uswds v3.13.0 */` important-comment marker becomes a regular `/* ... */`, and `@charset "UTF-8"` is dropped. If any downstream tool relies on `/*!` surviving a second minification pass or on the charset line being present, this would break it. Low risk but worth noting for anyone re-minifying the file.

### Rendered-output verification

Beyond the static CSS diff, I ran a computed-style comparison across each of the Storybook components (255 stories × 3 viewports = 765 harness files) using Playwright chromium. For each story, extracted the rendered DOM from Storybook, re-rendered it against each variant's `uswds.min.css`, and diffed `getComputedStyle()` for every element plus `::before`/`::after`.

| Result | Count |
|---:|---|
| Stories with byte-identical computed styles | **197 / 255** |
| Stories with only classified-superficial mismatches | **58 / 255** |
| Stories with real regressions | **1 pattern** (monospace, ripples through ~9 stories) |
| Stories with unclassified / unexplained mismatches | **0** |

Classified mismatch categories:

| Count | Category | Status |
|---:|---|---|
| 279 + 4 ripple | `monospace-hack-collapse` | **Item 3 above** |
| 600 | `bg-position-formatting` | Superficial — `getComputedStyle` serializes `0px 0px` vs `0% 0%` identically to the GPU but differently in its output string |
| 197 | `sub-pixel-layout-ripple` | Superficial — dimension differences < 0.5px from lightningcss truncating fractional percentages (e.g. `33.3333333333%` → `33.3333%`). |

Full report and per-story reproductions under `vrt/computed-style/report/` (see `_summary.md`). Re-runnable by any reviewer via the scripts in `vrt/computed-style/scripts/`.

### Cascade-order check

**What this would look like if it failed:** Two rules with equal specificity that set the same property are resolved in source order — the *later* rule wins. Suppose csso emits:

```css
/* …earlier in file… */
.usa-button:hover { color: red; }
/* …later in file… */
.usa-button:hover { color: blue; }   /* wins → hover is blue */
```

and lightningcss emits the same two rules in reverse:

```css
.usa-button:hover { color: blue; }
.usa-button:hover { color: red; }    /* wins → hover is red */
```

Both files contain the same rules, both pass any per-component diff check, but the rendered color flips. That's a cascade inversion, and it's the failure mode the per-component split can't see.

Because the per-component split loses global rule ordering, I also ran a separate cascade check (`extract-sequence.mjs`): extract the top-level rule-head sequence from each min file and verify that, for selectors present in both, the first-occurrence order is preserved. Ran it two ways:

- **Strict** (sole-selector rules appearing exactly once in both files, 4,150 selectors) → **2 violations:**
- **Broad** (any selector, first-occurrence index, 4,345 selectors) → **11 violations:**

**Zero of the 11 are real cascade inversions:** They break down as:

- **1** `@keyframes` hoisting - csso moves `@keyframes` blocks to the top of the file, lightningcss leaves them in place. Keyframe internal selectors (`0%`, `to`) only match their parent animation and don't participate in the document cascade.
- **3** cross-element pairs (`.usa-textarea` ↔ `.usa-range`, `.usa-range` ↔ `.usa-input-group`, `.usa-input-group` ↔ `img`) — different elements, cannot co-match on any DOM node, so source order is irrelevant.
- **1** descendant-selector pair (`.usa-date-picker__button` vs `.usa-date-picker--active .usa-date-picker__button`) — the descendant variant has strictly higher specificity, source order doesn't decide.
- **1** disjoint-property pair on `.usa-in-page-nav__list` — the bare rule sets `color`, the `:focus` rule sets `outline-offset`. Both apply on focused elements.
- **5** button state-variant pairs of the form `.usa-X.usa-button--active` vs `.usa-X:hover` on `usa-accordion__button`, `usa-banner__button`, `usa-menu-btn`, `usa-nav__primary button`, and `usa-nav__close`. All five are symptoms of the rule-count difference described in the next section. lightningcss ships an extra, dead `:hover { color: A }` rule at an early position that is later overridden by `:hover { color: B }`. The "early position" triggers a swap flag against `.usa-button--active`, but the dead rule never participates in rendering.

### Why lightningcss has more rules than csso (6,362 vs 6,128)

csso does a dead-rule elimination pass that lightningcss doesn't. When two rules have equal specificity and set the same property, csso deletes the earlier (unreachable) one. USWDS SCSS produces this pattern often because multiple mixins layer state variants (`:hover`, `:active`, `:visited`) on the same selector chain with different values — e.g. one mixin emits `:hover { color: A }`, a downstream override emits `:hover { color: B }`, and the first rule becomes unreachable.

- csso's min keeps only the second rule.
- lightningcss's min keeps both; the first is dead code the browser never computes.

Confirmed occurrences of this pattern: at least 6 components (`usa-accordion__button`, `usa-banner__button`, `usa-menu-btn`, `usa-nav__primary button`, `usa-nav__close`, `usa-in-page-nav__list`), plus several more in other state-variant-heavy areas. This is the primary source of the 234-rule gap between the two min files.

**Rendered behavior is identical in both builds:** A browser never computes the dead values. The practical impact is that lightningcss ships a small amount of unused CSS that csso prunes, which is fully hidden inside the +24 B net-size delta because other lightningcss optimizations (shorthand collapse, `rgba()` → `#RRGGBBAA`, dropped legacy prefixes) offset the bloat.

Not a regression or a manual-QA target, just flagged so anyone who reviews the diff of the two files doesn't get confused by the extra rules.

### Where to focus manual QA

The automated comparison can't see visual regressions from color or layout rounding, and it can't see JS-driven states that aren't in the static CSS. Suggested areas to double-check in a browser: (some of these are already validating in the visual regression tests, but worth a sanity check as well)

1. **Responsive breakpoints on Safari 16.4+:** The Media Query Level 4 rewrite is the highest-impact change. Check at minimum:
   - `usa-banner` collapse/expand at mobile and tablet widths
   - `usa-header` / `usa-nav` desktop/mobile transition
   - `usa-footer` column reflow
   - `usa-prose > table` responsive behavior

2. **Components with the largest diffs:** Already spot-checked at the source level, but worth clicking through in Storybook:
   - `usa-prose`, `usa-icon-list`, `usa-date-picker`, `usa-range`, `usa-table`, `usa-banner`, `usa-button`, `usa-nav`, `usa-footer`

3. **Forms with form controls that used `-webkit-appearance`/`-moz-appearance`:** These prefixes were dropped. Even though they target old browsers, confirm the modern `appearance: none` reset still produces expected styling on:
   - `usa-select` / `<select>` in current Firefox, Safari, Chrome
   - `usa-radio`, `usa-checkbox` custom controls
   - `usa-file-input`

4. **Icon-masked elements:** The `mask-position: center center` → `50%` normalization touches `-webkit-mask-position` on icon pseudo-elements. Confirm icons still render centered in:
   - `usa-accordion__button` expand/collapse icon
   - `usa-alert` status icons
   - `usa-link--external` launch icon

5. **Print preview of `usa-checkbox` checked state and `usa-radio` selected state:** These are the only two `@media print` blocks. Both render identical declarations between builds; verification is just confirming the printable fallback glyph still appears in the browser print preview.

---

### How to verify these changes yourself

Everything referenced in this comment is committed to a companion branch, [`use-lightningcss-verification/vrt`](https://github.com/IHIutch/uswds/tree/use-lightningcss-verification/vrt). There are 3 directories, each self-contained with its own README:

- [`vrt/static-diff/`](https://github.com/IHIutch/uswds/tree/use-lightningcss-verification/vrt/static-diff/): text-level diff of the compiled `uswds.min.css` from each branch, with every changed line classified by category. Source of the byte counts, the category-line-counts table, and the cascade-order check.
- [`vrt/computed-style/`](https://github.com/IHIutch/uswds/tree/use-lightningcss-verification/vrt/computed-style/): renders every Storybook story in chromium against each variant's CSS and diffs `getComputedStyle()` per element. Source of the 197/255 byte-identical figure and the monospace regression evidence.
- [`vrt/pixel-regression/`](https://github.com/IHIutch/uswds/tree/use-lightningcss-verification/vrt/pixel-regression/): Vitest screenshot suite. Renders each story against develop's CSS, then against this branch's CSS, and fails on pixel diffs above the antialiasing threshold. Run with `npm run test:visual`.

Each subdirectory's README has the reproduction commands.

IHIutch:
Just wanted to come back to the concern of these issues: 

> - https://github.com/parcel-bundler/lightningcss/issues/231
> - https://github.com/parcel-bundler/lightningcss/issues/805
> - https://github.com/parcel-bundler/lightningcss/issues/1207
> - https://github.com/parcel-bundler/lightningcss/issues/572

None of these issues are related to changes made in this PR; there are no changes that introduce longhand properties before shorthand properties. We can be extra confident with this based on the outputs found in [`/computed-style`](https://github.com/IHIutch/uswds/tree/use-lightningcss-verification/vrt/computed-style/).

Issue #6569, opened 2026-03-05
USWDS - [Meta]: Clarify / publish open source model / external contribution methods


annepetersen:
**Tasks required / done when**
___

- [ ] Draft governance guidance 
- [ ] Create contributor ladder issue templates
- [ ] Identify and label further "Good First Issues" in repo 
- [ ] Identify where content will go and update in those locations

PR #6609, opened 2026-04-14
Minor / META - COMMUNITY.md edits

annepetersen:
# Summary

- added former roles / titles for @thisisdano and @heymatthenry 
- changed section headers to simplify
- bullet pointed the participation topline message
- moved the full contributor list placeholder to the end, since that's where that automation should end up

## Breaking change

## Related issue

Supports #6569 

## Solution

So the community knows more about other active participants and how to effectively participate


chandracarney (approved):
Great to get more folks added to the ladder! 👍

PR #6611, opened 2026-04-16
USWDS - Workflow: Add ci-docs.yml to identify broken links in markdown documentation files

natalialuzuriaga:
# Summary

This PR adds a new workflow file to the repository: `ci-docs.yml`. This workflow uses `github-action-markdown-link-check@1.0.15` to identify broken links in the repository's markdown documentation files every Monday at 00:00 UTC.

## Breaking change

**This is not a breaking change.**

However, the presence of broken links will cause the action workflow to fail, thus will cause a failed (but non-breaking) check in the CI pipeline.

## Related issue
N/A

## Related pull requests
N/A

## Preview link
I ran this action on the DSACMS fork and here are the results:
Preview link: https://github.com/DSACMS/uswds/actions/runs/22197348828/job/64201014836

## Problem statement

Repository maintenance and upkeep can be time-consuming and error-prone. In particular, links in documentation are hard to maintain and track since website URLs can change and become out-of-date. We seek to create an automated way to identify broken links.

## Solution

We would like to create a workflow that uses a GitHub Action to automate the identification of broken links in markdown files to assist with the repository's maintenance and upkeep. This approach is chosen because it is tried and tested - CMS OSPO uses the same workflow to maintain their numerous websites and documentation. I implemented the change by adding a new workflow file for this to be run as GitHub Action

## Major changes

## Testing and review

Tested this in the DSACMS fork where it ran successfully and and produced results here:
https://github.com/DSACMS/uswds/actions/runs/22197348828/job/64201014836

If you'd like, I can update the workflow file to include `workflow_dispatch` which allows you to manually trigger the workflow so you can test it in this repository!


annepetersen (approved):

.github/workflows/ci-docs.yml
```diff
@@ -0,0 +1,20 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  lint-markdown:
+    name: Lint markdown
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
ethangardner, line 16: Thanks for this contribution! Our repo (uswds) has a security requirement on the repository that enforces that GitHub actions are pinned to a specific commit hash as opposed to the version number. I find that a comment after the SHA is also helpful so the installed version is more readily identified. This would look like:

`- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # pin v6.0.2`

This also applies to the markdown link check action below. Would you mind making this change?

ethangardner (changes requested):

annepetersen (changes requested):
Thanks @natalialuzuriaga ! While you're doing that revision, could you see if there's a way to ignore errors we might get from sites that block scanning tools? ([npmjs.com](http://npmjs.com/) and [ssa.gov](http://ssa.gov/) both do that, for example)

Should be doable through a config file in .github/workflows/markdownlink-config.json

.github/workflows/ci-docs.yml
natalialuzuriaga: Gotcha, updated accordingly!

natalialuzuriaga:
> Thanks @natalialuzuriaga ! While you're doing that revision, could you see if there's a way to ignore errors we might get from sites that block scanning tools? ([npmjs.com](http://npmjs.com/) and [ssa.gov](http://ssa.gov/) both do that, for example)
> 
> Should be doable through a config file in .github/workflows/markdownlink-config.json

Gotcha, added the config file where 403 errors considered an alive code, thus will be ignored

natalialuzuriaga:
Ready for re-review! Tested this in the DSACMS fork where it ran successfully and and produced results here: https://github.com/DSACMS/uswds/actions/runs/25189223094/job/73854606974

ethangardner (approved):
LGTM. Thank you!

ethangardner (changes requested):
Apologies, @natalialuzuriaga . I had previously approved this, but I notice when I go to the `gaurav-nelson/github-action-markdown-link-check` repository. It indicates this action has been deprecated. 

> This repository is now ⛔️ deprecated and is no longer actively maintained. For support and further development, please use the maintained fork available at [Tcort GitHub Action Markdown Link Check](https://github.com/tcort/github-action-markdown-link-check).

Would you be able to swap these out? My previous comment about dependency pinning still applies, and hopefully this is a 1-to-1 compatible package swap since it was a fork.

natalialuzuriaga:
> Apologies, @natalialuzuriaga . I had previously approved this, but I notice when I go to the `gaurav-nelson/github-action-markdown-link-check` repository. It indicates this action has been deprecated.
> 
> > This repository is now ⛔️ deprecated and is no longer actively maintained. For support and further development, please use the maintained fork available at [Tcort GitHub Action Markdown Link Check](https://github.com/tcort/github-action-markdown-link-check).
> 
> Would you be able to swap these out? My previous comment about dependency pinning still applies, and hopefully this is a 1-to-1 compatible package swap since it was a fork.

Oops, good catch. I completely missed that the action has been archived. Updated to use tcort/github-action-markdown-link-check. Workflow runs successfully as shown here: https://github.com/DSACMS/uswds/actions/runs/25342653751/job/74303622832

ethangardner (approved):
Looks good. Thanks!

ethangardner:
@annepetersen - You had a previous change request, so this will need an approval from you as well if it looks good to you. The workflow is flagging some links, but that's to be expected until https://github.com/uswds/uswds/pull/6622 is merged.

annepetersen (approved):

PR #6613, opened 2026-04-20
Replace individual font links that 404

annepetersen:
- Update two font licenses with the most current stable links available for them

Fixing in advance of implementing a [broken link checker workflow](https://github.com/uswds/uswds/pull/6611)


ethangardner (approved):
LGTM

PR #6623, opened 2026-04-24
USWDS - USA-Banner: replace <br> "empty group" with CSS

ajfarkas:
# Summary

_Fixed a bug that caused screen readers to hear "empty group" in the Official gov website banner_ Previously, we used the line break element, `<br>`, to create a visual break in the banner detail. 

## Breaking change

## Problem statement

A bug in Edge causes Mac VoiceOver to read `<br>` as "empty group". VoiceOver users hear unneeded page structure data, which should be removed. This could cause confusion about whether the user is missing information.

## Solution

Our use of `<br>` causes this unneeded text to be read.  Because this is a visual separation and not a programmatic one, we should use CSS to style these elements.  
This change styles the separated text as its own block, which visually separates it.  

An alternative approach is to move the initial text into its own `<p>` element, and assign the following text the class `.margin-top-0`, but that doesn't offer significant benefit. We could also consider whether this is a header, or if there's a better HTML structure for this section.

## Testing and review

Tested on Mac Edge and Chrome with VoiceOver.  
With VO enabled, navigate to any USWDS page, as they all have this banner. Select "Here's how you know" to expand the accordion, and listen to the contents. The phrase "empty group" is read twice before this change, and not at all after.

Please suggest any HTML structure that is more appropriate.

- [x] Confirm that this code follows the [18F Front End Coding Style Guide](https://pages.18f.gov/frontend/) and [Accessibility Guide](https://guides.18f.org/accessibility/checklist/).
- [ ] Run `npm run prettier:sass` to format any Sass updates.
- [ ] Run `npm test` and confirm that all tests pass.

### DIFF
--- a/packages/usa-banner/src/usa-banner.twig
+++ b/packages/usa-banner/src/usa-banner.twig
@@ -30,13 +30,13 @@
           <img class="usa-banner__icon usa-media-block__img" src="./img/icon-dot-gov.svg" role="img" alt="" aria-hidden="true">
           <div class="usa-media-block__body">
             {# Will cause error without trim because it doesn't know what to do with new lines #}
-            <p><strong>{{ domain.heading }}</strong><br/>{{ domain.text | trim | raw }}</p>
+            <p><strong class="display-block">{{ domain.heading }}</strong>{{ domain.text | trim | raw }}</p>
           </div>
         </div>
         <div class="usa-banner__guidance tablet:grid-col-6">
           <img class="usa-banner__icon usa-media-block__img" src="./img/icon-https.svg" role="img" alt="" aria-hidden="true">
           <div class="usa-media-block__body">
-            <p><strong>{{ https.heading }}</strong><br/>{{ https.pretext | trim | raw }} ({{ lock }}) {{ https.posttext | trim | raw }}</p>
+            <p><strong class="display-block">{{ https.heading }}</strong>{{ https.pretext | trim | raw }} ({{ lock }}) {{ https.posttext | trim | raw }}</p>
           </div>
         </div>
       </div>


annepetersen:
Hey @ajfarkas: thanks for this. There might be a few typos that lose some important info in this, like 
"Our use of
causes"...

Your use of... what? was my question.

ethangardner:
I edited the original submission. There were missing tick marks around one of the `<br>` tags in the description, and it made it look like there was a word missing.

ajfarkas:
I discussed this a bit with Chandra, and it seems like we should close this. I'll give her the opportunity to comment first.

chandracarney:
Thanks @ajfarkas! We just talked about potential issues coming up for teams regarding `display-block`. If teams use default USWDS, they get utility classes (`display-block`). If they don’t use the default, they don’t get the utility classes. Because banner is so widely used, we don't want to inadvertently  break things. So we'll close this for now.

Pulling some of our prior discussion for context into the repo here, too.

> Broadly, my philosophy has always been to put pressure on the system or product that's causing the issue, and generally I'd recommend reporting the issue to the company or product's support teams through an available avenue. This is, of course, assuming we've tested our own system and verified it isn't us causing the issue 
> 
> If we were to fix the issue for an edge case (hah! Edge case), that could be a problem in two ways:
> 
> - While it might alleviate the pain for that user group, we might accidentally introduce issues for users of other combinations of browsers (Safari, Chrome, Firefox) and screen readers (NVDA, TalkBack, JAWS). While we could test using a sampling of those combinations, we likely wouldn't cover all of them. We might then have to keep fixing those downstream issues, if they're ever reported. 
>   - There are occasional cases, of course, where making a change would fix things because the change introduces different, appropriate, valid semantic HTML/JS/CSS
> - If we made the change and Edge fixes the issue, it could cause issues for the original user group. It's unlikely we'd hear about the fix in a timely manner (if ever), and if we did hear about the fix, we don't want to depend on our availability to immediately be able to revert the change.

Issue #5239, opened 2023-04-14
USWDS: Update storybook to 7.0.x


mahoneycm:
## Description

Dependabot flagged some dependency descrepencies that could be resolved by updating Storybook to the latest version.

Storybook offers an automigrate feature but needs some additional manual updating before it'll be completely usable on our repo.

mejiaj:
This could unblock #5010.

Coop9200:
I'm getting a lot of 'unable to resolve path to module' and 'missing file extension' after I updated storybook to v7.  Any thoughts on how to resolve?

ethangardner:
Update to the above and scope of this issue as it was opened some time ago. A recent update was done to go from v6 to v7, but the current version as of this writing is v10. I think it would be best to go incrementally to v8 then 9 followed by 10 to hopefully keep the PR scope smaller and easier to review.

PR #6638, opened 2026-05-12
Remove Unmaintained Dependency

akuny:
# Summary

Remove an unmaintained dependency. This is not a breaking change.

## Problem statement

Ahead of starting work on https://github.com/uswds/uswds/issues/5239, we found that twigjs-loader is not actively maintained. Leaving it in place during the Storybook 6 → 7 upgrade would mean debugging two unfamiliar variables at once if anything broke (the loader plugs into Storybook's Webpack config, which is rewired in v7). Since this package is essentially a thin wrapper around the twig npm package, [vendoring](https://macwright.com/2021/03/11/vendor-by-default) this dependency's functionality cleans up our direct dependencies and should clear the way for iterative Storybook version upgrades.

## Solution

Vendor the twig webpack loader (added as tasks/webpack-twig-loader.js, adapted from twigjs-loader, MIT) and depend on the twig npm package directly.
 
## Dependency updates

| Dependency | Previous version | New version |
|---|:---:|:---:|
| twigjs-loader | 1.0.3 | (removed) |
| twig | — | 1.17.1 (new) |


ethangardner (approved):
I think this makes sense to get us a cleaner upgrade path. Thanks!

Issue #4644, opened 2022-04-24
Utilizing CSS variables


morganspencerdev:
**Is your feature request related to a problem? Please describe.**

The CSS styles currently have some limitations, mainly that SASS is required for customization with the tokens and many other features. This tends to overcomplicate things when implementing the styles.

**Describe the solution you'd like**

Now that support for IE11 has been dropped in version 3.0, I'm proposing the use of CSS variables to simplify implementation and improve customization. Many modern frameworks utilize the robustness of these variables to provide flexibility both in development and to end users.

This would, of course, be a large undertaking and some things cannot be solved fully with this solution. However, I think it provides great benefits to users of USWDS.

I'm happy to assist and I can even make a small mockup of the proposed solution if that might help.

**Describe alternatives you've considered**

I haven't found an alternative that would provide the same flexibility and customization that CSS variables would.

**Additional context**

CSS variables support: https://caniuse.com/css-variables

mejiaj:
Hey @morganmspencer, thanks for submitting this issue!

Could you explain a bit more on how CSS variables would simplify implementing styles? Do you mean that we'd get rid of the SASS requirement? We'd need to make sure this proposal covers the browsers above 2% in https://analytics.usa.gov/

A small mockup of the proposed solution would be helpful.

esheehan-gsl:
I'd love to see the design tokens provided as CSS custom properties similarly to how [Open Props](https://open-props.style/) is structured. It'd be really handy to be able to load a couple of CSS files to use the tokens without having to set up Sass. A person could then build off of the USWDS tokens without opting in to all of the components simply by importing the relevant custom props CSS.

CullenJWebb:
I strongly support this suggestion.

SASS is the best way to customize a front-end framework like this but sometimes that is unfortunately not an option. Just recently I was working with a team on a project and we were restricted to vanilla CSS (for silly reason, admittedly). 

There are other benefits , such as easier theming with tools such as the new WordPress Site Editor. With block based themes being able to edit vanilla CSS variables on the fly, but not SASS, they are presently locked out of USWDS customization.

It also unlocks more granular control with vanilla CSS. Similar to editing the variables with :root we could also/instead use a .modifier class and only change the variables in specific circumstances:

```
/* User defined variables */
:root {
    /* Replaces USWDS defined variable value */
    --usa-theme-color__info: var(--usa-color__blue-cool-30v);
}

/* Dark Mode */
.dark {
[...]

PR #6650, opened 2026-05-14
expose theme color values as css variables for consumers of uswds.css

nektro:
# Summary

Exposes CSS Variables parallel to the existing `$theme-color-` SCSS ones.
Does not use them internally yet.

While there were more variables listed in `packages/uswds-core/src/styles/settings/_settings-color.scss` I chose to for now only add variables for the tokens listed on https://designsystem.digital.gov/design-tokens/color/theme-tokens/

## Breaking change

## Related issue

Related: https://github.com/uswds/uswds/issues/4644

## Related pull requests

N/A.

## Preview link

image

## Problem Statement

The way I use USWDS is by manually importing the `uswds.css` in my HTML like so:

```html
<link rel="stylesheet" href="/-/npm/@uswds/uswds@3.11.0/dist/css/uswds.css" integrity="sha384-G0llpcOOfpdIocwce8kYFUqkiabzRMkzBn3zBHHRag3s/1wb6JKvEnArq0RXjZUT" crossorigin="anonymous" />
```

As a result, one of the things I lose out on is the ease of theming that comes with the sass build system. Now migrating to use these variables may be too soon to migrate internally ([Can I Use](https://caniuse.com/css-variables) reports a 95% support), but it will allow for easier consumption for folks like me, and an easier migration later.

### DIFF
--- a/packages/uswds-elements/src/styles/_color.scss
+++ b/packages/uswds-elements/src/styles/_color.scss
@@ -8,7 +8,7 @@
     --theme-color-base-dark: #{color("base-dark")};
     --theme-color-base-darker: #{color("base-darker")};
     --theme-color-base-darkest: #{color("base-darkest")};
-    // --theme-color-base-ink: #{color("base-ink")};
+    --theme-color-base-ink: #{color("ink")};
 }
 :root {
     --theme-color-primary-lighter: #{color("primary-lighter")};
--- a/packages/uswds-elements/src/styles/_color.scss
+++ b/packages/uswds-elements/src/styles/_color.scss
@@ -1,0 +1,42 @@
+@use "uswds-core/src/styles/functions/utilities" as *;
+
+:root {
+    --theme-color-base-lightest: #{color("base-lightest")};
+    --theme-color-base-lighter: #{color("base-lighter")};
+    --theme-color-base-light: #{color("base-light")};
+    --theme-color-base: #{color("base")};
+    --theme-color-base-dark: #{color("base-dark")};
+    --theme-color-base-darker: #{color("base-darker")};
+    --theme-color-base-darkest: #{color("base-darkest")};
+    // --theme-color-base-ink: #{color("base-ink")};
+}
+:root {
+    --theme-color-primary-lighter: #{color("primary-lighter")};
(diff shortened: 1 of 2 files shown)


caseywatts:
There's another effort to do USWDS tokens as css variables that you might like to know about!
https://github.com/open-government-design-system/ogds-tokens

This tokens project is being led by Matt Henry, who was the former technical lead of USWDS.

There's also another Slack space you might be interested in, that's where this project is being discussed:
https://docs.google.com/document/d/10XN6ZsRt9aAOH1PUmtfcCCEBZQUSgs1cCI5krYBOoi4/edit?tab=t.0

caseywatts:
For browser support, "tracked browsers" is the setting I would reference in CanIUse.com this -- that removes a bunch of bot accounts and scripts. 99%+ is pretty good! That meets the USWDS criteria of 98% of browsers being supported.

image

And here's a snapshot from last year for how my team is thinking about browser support, generally:
https://www.caseywatts.com/blog/browser-support-2025/

ethangardner:
Thanks for this PR! I agree there's a need to provide the tokens as CSS custom properties. As @caseywatts mentioned, there is some work being done on this in USWDS and its surrounding community. We have some of the token work started in https://github.com/uswds/uswds-elements repo in the `tokens` directory. If you wanted to contribute to that effort, we're using [style dictionary](https://styledictionary.com/) as the source of truth, and this is on our [roadmap](https://github.com/orgs/uswds/projects/8/views/59).

I'll close this PR since it's diverging a bit from our work in progress, but we'd welcome your contributions in the repo mentioned above.

nektro:
I don't currently plan on migrating to uswds-elements since it uses web components

ethangardner:
@nektro - I understand. You're welcome to use whatever fits your needs. I will add that it would be possible with what I proposed to only use the style tokens from elements and not use the web components themselves. Some of the artifacts we produce in the package are CSS files that contain token definitions. 

You can see these here if you drill into the `build/css` directory: https://www.npmjs.com/package/@uswds/elements?activeTab=code

nektro:
thanks!

PR #6651, opened 2026-05-15
[uswds-core]: ink assignments theme color was referencing wrong variable

nektro:
in the default theme `base-darkest` and `base-ink` are the same hex code (`gray-90` `#1b1b1b`) but anyone overriding their own theme colors using this variable would not see the value reflected.


ethangardner (approved):
Thanks for the contribution! LGTM.

PR #6659, opened 2026-05-19
USWDS - Range slider: Improve border visibility

### LINKED ISSUE #6444: USWDS - Feature: Improve color contrast and focus outline on range slider

Both [end-user feedback](https://github.com/uswds/uswds/issues/6117#issuecomment-2478954889) and [a11y expertise](https://github.com/uswds/uswds/issues/6117#issuecomment-2725823497) recommend the following:

- Thicken the focus indicator border line around the slider
- Increase the contrast of the slider edge and interior color. Currently it barely passes 1.4.11

Image

1. Thicken the border (not the focus border but the border of the range slider itself) by at least 1-2 pixels to increase visibility for low vision users.

2. Increase the contrast between the background color of the range slider and the border around it. Try using hex color #646464 instead of hex color #757575

_No response_

### Additional context

_No response_

### Code of Conduct

manichandra:
# Summary

Improved range slider visibility for low-vision users. The range slider border now uses a thicker `2px` width and darker `#646464` color for stronger contrast against the slider background.

## Breaking change

## Related issue

Closes #6444

## Related pull requests

None.

## Preview link

Preview link: N/A

## Problem statement

The range slider border was difficult to see because the default border was thin and had limited contrast against the slider background. This could make the component harder for low-vision users to perceive.

## Solution

Updated the range slider border styling to use a `2px` border and `#646464`, matching the issue recommendation. This applies to the track border and thumb border treatment while leaving the existing focus behavior unchanged.

## Major changes

- Increased the range slider track border from `1px` to `2px`.
- Updated the range slider border color to `#646464`.
- Updated the thumb border treatment to use the same `2px #646464` styling.

## Testing and review

Tested locally with:

```bash
npx prettier --check packages/usa-range/src/styles/_usa-range.scss
npx stylelint packages/usa-range/src/styles/_usa-range.scss
npx gulp compileSass
npm run lint:sass
git diff --check

### CODE REVIEW

**Comment by manichandra** [comment]
image
 
 after the fix

**Comment by chandracarney** [comment]
Thank you @manichandra! Accessibility improvement-wise, this is looking good to me. I've added one of our engineers to give a review and sign-off.

File: packages/usa-range/src/styles/_usa-range.scss [hunk 1]
```diff
@@ -1,5 +1,8 @@
 @use "uswds-core" as *;
 
+$range-border-color: #646464;
```
ethangardner: The before diff uses a [theme color token](https://designsystem.digital.gov/design-tokens/color/theme-tokens/), `color("base")`. Since the intent is to go darker for more contrast, my suggestion is to keep the color themeable as it was before. 

You might try `color("base-darker")` as a replacement. Also, I would suggest making this variable private by using the leading dash (`-`) convention or simply by using `color("base-darker")` directly where it appears below. I have a personal preference for the latter.

packages/usa-range/src/styles/_usa-range.scss
```diff
@@ -1,5 +1,8 @@
...
+$range-border-width: units(2px);
```
ethangardner: Looking elsewhere in the codebase, the convention we follow in https://github.com/uswds/uswds/blob/6ca2a6f701b036ad932bc023c4489fe0014652fc/packages/usa-nav/src/styles/_usa-nav.scss is to define only the value as a variable and to not assign the function result to the variable. i.e. this would become `$-range-border-width: 2px;`, and then its usage below would be `units($-range-border-width)`. The same comment I left about the variable being private or sticking to direct usage applies to this as well.

ethangardner (changes requested):
Thanks for the contribution! I left some comments related to conventions we follow elsewhere in the codebase and there are two stylistic changes I requested.

packages/usa-range/src/styles/_usa-range.scss
manichandra: reaplced with color("base-darker")

packages/usa-range/src/styles/_usa-range.scss
manichandra: have made the change

ethangardner (approved):
LGTM! Thank you!

PR #6670, opened 2026-05-21
Add range limit labels example

### LINKED ISSUE #4504: Add example markup for range limit labels
There is usability guidance on the [range slider component](https://designsystem.digital.gov/components/range-slider/) page that states:

```
When appropriate, label the ends of the slider with the limits 
of the range (for example: “0/100”, “small/large” or 
“less expensive/more expensive”).
```

but there is no example markup on what the end labels should use. This makes it difficult to know how to add these labels so they display consistency, and in line with the USWDS. 

It is also unclear if there is accessibility concerns to using particular markup. For instance, are there `aria` attributes that should be included with this markup? 

Add an example to the component documentation page of the end label markup.

tatakaisun:
## Summary
- Adds a range slider example variant with visible minimum and maximum labels.
- Keeps the labels hidden from assistive technology because the same limits are already exposed through the range input attributes.

## Why
Closes #4504.

## Changes
- Adds optional `range_min_label` and `range_max_label` markup to the range Twig template.
- Adds a Storybook example with `min="0"`, `max="100"`, and matching endpoint labels.

## Testing
- `git diff --check`

## Known Issues
- Full Storybook/build checks were not run locally because dependencies were not installed in this fresh clone.

### DIFF
--- a/packages/usa-range/src/usa-range.stories.js
+++ b/packages/usa-range/src/usa-range.stories.js
@@ -16,6 +16,14 @@ export default {
       name: "Preposition (of, for, in, de, etc.)",
       control: { type: "text" },
     },
+    range_min_label: {
+      name: "Minimum label",
+      control: { type: "text" },
+    },
+    range_max_label: {
+      name: "Maximum label",
+      control: { type: "text" },
+    },
     min: {
       name: "Min",
       control: { type: "number" },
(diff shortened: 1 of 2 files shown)


ethangardner (commented):
Thanks for the contribution! Based on other findings in #6681, we are going to hold off on this until we decide internally how we can support current value label, max and min labels in a way that scales well for the different device types. 

The USWDS team needs to have a discussion before we weigh on on the approval status of this work.

ethangardner:
Thanks for the contribution. There are some other design tweaks that we would like to make to this component to reflect the min/max and current values of the range slider component. We're going to hold off on this until we have the design tweaks sorted out, but we appreciate the contribution nonetheless.

Issue #6443, opened 2025-03-14
USWDS - Bug: Range slider is missing instructions


amycole501:
### Describe the bug

The range slider has no visible instructions (and therefore no screen reader accessible instructions) for how to operate the range slider.

### Steps to reproduce the bug

1. Go to https://designsystem.digital.gov/components/range-slider/
2.  Note there are no instructions on how to select an amount.

### Expected Behavior

The range slider needs visible instructions, which are also screen reader accessible, to pass [WCAG 3.3.2 labels or instructions.](https://www.w3.org/TR/WCAG21/#labels-or-instructions)

Above the range slider element add the instructions "Move the button to the right to increase the amount and to the left to decrease it."

Ensure the instructions are associated programmatically with the range slider to they are also accessible to a screen reader to comply with [1.3.1 Info and relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships).

### System setup

All devices and browsers.

### Additional context

Ideally, this needs to be done prior to May 2025.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [x] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

amycole501:
The sample instructions should read "Use the control button to drag the slider to the left or right to select an amount."

daresTheDevil:
Took a look at this. Added a `usa-hint` span with `aria-describedby` on the input — same pattern used in character-count, memorable-date, and other USWDS form components. The hint text defaults to the instructions from the issue but is configurable via a Twig variable (`hint_text`) for localization or custom slider implementations (vertical, etc.).

Updated all test fixtures and added a unit test for the `aria-describedby` → hint association. Full test suite passes (742/742).

PR incoming.

chandracarney:
After discussing range slider with the VA team, we've decided to create a fresh issue describing this bug in #6614. I'll close this issue out and continue discussions in the new issue. Thanks @amycole501 for all your work on improving range slider!

PR #6673, opened 2026-05-23
fix: add visual hint instruction to range slider  - 6614

### LINKED ISSUE #6614: USWDS - Bug: Introduce visual instruction for range slider
### Describe the bug

The range slider component should have instructions to ensure visual users know how to operate the slider.

In partnership with VA teams (including @coforma-jamie, @artsymartha68), we are re-describing minor issues with range slider under this issue and closing out a prior issue, in order to avoid confusion. This issue closes and replaces #6443 and is related to #6116.

2. Note there's no visual instruction indicating how to operate the component

### Expected Behavior

The range slider component and example on the USWDS website should both have visual instruction indicated through a hint in the component example:
1. The instruction should read: “Move the slider to change the value”
2. The color of this text can be presented in the dark grey `class="usa-hint"`, which is used in both [character count](https://designsystem.digital.gov/components/character-count/) and [date range picker](https://designsystem.digital.gov/components/date-range-picker/) components.

### Related code

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

Full instruction on how to operate a range slider is available to screen reader users through a screen reader's inherent behavior navigating semantic HTML. Using both VoiceOver and JAWS screen readers, we verified they each describe how to use the range slider as-is, when no hints or ARIA are added on. Because instruction slightly differs across screen readers, we want to ensure we aren't overly prescriptive in applying screen reader-only hints (excess ARIA was removed in #5413).

### Code of Conduct

ravitejapioneerblaze-code:
## Summary
Added a visual hint to the range slider component so sighted users 
know how to operate the slider. Previously, only screen reader users 
received this instruction through VoiceOver/JAWS behavior.

## Breaking change

## Related issue
Closes #6614

## Related pull requests
None

## Problem statement
The range slider component had no visual instruction for sighted users on how to operate it. Screen reader users already received guidance through VoiceOver and JAWS, but visual users had no such hint, creating an inconsistent experience.

## Solution
Added a `usa-hint` span element below the range slider input with the text "Move the slider to change the value". This matches the existing hint pattern already used in the character-count and date-range-picker components.

## Major changes
* Added `<span class="usa-hint" id="range-hint">Move the slider to change the value</span>` to `usa-range.twig`
* Added `aria-describedby="range-hint"` to the range slider `<input>` in `usa-range.twig`
* Updated `template.html` to match
* Added automated test in `range.spec.js` to verify hint exists, contains correct text, has correct id, and input has correct aria-describedby

## Testing and review
* Added unit test verifying `.usa-hint` element exists, contains correct text, has `id="range-hint"`, and range input has `aria-describedby="range-hint"`
* All 6 tests passing
* Run: `npx mocha --require jsdom-global/register packages/usa-range/src/test/range.spec.js`


package.json
```diff
@@ -161,11 +161,11 @@
     "gulp-svgstore": "9.0.0",
     "gulp-uglify": "3.0.2",
     "html-webpack-plugin": "5.6.7",
-    "jsdom": "22.1.0",
+    "jsdom": "^22.1.0",
```
ethangardner: Would you mind rolling back the updates to package.json and package-lock.json that were included with this PR?

They don't appear to be necessary for the a11y improvement. If they are necessary, could you indicate why in the PR description?

packages/usa-range/src/test/range.spec.js
```diff
@@ -64,14 +63,27 @@ tests.forEach(({ name, selector: containerSelector }) => {
...
-        // change the span element, make sure it updated and that the span and the range are equal.
         assert.strictEqual(
           slider.value,
           spanElement.textContent,
           "slider value does not match span value",
         );
       });
+
+      // NEW TEST — Added for issue - 6614
+      // Verifies visual hint is present for sighted users
```
ethangardner: Can we remove these comments? I think the description in the `it` function argument sufficiently captures it, and `git blame` or `annotate` sufficiently captures what commit or issue this was part of.

packages/usa-range/src/test/range.spec.js
```diff
@@ -64,14 +63,27 @@ tests.forEach(({ name, selector: containerSelector }) => {
           "40",
           "range slider value is not set to the value in the test.",
         );
-
-        // change the span element, make sure it updated and that the span and the range are equal.
```
ethangardner, line 68: Can you restore this comment please? I think it's helpful to understand the intent of the test assertion and desired behavior.

ethangardner (changes requested):
Thanks for the contribution! This looks very close, but there are a few minor things that I noticed and a question about dependencies that I had on this work.

packages/usa-range/src/test/template.html
```diff
@@ -9,4 +9,5 @@
     value="20"
     data-text-unit="percent"
   >
+  <span class="usa-hint">Move the slider to change the value</span>
```
ethangardner: @chandracarney - I'd like your opinion on this. Looking at what is here, my gut tells me it might be a good idea to add `id` on the `<span>` and `aria-describedby` on the `<input>` to indicate that the instructions correlate to the range slider. What do you think?

I tested it with NVDA, and the readout I got made more sense with the id and aria than it did without it.

package.json
ravitejapioneerblaze-code: You're Right. Done! Rolled back package.json and package-lock.json to upstream versions.

packages/usa-range/src/test/range.spec.js
ravitejapioneerblaze-code: That makes sense. Removed the spec comments as requested.

packages/usa-range/src/test/range.spec.js
ravitejapioneerblaze-code: Thanks, restored the spec comments as requested.

packages/usa-range/src/test/template.html
chandracarney: I think that makes sense to me as well, though I'd like to tag in @jonathanbobel for ARIA discussion here.

jonathanbobel: I agree with the semantics here regarding the aria-describedby and associated id relationship you proposed, @ethangardner. I would suggest revising the message itself, however, as it currently only identifies the control as a slider, which assistive technology users will already know from the element’s role announcement, rather than explaining its purpose or how to interact with it.

Since USWDS allows developers to define values such as `min`, `max`, and `step`, a more useful hint might provide interaction guidance, such as informing users that they can use the arrow keys to increase or decrease the value.

In the USWDS guidance, it may also be worthwhile to encourage developers to provide information about the slider’s range (minimum and maximum values) and increment size when that information is important to understanding the control. That could be addressed as a follow-up issue.

chandracarney: Thanks @jonathanbobel and sorry I didn't include some more context here about the current text recommendation. There was an older issue #6443 that recommended to include a more verbose and precise hint, however, we recently discussed it with VA teams and decided the hint should be more generalized to account for mobile gestures, native screen reader-based instruction, as well as potential orientation (vertical or horizontal) presentation choices. We closed #6443 and replaced with this parent issue #6614. (I also uncovered an open issue to consider `min` and `max` values: #4504 and we can aim to discuss in the next week.)

Thank you so much for your input about aria, and again, sorry I didn't share all that context with you!

jonathanbobel: That makes sense @chandracarney. I would still recommend a **guidance** update on the USWDS component to help developers make the hint more useful depending on their use, and what attributes are set with their specific implementation. That recommendation doesn't interfere with this approval, gtg after the update.

chandracarney: @jonathanbobel thanks, great! We can discuss guidance updates in the next week, too 🥳

ethangardner (approved):
Thank you so much! This looks good.

ravitejapioneerblaze-code:
@ethangardner , Thank you for the review and the merge, really appreciate it! Looking forward to contributing more to USWDS.

Issue #6678, opened 2026-05-26
USWDS - Feature: Upgrade ESLint from 8.x to 10.x


IHIutch:
The current stable line is **10.x** (10.4.0 as of writing). v8 stopped receiving non-security fixes in October 2024,  v9 is in maintenance mode (`9.39.4`, tagged `maintenance`).

Upgrading unlocks:

- **Modern first-party plugins** that target the flat-config `language` API:
  - `@eslint/json`: lints `.json` and `.jsonc` files. Would let us validate `packages/*/src/content/*.json` (currently only implicitly checked by `npm run test:html` via `JSON.parse`)
  - `@eslint/markdown`: lints code blocks inside Markdown
  - `@eslint/css`: first-party CSS linter
- Security and bug fixes that aren't being backported to 8.x
- Performance improvements in the flat-config code path
- Removes `ESLINT_USE_FLAT_CONFIG=true` from `lint:js`

### Dependencies

| Package | Current | Target | Notes |
|---|---|---|---|
| `eslint` | 8.56.0 | ^10.x | Major bump (skipping 9) |
| `eslint-config-prettier` | 9.1.0 | ^9.x or ^10.x | Already supports 10 |
| `eslint-plugin-no-unsanitized` | 4.1.2 | ^4.x | Already declares `eslint: ^9 \|\| ^10` |
| `eslint-plugin-lit` | 2.2.1 | ^2.x | Peer `eslint: ">= 8"` — should work; verify |
| `eslint-plugin-import` | 2.32.0 | **drop, replace with `eslint-plugin-import-x`** | Latest `eslint-plugin-import` peer caps at `^9`. `eslint-plugin-import-x` is the maintained fork with ESLint 10 support; same rule names and config shape |
| `eslint-config-airbnb-base` | ^15.0.0 | **drop, replace** | Pinned to `^7 \|\| ^8`; unmaintained. Already incompatible with 9. See "Base config replacement" below |
| `eslint-plugin-airbnb-base` | ^0.0.1-security | **drop** | Security placeholder, not a real plugin |

### `eslint-config-airbnb-base`

`eslint-config-airbnb-base` isn't supported in v10, recommendation is to replace it with `eslint:recommended` and manually add any project specific rules.

### Config (`eslint.config.mjs`)

- Replace `import importPlugin from "eslint-plugin-import"` with `import importPlugin from "eslint-plugin-import-x"` (rule names and config keys are identical)
[...]

PR #6679, opened 2026-05-26
USWDS - Build: Replace webpack-based build:html with a direct Twig.js test:html validator 

IHIutch:
## Summary

Replaced the webpack-based `build:html` script with a small Node script that uses `twig` directly to validate templates, unblocking `npm run prettier` for contributors on a fresh checkout and removing a gitignored build artifact (`html-templates/`) that nothing consumed.

## Breaking change

Two scripts are affected. Any external CI, automation, or contributor workflow that invokes them directly will need to update:

- `npm run build:html` renamed to `npm run test:html`. Same purpose (validate every `.twig` + `content/*.json` pair renders), faster execution, no on-disk output.
- `npm run prettier:html` removed. Its target directory (`html-templates/`) is no longer generated by any script. `npm run prettier` no longer chains it.

No runtime, package API, markup, or component behavior changes.

## Related issue

n/a

## Problem statement

The PR template asks authors to run `npm run prettier` before submitting. That script previously chained `prettier:js`, `prettier:sass`, and `prettier:html`. `prettier:html` targets `./html-templates/**/*.html` which only exists after running `npm run build:html`. On a fresh checkout, if you run `npm run prettier` you get an error message:  `"No files matching the pattern were found: "./html-templates/**/**/*.html"`. 

`build:html` wrote HTML to a gitignored directory **that nothing in the repo or published package consumed**. It _appears_ to be a CI test that only validates that the Twig templates and JSON content render. I'd think `storybuild build` would already suffice for this use case.

## Solution

Replaced the entire webpack pipeline with a Node script, `tasks/check-twig.js`, that uses the `twig` package directly. No files are written, no output directory is created.

**Output parity verified:** both pipelines were run side-by-side against all 213 template/data pairs for comparison. With `html-webpack-plugin`'s default HTML minification disabled, the outputs are identical.

**Storybook unaffected:** the `.storybook/main.js` webpack config uses a separate vendored loader (`tasks/webpack-twig-loader.js`) to compile `.twig` imports in `.stories.js` files. That pipeline is untouched.

## Major changes

- **New** `tasks/check-twig.js` direct Twig.js validator
- **Deleted** `webpack.twig.config.js` sole consumer was `build:html`
- **Deleted** `tasks/twig-entry.js` empty file that existed only to satisfy webpack's entry requirement
- **`package.json`**
  - Renamed `build:html` to `test:html` (now `node ./tasks/check-twig.js`)
  - Updated `test:ci` to call `test:html`
  - Removed `prettier:html`
  - Removed `html-webpack-plugin` from `devDependencies`
  - Removed `webpack.twig.config.js` from the `files` array
- **`.gitignore`** removed `html-templates/` which is no longer generated

## Testing and review

1. `npm install` to pick up the removed `html-webpack-plugin` dep
2. `npm run prettier` on a fresh checkout, should complete without errors
3. `npm run test:html` should report `OK — rendered 213 twig templates` and exit 0
4. `git status` after running scripts`html-templates/` should not appear in the working tree
5. `npm run test:ci` to check full CI pipeline; `test:html` runs as the final step in place of `build:html`
6. Storybook still builds and renders all twig stories: `npm run build:storybook`

## Dependency updates

| Dependency name | Previous version | New version |
| --- | :---: | :---: |
| html-webpack-plugin | 5.6.7 | -- |

## Feedback

Are there any external workflows (downstream CI, contributor scripts, internal tooling) that call `npm run build:html` or `npm run prettier:html` directly and would need a heads-up about the rename/removal?

I'd think `storybuild build` would already suffice for this use case. It might just be better to remove this check altogether? Realistically, it's only validating `packages/*/src/content/*.json`, which could just be validated with `@eslint/json` (although most editors validate JSON natively already), but that requires ESLint ^9.x (Related #6678).


ethangardner (changes requested):
Thanks for the contribution! There's a non-obvious wrinkle with twig and how it's used. The site in the https://github.com/uswds/uswds-site repo consumes the code produced by Twig to make the example code that's live on https://designsystem.digital.gov, which is also why these are formatted with Prettier.

What you pointed out is a bug where prettier is run against a non-existent directory. If you wanted to fix that part, think a node script that does a quick check before the `prettier:html` task would be a good addition that wouldn't have as big of an impact on the example code that's on the site.

IHIutch:
Wouldn't it make more sense for the site repo to run `build:html` from its own repo root? It's not clear why the USWDS source repo needs to maintain a single function that's used by the downstream site repo. Separation of concerns and all, that feels fragile and unnecessary.

Calling `build:html`or `prettier:html` as a maintainer of this repo doesn't actually do anything meaningful. All the code in `./html-templates` is gitignored. Validating the twig files (which `test:html` now does and replaces `build:html` in CI) in this repo makes downstream use safe, including in the site repo.

If it helps, I'd be happy to write a complementary PR in the site repo to avoid this dependency. Then the prettier formatting could/should live there too, as necessary.

ryparker:
@coderabbitai full review

ryparker:
The new validation task in `2ed10b1e` doesn't replace everything `build:html` does. Removing `build:html`, `prettier:html`, and the published Webpack Twig configuration breaks existing commands and stops generating `html-templates`. The replacement [`test:html` discards the rendered HTML](https://github.com/uswds/uswds/blob/2ed10b1edc3a63302be020493e7ca582fa3706be/tasks/check-twig.js#L78). Earlier maintainer feedback identifies the documentation site's example-code workflow as one user of that output.

Please keep the existing commands and generated files, either directly or through compatible wrappers. The missing-directory formatting fix could be a smaller, separate change. We need to check both template validation and downstream example generation before removing the breaking-change label.

PR #6681, opened 2026-05-26
Fix: Prevent layout shift in range slider value display

### LINKED ISSUE #6465: USWDS - Range Slider Visual Refinements
> Just two minor styling fixes! Let me know if you have any questions

_Originally posted by @mahoneycm in https://github.com/uswds/uswds/pull/6302#pullrequestreview-2637417638_

ravitejapioneerblaze-code:
## Summary
Fixed layout shift in the range slider value display by replacing 
percentage-based inline sizing with fixed USWDS design tokens. 
Added text-align: center for consistent visual presentation.

## Breaking change

## Related issue
Closes #6465

## Problem
The .usa-range__value element used percentage-based min/max inline 
sizes (5%) causing layout shift when the displayed value changed 
between different digit lengths (single, double, triple digits). 
Hardcoded px values were also inconsistent with USWDS token standards.

## Solution
Replaced percentage-based sizing with fixed units(6) tokens ensuring 
consistent width regardless of digit count. Replaced hardcoded px 
padding/margin values with USWDS units() tokens. Added text-align: 
center for consistent value positioning.

## Changes
- Updated min/max inline-size from 5% to units(6) in _usa-range.scss
- Replaced hardcoded px values with USWDS design tokens
- Added text-align: center to value display
- Added 3 automated tests verifying stable value display across 
  digit changes

## Testing
- All 10 tests passing
- Run: npx mocha --require jsdom-global/register 
  packages/usa-range/src/test/range.spec.js --timeout 10000

## References
- Closes #6465
- Related to #6302

### DIFF
--- a/packages/usa-range/src/styles/_usa-range.scss
+++ b/packages/usa-range/src/styles/_usa-range.scss
@@ -40,10 +40,11 @@
 
 .usa-range__value {
   @extend %block-input-general;
-  max-inline-size: 5%;
-  min-inline-size: 5%;
-  padding-top: 5px;
-  margin-left: 5px;
+  max-inline-size: units(6);
+  min-inline-size: units(6);
+  padding-top: units(1);
+  margin-left: units(1);
+  text-align: center;
 }
 
 .usa-range {
--- a/packages/usa-range/src/test/range.spec.js
+++ b/packages/usa-range/src/test/range.spec.js
@@ -7,11 +7,6 @@ const TEMPLATE = fs.readFileSync(path.join(__dirname, "./template.html"));
 
 const EVENTS = {};
 
-/**
- * send a change event
- * @param {HTMLElement} el the element to sent the event to
- */
-
 EVENTS.change = (el) => {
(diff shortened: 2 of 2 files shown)


ethangardner (commented):
Thanks for the contribution. We're currently having some discussion around range slider and its design. 

I feel like this is a step in the right direction, but I think we (the USWDS team) needs to revisit this overall layout. I think having the current value on the right of the range slider presents some challenges, which isn't a reflection on this PR as I know you were just adding space to the existing label. For the time being, I do want to hold off on any further review of this until we have the discussion internally.

ethangardner:
Thanks for the contribution. There are some other design tweaks that we would like to make to this component to reflect the min/max and current values of the range slider component. We're going to hold off on this until we have the design tweaks sorted out, but we appreciate the contribution nonetheless.

ravitejapioneerblaze-code:
Hi @ethangardner , completely understand. Thank you for the review and for keeping me updated. Happy to help once the design direction is finalized.

PR #6691, opened 2026-06-04
USWDS - Button Group: Fixed margin issues for button groups within usa-form

### LINKED ISSUE #5883: USWDS - Bug: buttons in a button group in a form display issue
### Describe the bug

When a button is inside a button group that inside a form, the button height becomes too tall.

1. Create a form element with the `usa-form` class
2. Add a button group as a child of the form element with the `usa-button-group` and `usa-button-group__item` classes
3. View buttons

### Expected Behavior

Buttons are the correct height for their content as in previous versions (less than 3.8.0)

### Related code

```
<form class="usa-form">
    Form Content
    
    <ul class="usa-button-group">
       <li class="usa-button-group__item">
           <button
               type="button"
               class="usa-button usa-button--outline">
               Back
           </button>
       </li>
       <li class="usa-button-group__item">
           <button
               type="submit"
               class="usa-button">
               Continue
           </button>
       </li>
    </ul>
</form>
```

### Issue
image
### Expected
image

### System setup

- USWDS version: 3.8.0

### Additional context

_No response_

### Code of Conduct

olsonap:
# Summary

**Fixed margin issues for button groups within usa-form.** Ensured margins for button(s) within form are applied to either the usa-button-group or the usa-button, not both.

## Breaking change

Not a breaking change.

## Related issue

Closes [#5883](https://github.com/uswds/uswds/issues/5883)

## Related pull requests

_None_

## Preview link

_Not available_

## Problem statement

It's clear the intended spacing between the form content and the button(s) is 1 unit on small screens and 3 units at breakpoint mobile-lg.
This rule was applied to .usa-button but didn't handle the case where .usa-button was within .usa-button-group.
As a result, the buttons within the button group became stretched.
Without the fix, the buttons are noticeably oversized and users will question the reliability of the website.

## Solution

First, I've modified the current rule in .usa-form to apply only when the .usa-button is not within a .usa-button-group. In the case where it is within .usa-button-group, no additional margin should be applied to the button.
Second, I've added a separate rule in .usa-form to apply a margin to .usa-button-group. This margin is smaller than the respective rules for .usa-button because .usa-button-group already applies a .5 units of margin to its .usa-buttons independent of the form.
As a result of these changes, the following is always true:
1. There is 1 unit of space between form content and the button(s) by default.
2. At breakpoint mobile-lg, the space between form content and the button(s) increases to 3 units.

I've also added tests/stories to inform the intent of these rules and help prevent regression.

Alternative:
```
.usa-button, .usa-button-group {
  margin-top: units(1);

  @include at-media("mobile-lg") {
    margin-top: units(3);
  }
}

.usa-button-group .usa-button {
  margin-top: 0;
}
```
Something like this would have been an alternative but I feel it's more fragile.
The :not(:has(.usa-button-group)) approach scopes the rule to only the context where it actually applies. Each stylesheet remains responsible for its own component's spacing, and no style needs to be "undone" by another.

## Major changes

_Not applicable_

## Testing and review

Clone the repo, checkout this branch, execute `npm install` && `npm start`.
Open Storybook in the browser and navigate to the Button and Button Group components.
Both Button and Button Group have new stories called "Test Within Form".
Adjust the window size to demonstrate the difference in spacing on small screens vs large screens.
Confirm the stories under Patterns/Forms are unaffected by this change.

Here are screenshots to show it's resolved:
image
image

Not negatively impacted:
image


packages/usa-form/src/styles/_usa-form.scss
```diff
@@ -24,13 +24,23 @@
     max-width: none;
   }
 
-  .usa-button {
+  // Intended space between the form content and the button(s) is 1 unit.
+  // _usa-button-group.scss already sets .5 units of margin, so only
+  // .5 units are added here to total 1 unit.
+  &:not(:has(.usa-button-group)) .usa-button {
```
heymatthenry, line 30: Unless I misunderstand, I think this does a little more than it's intended to do. The `:has` will apply to the whole form, so once that rule applies it'll hit any `.usa-button`s in that form whether or not that particular button is in a group or not.

packages/usa-form/src/styles/_usa-form.scss
```diff
@@ -24,13 +24,23 @@
...
     margin-top: units(1);
 
     @include at-media("mobile-lg") {
       margin-top: units(3);
     }
   }
+  .usa-button-group {
```
heymatthenry, line 37: I think this might need a `:where(.usa-button-group)` to keep the specificity pattern established elsewhere?

packages/usa-form/src/styles/_usa-form.scss
heymatthenry: This also increases the specificity so it might have unintended consequences for users who have made their own tweaks to this (admittedly probably a small group, but it's arguably a breaking change).

heymatthenry (changes requested):
Thank you @olsonap! A few things to consider in here

ryparker:
@coderabbitai full review

ryparker:
At `8613772f`, the spacing fix affects buttons outside button groups too. The [new selector](https://github.com/uswds/uswds/blob/8613772fa8bca5f73343f2b54f7ea896da37227c/packages/usa-form/src/styles/_usa-form.scss#L30) removes the existing form margin from every standalone button whenever any button group appears elsewhere in the form.

It also raises specificity from `(0,2,0)` to `(0,3,0)` because `:not(:has(.usa-button-group))` contributes the group class. A later `.consumer-form .usa-button { margin-top: 0; }` override previously won and now loses. Those changes are why this has the breaking-change label.

Please limit the spacing exception to buttons inside a button group and preserve the previous specificity, potentially with a zero-specificity condition. Tests should include a form with both grouped and standalone buttons and a later custom override. That would address the stretching problem without changing unrelated buttons.

PR #6447, opened 2025-03-20
USWDS - Storybook: Update storybook to v8

anmazz:
# Summary

Updates Storybook to version 8.6.7 (latest non-beta version as of Mar. 20, 2025).

## Breaking change

:warning: This is potentially a breaking change.

## Related issue
Closes #5239 

## Problem statement
Storybook was using an old version and can be updated to support the latest, version 8.

## Solution
I followed the automatic migration as well as the migration guide here: https://storybook.js.org/docs/9/migration-guide
I tried maintaining as much existing code as possible. I believe, for instance, that the `"../packages/**/*.stories.mdx",` could be removed from `main.js`, but kept it in for now pending comment.

## Major changes
- Storybook upgraded to version 8.6.7
- Scripts updated to reflect new storybook script syntax
- Adds explicit framework `@storybook/html-webpack5` and dependencies

## Testing and review
Run storybook and verify that storybook version has been upgraded and works as usual. 

## Dependency updates

| Dependency name              | Previous version | New version |
| ---------------------------- | :--------------: | :---------: |
| @chanzuckerberg/axe-storybook-testing |     6.3.1      |   8.2.2   |
| @storybook/addon-a11y | 6.5.16 | 8.6.7 |
| @storybook/addon-essentials | 6.5.16 | 8.6.7 |
| @storybook/addon-links | 6.5.16 | 8.6.7 |
| @storybook/html | 6.5.16 | 8.6.7 |
| @storybook/html-webpack5     |        --        |   8.6.7   |
| @storybook/addon-webpack5-compiler-babel | -- | 3.0.5 |
| @storybook/builder-webpack5 |     3.0.5    |     --      |
| @storybook/manager-webpack5 |     6.5.16    |     --      |

PR #6693, opened 2026-06-05
USWDS - Storybook: Converted deprecated argTypes defaultValues to args attributes

olsonap:
# Summary

**Fixed story default values following upgrade from Storybook 6 to 7.** Converted the deprecated argTypes.defaultValues to args attributes to align with Storybook 7 functionality.

## Breaking change

This is not a breaking change.

## Related issue

Closes [#6692](https://github.com/uswds/uswds/issues/6692)

## Related pull requests

_Not applicable_

## Preview link

_Not available_

## Problem statement

Many stories are meant to show a non-standard state like "disabled" or "error_state", etc.
Previously, developers would use the following syntax:
```
<story name>.argTypes = {
  <control name>: {
    defaultValue: <control value>,
  },
}
```
This syntax, and more specifically the defaultValue attribute is deprecated in Storybook 7 so the stories are no longer rendering with the control in effect.
This makes it particularly confusing navigating Storybook, especially in cases where the control was disabled while also setting the defaultValue. For example, there's no way to toggle the error state on the Test Error Form Elements story despite the control being defined.

## Solution

All variations of this:
```
<story name>.argTypes = {
  <control name>: {
    defaultValue: <control value>,
  },
}
```
...have been converted to some variation of this:
```
<story name>.args = {
  <control name>: <control value>,
}
```

## Major changes

_Not applicable_

## Testing and review

Clone the repo, checkout this branch, execute npm install && npm start.
Open Storybook in the browser and navigate to the various component stories touched by this PR.
In a separate tab, compare the same stories using the Storybook preview [here](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/patterns-forms--test-error-form-elements&globals=backgrounds.grid:!false;backgrounds.value:!hex(F8F8F8)).
Confirm the stories render in their appropriate state. The titles should hint at their behavior.
There are many components impacted, so if asked, I'll provide the itemized list of stories affected by this.

Screenshots of the current state are provided in the GH issue.
Here is evidence of resolution:
image
image
image
image

PR #6697, opened 2026-06-05
USWDS - Storybook: Update Storybook to v8.6

### LINKED ISSUE #5239: USWDS: Update storybook to 7.0.x

ericsorenson:
# Summary

**Updated Storybook from 7.6.24 to 8.6.18.** Upgraded all Storybook packages to the latest stable v8 release, upgraded React to 18.x (required by Storybook 8's manager UI), migrated deprecated `argTypes.defaultValue` to `args` across 12 story files, removed unused `@storybook/addon-links`, and cleaned up deprecated config options.

## Breaking change

## Related issue

closes #5239

Supersedes #6447 — this PR goes further by upgrading to 8.6.18 (vs 8.6.7), migrating all deprecated `argTypes.defaultValue` usages, removing dead config, and fixing the `options` placement in Button stories.

## Problem statement

Storybook 7.x is no longer receiving feature updates and has known security vulnerabilities in its dependency tree. The project needed to upgrade to stay secure and access the latest Storybook features. Additionally, 38 stories used the deprecated `argTypes.defaultValue` pattern which silently breaks in Storybook 8 (controls render with `undefined` defaults).

## Solution

- Upgraded all `@storybook/*` packages from 7.6.24 to 8.6.18
- Upgraded `react` and `react-dom` to 18.x (Storybook 8 requirement for its internal UI — not used in component code)
- Upgraded `@chanzuckerberg/axe-storybook-testing` from 7.2.0 to 8.2.3
- Removed `@storybook/addon-links` (confirmed unused — zero imports in any story file)
- Removed deprecated `argTypesRegex` from `preview.js`
- Removed dead `.stories.mdx` glob and `docs.autodocs` from `main.js`
- Migrated 38 `argTypes.defaultValue` usages to `args` across 12 story files
- Fixed `options` placement inside `control` object in Button stories (deprecated in SB8)
- Cleaned up dead `#root` CSS selectors in `preview-head.html` (Storybook uses `#storybook-root` since v7)
- Bump Node.js to 24.16.0 (current LTS)

## Dependency updates

| Dependency | Previous | New |
|-----------|----------|-----|
| `storybook` | 7.6.24 | 8.6.18 |
| `@storybook/html-webpack5` | 7.6.24 | 8.6.18 |
| `@storybook/addon-a11y` | 7.6.24 | 8.6.18 |
| `@storybook/addon-essentials` | 7.6.24 | 8.6.18 |
| `@storybook/addon-links` | 7.6.24 | **Removed** |
| `@chanzuckerberg/axe-storybook-testing` | 7.2.0 | 8.2.3 |
| `react` | 17.0.2 | 18.3.1 |
| `react-dom` | 17.0.2 | 18.3.1 |

## How to test

1. `npm install`
2. `npm run start:storybook` — verify Storybook launches at localhost:6006
3. Spot-check these components:
   - **Accordion**: expand/collapse panels (tests JS decorator timing)
   - **Date Picker > Default Date**: verify default date "03/06/1995" appears (tests `args` migration)
   - **Button**: open Controls panel, enable "Add icon", verify Icon Name dropdown shows options (tests `options` fix)
   - **Range**: verify min=0, max=100, step=10 defaults (tests `args` migration)
   - **Modal**: open/close, verify focus trap
   - **Banner**: renders USA banner correctly
4. `npm run build:storybook` — verify production build succeeds
5. `npm run test:a11y` — verify accessibility tests pass

## Known technical debt (out of scope, tracked separately)

- Storybook 9 will drop `@storybook/html-webpack5` in favor of `@storybook/html-vite` — a separate Vite migration will be needed
- Stories use CSF 2 (`.bind({})` pattern) — still supported in SB8, no migration needed now
- No explicit compiler addon added — SB8's built-in fallback works; add `@storybook/addon-webpack5-compiler-babel` before SB9


packages/usa-radio/src/usa-radio.stories.js
```diff
@@ -3,6 +3,9 @@ import Tile from "./usa-radio--tile.twig";
 
 export default {
   title: "Components/Form Inputs/Radio",
+  args: {
+    indeterminate_state: false,
```
ethangardner, line 8: Let's add `disabled_state: "none"`

packages/usa-checkbox/src/usa-checkbox.stories.js
ethangardner: Should this property be updated to `indeterminate_state`?

.storybook/preview.js
```diff
@@ -4,7 +4,10 @@ import "../packages/uswds-core/src/js/start";
 /** @type { import('@storybook/html').Preview } */
 const preview = {
   parameters: {
-    actions: { argTypesRegex: "^on[A-Z].*" },
```
ethangardner, line 7: I noticed the PR indicates this parameter is deprecated, but I wasn't able to find deprecation info in the source docs. I still see it in the Storybook documentation https://storybook.js.org/docs/essentials/actions#argtypesregex. The docs note there are limitations to the approach, but it appears it is still valid. Is there a source you can point me to or should this be restored?

ethangardner (changes requested):
@ericsorenson - There were a few very minor things and questions. Otherwise, this looks great. Thanks for the contribution.

olsonap:
Can I suggest using PR #6693 which predates this PR and includes many of these changes?
The deprecation of argTypes.defaultValue happened in the switch from Storybook 6 -> 7, not 7 -> 8.
PR #6693 would handle the deprecation from 6->7 and this PR can be focused on the 7->8 upgrade.

packages/usa-radio/src/usa-radio.stories.js
ericsorenson: Updated!

packages/usa-checkbox/src/usa-checkbox.stories.js
ericsorenson: Yes, good catch, updated!

ethangardner:
@olsonap - Fair point. One of the challenges we're facing right now is managing overlapping PR submissions. I'm doing my best to review and offer feedback on changes that move the design system forward.

In this case, I had a direct conversation with the submitter of this PR who agreed to take this work on, which is why I jumped on this review when I saw it come in. I acknowledge that the scope between upgrading to v8.6 and fixing issues from our recent v7 upgrade got a bit nebulous. I can definitely see how both interpretations were valid.

At this point, I've reviewed this PR, which is very close to merge-able, and I regrettably haven't started on a review for the one you submitted. I see that you've been submitting PRs for other issues both in this repo and on `uswds-site` that I owe you reviews on. I'll try to be better at leaving comments to acknowledge the time and effort that people put forth to submit PRs, even if the PRs can't be reviewed right away. The contributions and support we're getting truly is appreciated. Thanks for understanding.

.storybook/preview.js
ericsorenson: You're right, argTypesRegex isn't formally deprecated. It's still documented and functional. The automigrate tool flagged it as a "manual migration" because the fn() approach is now recommended, but it's not a removal.

That said, it has no practical effect in this project, twig-rendered HTML stories don't emit on* event args through Storybook's action system, so the regex never matches anything. I removed it as dead code rather than for deprecation reasons. Happy to restore it if you'd prefer to keep it for future use.

ethangardner: @ericsorenson - I looked into this more. I agree with your assessment that it's not used. It appears that removing is fine to do in our case.

ethangardner (approved):
Good to go! Thanks for the contribution!

PR #6703, opened 2026-06-12
USWDS - Modal: Fix for content reading twice in screen readers

### LINKED ISSUE #6427: USWDS - Bug: Modal content is read twice for screenreaders
### Describe the bug

For screenreaders, Modals are reading their content twice irregardless of configuration.

1. Open a modal with a screenreader, such as Narrator for windows.
2. Note that the modal and its content gets announced first, then its content is re-read.

### Expected Behavior

The modal should not read the same content twice.

### Related code

You can replicate this issue on [USWDS' modal docs](https://designsystem.digital.gov/components/modal/) under component preview

### Screenshots

_No response_

### System setup

USWDS version: 3.8.1
Device: Laptop
Operating system: Windows
Browser: Chrome 133
(note: this issue was replicated on MacOS)

### Additional context

[This issue was originally identified on the .gov registrar on MacOS](https://github.com/cisagov/manage.get.gov/issues/1612). 

### Code of Conduct

jonathanbobel:
USWDS - Modal: Fix for content reading twice in screen readers

# Summary
When a modal opened, the focus trap and modal both moved focus programmatically, and focus landed on the non-interactive `.usa-modal` wrapper, causing NVDA to repeat the dialog title and scan child content. This change lets the modal own open focus (first footer button by default), disables trap auto-focus, and sets aria-modal on the wrapper so Safari VoiceOver and other AT get clearer modal semantics.

## Breaking change

## Related issue
Closes #6427 

## Problem statement
When a modal opened, screen readers (especially NVDA) often announced the dialog title multiple times. Two things contributed:
1. **Double programmatic focus** - the focus trap auto-focused the first tab stop, then modal code focused the inner `.usa-modal` container (`tabindex="-1"`).
2. **Container focus** - focusing the non-interactive wrapper caused NVDA to scan child content (e.g. “3 items, group”) in addition to the dialog name from `aria-labelledby`.
In Safari + VoiceOver, the title could also fail to announce when only the footer button received focus without a clear modal context.

## Solution
1. **`autoFocus` option on the shared focus trap** (default `true` for other components). Modal passes `{ autoFocus: false }` so the trap only handles Tab cycling; the modal controls initial focus.
2. **Open focus on the first footer button** (fallback: `[data-focus]`, then any `button`) instead of the `.usa-modal` container — aligned with action-oriented confirmation modals.
3. **`aria-modal` lifecycle on the dialog wrapper** - `aria-hidden="true"` when closed; on open, remove `aria-hidden` and set `aria-modal="true"` so assistive technologies treat the dialog as modal and expose the name from `aria-labelledby`.

## Testing and review
- [ ] Open modal — focus on first footer button
- [ ] Tab / Shift+Tab stay inside dialog
- [ ] Escape closes non-forced modals
- [ ] Focus returns to opener on close
- [ ] NVDA + Chrome/Edge: dialog title announced on open, announces header later in relation to focusable element(s)
- [ ] Safari + VoiceOver: dialog title announced on open
- [ ] iOS Safari + VoicerOver: first button announced on open, user can scan modal
- [ ] Unit tests pass (`packages/usa-modal/src/test/modal.spec.js`)

## Known Limitations
This work was tested using NVDA + Chrome and Edge, and VoiceOver + Safari, Chrome, and iOS Safari. While there are variances in experience depending on the assistive technology and browser combination, one caveat is worth noting: in iOS Safari, only the first focusable element is read on modal open, rather than the header followed by the first focusable element. Users will need to use iOS VoiceOver's navigation strategies to explore the modal and reach the header. This is a trade-off from fixing the original bug, where headers were read multiple times. The modal still follows [WAI-ARIA APG guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#examples), follows their modal example, and the modal's semantic HTML labeling is correct throughout.

---
### WAI-ARIA APG
- [Dialog (Modal) Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) — focus moves into the dialog on open; `aria-modal` marks background inert
- [Modal Dialog Example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/) — `aria-labelledby`, `aria-modal="true"`, initial focus on a primary control where appropriate
### WCAG 2.2
- [2.4.3 Focus Order (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html) — open focus moves into the dialog in a meaningful order (footer action, not a hidden wrapper)
- [2.4.7 Focus Visible (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html) — focus lands on a visible, focusable control
- [4.1.2 Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html) — dialog exposes name via `aria-labelledby`, role via `role="dialog"`, modal state via `aria-modal`
---


packages/usa-modal/src/index.js
```diff
@@ -141,39 +139,57 @@ function toggleModal(event) {
   }
 
   // Handle the focus actions
-  if (safeActive && openFocusEl) {
-    // The modal window is opened. Focus is set to close button.
+  if (safeActive) {
+    const openFocusEl =
```
ethangardner, line 143: Regarding test coverage for this, I see a test for the 2nd condition of this assignment at https://github.com/uswds/uswds/pull/6703/changes#diff-dbcd78c0e686fd34637d48ec60d2ee83e50b8b91dee5ab7bf34e9a93dcdb9794R113. I think it would be good to  add test cases for:

- Custom data-focus element taking priority over default (condition 1)
- Fallback to any button when no footer button exists (condition 3)

packages/usa-modal/src/index.js
```diff
@@ -141,39 +139,57 @@ function toggleModal(event) {
...
         Escape: onMenuClose,
       });
     }
 
     // Handles focus setting and interactions
     modal.focusTrap.update(safeActive);
-    openFocusEl.focus();
+
+    if (openFocusEl) {
+      openFocusEl.focus();
```
ethangardner: The focus logic here could use optional chaining:

```js
// Current
if (openFocusEl) {
  openFocusEl.focus();
}
```

```js
// Suggestion
openFocusEl?.focus();
```

This keeps the defensive focus handling more concise and readable in my opinion. This is far from being a blocker, but it's worth considering if you're touching this code again.

packages/usa-modal/src/index.js
```diff
@@ -141,39 +139,57 @@ function toggleModal(event) {
...
 
   // Handle the focus actions
-  if (safeActive && openFocusEl) {
-    // The modal window is opened. Focus is set to close button.
+  if (safeActive) {
+    const openFocusEl =
+      customFocusEl ||
+      targetModal.querySelector(".usa-modal__footer button:not([disabled])") ||
+      targetModal.querySelector("button:not([disabled])");
+
```
ethangardner, line 147: Before, the code in the conditional beginning at L142 did not executed if the `openFocusEl` was falsy. Just doublechecking that the rest of the code should execute if `openFocusEl` did does not exist. I'm wondering if there should be an early return statement here.

This would also impact my advice if the early return is added regarding test coverage for `openFocusEl`

packages/usa-modal/src/test/modal.spec.js
```diff
@@ -106,8 +110,19 @@ tests.forEach(({ name, selector: containerSelector }) => {
         assert.strictEqual(isVisible(modalWrapper), true);
       });
 
-      it("focuses the modal window when opened", () => {
-        assert.strictEqual(document.activeElement, modalWindow);
+      it("focuses the first footer button when opened", () => {
```
ethangardner, line 113: See comment in `packages/usa-modal/src/index.js` related to additional tests.

ethangardner (changes requested):
This seems like a great start! I have a question regarding a behavior change I noticed in the `index.js` file for the modal, and a few additional tests.

Don't take any of this for absolute. It's all open for discussion if you disagree with any of it or have feedback. 

I'm also curious what your thoughts are about adding test coverage for the Language Selector component that also relies on the `focus-trap` utility. I think it would be good to have a test to verify backward compatibility on it since the focus trap could have an impact there now or in the future.

jonathanbobel:
> This seems like a great start! I have a question regarding a behavior change I noticed in the `index.js` file for the modal, and a few additional tests.
> 
> Don't take any of this for absolute. It's all open for discussion if you disagree with any of it or have feedback.
> 
> I'm also curious what your thoughts are about adding test coverage for the Language Selector component that also relies on the `focus-trap` utility. I think it would be good to have a test to verify backward compatibility on it since the focus trap could have an impact there now or in the future.

Added tests for focus-trap.js to make sure the default behavior still works (autoFocus: true) and that autoFocus: false does what we want for the modal. Also checked that Escape and Tab still work. Modal-specific stuff is already in modal.spec.js. I didn’t add Language Selector tests because that component creates a focus trap but never actually turns it on.

packages/usa-modal/src/index.js
jonathanbobel: Solid question. 

The old code only ran the trap and background hiding when openFocusEl existed, but that was mostly hidden because it always fell back to` .usa-modal`. I split those: opening always sets up aria-modal, the trap, and background hiding. Only the initial focus is optional now `(openFocusEl?.focus())` after your previous recommendation. 

An early return here would leave a visible modal with no trap, so I kept it as is.

ethangardner (approved):
LGTM! @chandracarney do you want to take a look at this for final approval?

chandracarney:
> LGTM! @chandracarney do you want to take a look at this for final approval?

Thanks! We'll just need to conduct the [final accessibility checks](https://docs.google.com/spreadsheets/d/1fAJaObQHv7aIaeUpH4fTLmauhmugMtNCrbpZe4BJcPY/edit?gid=97481116#gid=97481116&range=A358) (internal link 🔒) against the modal component to make sure no accidental breakage happened. 🥳

chandracarney (approved):
Excited to get this change in, thanks for all your work to lay out the changes and the side effect to iOS VO + Safari (minor in comparison with the current issues) 🚀

PR #6713, opened 2026-06-16
Fix Escape handling in language selector

### LINKED ISSUE #6496: USWDS - Bug: Language Selector calls nonexistent method
### Describe the bug

The `onLanguageClose` function for usa-language-selector has the following code:
```
const onLanguageClose = () =>
  languageSelector.toggleLanguage.call(languageSelector, false);
```
However, there is no toggleLanguage function.

This line was added in commit 2d5a382855f6a537c9744446b229818c8c267a8b, which also added the toggleLanguage function, which was subsequently removed.

I suspect that this should be replaced by the following:
```
const onLanguageClose = () => toggle(languageSelector, false);
```

1. Create a language selector with at least 3 languages.
2. When the selector is open, hit the Escape key.
3. Observe error in the console.

### Expected Behavior

No JS error should occur.

### Related code

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

arpitjain099:
Fixes #6496

**Language selector no longer calls a removed method on Escape.** This replaces the stale toggleLanguage call in the focus trap Escape handler with the same dropdown-close path used elsewhere in the component.

I also added a regression test that dispatches Escape from a submenu link and asserts there is no throw and the menu closes.

How to test:
- Open a language selector with a submenu.
- Open the menu and focus a submenu link.
- Press Escape and confirm no console error occurs and the menu closes.

Verification run locally:
- npx eslint packages/usa-language-selector/src/index.js packages/usa-language-selector/src/test/language-selector.spec.js
- node jsdom harness for submenu Escape path (no throw + aria-expanded false)


packages/usa-language-selector/src/test/language-selector.spec.js
```diff
@@ -74,6 +74,12 @@ describe("language selector component", () => {
     assert.strictEqual(languageButton.getAttribute("aria-expanded"), "false");
   });
 
+  it("collapses dropdown when Escape is pressed from a submenu link", () => {
```
ethangardner: Thanks for adding a test, however, I noticed this test passes even on the base branch before this fix, which means it doesn't capture potential regressions. 

Could you update the test so it exercises the FocusTrap Escape callback or captures that the FocusTrap path no longer throws and closes the dropdown? Something like this would work:

```js
it("collapses dropdown via the FocusTrap Escape callback", () => {
  languageButton.click();
  // Activate the FocusTrap so its keydown/Escape binding is attached and
  // fires when Escape is dispatched from within the submenu.
  LanguageSelector.focusTrap.update(true);

  // jsdom swallows errors thrown inside event listeners; use `window.onerror`
  // to surface them so a broken Escape callback fails this test.
  let listenerError = null;
  const prevOnError = window.onerror;
  window.onerror = (_msg, _src, _line, _col, err) => {
    listenerError = err;
    return true;
  };

  try {
    EVENTS.escape(languageLink);
  } finally {
    window.onerror = prevOnError;
  }

  assert.strictEqual(
    listenerError,
    null,
    `FocusTrap Escape callback threw an error: ${listenerError}`,
  );
  assert.strictEqual(languageButton.getAttribute("aria-expanded"), "false");
});
```

ethangardner (changes requested):
Thanks for contributing this fix. The runtime code change itself looks good. Replacing that stale callback with the existing dropdown close path makes sense.

I do have a comment on here regarding an update to the test you added. If you can make that change, I think this will be ready to merge.

packages/usa-language-selector/src/test/language-selector.spec.js
arpitjain099: Good catch, you're right. The old test went through the component's own `keydown` binding on `.usa-language__primary`, which never touches the FocusTrap callback, so it passed on develop too.

Pushed your version in f66c2fc, with one addition: `LanguageSelector.focusTrap.update(false)` in the `finally` block, so the trap's document-level keydown listener doesn't stay attached for the rest of the suite.

Checked it both ways this time. With `packages/usa-language-selector/src/index.js` reverted to develop it fails with `TypeError: Cannot read properties of undefined (reading 'call')` thrown from `onLanguageClose`, and it passes with the fix in place.

ethangardner (approved):
LGTM! Thanks for the contribution.

PR #6714, opened 2026-06-17
Fix toggle lookup to respect component root

### LINKED ISSUE #6244: USWDS - Bug: Toggle utility looks for elements in document, not in user-defined root
### Describe the bug

In the [documentation on JS customization](https://designsystem.digital.gov/documentation/developers/#js-customization), the `on` function is described as accepting a root element with `window.document` given as an example. 
When attempting to use the `usa-banner` element inside a Lit web component, the JS functionality broke because the tags necessary for toggling existed in the ShadowDOM. I assumed that by passing the root of the ShadowDOM to the `on` function in the `banner` module, the Javascript would be aware of the shadowDOM. This was partially true and some of the JS did start to work but the `toggle.js` function started to throw an error saying it could not find `gov-banner-default`. That's because [this line]( https://github.com/uswds/uswds/blob/c512a8d408dddac083216d5c14088e20c58e6633/packages/uswds-core/src/js/utils/toggle.js#L15  ) statically refers to the document in a way that is blind to the ShadowDOM. 

1. Render the `usa-banner` component inside a Lit element
2. Import the `banner` package and call `banner.on(this.renderRoot)` in the `firstUpdated` hook for the LitElement. 
3. See that the banner does not open or close properly when you render the component. 

### Expected Behavior

I would expect that if a different root is passed during the call to `on`  that the `toggle` method would respect that and use that to find the elements to show and hide. 
OR 
I would expect the documentation to be clearer on the limitations of passing in a root. 

### Related code

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

I'm aware that the upcoming web components work will probably make this unnecessary but I might file a PR to fix this at some point. 

### Code of Conduct

arpitjain099:
Hi team - this updates toggle so it resolves aria-controls from the control root node instead of always using document.

That lets banner toggles work when banner.on(root) is used with a shadow root.

I also added a banner regression test that initializes inside a shadow root and verifies open/close behavior.

Closes #6244.


ethangardner (approved):
LGTM. Thanks for the contribution.

PR #6715, opened 2026-06-17
USWDS - Storybook: Upgrade to Storybook 9 and migrate from Webpack to Vite

ericsorenson:
**Upgrades Storybook from 8.6 to 9.1 and migrates the build pipeline from Webpack to Vite.** Storybook 9 removed `@storybook/html-webpack5`, so this migrates the framework to `@storybook/html-vite`. Because USWDS renders Twig templates client-side and ships component JavaScript as CommonJS, two custom Vite plugins were written to preserve existing behavior without modifying component source files.

### What changed

**Framework & dependencies**
- Upgraded `storybook` 8.6 → 9.1.20
- Replaced `@storybook/html-webpack5` with `@storybook/html-vite`
- Removed `@storybook/addon-essentials` (controls, viewport, backgrounds, actions are now built into Storybook 9 core)
- Bumped `@storybook/addon-a11y` and `@storybook/test-runner` (0.21 → 0.23) to SB9-compatible versions
- Removed now-unused Webpack loaders: `css-loader`, `file-loader`, `postcss-loader`, `resolve-url-loader`, `sass-loader`, `style-loader`

**New build tooling (`tasks/`)**
- `vite-plugin-twig.mjs` — compiles `.twig` templates into ESM render functions; resolves `@components`/`@templates` namespace aliases and `{% include %}`, `{% extends %}`, `{% embed %}`, `{% from %}`, `{% import %}` dependencies; cascades HMR through the template dependency graph
- `vite-plugin-uswds-cjs.mjs` — transforms USWDS component CJS (`require`/`module.exports`) into ESM at build time (with source maps via MagicString), avoiding a Rollup commonjs-plugin crash on the circular `uswds-core` dependency graph
- `twig-esm-wrapper.mjs` — provides ESM-compatible named exports for the CJS-only `twig` package in production builds

**Configuration**
- `.storybook/main.js` — rewritten for `html-vite` with a `viteFinal` hook (Twig plugin, CJS plugin, SCSS asset-path overrides, `optimizeDeps`, resolve aliases)
- `.storybook/preview.js` — updated a11y parameter from deprecated `a11y.manual` to `a11y.test.disable`

**Notes / out of scope**
- The separate `build:html` Webpack pipeline (`webpack.twig.config.js`, `tasks/webpack-twig-loader.js`) is unchanged
- `.storybook/main.js` remains CommonJS because Storybook 9 loads config via `esbuild-register`, which cannot evaluate native ESM config files (this is resolved in Storybook 10)
- CSF2 → CSF3 story migration is intentionally deferred to a separate PR to keep this diff focused on the framework migration

### How to Test

1. Check out this branch and run a clean install:
   ```sh
   npm install
   ```
2. **Dev server** — start Storybook and confirm it boots on `html-vite`:
   ```sh
   npm start
   ```
   - Storybook should report `Storybook 9.1.20 for html-vite started`
   - Open `localhost:6006` and verify the sidebar shows all categories: Design Tokens, Components, Patterns, Pages
3. **Verify rendering across template patterns** — open and confirm each renders with full USWDS styling and no console errors:
   - Simple template: **Components → Alert → Default**
   - `{% include %}`: **Components → Header → Default**
   - `{% from %}` macros: **Components → Pagination → Default**
   - `{% embed %}` blocks: **Components → Footer → Big Footer**
4. **Verify interactive JS components** work (toggle/open behavior):
   - **Accordion** (expand/collapse), **Date Picker** (calendar opens), **Combo Box** (dropdown opens), **Modal** (opens on trigger)
5. **HMR** — with the dev server running, edit a `.twig` file and confirm the story updates
6. **Production build** — confirm the static build succeeds:
   ```sh
   npm run build:storybook
   ```
7. **Accessibility tests** — confirm the a11y suite passes against the built output:
   ```sh
   npm run test:a11y
   ```
8. Confirm the separate HTML build is unaffected:
   ```sh
   npm run build:html
   ```


ericsorenson:
> Hi @ericsorenson — thanks for your contribution!
> 
> There is one thing that may be holding this up: for security reasons, USWDS requires that all commits to this repo have a [verified commit](https://github.com/uswds/uswds/blob/develop/CONTRIBUTING.md#setting-up-verified-commits). The following commit(s) on this PR aren't verified:
> 
> - `2dcf683` — updating contributors list
> - `c0aa47e` — fix: add role=presentation to contributors table and update contributor count
> 
> Here's what's needed:
> 
> 1. **Set up signature verification in GitHub.** Use one of the following GitHub guides to set up your verification signature:
>    - [GPG commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)
>    - [SSH commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification)
> 
> 2. **Rebase your commits with your verified signature.** Once your signature is set up, complete the following steps in the terminal to update the verification status of your previous commits:
>    ```sh
>    git checkout feat/storybook-9-upgrade
>    git rebase origin/develop
>    git push --force-with-lease
>    ```
> 
> Please let us know if you have any questions. We appreciate your help here. This comment will update automatically once your commits are verified.
> 

These aren't commits from me?

ethangardner:
@ericsorenson - You're right. I'm not sure what happened there. There's an open issue with the workflow (bot) that created the commits, and it's paused until we get it sorted out. Would you mind rebasing and dropping those commits? You can force push with the `--force-with-lease` flag once those bot commits are dropped. We can pair on it if you'd like to get it figured out together. I'll message you.

ericsorenson:
> @ericsorenson - You're right. I'm not sure what happened there. There's an open issue with the workflow (bot) that created the commits, and it's paused until we get it sorted out. Would you mind rebasing and dropping those commits? You can force push with the `--force-with-lease` flag once those bot commits are dropped. We can pair on it if you'd like to get it figured out together. I'll message you.

This should be good to go.

ethangardner:
Thanks, @ericsorenson. I'm teeing this one up for review. I'll give you input within the next day or two.

tasks/vite-plugin-twig.mjs
```diff
@@ -0,0 +1,300 @@
...
+ * graph that Rollup's commonjs plugin cannot statically analyze (it produces an
+ * empty module, causing "twig is not a function" at runtime). To get reliable
+ * CJS→ESM interop in BOTH dev and production builds, this plugin pre-bundles
+ * twig with esbuild (the same engine Vite's optimizeDeps uses in dev) and serves
+ * it through a virtual module that `.twig` files import.
+ */
+
+import path from "node:path";
+import { createRequire } from "node:module";
+import { buildSync } from "esbuild";
```
ethangardner: Since this is being imported, I'd like it to make sure it's also added to `package.json` as a `devDependency` to make sure we're not relying on a transitive dependency from Storybook.

tasks/vite-plugin-twig.mjs
```diff
@@ -0,0 +1,300 @@
...
+  const twigEntry = require.resolve("twig");
+  const result = buildSync({
+    entryPoints: [twigEntry],
+    bundle: true,
+    format: "esm",
+    platform: "browser",
+    write: false,
+    // twig references Node built-ins in its fs loader; stub them since
+    // Storybook renders templates from inline data, not the filesystem.
+    external: ["fs", "path", "module"],
```
ethangardner: When running this locally with `npm run start:storybook`, I see `Missing fs and path modules. Error: Dynamic require of "fs" is not supported` in the console. Can you explore a little and see if there is a way to clean up the error in a way that works both when running via dev server and when it gets built for production?

tasks/vite-plugin-uswds-cjs.mjs
ethangardner, line 1: Is there a way to get test coverage over this plugin? An idea would be to take some of the common patterns from the `packages/` directory and get them into a test. Particularly with this file mentioning the ReDoS risk, I think it would be a good thing to make sure we're capturing with a test to make sure there are no regressions.

tasks/vite-plugin-twig.mjs
ethangardner, line 1: I'd like to see a small, focused test for this plugin.

.storybook/main.js
```diff
@@ -1,132 +1,68 @@
...
-        type: "javascript/auto",
-        use: {
-          loader: "file-loader",
-          options: {
-            name: "[name].[ext]",
-          },
+      }),
+    );
+
+    // SCSS configuration: override asset paths to use staticDirs-served paths.
```
ethangardner: Question about this. When this is in dev mode, serving from the root directory is reliable. However, when this is built and deployed, it ends up on a preview URL like this: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/develop/?path=/story/components-button-group--test-text-wrapping. When the storybook assets are going to be deployed to a non-root directory, can we expect the paths to resolve correctly or are there other changes that need to be made to support this?

ethangardner (changes requested):
Thanks for the contribution! I appreciated the comments you added throughout. They made it much easier for me to review. I like the general approach here. There were some questions and changes I had in this but this is overall a very solid start.

.storybook/main.js
ericsorenson: Good catch investigating this. I tested it and the deployed preview resolves assets correctly, but digging in surfaced that the $theme-image-path: "/img" override wasn't actually doing what the comment claimed, so I've removed it.

What's happening: the USWDS SCSS defaults ($theme-image-path: "../img", $theme-font-path: "../fonts") emit relative url()s. Vite rebases those relative to the emitted CSS file (_site/assets/), so they resolve to _site/img and _site/fonts — exactly where staticDirs: ["../dist"] serves them. Because the URLs stay relative rather than domain-absolute, they resolve correctly under a non-root deploy path too.

I verified this by serving a fresh build under the exact preview sub-path (/preview/uswds/uswds/develop/) and confirming the compiled CSS, images (../img/...), and fonts (../fonts/...) all return 200.

I also confirmed the /img override was a no-op: rebuilding with it removed produced byte-identical asset URLs (50 unique url()s, clean diff). So I dropped it and rewrote the comment to accurately explain the relative-path behavior. Less config, and no misleading rationale for the next reader.

tasks/vite-plugin-twig.mjs
ericsorenson: Fixed, and it's now clean in both dev and production. Thanks for flagging, the message was harmless (twig catches it and degrades gracefully) but it shouldn't be there.

Root cause: the runtime bundle marked fs, path, and module as external, which left require("fs") calls in the ESM output. esbuild's ESM interop turns those into a __require shim that throws Dynamic require of "fs" is not supported. twig's optional filesystem loader wraps the requires in a try/catch and logs Missing fs and path modules. ... on every load, that's the console message you saw.

Storybook renders from inline template data, so twig's fs loader is never actually used. Instead of marking the built-ins external, I added a small esbuild plugin that stubs fs/path/module to export default undefined. Now twig's try block succeeds silently, and its own if (!fs || !path) guard still correctly rejects any real filesystem load — so behavior is unchanged, just without the spurious warning.

One implementation note: esbuild plugins aren't allowed in the synchronous buildSync API, so bundleTwigRuntime() is now async (build()), the virtual-module load hook awaits it, and I cache the bundling promise rather than the resolved string so concurrent loads dedupe onto a single esbuild run.

tasks/vite-plugin-uswds-cjs.mjs
ericsorenson: Added test coverage for this plugin. New spec at tasks/vite-plugin-uswds-cjs.spec.mjs with 16 tests driving the transform hook directly:

    require() imports — default require, single-line destructure, rename syntax ({ prefix: PREFIX }), multi-line destructure collapsing to one line, multiple destructures getting unique _modN vars, and a guard that an object-literal assignment (const obj = { a: 1 }) is not mistaken for a destructure require.
    exports — module.exports = x → export default, exports.name = value → named export, aliased exports.name = other, and the UMD/IIFE wrapper case.
    skip rules — files outside packages/, .stories. files, already-ESM files, and non-.js files.
    ReDoS regression — feeds the exact pathological input from the CodeQL alert (const {{\n + 50k newlines) plus a long unterminated destructure, asserting the transform completes well under a 1s budget. If the nested-quantifier shape ever comes back, these fail fast.

I based the require/export cases on the real shapes in packages/ (e.g. const { CLICK } = require("../events"), const { prefix: PREFIX } = require("../config")), as you suggested.

On wiring: I added a tasksTests gulp task in tasks/test.js and folded it into the existing gulp test series (right after unitTests), so it runs through the same gulp-mocha runner as the component and sass tests and is picked up by both npm test and CI's npm run test:ci. test:tasks mirrors test:sass/test:unit (gulp tasksTests). Full gulp test run is green: 106 sass + 771 unit + 16 tasks passing.

tasks/vite-plugin-twig.mjs
ericsorenson: Done — added "esbuild": "^0.25.12" to devDependencies so we're not relying on the transitive copy from Storybook/Vite. I pinned it to the version they already resolve so npm dedupes onto a single esbuild install (it enforces version matching between its JS wrapper and native binary). Confirmed npm ls esbuild shows one deduped 0.25.12 across all consumers, and the build still passes.

tasks/vite-plugin-twig.mjs
ericsorenson: Added tasks/vite-plugin-twig.spec.mjs — 11 focused tests covering the plugin's core hooks:

    resolveId — the virtual runtime id, @components/@templates namespace aliases, and pass-through of relative/non-twig sources.
    transform — null for non-.twig ids, a plain .twig compiling to an ESM render function that imports the virtual runtime, and {% include %} emitting a dependency import.
    load — the virtual runtime returns bundled twig code, returns null for unrelated ids, and (guarding the fs-loader fix from the other thread) asserts the bundle carries no throwing require("fs") shim.

Runs via the same gulp tasksTests task as the CJS-plugin tests, so it's picked up by both npm test and CI.

tasks/vite-plugin-twig.mjs
```diff
@@ -0,0 +1,334 @@
...
+import { build } from "esbuild";
+
+// twig.js is CJS-only; use createRequire to load it in an ESM context.
+const require = createRequire(import.meta.url);
+const Twig = require("twig");
+
+// Disable Twig's internal template cache. Vite handles caching at the
+// module level; leaving Twig's cache on causes "There is already a template
+// with the ID" errors during HMR.
+Twig.cache(false);
```
ethangardner, line 33: I see this comment, but there's an error with HMR currently. Steps to reproduce:

1) start storybook with `npm run start:storybook`
2) Navigate to a component in the Storybook UI. I used the alert component.
3) Open the browser console.
4) Edit the corresponding `.twig` file for the component you have open in the storybook UI
5) See error below in the browser's console.

```
{
    "message": "There is already a template with the ID packages/usa-alert/src/usa-alert.twig",
    "name": "TwigException",
    "type": "TwigException"
}
```

ethangardner (changes requested):
Thanks, @ericsorenson! I spotted one issue for HMR that I uncovered in testing and added the steps to reproduce that. If you can resolve that, I think we'll be all set here.

Looking at the cjs plugin you wrote, I'm going to suggest a follow up PR once and issue so we can remove the UMD-related checks from the plugin. There's only 1 UMD file in the repo at this point, and I think that'll help us dial down the complexity for things we're adding. I'll create the issue for it and hold off on doing it so you're not dealing with conflicts or things changing while this work is in progress.

tasks/vite-plugin-twig.mjs
ericsorenson: Good catch, this was a real bug, and thanks for the precise repro. Root cause: the Twig.cache(false) I had was running on the plugin's Node-side (build-time) twig instance, but the exception is thrown by the browser runtime, the esbuild-bundled virtual:uswds-twig-runtime, which is a separate copy. Each .twig module registers itself with Twig.twig({ id, data }) when it executes; on HMR, Vite re-executes the changed module against the still-alive browser runtime that already has that id registered, hence "There is already a template with the ID …". My old cache(false) never touched that instance.

Fix: bundleTwigRuntime() now applies cache(false) to the actual instance the browser imports (rebinding esbuild's export default <expr> so we can call it on the exported twig object), and I removed the ineffective build-time call. Verified by running your exact steps (Alert story open, edit usa-alert.twig, 3× edit/revert): no TwigException, component keeps rendering. Confirmed the patch survives into the production chunk (.cache(!1)) too, and added two regression tests (vite-plugin-twig.spec.mjs) — one asserting the bundled runtime disables the cache, one importing the runtime and proving re-registering an id doesn't throw.

ethangardner: I'm still seeing the error with the new changes. Interestingly, the problem didn't happen with playwright mcp but it did happen when I tested it manually.

https://github.com/user-attachments/assets/f3324680-6845-4b18-9b42-d916543af94b

I did some brief troubleshooting, and if you keep the changes you made from this round and add back in the `Twig.cache(false)` you had on line 30, that fixes the problem. It looks like the cache needs to be disabled in both places. Once you do that, I think we'll be all set and can ship this work.

ericsorenson: Good call. I had only patched the browser-side instance but the Node-side one (used by transform() during HMR) also needs cache(false). Added it back, tests pass.

ethangardner (approved):
Thanks for the strong work on this and working through the issues. Thanks for your patience waiting for me to get this reviewed. This is ready to go and looks good to me!

PR #6722, opened 2026-06-21
USWDS - Breadcrumb: Wrap by default instead of truncating

### LINKED ISSUE #6689: USWDS - Bug: Breadcrumbs become truncated and lose meaning when text is enlarged
### Describe the bug

By default, USWDS breadcrumbs are truncated using ellipses rather than wrapping to the next line:

Image

This is frustrating for non-screen reader users, who don't have access to the entire text, but becomes an even bigger issue when a user increases the text size. This example is at 125%, note that more of the copy has been cut off:

Image

Per [this article](https://www.jamesjacobs.me/blog/the-accessibility-pitfalls-of-truncating-text/), this may fail WCAG 1.3.1, 3.3.2, 1.4.4, and other success criteria.

1. Go to https://designsystem.digital.gov/components/breadcrumb/
2. Set your browser width to 1280px
3. In your browser, press Cmd+ (Mac) or Ctrl+ (PC) to increase the size of the text
4. Observe the breadcrumbs in the "Component preview" 

### Expected Behavior

All copy in the breadcrumb should be visible to users of all AT, including magnification, as well as non-AT users. Ideally, the text would wrap to the next line.

### Related code

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

AKnassa:
# Summary

**In one breath:** Long breadcrumb trails used to get cut off with a "…". Now they wrap to the next line, so you can always read the whole path — even when you zoom in or make text bigger.

**Breadcrumbs now wrap by default instead of truncating.** The old single-line, cut-off style is still available as an opt-in `usa-breadcrumb--truncate` variant.

## Breaking change

 This is potentially a breaking change.

Existing breadcrumbs that showed one clipped line will now wrap to more than one line, so the component can get a little taller. Easy to adjust:

- Want the old single-line look back? Add `usa-breadcrumb--truncate`.
- Using `usa-breadcrumb--wrap` today? You can remove it — wrapping is the default now (the class still works, it just doesn't do anything).

## Related issue

Closes #6689

## Related pull requests

A small companion docs update in [uswds-site](https://github.com/uswds/uswds-site) is still needed (the component's variant list). Not opened yet.

## Problem

By default, long breadcrumbs were squeezed onto one line and cut off with a "…". When someone zoomed in or enlarged text, even more of the trail vanished — usually the most useful part (the current page and its parent). The hidden text stays in the page for screen readers, so this mostly hurts sighted people who zoom, magnify, or use voice control.

## Solution

Make wrapping the default. USWDS already shipped this behavior as the `usa-breadcrumb--wrap` variant, so this just promotes it to the default and moves the old truncation behind a new opt-in `usa-breadcrumb--truncate` class. It's CSS only — markup and ARIA are unchanged.

## Testing

- `npm run lint`, `npm run test:sass`, and `prettier` all pass.
- New Storybook `play` tests pass, and axe-core finds **0 violations** on both stories.
- To check by hand: `npm start` → Components → Breadcrumb. Zoom or narrow the **Default** story → it wraps and nothing is cut off. The **Truncate** story still clips, on purpose.


iamjolly (commented):
This looks like really good work @AKnassa, and I appreciate (and agree with) the thinking behind this PR to change the default to be more accessible and usable. 

Since this is a potential breaking change, I want us to be deliberate and thoughtful about introducing this into the system. I don't know if this will cause problems or headaches for many, so I'll look for input from @annepetersen, @chandracarney, @ethangardner, and others before this is fully approved and merged. 

Some questions I'm thinking about are:

- Do we have any numbers on sites using USWDS (this version) that may be impacted?
- How should we communicate this change so that there is clarity for implementors on problems to look for and how to address them?
- If the number of sites using the current version is below a certain threshold, should we publish now with the appropriate caveats/documentation addressed in the previous question?

annepetersen:
Agreed with @iamjolly: thanks for this, and I also don't love truncation. Truncation, as Karen McGrane so memorably puts it, is not a content strategy.

But I also agree with @iamjolly in terms of needing to anticipate possible pain points for teams adopting versions of the design system that don't behave like they used to. In the interest of that: 
  
> * Do we have any numbers on sites using USWDS (this version) that may be impacted?

No firm numbers that I have on hand at the moment, but to broad-brush this, we know that a majority of the federal-level sites using USWDS we can track _that list a version number_ (the vast majority do not, I should note) are using version 2.0 or earlier. (I know, I know: I'd very much like that to be different too.) 

Then keep in mind that not every team will update right away — there seems to be a tendency to batch version changes, from my memory of the research on the subject. _Then_ keep in mind that many teams that _could_ ingest this may be shortstaffed, which speaks both to updating being potentially a low priority, and also the care with which they may look at the changes involved (this item could go either way, to be honest). 

> * How should we communicate this change so that there is clarity for implementors on problems to look for and how to address them?

Great question: in the past USWDS has had more options available. The monthly call was certainly a primary venue, plus the newsletter, plus the [public Slack channel](https://designsystem.digital.gov/about/community/#join-the-community-2), plus of course the release notes. It's possible to announce through some but not all of these, considering current staffing, unless someone in the community wants to step in to help. 

> * If the number of sites using the current version is below a certain threshold, should we publish now with the appropriate caveats/documentation addressed in the previous question?

So just now I pulled up the [metrics that I do have available](https://docs.google.com/spreadsheets/d/1HjphFhSrL-aN8CBWMekVdmixqdYjpUS_7v-M36XAxRQ/edit?gid=132557407#gid=132557407) (that's a protected link, but anyone could fetch the info separately: thank you [Site Scanning](https://github.com/GSA/site-scanning)), and again, big caveat that these are only the sites _that list a version number_, but the grand total of sites tracked there that are using 3.10.0 or later is... **49**. And a solid majority of those are TTS- or GSA-based sites, whose teams are _much_ easier to communicate with.

So, long and short, definitely glad we're considering this together, but in this case I think the stats point toward being in fairly good shape to make this change, as long as there's some careful consideration to communication involved. (Which is, in the end, my responsibility.)

annepetersen (dismissed):
Per [my previous novel of a comment](https://github.com/uswds/uswds/pull/6722#issuecomment-4860801776) :laughing:, I'm prepared to move forward on this one as long as @ethangardner or @jonathanbobel can also give it a thorough review

jonathanbobel:
I think this looks good. I appreciate the `.usa-breadcrumb--truncate` fallback. Approved.

jonathanbobel (dismissed):

annepetersen:
Will just need to resolve these conflicts first 👇 
```
package-lock.json
package.json
```

jonathanbobel:
@annepetersen need your approval one more time since I did the last commit

annepetersen (approved):

PR #6728, opened 2026-06-23
Backfill test coverage for character count fix.

### LINKED ISSUE #6727: Write a test for the fix in pull request #6385
There was a contribution that sat waiting for review for a long time that was just merged. It wasn't fair to ask the original author to write the test after waiting such a long time for review, so this ticket backfills test coverage.

ethangardner:
## Summary

Added test coverage for the fix in PR #6385 that ensures the character count component handles input IDs containing special characters like colons.

## Breaking change

## Related issue

Closes #6727

## Problem statement

PR #6385 fixed a bug where the character count component would fail if an input element ID contained special characters   (like colons) that make CSS selector queries invalid. The fix was merged without test coverage to avoid delaying the   contributor further after a long review wait.

## Solution

Added a test suite that verifies the character count component works correctly when input IDs contain special characters   (specifically colons, which are common in React-generated IDs like :r0:). The tests confirm:

1. The component initializes without throwing errors when the input ID contains colons
2. The label is correctly found via its for attribute and receives the error class when validation fails

### DIFF
--- a/.github/workflows/verify-commit-signatures.yml
+++ b/.github/workflows/verify-commit-signatures.yml
@@ -80,7 +80,7 @@ jobs:
         with:
           header: verify-commits
           message: |
-            ✅ All commits on this PR now have a verified signature. Thanks for making that change!
+            ✅ All commits on this PR have a verified signature. Thanks!
 
       - name: Add label when commits are unverified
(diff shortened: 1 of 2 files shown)


jonathanbobel (approved):
Looks good. Tests pass. 🚀

Issue #6206, opened 2024-11-15
USWDS - Feature: File input support for translated strings


mejiaj:
### Is your feature request related to a problem? Please describe.

File input has hardcoded strings in English, which makes it difficult for localization.

### Describe the solution you'd like

Component should support custom translated strings.

### Additional context

Related to #3445.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [x] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Feature+Request%22) for duplicate feature requests.

amyleadem:
We should keep this issue in mind as we explore solutions in #5616. The strings currently are fairly complex to translate. When we work on #5616, we should see if we can simplify the language structure to be more easily translatable. 

Marking this issue as on hold pending the update in #5616.

mgifford:
Just looking at:
https://designsystem.digital.gov/components/file-input/

Comparing it with UK's GDS:
https://design-system.service.gov.uk/components/file-upload/

But also with what I know about how this works in Drupal.

I think that the easiest way to do this wouldn't be in the design system as such. You would want to extract the language metadata from the file (assuming it is a .docx or .pdf file) and then associate that metadata with the file. Other metadata could be the date, title, file size.

In looking at this with Gemini, maybe creating a `locales.yaml` with

```
en:
  file-input:
    instructions: "Drag file here or choose from folder"
    no-file: "No file selected"
    change-file: "Change file"
  search:
    placeholder: "Search"

es:
  file-input:
    instructions: "Arrastre el archivo aquí o elíjalo de una carpeta"
    no-file: "Ningún archivo seleccionado"
    change-file: "Cambiar archivo"
  search:
    placeholder: "Buscar"

fr:
  file-input:
    instructions: "Glissez le fichier ici ou choisissez dans un dossier"
    no-file: "Aucun fichier sélectionné"
    change-file: "Modifier le fichier"
```



Which could then be called by:

```
<input{{ attributes.addClass(classes) }} 
  data-instructions="{{ 'Drag file here or choose from folder'|t }}"
  data-no-file="{{ 'No file selected'|t }}"
  data-change-file="{{ 'Change file'|t }}"
/>
```

We'd need a uswds-file-input-i18n.js

```
/**
 * USWDS File Input Localization Polyfill
[...]

Issue #5616, opened 2023-11-14
USWDS - File input: Multiple interactions within a single button is confusing.


jaclinec:
# Summary
Having multiple interactions (i.e. "Drag file here" and "Choose from folder") within a single button may be confusing to screen reader users. 

# Observations
One person thought it might make more sense for “Drag file here” and “Choose from folder” to be two separate elements. She thought it was a little confusing that they were combined into one button. "That should split into two different things that you can interact with but it doesn't." She was using Fusion (screen reader + screen magnification) and said that while she herself wasn't confused by it since she has some vision, she thinks it could be confusing to people who rely heavily on feedback they get from screen readers since what its announcing is two separate interactions within one button.

[Video clip](https://drive.google.com/file/d/1rzZXAUprz7p08ko7Ye0CwmkT5SvUZkGj/view?usp=sharing) :lock:

# Affected user groups
- Screen reader users

# Research method
Usability testing with 5 visually impaired participants. See [findings report ](https://github.com/uswds/uswds/wiki/Usability-research-findings-for-Q3-2023)for details.

# Recommended next steps
Consider solutions to make each interaction its own element. For example, Goldman Sachs DS offers three different options for its file upload component. Input only, Input with drag and drop, and drag and drop only (which still appears to be both)

![image](https://github.com/uswds/uswds/assets/109540526/ae24025a-e88d-4025-9d42-13259a3b5104)

alex-hull:
This issue relates directly to an issue with the a11y audit. Adding the audit passes with exceptions label

amyleadem:
Next step:
- Explore if there are low effort solutions that will allow the component to meet accessibility criteria.

amyleadem:
## Possible consideration
We should evaluate our use of `aria-label` in this component. We should see if there is a way to build the component without overriding the `<label>` element.

chandracarney:
This issue is closed via work in #3567.

PR #6738, opened 2026-06-29
Add data-attribute support for custom file-input strings

hasnan-khan:
# Summary

Added optional `data-text-*` attributes to the file input so its instruction strings can be translated, falling back to the English defaults when unset.

## Breaking change

## Related issue

Related to #6206

## Problem statement

The file input renders its instruction strings ("Drag file here or", "choose from folder", "Selected file", "Change file", and the no file status) from hardcoded English in the JavaScript. Since they aren't exposed for overriding, apps serving non-English users can't translate them, so the input stays in English on otherwise translated pages.

## Solution

Added optional `data-text-*` attributes that let a consumer pass translated strings, defaulting to the existing English text when absent. This matches the `dataset.x || default` pattern already used in the codebase (the range component's `data-text-*` attributes and the file input's own `data-errormessage`), so it's fully backward compatible. The custom values also flow through to the `aria-label`.

This PR covers the non counted strings: drag text, choose text, "Selected file", "Change file", and the no file status.

I left the counted strings out ("N files selected" and the "You have selected N files" status). They can't be localized correctly across languages without plural rules support, since the component branches only on singular vs plural, while many languages (Polish, Russian, Arabic) need several plural forms selected by the number itself. That fits better with the larger restructure in #6206 / #5616, so I scoped this to the strings that translate soundly today.

## Major changes

- Added `data-text-drag`, `data-text-choose`, `data-text-change-file`, `data-text-selected-file`, and `data-text-no-file-status` overrides in `index.js`, each falling back to the English default.
- Added the matching conditional attributes to `usa-file-input.twig`.
- Added a `CustomTranslatedString` Storybook story demonstrating the strings in Spanish.
- Added unit tests for each override and the `aria-label`, using the existing `mockMatchMedia` pattern.

## Testing and review

1. Run `npm start` and open Components → Form Inputs → File Input.
2. Open the **Custom Translated String** story. The instructions render in Spanish, and selecting a file shows the translated "Selected file / Change file" text.
3. Confirm the **Default** story still renders the English defaults unchanged.
4. Run `npm test` and confirm all file input tests pass.

### DIFF
--- a/packages/usa-file-input/src/index.js
+++ b/packages/usa-file-input/src/index.js
@@ -191,13 +191,13 @@ const createTargetArea = (fileInputEl) => {
 const updateVisibleInstructions = (fileInputEl, instructions) => {
   const instructionsEl = instructions;
   const itemsLabel = getItemsLabel(fileInputEl);
-  const dragText = `Drag ${itemsLabel} here or`;
+  const dragText = fileInputEl.dataset.textDrag || `Drag ${itemsLabel} here or`;
   const dragTextEl = document.createElement("span");
   const chooseTextEl = document.createElement("span");
   const dragTextIsVisible = shouldShowDragText();
-  const chooseText = dragTextIsVisible
-    ? "choose from folder"
-    : "Choose from folder";
+  const chooseText =
+    fileInputEl.dataset.textChoose ||
+    (dragTextIsVisible ? "choose from folder" : "Choose from folder");
 
   // Create instructions text for aria-label
   DEFAULT_ARIA_LABEL_TEXT = dragTextIsVisible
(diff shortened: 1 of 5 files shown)


ethangardner:
Hi @hasnan-khan - thanks for the contribution. I agree with the need to provide translations for the strings that are baked into some of our interactive components like the file input you worked on here. However, I think the direction we take with USWDS some planning from our team to consider how we might do this in a way that supports all of our components, the future needs of the project and its consumers, and how we might provide a single interface to do so. 

While the `data` attribute might work in some cases, I think we need to have some discussion about pros and cons of this approach and weigh its merits against other approaches like providing i18n hooks and a translation files that might offer different benefits. I'm going to close this PR for now, but I did want to take the time to say thanks for taking the time to work on this. If this is the approach we end up deciding on, we'll reopen the PR and give this a full review.

PR #6746, opened 2026-07-06
USWDS: fix deprecation warnings for global functions

### LINKED ISSUE #6216: USWDS - Sass 3.0.0: `@import` and global built-in functions deprecation

ethangardner:
## Summary

Fixes Sass deprecation warnings for `length()`, `unquote()`, and `unit()` by replacing global built-in function calls with their namespaced module equivalents.

## Breaking change

This is not a breaking change

## Related issue

Closes #6216

## Problem statment

Sass is deprecating global built in functions like `length`, `unquote`, and `unit` and moving them to namespaced equivalents. Output of a sass build shows warnings in the terminal and has caused some confusion with downstream consumers.

## Solution

Replaced deprecated global function calls with their modern equivalent to fix the root cause of the warnings and prepare for future updates.

## Changes

- `rem-to-user-em.scss`: Replace `unit()` with `math.unit()`
- `add-bar.scss`: Remove unnecessary `unquote()` calls on interpolated string properties (interpolation already produces unquoted output)
- `layout-grid.scss`: Replace `unquote()` with `string.unquote()`, replace `length()` with `list.length()`, add `@use "sass:string"`
- `_grid-container.scss`: Replace `length()` with `list.length()`

## Testing and review

Run `npm run test:sass` to confirm tests pass. Run `npm run build` to verify that the only deprecation warnings remaining are releated to the `if` which will be updated in a subsequent PR.

### DIFF
--- a/packages/uswds-core/src/styles/functions/units/rem-to-user-em.scss
+++ b/packages/uswds-core/src/styles/functions/units/rem-to-user-em.scss
@@ -11,7 +11,7 @@ queries
 @use "sass:math";
 
 @function rem-to-user-em($grid-in-rem) {
-  @if unit($grid-in-rem) != "rem" {
+  @if math.unit($grid-in-rem) != "rem" {
     @error 'This value must be in rem';
   }
   $rem-to-user-em: (math.div($grid-in-rem, 1rem)) * 1em;
--- a/packages/uswds-core/src/styles/mixins/general/add-bar.scss
+++ b/packages/uswds-core/src/styles/mixins/general/add-bar.scss
@@ -28,12 +28,12 @@
       bottom: units($offset-y);
       top: units($offset-y);
       width: units($weight);
-      #{unquote($side)}: units($offset-x);
+      #{$side}: units($offset-x);
     } @else {
       height: units($weight);
       left: units($offset-x);
       right: units($offset-x);
-      #{unquote($side)}: units($offset-y);
+      #{$side}: units($offset-y);
 
       @media (forced-colors: active) {
         background-color: ButtonText;
--- a/packages/uswds-core/src/styles/mixins/general/layout-grid.scss
+++ b/packages/uswds-core/src/styles/mixins/general/layout-grid.scss
@@ -1,5 +1,6 @@
 @use "sass:list";
 @use "sass:map";
+@use "sass:string";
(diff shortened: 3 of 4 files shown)


jonathanbobel (approved):
Thanks for this. 🚀

PR #6756, opened 2026-07-13
USWDS - Core: Fix Ctrl modifier not being detected in keymap util

### LINKED ISSUE #6757: USWDS Core - [Bug]: keymap.js uses "Ctrl" instead of "Control" for getModifierState()
## Description

`keymap.js` used `"Ctrl"` as a modifier name in its internal array passed to `KeyboardEvent.getModifierState()`. However, the DOM spec requires `"Control"`. `getModifierState("Ctrl")` always returns `false`, so any key combo using the Control modifier would silently never fire.

The fix is minimal: replace `"Ctrl"` with `"Control"` in the modifier array in `packages/uswds-core/src/js/utils/keymap.js`, and update the corresponding test combos from `"Ctrl+..."` to `"Control+..."`.

I confirmed that there are no components that register a shortcut that uses the control key, so this is solely to prevent a future issue in the event that one is registered in the future. See https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState#syntax. There wasn't anything to reproduce the bug here other than checking the modifier names in the docs.

## Additional information

- `KeyboardEvent.getModifierState()` spec: https://www.w3.org/TR/uievents-key/#keys-modifier
- Fix is in `packages/uswds-core/src/js/utils/keymap.js` (one-line change) and `packages/uswds-core/src/js/utils/test/keymap.spec.js` (combo string updates).

ethangardner:
## Summary

Fixed the `keymap` utility so that `Ctrl`-prefixed key combinations (e.g. `"Ctrl+Tab"`) are correctly detected, and added a regression test to cover it.

## Breaking change

## Related issue

Closes #6757

## Preview link

N/A. This was a silent failure that safeguards us against any future modifier key combos that use the control key.

## Problem statement

`packages/uswds-core/src/js/utils/keymap.js` was rewritten since the last release. It had an array for the modifier keys that had `"Ctrl"`. The correct modifier name per the spec is `"Control"`

## Solution

Updated the keymap to contain the correct modifier key names: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState#syntax

## Testing and review

1. `npm run test:unit` - all tests pass
2. Verify the correct event name is "Control" - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState#syntax
3. No linting or formatting issues exist in the changed code.

### DIFF
--- a/packages/uswds-core/src/js/utils/keymap.js
+++ b/packages/uswds-core/src/js/utils/keymap.js
@@ -17,6 +17,16 @@
  * @param {KeymapConfig} map
  * @return {KeyboardEventHandler}
  */
+// Maps the authoring-facing modifier name used in combo strings to the name recognized by
+// `KeyboardEvent.getModifierState()`. Every name matches except "Ctrl", whose DOM name is
+// "Control".
+const MODIFIER_STATE_NAMES = {
+  Shift: "Shift",
+  Alt: "Alt",
+  Ctrl: "Control",
+  Meta: "Meta",
+};
+
 module.exports = (map) => (event) => {
   // Bail out if this is not a KeyboardEvent (e.g. InputEvent from datalist
   // selection). Only KeyboardEvents have `key` and `getModifierState`.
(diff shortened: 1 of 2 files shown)


jonathanbobel (approved):
Good catch. 🚀

Issue #6752, opened 2026-07-08
USWDS Core - Update sanitizer.js to modern syntax


ethangardner:
To move forward with storybook, we're taking on some additional complexity with the plugins introduced in https://github.com/uswds/uswds/pull/6715. The `cjs` file has regex in it to handle UMD files, however there is only 1 UMD file in the repository, `packages/uswds-core/src/js/utils/sanitizer.js`. Looking at this file's usage, only one of the methods `escapeHTML` is used in the codebase. We have the opportunity to simplify this utility and reduce the complexity of the vite plugin at the same time.

## Acceptance Criteria
- [x] `packages/uswds-core/src/js/utils/sanitizer.js` is converted to commonjs format
- [x] `escapeHTML` continues to be used as a tagged template literal
- [x] Vite plugin code used for storybook is simplified to drop the checks for UMD and associated handling
- [x] Tests are written for escapeHTML

## Out of scope
- Converting JS to ESModule syntax

PR #6760, opened 2026-07-15
USWDS - Core: Convert sanitizer.js to CommonJS and fix falsy-value bug

ethangardner:
# Summary

Modernized the internal `Sanitizer` utility to plain CommonJS, removed its unused `createSafeHTML`/`unwrapSafeHTML` methods, fixed a bug that silently dropped falsy interpolated values (`0`, `false`) from escaped output, and added unit test coverage for `escapeHTML`.

## Breaking change

## Related issue

Addresses #6752

## Related pull requests

#6715 must merge first; a follow-up PR should then simplify `tasks/vite-plugin-uswds-cjs.mjs` to drop its UMD detection now that `sanitizer.js` is the only UMD module and has been converted.

## Preview link

Preview link: https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/task/update-sanitizer-js/?path=/story/design-tokens-fonts--fonts

## Problem statement

`packages/uswds-core/src/js/utils/sanitizer.js` was written as a UMD module wrapped in an IIFE, with `/* eslint-disable */` at the top of the file so it was excluded from linting. It also shipped `createSafeHTML` and `unwrapSafeHTML` helpers that are unused anywhere in the codebase, and had no unit tests. Additionally, `escapeHTML` used `arguments[i + 1] || ""` to default missing interpolated values, which meant falsy-but-valid values like `0` and `false` were silently replaced with an empty string instead of being rendered.

## Solution

Rewrote `sanitizer.js` as a plain CommonJS module (matching how every consumer already imports it via `require`), removed the dead `createSafeHTML`/`unwrapSafeHTML` methods, and re-enabled linting on the file. `escapeHTML` now only substitutes an empty string when a value is `null` or `undefined`, so `0` and `false` are preserved. Added a `sanitizer.spec.js` test suite covering entity escaping, multiple interpolations, falsy values, and `null`/`undefined` handling.

## Major changes

- Converted `sanitizer.js` from a UMD/IIFE pattern to CommonJS (`module.exports = Sanitizer`).
- Removed unused `createSafeHTML` and `unwrapSafeHTML` methods.
- Removed `/* eslint-disable */` so the file is linted like the rest of the codebase.
- Fixed `escapeHTML` to preserve falsy interpolated values (`0`, `false`) instead of dropping them.
- Added `packages/uswds-core/src/js/utils/test/sanitizer.spec.js` with unit tests for `escapeHTML`.
- Simplifies the `cjs` plugin for Vite that's used in Storybook now that the codebase no longer contains UMD-formatted modules

## Testing and review

1. All tests sgould pass in CI. New tests were added to cover the changes.
2. Linting and prettier formatting are clean in CI.
3. Reviewers should confirm no other package relies on the removed `createSafeHTML`/`unwrapSafeHTML` methods (a repo-wide search found none) and spot-check `escapeHTML` call sites (`usa-file-input`, `usa-date-picker`, `usa-table`, `usa-in-page-navigation`, `usa-combo-box`) still render correctly in Storybook with `npm run start:storybook` on your local environment or the [pages preview link](https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/task/update-sanitizer-js/?path=/story/design-tokens-fonts--fonts).

### DIFF
--- a/packages/uswds-core/src/js/utils/sanitizer.js
+++ b/packages/uswds-core/src/js/utils/sanitizer.js
@@ -36,7 +36,10 @@ const Sanitizer = {
     for (let i = 0; i < strings.length; i++) {
       result += strings[i];
       if (i + 1 < arguments.length) {
-        const value = arguments[i + 1] || "";
+        let value = arguments[i + 1];
+        if (value === null || value === undefined) {
+          value = "";
+        }
         result += String(value).replace(Sanitizer._entity, Sanitizer.getEntity);
       }
     }
--- a/packages/uswds-core/src/js/utils/sanitizer.js
+++ b/packages/uswds-core/src/js/utils/sanitizer.js
@@ -1,14 +1,12 @@
 /**
- * A simple library to help you escape HTML using template strings.
+ * A simple library to help you escape HTML using template strings. This originated
+ * from https://web.archive.org/web/20150910040110/https://developer.mozilla.org/en-US/Firefox_OS/Security/Security_Automation
  *
- * It's the counterpart to our eslint "no-unsafe-innerhtml" plugin that helps us
- * avoid unsafe coding practices.
- * A full write-up of the Hows and Whys are documented
- * for developers at
- *  https://developer.mozilla.org/en-US/Firefox_OS/Security/Security_Automation
- * with additional background information and design docs at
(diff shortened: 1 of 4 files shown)


jonathanbobel (approved):
LGTM 🚀

Issue #5852, opened 2024-04-03
USWDS - Character count: Create an MVP for an audible cue when the character limit is exceeded


amyleadem:
## Summary
Implement an audible cue, such as a ding via ARIA, when the character limit is exceeded. It should only ding one time once the first character over the limit is typed, not ding every time they type over the limit. Pair this audible cue with the message "you are x characters over the limit" for screen readers. 

### Details
We should explore the options for immediate audible cues when the character count max is reached. This should work alongside the current "polite" message that explains how many characters over the limit they are. Possible options are a short phrase like "over limit" or a distinct and easy-to-understand sound effect.

amycole501:
My recommendations are ready for review https://docs.google.com/document/d/17R5GUS9K3JuakghiPquTQ4CP8VhFcothU9YsP_ASc64/edit?usp=sharing 

Passing this over to @jaclinec for usability research on which approaches may work best.

jaclinec:
My recommendation for the MVP is to implement a verbal cue instead of the 'ding' per Amy's research. The verbal cue should notify users immediately once the character limit has been reached by announcing "You have reached the character limit" or something similar.

Passing back to engineering @amyleadem and @mahoneycm

chandracarney:
This issue has been closed via [#6767](https://github.com/uswds/uswds/pull/6767) 🎉

PR #6767, opened 2026-07-17
USWDS - Character count: Improve over-limit screen reader announcements

jonathanbobel:
USWDS - Character count: Improve over-limit screen reader announcements

# Summary
Improves screen reader feedback when users exceed the character limit, while preserving production’s debounced announcement order (character echo first, then count status). Adds an assertive alert after typing pauses while over the limit, faster recovery when backspacing under the limit, and defers `setCustomValidity` updates so Safari VoiceOver does not re-read the entire field context on validity changes. Restores deferred `aria-live` initialization (#6541), prevents duplicate status elements on double-init (Storybook + auto-init), and fixes label lookup for input IDs containing special characters.

## Breaking change

## Related issue
Related to #5852

Also preserves fix from #6541 (deferred `aria-live` on init to prevent iOS VoiceOver false announcement on page load).

## Problem statement
**Desired:** When a user exceeds the character limit, screen readers should clearly announce that the limit was exceeded and how many characters they are over, without interrupting character echo or re-reading the full field label/role/hint. When correcting with backspace, users should hear recovery feedback quickly and naturally. Mid-word typing should not trigger premature or stale over-limit announcements.

**Actual (before):** Production used a single 1s debounced polite live region for all count updates. There was no distinct over-limit alert (no assertive “ding”), and the message did not explicitly say “Character limit exceeded.” `setCustomValidity()` ran synchronously on each keystroke; on Safari + VoiceOver, clearing validity when backspacing under the limit often caused VO to re-announce the entire control (“text field… label… hint…”). Label error styling could fail when the input `id` contained characters that break unquoted CSS selectors (e.g. colons). Storybook could double-init the component, appending duplicate status live regions.

**Consequences:** Over-limit feedback was easy to miss or delayed. Fast typing could cancel or replay stale assertive messages. Backspacing to fix over-limit input felt disruptive in VoiceOver. Some implementations (e.g. React-generated IDs) could not associate error styling with the label.

## Solution
Minimal changes on top of production behavior in `packages/usa-character-count/src/index.js`:

Implementation: `data-character-count-over-limit`, `crossedUnderLimit` in `updateCountMessage()`, cancellable debounced `srUpdateStatus()` (assertive when over) vs `srUpdateStatusRecovery()` (fast polite recovery).

**Assertive announcement after typing pauses while over the limit**
While the user is over max length, the debounced SR update (1200ms after last keystroke) uses `aria-live="assertive"` and the message `"Character limit exceeded. N character(s) over limit"`. This waits for a typing pause so mid-word entry (e.g. finishing “biscuits”) does not fire a premature alert on the first over-limit character. Normal typing under the limit remains debounced polite (1200ms), preserving production’s echo-then-count order.

**Faster recovery when backspacing under the limit**
When transitioning from over → under, a separate 100ms debounced polite update announces the recovered count (e.g. `"1 character left"` or `"20 characters allowed"`) instead of waiting the full debounce. Pending over-limit timers are cancelled so stale assertive messages do not replay after backspacing to empty.

**Deferred constraint validation sync**
Clearing `setCustomValidity()` is deferred 100ms via `scheduleValiditySync()` so invalid→valid changes do not fire on the same tick as backspace, reducing Safari VO full field re-reads. Setting invalid when going over remains immediate. Visual error classes still update immediately.

**Deferred `aria-live` on init (#6541)**
`aria-live="polite"` is applied 100ms after the SR status element is inserted, preventing iOS VoiceOver from announcing the default count on page load.

**Idempotent initialization**
`data-enhanced` on the input and a guard in `createStatusMessages()` prevent duplicate `.usa-character-count__status` / `__sr-status` elements when USWDS auto-init and Storybook both call `characterCount.on()`.

**Quoted label selector**
`label[for="${inputEl.getAttribute("id")}"]` correctly resolves labels when IDs contain special characters.

**Why this approach:** Keeps production’s proven debounce timing for normal typing; uses assertive only after a conversational pause while over limit; cancels stale timers on state transitions; avoids immediate live-region updates that caused count-before-echo race conditions in earlier iterations.

## Limitations / not in scope / still TBD

- **Storybook iframe:** “Inside a frame” language may still appear in Storybook preview/new tab; not introduced by this component.
- **Repeated over-limit typing:** Assertive re-fires after each typing pause while still over (with updated count), not on every keystroke.
- **Dynamic count in `aria-describedby`:** Count/error status is still announced via the `.usa-character-count__sr-status` live region, not wired into `aria-describedby` on the input.
- **`setCustomValidity` delay:** Clearing validity syncs ~100ms after keystroke when going under limit; visual invalid styling is immediate. Very fast submit within that window could briefly disagree with validity state (edge case).
- **Shared debounce across inputs:** Over-limit and under-limit debouncers are module-level; multiple character count fields on one page share timers (existing production pattern).
- **Immediate threshold announcements at exactly max (0 left):** Not added; assertive treatment applies only when over the limit.
- **NVDA: Debounced assertive live-region announcements while focused in the field are unreliable (forms mode; matches production):** NVDA may instead surface the field’s invalid state (e.g. “invalid entry”) when tabbing into or out of an over-limit input/textarea via setCustomValidity, rather than reading the “Character limit exceeded…” live-region text. VoiceOver remains the primary verified path for that assertive message after a typing pause.

## Major changes
- **`index.js`**
  - `LIMIT_EXCEEDED_MESSAGE`, `SR_STATUS_DEBOUNCE_MS` (1200), `AT_DEFER_MS` (100)
  - `updateSrStatus()` - sets `aria-live` and message text
  - `createDebounced()` - debounce with `.cancel()` for stale timer cleanup
  - `srUpdateStatus()` - assertive when over limit, polite when under (1200ms debounce)
  - `srUpdateStatusRecovery()` - 100ms polite recovery path
  - `scheduleValiditySync()` - deferred clearing of `setCustomValidity`
  - `updateCountMessage()` - over/under threshold detection (`crossedUnderLimit`); cancels opposing debouncers on transition
  - `createStatusMessages()` - restores deferred `aria-live` (#6541); skips if SR status already exists
  - `enhanceCharacterCount()` - `data-enhanced` guard; initialize over-limit state on init
  - `getMaxLength()` - shared max-length lookup
  - Quoted `label[for="..."]` selector
- **`character-count.spec.js`**
  - Assertive over-limit announcement after debounce
  - Assertive waits for typing pause (rapid over-limit keystrokes → final count)
  - No stale over-limit replay after backspacing to empty
  - Polite under-limit debounced announcements
  - Fast recovery when backspacing under limit
  - Deferred `aria-live` on init (#6541)
  - No duplicate status messages on double init
  - Special-character input ID tests (colon in `id`)

## Testing and review
- [ ] **Safari + VoiceOver:** Type under limit — hear typed characters first, then count status ~1.2s after pause
- [ ] **Safari + VoiceOver:** Type past max (including mid-word), pause ~1.2s — hear assertive ding + `"Character limit exceeded. N character(s) over limit"`
- [ ] **Safari + VoiceOver:** Backspace under limit — character echo first; ~100ms later hear recovered count; no stale over-limit message after clearing field
- [ ] **iOS VoiceOver on load:** No spurious count announcement before user input (#6541)
- [ ] **Storybook:** One `.usa-character-count__status` and one `__sr-status` per field (no duplicates)
- [ ] **Visual feedback:** Red error styling appears immediately when over limit; clears immediately when under
- [ ] **Form validation:** Field blocks submit when over limit (after validity sync)
- [ ] **Input + textarea:** Both stories in Storybook behave consistently
- [ ] **Input ID with colons:** Error class applies to associated label
- [ ] **NVDA:** Live over-limit message may be silent while typing; optional “invalid …” on blur/focus when over limit is expected browser/AT behavior, not the live-region string

### Unit tests pass
- `packages/usa-character-count/src/test/character-count.spec.js`
- `packages/usa-character-count/src/test/valid-template-multiple-validators.spec.js`

## References

**Design system guidance**
- [GOV.UK Design System: Character count](https://design-system.service.gov.uk/components/character-count/): soft limit (does not block typing); separate visual and screen reader count behavior; debounced SR updates
- [Making a character count component more accessible (David Cox)](https://dav-idc.com/making-a-character-count-component-more-accessible/): split static limit, visible count, and debounced SR-only count

**WCAG 2.2**
- [4.1.3 Status Messages (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html): programmatic status for count/limit feedback
- [3.3.1 Error Identification (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/error-identification.html): over-limit state communicated to assistive technology
- [3.3.2 Labels or Instructions (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/labels-or-instructions.html): character limit guidance
- [4.1.2 Name, Role, Value (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html): label association for inputs with non-trivial IDs

**ARIA techniques (not a character-count pattern)**
- [ARIA19: Using ARIA role=alert or Live Regions to Identify Errors](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA19.html): assertive/live region for error/status feedback
- [WAI-ARIA APG: Alert pattern](https://www.w3.org/WAI/ARIA/apg/patterns/alert/): assertive announcements for important status changes


chandracarney (commented):
Thanks for all the detailed work, looking forward to getting this change in! I'll pass it off to @ethangardner for code review.

package.json
ethangardner, line 1: I think this should be removed so we can stick to NPM as our package manager. Switching to something else would require a larger discussion.

packages/usa-character-count/src/test/character-count.spec.js
```diff
@@ -177,6 +193,105 @@ tests.forEach(({ name, selector: containerSelector }) => {
         assert.strictEqual(childNode.nodeType, Node.TEXT_NODE);
       });
     });
+
+    it("announces assertively after debounce when over the limit", (done) => {
```
ethangardner: Thanks for adding these tests. I think we can DRY them out some if you extract the timeout part to a helper. Something along the lines of this might work.

```js
const assertSrAfterDelay = (delay, statusMessageSR, { text, ariaLive }, done) => {
  setTimeout(() => {
    assert.strictEqual(statusMessageSR.textContent, text);
    assert.strictEqual(statusMessageSR.getAttribute("aria-live"), ariaLive);
    done();
  }, delay);
};
```

That will help cut down on some of the repetition in the tests and keep them more concise.

packages/usa-character-count/src/test/valid-template-multiple-validators.spec.js
ethangardner, line 1: I think `AT_DEFER_MS + 50` could be its own constant. It's used in this spec as well as the one for character count.

Also, I'm wondering if for some of the timers in this PR, we might be able to use https://sinonjs.org/concepts/fake-timers/. Sinon is already a project dependency.

packages/usa-character-count/src/index.js
```diff
@@ -119,15 +130,76 @@ const getCountMessage = (currentLength, maxLength) => {
...
+/**
+ * @param {Function} callback
+ * @param {number} delay
+ * @returns {Function & { cancel: () => void }}
  */
-const srUpdateStatus = debounce((msgEl, statusMessage) => {
-  const srStatusMessage = msgEl;
-  srStatusMessage.textContent = statusMessage;
-}, 1000);
+const createDebounced = (callback, delay) => {
```
ethangardner: We have `packages/uswds-core/src/js/utils/debounce.js` in the repo already. It was removed in this PR. Would it be possible to refactor the existing debounce function so we don't inline it here and have to maintain multiple implementations of a debounce function?

ethangardner (changes requested):
Overall, nice work. I have a few suggestions to help with test setup and one question about the new debounce function that was added to the runtime.

jonathanbobel:
@ethangardner just wanted one more quick look at this one

packages/uswds-core/src/js/utils/debounce.js
```diff
@@ -1,18 +1,27 @@
...
  * @param  {number} delay - Milliseconds to wait before calling function
- * @returns {Function} A debounced function
+ * @returns {Function & { cancel: () => void }} A debounced function
  * @example const updateStatus = debounce((string) => console.log(string), 2000)
  */
 
 module.exports = function debounce(callback, delay = 500) {
   let timer = null;
-  return (...args) => {
+  const debounced = (...args) => {
```
ethangardner: I noticed the `apply` method uses `this`, and we're using an arrow function here. The arrow function was there in the code previously, but I think we should fix it now since it would be a one line change.

Using the arrow function is problematic because it would capture the module scope's `this` instead, silently dropping the call-site context passed to `.apply()`.

Changing this line to `const debounced = function debounced(...args) {` should be all we need to correct the bug.

packages/uswds-core/src/js/utils/debounce.js
ethangardner, line 1: I noticed this file doesn't have test coverage currently. I think we should add unit tests for this given the new cancel behavior and bug fix I mentioned on line 16. If you can write test that covers the `debounced`, `.cancel`, and then a regression test for the arrow function, we should be all set here.

ethangardner (changes requested):
Thanks for these changes. There was a minor bug I spotted and some test coverage to add on the `debounce` function. After that, we can get this merged. The rest all looked good to me.

jonathanbobel:
@ethangardner addressed your comments. one last look should do it.

ethangardner (approved):
Killer work! Thanks for addressing the suggestions I made. This is ready for prime time now.

PR #6786, opened 2026-08-05
USWDS - Modal: Restore page content when the opener has left the document

### LINKED ISSUE #6785: Bug: Modal leaves page content aria-hidden after close when its opener is gone
### Describe the bug

Opening a modal sets `aria-hidden="true"` on everything outside it. Closing is supposed to remove that, but the removal is wrapped in `if (menuButton && returnFocus)`.

`returnFocus` is looked up from the modal's `data-opener` attribute. If the opener element isn't in the DOM anymore when the modal closes, that check fails and the attributes are never removed.

So you end up with no modal on screen and the whole page still `aria-hidden`. Screen readers see an empty page, and it stays that way until reload. So opening always hides. Closing only sometimes unhides.

1. Page with some content, a modal, and a `[data-open-modal]` opener.
2. Click the opener. Modal opens, page content gets `aria-hidden="true"`. Correct so far.
3. Remove the opener from the DOM while the modal is open.
4. Close the modal, either with the close button or Escape.
5. Page content still has `aria-hidden="true"` and `data-modal-hidden`.

I tested four variations with Playwright on a static page:

- Opener left alone: restores fine
- Opener removed: leaks
- Clicked opener removed but a second opener still on the page: leaks
- Closed with Escape instead of the button: leaks

The third one is the useful one. `menuButton` was fine there because another opener existed, and it still leaked, so `returnFocus` is what's failing.

Reproduced on published 3.13.0 and on `main`.

### Expected Behavior

Closing a modal should always unhide the page.

Returning focus is a different thing. It can't work if the opener is gone, but that shouldn't stop the page from being readable again.

### Related code

`packages/usa-modal/src/index.js`. Open hides with no conditions:

```js
document.querySelectorAll(NON_MODALS).forEach((nonModal) => {
  nonModal.setAttribute("aria-hidden", "true");
  nonModal.setAttribute(NON_MODAL_HIDDEN_ATTRIBUTE, "");
});
```

Close only unhides inside the guard:

```js
if (menuButton && returnFocus) {
  document.querySelectorAll(NON_MODALS_HIDDEN).forEach((nonModal) => {
    nonModal.removeAttribute("aria-hidden");
    nonModal.removeAttribute(NON_MODAL_HIDDEN_ATTRIBUTE);
  });

  returnFocus.focus();
}
```

Moving the unhide out of the guard and leaving only `returnFocus.focus()` inside it would fix it. I can send a PR with a test if that works for you.

### Screenshots

_No response_

### System setup

_No response_

### Additional context

_No response_

### Code of Conduct

vssinghh:
# Summary

**Closing a modal now always restores screen reader access to the page.** If the element that opened the modal was gone by the time it closed, page content kept `aria-hidden="true"` and stayed invisible to assistive tech until reload.

## Breaking change

## Related issue

Closes #6785

## Problem statement

Opening a modal hides everything outside it from screen readers. Closing should undo that, but the undo sat inside `if (menuButton && returnFocus)`.

`returnFocus` comes from the modal's `data-opener` attribute. If that element isn't in the document anymore, the lookup returns null and the attributes never come off. You get no modal on screen and a page that's still entirely `aria-hidden`, with no way out except a reload.

Opening always hid. Closing only sometimes unhid.

## Solution

Moved the unhide out of the guard so it always runs, and left `returnFocus.focus()` inside it.

Focus return can't work if the opener is gone, but that shouldn't keep the page hidden.

I left the `menuButton` condition on the focus call alone. Dropping it might be fine too, but it would change focus behavior in a case unrelated to this bug. Happy to change it if you'd prefer.

## Testing and review

The test removes the clicked opener while the modal is open, closes it, then checks nothing is left with `data-modal-hidden`. A second opener stays on the page, so `menuButton` is still truthy and only `returnFocus` is missing.

`npx gulp unitTests`: 802 passing. Reverting the fix but keeping the test gives 800 passing, 2 failing, and the only failures are the new test. It runs twice because the spec covers a modal initialized at `document.body` and at the modal element.

I also checked this outside the unit tests, on a static page with Playwright, against published 3.13.0 and `develop`. Opener left alone restores fine, opener removed leaks, Escape leaks the same way. Can share that if useful.


ethangardner (commented):
This looks good to me. 

@jonathanbobel - this touches modal code that you wrote previously. This fixes an issue where the cleanup can still happen and focus is only restored if the opener element was gone from the DOM by the time the modal closed (e.g. programmatic deletion or React re-render).

Passing off to you to see if it makes sense for AT. If it looks good, go ahead and approve and merge. It passes my review.

jonathanbobel (approved):
LGTM

ethangardner:
Thanks for the contribution, @vssinghh!

PR #6789, opened 2026-08-05
USWDS - Accordion: Extending accordion to allow for start/end icon alignment

### LINKED ISSUE #6784: USWDS - Accordion: Introduce a left-justified expand/collapse control

Screen magnification users or manual zoom users don't easily understand an accordion element is interactive if the accordion expand/collapse control icon is on the right, so important information within the accordion element remains hidden for those users. 

While a right-justified control icon impacts low-vision users at a higher rate (as majority of low-vision users use screen magnification or other zoom methods), it also impacts broader user groups:
* Any mobile-web-based and web-based users who are manually zoomed in (using a pinch motion to zoom in) on any webpage
* Any web-based users who are programmatically zoomed in (using a `command` or `control` + `+` key combination) on a page that isn't appropriately reflowing at certain breakpoints

This issue is related to #6483, opened by (and discussed within) our community.

### Additional context
* Left-justified controls are understandable and usable to all user groups (not solely limited to magnification tool users) based on a [VA user research study targeting shifting to a left-justified expand/collapse control](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/teams/digital-experience/ADE/research/2026-01-save-in-progress-2/research-findings.md#key-findings) in February 2026.
* Left-justified controls resolve issues for people using magnification tools, "100% of our screen magnification users were able to identify that the accordion was an interactive element and they could interact with it," based on the same February 2026 VA study ([see comment](https://github.com/uswds/uswds/discussions/6483#discussioncomment-16182289) within #6483).
* [GOV.UK Design System updated to a left-justified control](https://github.com/alphagov/govuk-design-system-backlog/issues/1#issuecomment-995675898), in 2021 after various findings that controls on the left are "easier to see for users of screen magnifiers" and when "icons [were] to the right, people often missed the fact that it was an accordion."
* Since the USWDS accordion component was written, the HTML [details element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details) with related [summary element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/summary) has been released, which provides a left-justified control as default, setting a more standardized way to present accordion-like elements.
* WCAG provides supplemental guidance and techniques to its success criteria similar to the issue here, [positioning labels to maximize predictability of relationships](https://www.w3.org/WAI/WCAG21/Techniques/general/G162).

- [x] Change the primary (default) accordion control to be a left-justified control.
- [x] Allow a right-justified control (prior default) to easily be implemented by USWDS users.
- [x] Preserve right-justified control for the Banner component.

### Code of Conduct

jonathanbobel:
USWDS - Accordion: Extending accordion to allow for start/end icon alignment

# Summary
Extended the expand/collapse icon to default to the start, or "left" side in the accordion button so magnification and zoom users can more easily discover interactive accordion controls, while keeping "end" (right-aligned) icons as an option. 

## Breaking change

## Related issue
Closes #6784

## Related pull requests
Related discussion: Component proposal in #6483 (VA user research on left-justified accordion controls).
Documentation for the new modifier and `$theme-accordion-icon-position` has a follow-up branch

## Problem statement
**Desired:** Accordion expand/collapse controls are discoverable and understandable for users viewing content at high magnification or zoom levels, including screen magnification users and users who pinch-zoom or use programmatic zoom (e.g. Ctrl/Cmd + +).

**Actual:** Accordion expand/collapse icons are fixed on the right/end side of the button. When users zoom in, the control can fall outside the visible viewport or be harder to associate with the accordion heading, so users may not realize the accordion is interactive and important content can remain hidden.

**Consequence:** This disproportionately affects low-vision users who rely on screen magnification, but it also affects anyone zoomed in on mobile or desktop when content does not reflow as expected. VA user research (February 2026, referenced in #6483) found that left/start-justified expand/collapse controls improved discoverability: all screen magnification users in that study could identify the accordion as interactive. Left/start-justified controls were also reported as understandable and usable across broader user groups, not only magnification tool users.

## Solution
This PR adds an easy, supported way to default to the left/start side of the accordion trigger, while keeping the option for a user to set the icon to the right/end.

1. **Theme setting**: $theme-accordion-icon-position (default "start") sets icon placement globally. Use "end" to keep the previous right-side (LTR) look.
2. **Modifier classes**: .`usa-accordion--icon-start` and `.usa-accordion--icon-end` support per-instance placement in markup.
3. **Refactored styles**: Shared mixins use logical properties (padding-inline-*, text-align: start, RTL background-position flips) so start and end variants stay visually and accessibly consistent in LTR and RTL.
4. **Storybook**: Default stories use the start-side icon. Added an IconEnd story and content fixture for the previous end-side placement.

No JavaScript or markup structure changes were required; open/closed icon behavior continues to be driven by existing `aria-expanded` styles.

**Why this approach**: A theme token plus modifier classes matches USWDS patterns and uses logical start / end so placement follows writing direction. Default is now "start" (left in LTR, right in RTL), which matches VA research on left-justified controls. Teams that want the previous look can set $theme-accordion-icon-position: "end" or add .`usa-accordion--icon-end`.

**Limitations**: Placement is limited to "start" and "end". Teams should still verify layout at common zoom levels and with screen magnification tools in their target browsers, including RTL (dir="rtl").

## Major changes
- Added `$theme-accordion-icon-position` to `_settings-components.scss` (default: "start")
- Refactored accordion button padding and background positioning into shared mixins using logical properties
- Updated high contrast mode ::before icon positioning to respect icon side
- Added `.usa-accordion--icon-start` and `.usa-accordion--icon-end` modifiers
- Added `usa-accordion~icon-end.json` content and IconEnd Storybook story
- Set accordion button text-align: start so heading text follows RTL with the icon and body copy

## Testing and review

- [ ] Run npm test and confirm all tests pass

## Manual verification

Run npm start

1. **Default / IconEnd stories**: Confirm default shows the icon on start (left in LTR); IconEnd shows the previous end-side placement with correct spacing
2. **RTL**: Set `dir="rtl"` and confirm heading text, body copy, and icon all flip with start/end
3. **Zoom / magnification**: At 200%+ browser zoom (and with OS screen magnification if available), confirm the expand/collapse icon remains visible and clearly associated with the accordion heading
4. **Interaction**: Expand/collapse panels; confirm open/closed icon states update correctly
5. **Hover and forced-colors**: Verify hover states and high contrast icon placement on both sides
6. **Bordered variant**: Confirm bordered accordion follows the theme default unless a modifier is applied

## Additional notes
Banner and header accordions are unchanged. The “Here’s how you know” control and header navigation keep their existing icon placement.


packages/usa-accordion/src/usa-accordion.stories.js
```diff
@@ -35,6 +40,9 @@ Bordered.args = BorderedContent;
 export const Multiselectable = Template.bind({});
 Multiselectable.args = MultiContent;
 
+export const IconEnd = Template.bind({});
```
ethangardner, line 47: Since users can set a global default to make the icon appear on the right, I think it would be good to also add a story here for IconStart to show how they would execute the per-instance override in that case.

packages/usa-accordion/src/styles/_usa-accordion.scss
```diff
@@ -44,11 +48,40 @@ $accordion-button-open-hc-icon: map.merge(
...
+
+  @if $side == "start" {
+    padding-inline-start: $accordion-icon-padding;
+    padding-inline-end: $accordion-icon-gap;
+    background-position: left $accordion-icon-gap top calc(50% - 1px);
+
+    [dir="rtl"] & {
+      background-position: right $accordion-icon-gap top calc(50% - 1px);
+    }
+  } @else {
```
ethangardner, line 62: In the `@else` branches in this file, I think it would be good to add a `@warn` if the `$side` value is not `end` and that we're executing a fallback. That will help people catch typos in their config.

packages/usa-accordion/src/styles/_usa-accordion.scss
```diff
@@ -196,3 +230,27 @@ $accordion-button-open-hc-icon: map.merge(
 .usa-accordion__button[aria-expanded="false"] {
   @include accordion-button-unopened-styles;
 }
+
+.usa-accordion--icon-start {
```
ethangardner, line 244: The nav and the banner both use `.usa-accordion__button` as well. Let's add a safeguard to prevent a change in icon position if someone tried:

```html
<section class="usa-banner" aria-label="Official website of the United States government">
  <div class="usa-accordion usa-accordion--icon-start">...
  </div>
</section>
```

or 

```html
<nav aria-label="Primary navigation" class="usa-nav">
  <button type="button" class="usa-nav__close">
    <img src="./img/usa-icons/close.svg" role="img" alt="Close">
  </button>
  <ul class="usa-nav__primary usa-accordion usa-accordion--icon-start"><!-- rest of nav -->
```

ethangardner (changes requested):
Thanks for this. I know a lot of research and discussion went into this decision. For documentation purposes, can you update the PR body so it reflects that the new default is icon-start (left) in the "Solution" heading and indicate that this is a breaking change. It still has default right and non-breaking in the PR body.

Since our last release, I added a good deal of test coverage for branching logic in `*.spec.js` files using the `sass-true` package, and I think it would be good to do the same here for the branching logic.

There are also some comments inline with the files as well. The biggest one is the defensive coding to protect against anyone who tries to add a `usa-accordion--icon-start` class to the nav or banner.

packages/usa-accordion/src/styles/_usa-accordion.scss
jonathanbobel: For these two, you would have to strip out a lot of the wrapping classes on parent nodes to "break" the accordion and have the buttons show up on the new default "start" area. I think we're good in it's current implementation.

ethangardner: The concern I had about it was sufficiently addressed with the `:not` selectors you added. I agree we're all set now with no further changes on this.

ethangardner (approved):
Thanks for addressing the feedback. This looks really good! You did great on the handling of the `rtl` treatment as well.

chandracarney (commented):
Thanks for this, looks great! Let's add a short note about excluding banner in the PR comment. 

I added banner exclusion to the main issue #6784 just now, as I noticed it was missing there as well.

PR #6797, opened 2026-08-08
USWDS - Pagination: Normalize button element styles to match links

### LINKED ISSUE #6212: USWDS - Bug: Pagination styles inconsistent when using <button> instead of <a>
### Describe the bug

When using pagination with `<button>` elements, buttons inherit default styles which change the style to be different than when using `<a>` elements

https://stackblitz.com/edit/stackblitz-starters-4adpqy

### Expected Behavior

Pagination styles are the same regardless of `<button>` or `<a>` element

### Related code

_No response_

### Screenshots

Visual differences of pagination component when using button vs link elements

### System setup

USWDS: v3.10.0
Browser: Chrome Version 130.0.6723.117 (Official Build) (arm64)
OS: macOS 13.4.1

### Additional context

_No response_

### Code of Conduct

opensource-joe:
## Summary
Normalizes native `<button>` user-agent styles on pagination items so a `<button>` renders identically to an `<a>`.

## Problem
Pagination items can be authored as `<a>` or `<button>`. `.usa-pagination__link` / `.usa-pagination__button` assumed `<a>`, so `<button>` items inherited native button styling (background, font, appearance, cursor) and looked different.

Reproduction in the issue: https://stackblitz.com/edit/stackblitz-starters-4adpqy

## Solution
Add a small reset (`appearance`, `background-color`, `font-family/size`, `line-height`, `cursor`) to both classes. The active-page `.usa-current` background still wins on specificity/order.

## Testing
Full USWDS Sass compiles cleanly; the normalization emits as expected. No token/theme changes.

Fixes #6212


ethangardner (changes requested):
Before I dive too deep on this, could you add stories to this in our Storybook instance and update the relevant twig template? Reading through the code, this looks like a good start, but I want to make sure we document this pattern and also have a way to verify the work going forward. 

The issue this corresponds to indicates that it needs some consideration from our accessibility specialists, and I want to make sure they have what they need to do their testing, and I want to make sure the `<button>` option is visually identical to the the component when it uses `<a>`

opensource-joe:
Both added, and building the verification turned up something worth flagging: the reset I had was incomplete.

Measuring computed styles on the two elements, the arrows were still 16px wider as buttons, because `.usa-pagination__button` restates its own border and padding but `.usa-pagination__link` does not, so a `<button>` arrow kept the user-agent values. Nothing reset `text-decoration` either, so every button lost the underline. Nine differences in total. Fixed with `border: 0`, `padding: 0` and `text-decoration: underline` on the shared reset.

On the Storybook side there are two stories. `Buttons` is the component with `pagination.element: "button"`, and `TestElementComparison` renders links and buttons one above the other so a difference between them is visible rather than something you have to remember to check. The twig now takes `pagination.element`, defaulting to `a`, with the tag chosen at the top of the macro so the classes and ARIA stay in one place for both elements.

Measured after the change: all 16 computed properties identical for the page numbers, the current page and the arrows, and identical box dimensions.

The SCSS is a separate commit from the template and story work, in case you would rather review them apart.

Links above, buttons below, same content and same page:

Pagination rendered with links and with buttons, identical

ethangardner (approved):
Thanks Joe! I appreciate the changes and the tests you added. The visuals made it really easy to review and verify, so thanks for including the screenshot. I gave this a look and everything is buttoned (get it?) up nicely. Thanks for the contribution.

PR #6798, opened 2026-08-08
USWDS - Table: Match caption text size to table and body text

### LINKED ISSUE #6253: USWDS - Table: Change caption text size to match body and table text size
## Summary
Can you please fix the style of default font size for `<caption>` code within [tables](https://designsystem.digital.gov/components/table/)  so that they’re not smaller than all text in the table and on the page?

Here's an example of a page with this issue so you can see what we're talking about: https://digital.gov/about/manage-content/

> [!note]
> This request originated in this [Slack thread](https://gsa-tts.slack.com/archives/C050HRGN7/p1734095567505819).

opensource-joe:
## Summary
Sizes the table `<caption>` to match the table/body text instead of rendering smaller.

## Problem
The `caption` rule used `u-font("body", "xs")`, one step smaller than the table text (`$theme-body-font-size`, default `"sm"`), so captions read smaller than everything around them.

## Solution
Use `u-font("body", $theme-body-font-size)` so the caption matches the table/body size and stays theme-aware.

## Testing
Full USWDS Sass compiles cleanly; `.usa-table caption` now renders at the body size (1.06rem with defaults).

Fixes #6253


ethangardner (approved):
LGTM! Thanks for your contribution.

PR #6796, opened 2026-08-08
fix(focus-trap): handle hidden and dynamic controls

opensource-joe:
## Summary

Keep modal and navigation keyboard cycling on currently usable controls. The focus trap excludes hidden, disabled, inert, collapsed, and explicitly negative-tabindex controls, checks computed display through ancestors, and recalculates tab stops on activation and each Tab keypress. CSS-revealed elements carrying `hidden` remain reachable. Controls added, removed, or hidden while a modal stays open are handled without stale targets.

Closes #6056.
Closes #5830.

This is not a breaking change. No public API, consumer markup, or styling migration is required. `autoFocus: false` and Escape aliases are preserved.

## Consolidation

The branch incorporates the dynamic-focus work from #6262 through a merge preserving its contributor commits. Keep #6262 and both linked issues open until this combined change is verified and merged. One implementation and regression suite cover both issue scopes; no additional tracking issue is needed.

## Browser regression coverage

Three dedicated modal stories contain enhanced date-picker and combo-box controls. They exercise hidden internals, adding/hiding/removing controls, and CSS-revealed hidden controls, using actual Tab/Shift+Tab, Escape, focus return, open-state axe, and 320px reflow in the existing Storybook runner. Failures save screenshots and focus/accessibility evidence; CircleCI retains that directory as job artifacts.

## Current conflict resolution (2026-09-18)

Signed merge commit `c29b383eee3d0f867aeb7461e34191f8c1b38b56` incorporates develop `cf807302c36146a2fee203a831f1dc331e094106` while preserving contributor commits and the existing focus-trap implementation. The only content conflict was in the modal Storybook decorator. The resolution keeps both develop's close-before-reset behavior and open/default play coverage, plus this PR's readiness signal, enhanced controls, and three focus regression stories. The runner retains develop's full-body axe scan.

Fresh focused validation after resolution: 103 focus-trap/modal/navigation unit tests, 12 interaction-runner tests, and all three tagged Chromium focus stories pass. Storybook rebuild, targeted ESLint/Prettier, and `git diff --check` pass. Browser scenarios exercise hidden, dynamic, and CSS-revealed controls, both Tab directions, Escape, focus return, open-state axe, and 320px reflow. This refresh did not repeat the full suite or the earlier physical zoom spot-check. New-head CI and the named screen-reader verification below remain separate requirements.

## Earlier validation before the develop refresh

[...]

PR #6799, opened 2026-08-08
test(storybook): cover open modals in accessibility scans

### LINKED ISSUE #6787: Bug: a11y test suite can't see the modal, and no story exercises interactive state
### Description

There are two gaps in the Storybook a11y setup resulting in the bug in #6785 not being caught by the CI. Both look small to fix and I'm happy to send a PR.

**1. axe is scoped to `#storybook-root`, but the modal moves out of it.**

`.storybook/test-runner.js` calls `checkA11y(page, "#storybook-root", ...)`. On init, `packages/usa-modal/src/index.js:320` runs `document.body.appendChild(modalComponent)`, so `.usa-modal-wrapper`'s parent is `BODY`. The modal sits outside the scanned subtree in every state, meaning no axe run has ever covered it.

**2. No story puts a component into an interactive state.**

One of 65 story files uses `play`: `packages/usa-breadcrumb/src/usa-breadcrumb.stories.js`. Both of its play functions only read `getComputedStyle` and assert CSS properties. Nothing opens a modal, expands an accordion, or opens a nav menu, so the suite only ever scans initial render.

### Suggested fix

- Widen the axe scope to `body`, or include the modal wrapper in the selector.
- Add a `play` to the modal story that opens and closes it. That alone would have caught #6785.

opensource-joe:
## Summary

**Scan both the open modal and its closed state in Storybook accessibility tests.** Include portaled markup in the axe scan and keep the modal open until the open-state scan finishes.

## Breaking change

This is not a breaking change. Changes are limited to Storybook stories and the accessibility test runner; no component implementation changes.

## Related issue

Fixes #6787.

## Problem and solution

The modal moves its wrapper outside `#storybook-root`, so scans restricted to the story root miss it. Opening and closing within one play function also leaves the modal closed by the time `postVisit` scans the page.

Scan `body`, keep the Default story's open/close interaction coverage, and add a separate Open story whose play function leaves the dialog open. Before rendering another modal story, close the preceding modal through its public control before teardown; this prevents its page-hiding state from contaminating the next story.

## Validation

- Full Chromium Storybook accessibility suite: 65 suites and 260 tests pass.
- Negative control through Storybook's real setup and this runner's `postVisit`: an unnamed button inside the open, portaled modal is missed by the old root-only scan and rejected by the repaired runner (`button-name`). Removing the injected button restores a passing scan.
- Build, changed-file ESLint, and Prettier pass.
- Full local `npm test` passes.

This validates automated test coverage and isolation, not a new screen-reader behavior claim. Runtime teardown changes remain separate in #6831; related focus work remains in #6796.

### DIFF
--- a/.storybook/test-runner.js
+++ b/.storybook/test-runner.js
@@ -23,7 +23,10 @@ module.exports = {
     // Wait for any animations/transitions to settle
     await page.waitForTimeout(200);
 
-    await checkA11y(page, "#storybook-root", {
+    // Scan the whole document body, not just #storybook-root: some components
+    // (e.g. the modal) move their markup out to <body>, so a root-scoped scan
+    // never covers them. See #6787.
+    await checkA11y(page, "body", {
       detailedReport: true,
       detailedReportOptions: {
         html: true,
--- a/packages/usa-modal/src/usa-modal.stories.js
+++ b/packages/usa-modal/src/usa-modal.stories.js
@@ -31,16 +31,20 @@ Default.args = DefaultContent;
 // is queried from the document rather than the story canvas. See #6787.
 Default.play = async ({ canvasElement, step }) => {
   const opener = canvasElement.querySelector("[data-open-modal]");
-  const wrapper = document.querySelector(".usa-modal-wrapper");
+  // The modal initializes a frame after render, portaling .usa-modal-wrapper
+  // to <body>. Re-query the document each time rather than capturing a stale
+  // (possibly null) reference before init has run.
+  const getWrapper = () => document.querySelector(".usa-modal-wrapper");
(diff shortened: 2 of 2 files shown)


ryparker:
@coderabbitai full review

ryparker (approved):
The updated modal tests catch the missing coverage. An unnamed button in the open, portaled modal passes the old root-only scan but fails with the new `postVisit` hook. Removing it restores the passing result. Story cleanup uses the normal close control.

All 65 Storybook suites and 260 accessibility tests pass in Chromium, along with the full local suite. This is a test-only change. Ready to merge once refreshed CI passes.

PR #6800, opened 2026-08-08
USWDS - Banner: Remove redundant role from empty-alt icons

### LINKED ISSUE #5619: USWDS - Bug: Redundant role attribute on USA banner icons with empty alt text

opensource-joe:
## Summary
Removes `role="img"` from the two banner icons, which have `alt=""`.

## Problem
HTML validators (e.g. Nu Html Checker) flag both banner icons:
> An img element which has an alt attribute whose value is the empty string must not have a role attribute.

```html
<img class="usa-banner__icon usa-media-block__img" src="…" role="img" alt="" aria-hidden="true">
```

## Solution
Drop `role="img"` from both icons in `usa-banner.twig`. They keep `alt=""` and `aria-hidden="true"`, so they remain correctly hidden from assistive tech — the role was redundant and invalid. These were the only two matches repo-wide.

Fixes #5619


jonathanbobel (approved):
Thanks for this update

Tested in
- Mac OS Safari/Voiceover
- iOS Safari/VoiceOver
- Edge and Chrome/NVDA

jhancock532:
Thanks Joe! 🔥

PR #6811, opened 2026-08-13
[Task] Range Slider: Moved hint above the slider to match other component hints

### LINKED ISSUE #6614: USWDS - Bug: Introduce visual instruction for range slider

chandracarney:
# Summary
Minor follow-up after PR #6673. This fixes the Range Slider hint placement by moving the hint above the slider. Previously, the hint was introduced below the component, which doesn't match the system's pattern.

## Breaking change

## Related pull requests
Follow-up to #6673, which closed #6614.

## Screenshots
Before
Screenshot 2026-08-14 at 11 50 20

After
Screenshot 2026-08-14 at 11 51 06

### DIFF
--- a/packages/usa-range/src/test/template.html
+++ b/packages/usa-range/src/test/template.html
@@ -1,4 +1,5 @@
 <form class="usa-form">
+  <span class="usa-hint" id="range-hint">Move the slider to change the value</span>
   <input
     id="usa-range"
     class="usa-range"
@@ -10,7 +11,4 @@
     data-text-unit="percent"
     aria-describedby="range-hint"
   />
-  <span class="usa-hint" id="range-hint"
-    >Move the slider to change the value</span
-  >
 </form>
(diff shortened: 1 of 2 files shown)


ethangardner (approved):
I tested this in Storybook and compared it to how we handle other components. This was a good catch and was well executed.

PR #6823, opened 2026-08-18
USWDS - Focus trap: Ignore hidden inputs when it finds tab stops

### LINKED ISSUE #6056: USWDS - Bug: Focus trap treats hidden inputs as focusable
### Describe the bug

Hidden input elements are not focusable but the focus trap considers them to be.

When a hidden input is the first "focusable" element in a modal, it breaks tab cycling. Once the user has tabbed to the last element in the modal, focus gets stuck there. For example, if you're following the USWDS advice to put the close button at the end of the modal code, then focus gets stuck on the close button.

1. Go to https://codepen.io/sanason/pen/WNqmYrg.
2. Click "Open modal".
3. Press Tab 4 times.

### Expected Behavior

On the fourth tab press, focus returns to the "Tell us what you thought:" text input, instead of remaining on the X close button.

### Related code

_No response_

### Screenshots

_No response_

### System setup

_No response_

### Additional context

Touchpoints recently started using the USWDS modal which means a lot of forms inside modals. Touchpoints forms used to begin with several hidden inputs, so this issue was fairly visible. We've since moved the hidden inputs to the end of the form, which is an easy workaround.

### Code of Conduct

vssinghh:
Fixes #6056.

The `FOCUSABLE` list in `focus-trap.js` includes `input:not([disabled])`, which also matches `<input type="hidden">`. Hidden inputs can't take focus, so when one is the first match the trap treats it as the first tab stop and calling `focus()` on it does nothing.

What you actually see depends on the component. `usa-modal` passes `autoFocus: false` and sets the initial focus itself, so the symptom there is that tabbing forward from the last control never wraps and you get stuck, which is what @sanason reported from Touchpoints. `usa-header` and `usa-language-selector` use the default `autoFocus`, so they try to focus the hidden input on open and focus ends up on `<body>` instead of the first control.

The fix is to exclude hidden inputs from the selector.

@mejiaj asked whether this is only a focus trap issue. It is, and no component code changes. Two related things I left alone, happy to open separate issues if they're worth tracking:

- The selector matches elements hidden with `display: none`, `visibility: hidden`, or the `hidden` attribute the same way.
- `usa-header` and `usa-language-selector` build their tab stop list once in `init()`, so it goes stale if the container's contents change later.

I added three tests to `focus-trap.spec.js` covering auto focus, Tab from the last control, and Shift+Tab from the first. They fail before the change and pass after.

### DIFF
--- a/packages/uswds-core/src/js/utils/focus-trap.js
+++ b/packages/uswds-core/src/js/utils/focus-trap.js
@@ -4,7 +4,7 @@ const select = require("./select");
 const activeElement = require("./active-element");
 
 const FOCUSABLE =
-  'a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), iframe, object, embed, [tabindex="0"], [contenteditable]';
+  'a[href], area[href], input:not([disabled]):not([type="hidden"]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), iframe, object, embed, [tabindex="0"], [contenteditable]';
 
 const tabHandler = (context) => {
   const focusableElements = select(FOCUSABLE, context);
--- a/packages/uswds-core/src/js/utils/test/focus-trap.spec.js
+++ b/packages/uswds-core/src/js/utils/test/focus-trap.spec.js
@@ -105,4 +105,47 @@ describe("focus trap", () => {
       assert.strictEqual(document.activeElement, lastButton);
     });
   });
+  describe("hidden inputs", () => {
+    beforeEach(() => {
+      document.body.innerHTML = `
+        <button type="button" id="outside">Outside</button>
+        <div id="trap">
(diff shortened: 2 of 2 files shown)


vssinghh:
Just noticed https://github.com/uswds/uswds/pull/6796 is trying to do the same thing. Closing this in favor of that PR.

One note for whoever reviews #6796, since that PR is scoped to the modal: the same selector affects `usa-header` and `usa-language-selector` too. `usa-modal` passes `autoFocus: false` at both `FocusTrap` call sites and sets initial focus itself, so its symptom is the tab cycle dead-ending on the last control. Header and language selector pass only an `Escape` binding, so they take the default `autoFocus: true`, and since `init()` calls `firstTabStop.focus()`, a leading hidden input leaves focus on `<body>` instead of moving into the menu.

PR #6831, opened 2026-08-19
USWDS - Modal and header: Restore page content on teardown

### LINKED ISSUE #6830: USWDS - Bug: Modal and header leave the page aria-hidden after teardown
### Describe the bug

When a modal or the mobile navigation opens, USWDS sets `aria-hidden="true"` on all other body content. This is correct. It stops screen readers from reading the page behind the open component.

The close path removes those attributes. The teardown path does not. If `off()` runs while the component is still open, the attributes stay. The whole page remains hidden from assistive technology, and there is no recovery except a page reload.

This affects two components:

- `usa-header`: `teardown()` removes the resize listener and resets `navActive`. It does not call `showNonNavItems()`.
- `usa-modal`: `teardown()` calls `cleanUpModal()`, which restores the modal to its original place in the DOM. It does not remove `aria-hidden` or `data-modal-hidden` from non-modal elements.

For a screen reader user, the page reads as empty. This is a major functionality problem for that user, and it is silent for everyone else.

Workaround: close the component before calling `off()`.

Save this as `repro-short.js` in the repo root and run `node repro-short.js`.

```js
// Run from the repo root: node repro-short.js
const { JSDOM } = require("jsdom");

const dom = new JSDOM(`<!DOCTYPE html><body>
  <div class="usa-overlay"></div>
  <header class="usa-header">
    <div class="usa-nav-container">
      <button type="button" class="usa-menu-btn">Menu</button>
      <nav class="usa-nav">
        <button type="button" class="usa-nav__close">Close</button>
        <ul class="usa-nav__primary"><li><a href="#one">One</a></li></ul>
      </nav>
    </div>
  </header>
  <main id="main-content"><h1>Page content</h1></main>
</body>`, { pretendToBeVisual: true });

for (const k of ["window", "document", "navigator", "HTMLElement", "Element", "Event", "getComputedStyle"]) {
  global[k] = k === "window" ? dom.window : dom.window[k];
}

const navigation = require("./packages/usa-header/src/index.js");
const main = document.getElementById("main-content");
const show = (l) => console.log(`${l.padEnd(22)} aria-hidden=${main.getAttribute("aria-hidden")}`);

navigation.on(document.body);

navigation.toggleNav.call(navigation, true);
navigation.toggleNav.call(navigation, false);
show("closed normally");        // control: restored

navigation.toggleNav.call(navigation, true);
navigation.off(document.body);
show("off() while open");       // bug: still hidden
```

Output:

```
closed normally        aria-hidden=null
off() while open       aria-hidden=true
```

The first line is the control. A normal close restores the page. The second line is the bug.

I also tested `off()` while the nav is already closed. That is harmless. So the problem is specific to teardown while the component is open.

The same test on `usa-modal` gives the same result. Open the modal, call `modal.off(document.body)`, and non-modal content keeps `aria-hidden="true"`.

### Expected Behavior

`teardown()` should restore the page content that the component hid from assistive technology. Page content must never be left hidden after a component is removed.

`usa-modal` already does this correctly on close. The restore loop runs unconditionally, because the opener may be gone from the document by then. Teardown needs the same guarantee.

Only the assistive technology state needs restoring. `packages/usa-nav/src/test/navigation.spec.js` documents `off()` as removing event listeners without closing an open component, so visible state should be left alone.

### Related code

`packages/usa-header/src/index.js:258`. `teardown()` never calls `showNonNavItems()` (line 63).

`packages/usa-modal/src/index.js:392`. `teardown()` calls `cleanUpModal()`, which restores the modal's original DOM location but never removes `aria-hidden` or `data-modal-hidden`. The working restore loop is at line 180, in the close path only.

### Additional context

Calling `off()` in a component's destroy lifecycle is the normal pattern for framework users, and #5062 describes teams doing this in React, Vue, Angular, and Nuxt. An unmount or route change while a modal or the mobile menu is open triggers this.

Related to #6785 and PR #6786, which fixed the same class of problem on the modal close path. That fix does not cover teardown.

I have a fix for both components with regression tests and will open a PR shortly.

vssinghh:
## Summary

Restore page content after a modal or mobile header is torn down while open. Scoped cleanup now preserves an open component outside the requested root, disconnects obsolete focus handlers, and clears removed-modal body state. Header teardown tracks its initialization root so it can clean up after the navigation DOM has been removed. Authored `aria-hidden` attributes remain unchanged.

Closes #6830.

This is a nonbreaking bug fix. No public API, consumer markup, or styling migration is required.

## Browser regression coverage

Three dedicated stories exercise modal teardown, header teardown, and teardown of an unrelated subtree. They use the existing Storybook/Playwright/axe runner with actual Tab, Shift+Tab, Escape, focus return, accessible background reachability, and 320px reflow. Open-state contrast checks wait for the opening transition to finish. Failures save screenshots and focus/accessibility evidence; CircleCI retains that directory as job artifacts.

## Validation

- Full unit suite: 888 passing; focused modal/header suite: 72 passing (overlapping counts).
- Full Storybook accessibility suite: 262 tests passing across 65 suites in Chromium.
- All three new browser scenarios passed ten consecutive repetitions: 30/30.
- Known-bad controls: the develop implementation fails all three scenarios; the previously reviewed PR implementation passes whole-component teardown but fails scoped header isolation. Failures were inspected to confirm the intended regressions.
- ESLint, Prettier, and diff checks pass for changed files. No project dependency or lockfile changes. Browser validation includes generated fonts, icons, and SVG sprite; the interaction runner waits for fonts. The final 30 scenario repetitions reported no missing font/image responses.
- Contributor history preserved; new commits signed and verified by GitHub. Local checks use the existing workspace dependency installation; check fresh CI on the current head before merge.

## Remaining accessibility verification

Needs accessibility specialist assistive technology verification. Actual VoiceOver/Safari replay is prepared locally but remains inconclusive because Safari remote automation is disabled; no screen-reader assertions ran.

Verify keyboard-only operation and VoiceOver + Safari (or NVDA + Firefox): open the modal/mobile navigation, tear down while open, confirm background reading resumes, authored hidden content stays hidden, and focus no longer targets removed controls. Confirm unrelated teardown preserves the active component's isolation. Automated 320px reflow passes; actual 400% browser zoom remains to verify. No icon or pointer interaction changes are introduced.


ryparker:
@coderabbitai full review

PR #6842, opened 2026-08-24
USWDS - Combo box: Allow iOS VoiceOver to swipe combo box options

### LINKED ISSUE #6840: USWDS - Bug: iOS VoiceOver cannot swipe through open combo box options
### Describe the bug

When a combo box list is open, iOS VoiceOver swipe skips the listbox. Options are visible on screen but not in linear swipe order. You can tap a visible option to select it, but you cannot swipe through the list to discover what is available.

This affects major functionality for VoiceOver users on iPhone. It reproduces on the live site.

Collapsed default-value announcement is not this bug. Tabbing to a combo box with data-default-value already announces the current value (for example, “Select a fruit, Blackberry, text field, combobox, list box popup, collapsed”). That finding came from [uswds-site#3264](https://github.com/uswds/uswds-site/issues/3264).

1. Open https://designsystem.digital.gov/components/combo-box/ on an iPhone with VoiceOver on.
2. Swipe to a combo box and double-tap to open the list (or activate the toggle).
3. Swipe right.

The VoiceOver cursor jumps past the options to the next item in the DOM. The list is on screen but unreachable by swipe.

A side-by-side prototype confirmed the pattern: the USWDS combobox (focus on the input, aria-activedescendant on the option) is skipped. A version that moves real DOM focus into the options can be swiped.

### Expected Behavior

After the list opens, swipe right should move through each option (“Apple”, “Banana”, “Cherry”, and so on). Swipe left should move back through the list. Selecting an option should return focus to the input and announce the value.

### Related code

Live: https://designsystem.digital.gov/components/combo-box/
packages/usa-combo-box/src/index.js (input currently sets aria-owns and aria-controls to the listbox id)

`aria-owns` reparents the sibling listbox under the combobox in the accessibility tree. iOS VoiceOver then skips it during linear swipe. aria-controls already associates the list. Removing aria-owns restores swipe without moving focus into the options.

### Screenshots

_No response_

### System setup

USWDS version: current develop / live docs site
Device: iPhone 14 Pro
Operating system: iOS
Browser: Safari with VoiceOver

### Additional context

Found while investigating [uswds-site#3264](https://github.com/uswds/uswds-site/issues/3264) (combo box accessibility tests for default selection). That site issue should stay open for the docs update. This library issue is the swipe bug only.

Desktop check after a proposed aria-owns removal:

macOS VoiceOver + Safari: same as before
NVDA + Chrome: arrowing still announces the active option

### Code of Conduct

jonathanbobel:
USWDS - Combo box: Allow iOS VoiceOver to swipe combo box options

# Summary
Fixed a bug that prevented iOS VoiceOver users from swiping through open combo box options. Removing aria-owns from the combobox input keeps the listbox in linear swipe order. aria-controls still associates the list with the input.

## Breaking change

## Related issue
Closes #6840

## Preview link
https://federalist-3b6ba08e-0df4-44c9-ac73-6fc193b0e19c.sites.pages.cloud.gov/preview/uswds/uswds/bug/5316-remove-focusable-false/

## Problem statement
Desired: When a combo box list is open, iOS VoiceOver users can swipe through the available options to discover and select a value.

Actual: Swiping right from an open combo box jumped past the listbox to the next DOM item. Options were visible on screen but unreachable by swipe. This reproduced on the live site at https://designsystem.digital.gov/components/combo-box/. Collapsed-state announcements already included the current value, so 4.1.2 was met when the list was closed.

Consequence: VoiceOver users on iPhone could tap a visible option, but they could not browse the list by swipe, which is the primary VoiceOver navigation pattern.

## Solution
Removed `aria-owns` from the combo box input. The listbox is a sibling of the input, not a DOM child. `aria-owns` reparented it under the combobox in the accessibility tree, so iOS VoiceOver skipped it during linear swipe.

`aria-controls` already points at the list id, and `aria-activedescendant` still marks the highlighted option while focus stays on the input. That is enough for desktop screen readers to follow arrow-key highlighting.

This approach was chosen over moving real DOM focus into the options on open. Focusing options would likely dismiss the iOS keyboard and break type-to-filter on desktop.

Limitations: macOS VoiceOver + Safari combobox behavior is unchanged (still imperfect). iOS VoiceOver swipe is the path this PR fixes.

## Major changes

- Removed aria-owns from the combo box input in packages/usa-combo-box/src/index.js
- Left aria-controls in place to associate the listbox with the combobox

## Testing and review
- Run npm test and confirm all tests pass

### Manual verification

- [ ] Run npm start and open Components / Combo box
- [ ] iOS VoiceOver: Open the list. Swipe through options. Confirm each option is reachable. Select one. Confirm the collapsed input still announces the value.
- [ ] iOS VoiceOver: Type to filter, then swipe the reduced list. Close with the toggle and with a selection.
- [ ] Keyboard, no SR: Arrow from the input, Enter to select, Escape to close. Type-to-filter still works with focus on the input.
- [ ] macOS VoiceOver + Safari: Tab to the combobox. Open and arrow through options. Behavior should match the live site (same as before this change).
- [ ] NVDA + Chrome: Open the list and arrow. NVDA should still announce the active option.

## References
https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

### DIFF
--- a/packages/usa-combo-box/src/index.js
+++ b/packages/usa-combo-box/src/index.js
@@ -222,7 +222,6 @@ const enhanceComboBox = (_comboBoxEl) => {
   // sanitize doesn't like functions in template literals
   const input = document.createElement("input");
   input.setAttribute("id", selectId);
-  input.setAttribute("aria-owns", listId);
   input.setAttribute("aria-controls", listId);
   input.setAttribute("aria-autocomplete", "list");
   input.setAttribute("aria-expanded", "false");


chandracarney (approved):
Looks good to me, thanks!

PR #6880, opened 2026-08-28
USWDS - Testing: Verify compiled CSS release artifacts

### LINKED ISSUE #6877: USWDS - Testing: Replace build.spec.js with a dist smoke test in release-prep verification
# Description

The mocha-based build test (`src/test/build.spec.js`) is timing-dependent and unreliable when run across different environments. We should replace the approach with a deterministic smoke test that runs as part of the `.github/workflows/release-prep.yml` verification workflow, where a bad build would actually block a release.

> [!NOTE]
> This issue supersedes the approach taken in **#6413** and **#6417**.

## Background and problem statement

1. **Mocha timeouts are the wrong tool for build output verification:** `src/test/build.spec.js` relies on running gulp tasks inside a mocha harness and bumping timeout budgets (`.mocharc.json`), which introduces flakiness depending on runner speed and cold caches.
2. **Build error swallowing:** In `tasks/sass.js` (and `tasks/javascript.js`), errors emit an `end` event and log, but can allow the pipeline to exit 0 while producing empty or truncated artifacts.
3. **Stale pre-bump artifacts:** `npm ci` in `release-prep.yml` triggers `prepare` (`npm run build`), populating `dist/` with pre-bump banners before `npm version` executes. A failed subsequent build can leave a stale `dist/` that still appears populated unless the version string is explicitly verified.

## Expected behavior

Add an inline `run:` step to `.github/workflows/release-prep.yml` immediately following the `Bump version` step (before `Verify tarball hash matches generated security file`).

The smoke test step must fail the workflow (`::error::` and `exit 1`) if any of the following conditions are not met:

### 1. Correct version string
- The version banner `/*! uswds v${NEW_VERSION} */` must be present in both `dist/css/uswds.css` and `dist/css/uswds.min.css`.
- *(Note: Version banners are injected during Sass compilation at `tasks/sass.js`; JavaScript bundles do not currently contain version banners, so this check is scoped to CSS).*

### 2. Valid compiled CSS content
- Verify the CSS is not empty and contains actual compiled rules by checking for a known selector sentinel (such as `.usa-button`).

### 3. Size floors and inequality check
- `dist/css/uswds.css` must be at least **100 KB** (102,400 bytes).
- `dist/css/uswds.min.css` must be at least **100 KB** (102,400 bytes).
- `dist/css/uswds.css` and `dist/css/uswds.min.css` must be **different sizes from one another** (confirming minification ran and produced differentiated output).

## Implementation guidance

- Collect all verification failures and print them before calling `exit 1`, rather than exiting on the first failure.
- Output the verified sizes and version string in the step logs for easy review.
- Append a summary section to `$GITHUB_STEP_SUMMARY`.
- Leave existing files under `src/test/` (e.g., `build.spec.js`, `include.spec.js`) untouched as they are not wired to any existing task/script on `develop`.

## Non-goals

- No broad directory presence checks across all `dist/` subfolders (tarball verification handles package integrity).
- No upper size ceiling (deltas are already tracked via `.github/workflows/build-diff.yml`).
- No JS bundle version checks (banners are not generated in JS).

## Testing and review

1. Dispatch `release-prep.yml` from a test release branch and confirm the step succeeds on a clean build.
2. Confirm the step fails if `dist/css/uswds.min.css` is truncated below 100 KB.
3. Confirm the step fails if `dist/css/uswds.min.css` is overwritten with `dist/css/uswds.css` (equal sizes).
4. Confirm the step fails if the version string does not match `NEW_VERSION`.

## Code of Conduct

ANONYMOUSZED-beep:
## Summary
- add a release-prep smoke test for the compiled and minified CSS artifacts
- verify the bumped version banner, `.usa-button` compiled-rule sentinel, 100 KB size floors, and distinct minified output
- collect all failures as workflow error annotations and include artifact sizes and status in the job summary

## Release notes
**Release preparation now verifies compiled CSS artifacts before packaging.** The workflow blocks releases with missing, stale, empty, undersized, or unminified CSS output.

## How to test
- `npx --yes yaml-lint@1.7.0 .github/workflows/release-prep.yml` — passed
- extracted inline shell: `bash -n` — passed
- generated USWDS 3.14.0 CSS: verification passed (`uswds.css` 660,311 bytes; `uswds.min.css` 527,075 bytes)
- truncated `uswds.min.css`: failed as expected with banner, selector, and size errors
- copied `uswds.css` over `uswds.min.css`: failed as expected with equal-size error
- tested with version `9.9.9`: failed as expected with both stale-banner errors
- `git diff --check` — passed

`npm ci` generated the CSS used above successfully. Its subsequent, unrelated web-component phase cannot complete under Windows because the existing single-quoted Vite config path is interpreted literally by `cmd.exe`; release CI runs Ubuntu.

Closes #6877

### DIFF
--- a/.github/workflows/release-prep.yml
+++ b/.github/workflows/release-prep.yml
@@ -66,6 +66,74 @@ jobs:
           echo "NEW_VERSION=$NEW_VERSION" >> "$GITHUB_ENV"
           echo "Bumped to version: $NEW_VERSION"
 
+      - name: Verify compiled CSS artifacts
+        run: |
+          CSS_FILE="dist/css/uswds.css"
+          MIN_CSS_FILE="dist/css/uswds.min.css"
+          EXPECTED_BANNER="/*! uswds v${NEW_VERSION} */"
+          MIN_SIZE=102400
+          CSS_SIZE=0
+          MIN_CSS_SIZE=0
+          ERRORS=()
+
+          if [ -f "$CSS_FILE" ]; then
+            CSS_SIZE=$(stat -c%s "$CSS_FILE")
+          else
+            ERRORS+=("Missing compiled CSS artifact: $CSS_FILE")
+          fi
+
+          if [ -f "$MIN_CSS_FILE" ]; then
+            MIN_CSS_SIZE=$(stat -c%s "$MIN_CSS_FILE")
+          else
+            ERRORS+=("Missing compiled CSS artifact: $MIN_CSS_FILE")
+          fi
+
+          for FILE in "$CSS_FILE" "$MIN_CSS_FILE"; do
+            if [ -f "$FILE" ] && ! grep -Fq "$EXPECTED_BANNER" "$FILE"; then
+              ERRORS+=("$FILE does not contain the expected version banner: $EXPECTED_BANNER")
+            fi
+            if [ -f "$FILE" ] && ! grep -Fq ".usa-button" "$FILE"; then
+              ERRORS+=("$FILE does not contain the .usa-button compiled-rule sentinel")
+            fi
+          done
+
+          if [ "$CSS_SIZE" -lt "$MIN_SIZE" ]; then
+            ERRORS+=("$CSS_FILE is below the 102,400-byte size floor ($CSS_SIZE bytes)")
+          fi
+          if [ "$MIN_CSS_SIZE" -lt "$MIN_SIZE" ]; then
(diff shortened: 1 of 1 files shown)


ethangardner (approved):
This is great! Thanks for your contribution. This smoke test is just what we need to help improve our deployment process, and I appreciate how you added the summary at the end and the helpful `echo` logging throughout. That'll be very easy to parse the results when we go to cut another release. Approved!

Issue #4307, opened 2021-09-01
Limit use of ARIA in SVGs


mgifford:
In looking at existing SVGs https://github.com/uswds/uswds/pull/4304

I ran into `aria-hidden="true" focusable="false”` within the SVG. I have no idea why this would be included, but really would like to remove ARIA unless it has a clear specific role.

Defining `role="img"` is useful. Looking at [Carie Fisher's work on Smashing Mag](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/) as an example:

```
<svg role="img" aria-labelledby="fox11 description11" ...>
   <title id="fox11">What does the fox say?</title>
   <desc id="description11">Will we ever know?</desc>
   [design code]
</svg>
```

We might want the images to look a bit more like:

```
<svg xmlns="http://www.w3.org/2000/svg" width="64" height="39" viewBox="0 0 64 39" role="img" aria-labelledby="title-arrow1 description-arrow1">
  <title id="title-arrow1">Down angle arrow</title>
  <description id="description-arrow1"></description>
  <style>
    path {
      fill: #005EA2;
   }
   @media (prefers-color-scheme: dark) {
     path {
        fill: #58b4ff;
      }
    }
  </style>
  <g>
    <path d="M63.36 4.004L60.155.741c-.427-.436-.919-.654-1.475-.654-.555 0-1.047.218-1.475.654L32 26.4 6.798.74C6.37.306 5.878.088 5.323.088c-.556 0-1.048.218-1.476.654L.642 4.005C.213 4.44 0 4.941 0 5.507c0 .565.214 1.066.642 1.501l29.883 30.427c.427.435.92.652 1.475.652.556 0 1.047-.217 1.474-.652L63.361 7.008c.427-.435.639-.936.639-1.501 0-.566-.212-1.067-.64-1.503z" />
  </g>
</svg>
```

thisisdano:
I would like to address this, but I know there are some compatibility reasons for our SVG being the way it is and we'll need to research and document the why of our current state before we make any decision here

thisisdano:
So, I believe this code is here because for a couple reasons, initially outlined here: https://github.com/uswds/uswds/pull/3892

[...]

Issue #6749, opened 2026-07-06
USWDS - Feature: accessibility skill for coding agents


ethangardner:
Per the [issue comment](https://github.com/uswds/uswds/pull/6748#issuecomment-4896813158) on PR #6711 and a follow up Slack discussion, adding an accessibility skill to the repo would be a welcome addition for coding agents.

PR #6886, opened 2026-09-01
USWDS - Icons: Remove legacy `focusable="false"` from SVGs

### LINKED ISSUE #5316: USWDS - Icons: Confirm if legacy attributes still required
Confirm if SVGs still require aria-hidden and focusable="false" (previously for legacy browsers). From this [comment in issue](https://github.com/uswds/uswds/issues/4307#issuecomment-1559410923).

jonathanbobel:
USWDS - Icons: Remove legacy `focusable="false"` from SVGs

# Summary

Removed the legacy `focusable="false"` attribute from inline SVGs. Decorative icons still use `aria-hidden="true"`. Modern browsers do not put these SVGs in the tab order, so the Internet Explorer-only attribute is no longer needed.

## Breaking change

## Related issue

Closes #5316

Related discussion: https://github.com/uswds/uswds/issues/4307#issuecomment-1559410923

## Related pull requests

Documentation examples on [uswds-site](https://github.com/uswds/uswds-site) that still show `focusable="false"` on SVGs may need a follow-up PR.

## Preview link

Preview link: N/A

## Problem statement

**Desired:** Inline USWDS SVGs use only the attributes current browsers need. Decorative icons stay hidden from assistive technology. They are not in the tab order.

**Actual:** Markup still included `focusable="false"` on SVGs, a workaround for Internet Explorer putting inline SVGs in the tab order. USWDS no longer supports IE. The extra attribute added noise to examples and did not change behavior in current browsers.

**Consequence:** Teams copy outdated markup, and it is unclear whether `aria-hidden="true"` is enough.

## Solution

Removed `focusable="false"` from every inline SVG in component Twig and related test templates. Left `aria-hidden="true"` (and existing `role="img"`) in place for decorative icons.

The banner lock SVG is labeled with `aria-labelledby` and is not `aria-hidden`. It also no longer has `focusable="false"`.

This matches current browser behavior: an SVG without `tabindex` is not keyboard-focusable, so the IE attribute is unnecessary.

## Major changes

- Removed `focusable="false"` from SVGs in:
  - `packages/usa-banner/src/usa-banner.twig`
  - `packages/usa-button/src/usa-button.twig`
  - `packages/usa-icon/src/usa-icon.twig`
  - `packages/usa-icon/src/usa-icon--sizes/usa-icon--sizes.twig`
  - `packages/usa-input-prefix-suffix/src/usa-input-prefix.twig`
  - `packages/usa-modal/src/usa-modal.twig`
  - Modal and form test templates

## Testing and review

- [ ] Run `npm test` and confirm all tests pass

**Manual verification**

1. Run `npm start:storybook` and spot-check **Icon**, **Button** (with icon), **Banner**, **Modal**, and **Input prefix**.
2. Confirm decorative icons are still skipped by the screen reader (`aria-hidden="true"`).
3. Tab through each component. Confirm the SVG itself does not receive focus. Focus should stay on the control (button, input, banner accordion, modal close).
4. **VoiceOver + Safari** (macOS) and **VoiceOver on iOS:** icons remain decorative. Tab/swipe order is unchanged.
5. **NVDA + Chrome** and **NVDA + Edge:** same as above.

No visual change is expected. The review is markup and focus order only.

### DIFF
--- a/packages/usa-banner/src/usa-banner.twig
+++ b/packages/usa-banner/src/usa-banner.twig
@@ -1,6 +1,6 @@
 {% set lock %}
   <span class="icon-lock">
-    <svg xmlns="http://www.w3.org/2000/svg" width="52" height="64" viewBox="0 0 52 64" class="usa-banner__lock-image" role="img" aria-labelledby="banner-lock-description" focusable="false">
+    <svg xmlns="http://www.w3.org/2000/svg" width="52" height="64" viewBox="0 0 52 64" class="usa-banner__lock-image" role="img" aria-labelledby="banner-lock-description">
       <title id="banner-lock-title">Lock</title>
       <desc id="banner-lock-description">Locked padlock icon</desc>
       <path fill="#000000" fill-rule="evenodd" d="M26 0c10.493 0 19 8.507 19 19v9h3a4 4 0 0 1 4 4v28a4 4 0 0 1-4 4H4a4 4 0 0 1-4-4V32a4 4 0 0 1 4-4h3v-9C7 8.507 15.507 0 26 0zm0 8c-5.979 0-10.843 4.77-10.996 10.712L15 19v9h22v-9c0-6.075-4.925-11-11-11z"/>
--- a/packages/usa-button/src/usa-button.twig
+++ b/packages/usa-button/src/usa-button.twig
@@ -19,7 +19,7 @@
 {% set icon %}
   <!-- @TODO: Allow size to be passed as a argument. -->
   <!-- Example: usa-icon--size-{3-9} -->
-  <svg class="usa-icon" aria-hidden="true" focusable="false" role="img">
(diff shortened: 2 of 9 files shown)


.agents/skills/uswds-accessibility/SKILL.md
```diff
@@ -0,0 +1,145 @@
+---
```
chandracarney: Looks like a couple files might be crossover from the skill PR

ethangardner (changes requested):
@jonathanbobel - I'm seeing the accessibility skill in this changeset that you added for #6749. It's possible you might have branched from that work in progress when you started this work. Let's remove that from this PR and the rest should be good to go.

chandracarney (approved):
This is an exciting update to improve things! Thank you.

ethangardner (approved):
Thanks. This looks good now!

PR #4653, opened 2022-04-28
Move 3.0 branch into active develop

thisisdano:
This updates `develop` to be current with the active 3.0 development work

PR #6896, opened 2026-09-04
replace map-deep-get with sass:map.get

nektro:
the builtin function already supports variadic arguments

### DIFF
--- a/packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
+++ b/packages/usa-layout-grid/src/styles/_usa-layout-grid.scss
@@ -6,7 +6,7 @@
 
 $namespace-grid: ns("grid");
 
-$custom-breakpoints: map-deep-get($system-properties, breakpoints, extended);
+$custom-breakpoints: map.get($system-properties, breakpoints, extended);
 $all-breakpoints: map-collect($system-breakpoints, $custom-breakpoints);
 
 // basic container
@@ -51,7 +51,7 @@ $all-breakpoints: map-collect($system-breakpoints, $custom-breakpoints);
   &.#{$namespace-grid}gap {
     @include grid-gap-responsive;
   }
-  @each $gap-key, $gap-val in map-deep-get($system-properties, gap, standard) {
+  @each $gap-key, $gap-val in map.get($system-properties, gap, standard) {
     &.#{$namespace-grid}gap-#{$gap-key} {
       $props: append-important($grid-global, $gap-key);
       @include grid-gap($props);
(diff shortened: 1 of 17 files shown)


ethangardner:
@nektro - Thanks for this! I think you're onto a possible simplification that we could make, but I'm having a difficult time determining what issue this PR is for or what problem this is solving. Would you mind referencing the issue this addresses in the PR body, and also using the [PR template](https://github.com/uswds/uswds/blob/develop/.github/PULL_REQUEST_TEMPLATE.md) to help me know what issue this addresses and how to go about my review of this contribution?

nektro:
hi ethan, this is not connected to an existing issue. i was browsing the code and noticed this possible enhancement and thought to send it off.

## Breaking change

this is not a breaking change

## Related issue

this does not close an existing issue

## Related pull requests

this not reference any other pull requests. it has been unchanged since its introduction in https://github.com/uswds/uswds/pull/4653

## Problem statement

simplifies the codebase by removing a utility that is provided by native functionality of the language

## Solution

it was an exact match so a simple find+replace made all existing code build successfully

## Testing and review

i ran `npm run build` when testing

mgifford:
Using AI I took a look at this code, not being terribly familiar with it.

> - Replaces map-deep-get() calls across 17 Sass files.
> - Deletes the 30-line custom implementation.
> - Removes it from the general-functions index.
> - Adds @use "sass:map" where necessary.
> Should produce the same compiled CSS.
> 
> The 67 additions and 96 deletions therefore represent a net simplification, not new behaviour.

It also highlighted:

> The deleted function is explicitly marked @access public, is forwarded through functions/general/_index.scss, and USWDS exports its Sass function paths through package.json which might break:
> `general.map-deep-get(...)`

The AI suggested just depreciating it first (which makes sense to me).

```
@function map-deep-get($map, $keys...) {
  @warn "map-deep-get() is deprecated. Use sass:map.get() instead.";
  @return map.get($map, $keys...);
}
```

@nektro this make sense to you?

nektro:
yes that makes sense, i can edit the commit

nektro:
bump

ryparker:
@coderabbitai full review

ryparker:
The wrapper at `ad5cf586` breaks an existing call with no keys:

```scss
map-deep-get((brand: #123456))
```

This previously returned the map unchanged. The [new wrapper](https://github.com/uswds/uswds/blob/ad5cf586c6e00414929bd3dc2066445389df9fff/packages/uswds-core/src/styles/functions/general/map-deep-get.scss#L14) passes no key to `map.get()`, which requires one. Comparing compilation before and after confirms `Missing argument $key`. Helpers that forward an empty key list would therefore stop compiling.

We can keep the simplification by returning `$map` when the key list is empty and using `map.get()` otherwise. Please retain the public export and add zero-key, one-key, and nested-key tests. That addresses the breaking-change label; the deprecation warning is a separate question.

PR #6899, opened 2026-09-05
USWDS - Core: Add characterization test for package-exports surface

### LINKED ISSUE #6848: USWDS - JS: Add a characterization test for the published package-exports surface
## Summary

Add an automated test that resolves `package.json`'s `exports` map the same
way a real consumer's `require()`/`import` would, and asserts the resulting
values have the expected shape. This is the safety net for a later change
(see the parent issue #6845) that repoints those exports from raw source files to
built artifacts, and it must keep passing, completely unchanged, before and
after that repoint -- that unchanged-ness is what proves the repoint didn't
break anything observable.

## Why this is needed

`package.json#exports` currently maps `"./js/*"` directly to
`"./packages/*/src/index.js"` -- i.e. consumers who do
`require("@uswds/uswds/js/usa-accordion")` get the raw source file, not a
built artifact. Nothing in this repo's test suite currently verifies that
consumer-facing resolution path works at all.

## Scope

- [ ] New file: `packages/uswds-core/src/test/package-exports.spec.js`
      (Mocha spec).
- [ ] Wired into the same `distTests` gulp task added in the "behavioral
      characterization test for the built JS bundle" issue (#6847; it also needs a
      built `dist/` directory to resolve some export targets).

## What the spec must assert

Use Node's `Module.createRequire()` API to build a `require` function
resolved against this package's own `package.json`, and a dynamic
`import()` for the ESM-condition check, so the test exercises Node's real
`exports`-map resolution logic (including condition selection) rather than
hand-constructing file paths:

```js
const { createRequire } = require("module");
const path = require("path");
const requireFromRoot = createRequire(
  path.resolve(__dirname, "../../../../package.json")
);
```

Then, for **each of the 22 component subpaths** (`@uswds/uswds/js/usa-accordion`,
`@uswds/uswds/js/usa-banner`, ... -- the full component list is in the
barrel test from the "behavioral characterization" issue (#6847)), assert:

1. `requireFromRoot("@uswds/uswds/js/<name>")` resolves without throwing,
   and the resulting value (unwrap `.default` if the module has
   `__esModule: true`, to make this test pass unchanged both before and
   after the module becomes ESM internally) has callable `.on` and `.off`
   methods.
2. The same resolution via `import()` (`await
   import("@uswds/uswds/js/<name>")`) also resolves to a value with a
   callable `.on`.

Additionally assert the **two known cross-component API surfaces** that a
naive refactor is likely to drop:
- `usa-date-picker`'s resolved value has callable
  `.getDatePickerContext`, `.isDateInputInvalid`, and
  `.updateCalendarIfVisible` methods (these are used by `usa-date-range-picker`).
- `usa-combo-box`'s resolved value has a `.COMBO_BOX_CLASS` string property
  and a callable `.enhanceComboBox` method (used by `usa-time-picker`).

Finally, assert the **barrel exports** (`@uswds/uswds/js` and
`@uswds/uswds/src/js/components`, both currently pointing at
`packages/uswds-core/src/js/index.js`) resolve to an object with exactly
the 22 keys listed in the "behavioral characterization" issue (#6847), each with a
callable `.on`.

## Verification
```
npm ci
npm run lint
gulp typecheck
gulp test
npm run test:dist
```
plus `npm run test:dist`.

## Depends on
Nothing in `packages/` source. Should land alongside or after the "add a
behavioral characterization test for the built JS bundle" issue (#6847; they share
the `distTests` gulp task).

ravitejapioneerblaze-code:
## Summary

Added a Mocha characterization test that verifies all 22 public component
subpaths resolve correctly via Node's real exports-map resolution, ensuring
no consumer-visible regressions occur during the CommonJS-to-ESM migration.

## Breaking change

## Related issue

Closes #6848

## Problem statement

Nothing in the test suite currently verifies that the consumer-facing
package.json exports map resolves correctly. When the exports are later
repointed from raw source files to built artifacts, there is no automated
safety net to confirm nothing observable changed for consumers.

## Solution

Added packages/uswds-core/src/test/package-exports.spec.js using
Node's Module.createRequire() API to resolve all 22 component subpaths
the same way a real consumer would. The unwrap helper ensures every
assertion passes unchanged both before and after a component's source is
converted from CommonJS to ESM internally.

## Major changes

- New file: packages/uswds-core/src/test/package-exports.spec.js
- No existing files modified

## Testing and review

Run the spec directly:
npx mocha packages/uswds-core/src/test/package-exports.spec.js --require jsdom-global/register
Result: 48 passing, 0 failing

Full suite with no regressions:
npx mocha "packages/usa-*/**/*.spec.js" "packages/uswds-*/**/*.spec.js" --ignore "packages/uswds-core/src/test/sass.spec.js" --ignore "packages/usa-accordion/src/test/accordion-icon.spec.js" --require jsdom-global/register
Result: 888 passing, 0 failing

The spec covers:
- All 22 component subpaths resolve via require() with callable .on and .off
- All 22 component subpaths resolve via import() with callable .on
- usa-date-picker cross-component API: getDatePickerContext, isDateInputInvalid, updateCalendarIfVisible
- usa-combo-box cross-component API: COMBO_BOX_CLASS string and callable enhanceComboBox
- Both barrel paths resolve to an object with exactly 22 keys each with callable .on

### DIFF
--- a/packages/uswds-core/src/test/package-exports.spec.js
+++ b/packages/uswds-core/src/test/package-exports.spec.js
@@ -1,0 +1,186 @@
+"use strict";
+
+const assert = require("node:assert");
+const { createRequire } = require("module");
+const path = require("path");
+
+// Resolve all imports the same way a real consumer would —
+// using Node's real exports-map resolution against the root package.json.
+const requireFromRoot = createRequire(
+  path.resolve(__dirname, "../../../../package.json"),
+);
+
+// Helper: unwrap .default if the module has the __esModule marker.
+// This makes every assertion pass unchanged both before and after
+// a module's source is converted from CommonJS to ESM internally.
+function unwrap(mod) {
+  return mod.default !== undefined ? mod.default : mod;
+}
+
+// The 22 component subpaths exposed via package.json#exports["./js/*"].
+// Derived from packages/uswds-core/src/js/index.js barrel and the
+// packages/usa-*/src/index.js directory scan.
+const COMPONENT_SUBPATHS = [
+  "usa-accordion",
+  "usa-banner",
+  "usa-button",
+  "usa-character-count",
+  "usa-combo-box",
+  "usa-date-picker",
+  "usa-date-range-picker",
+  "usa-file-input",
+  "usa-footer",
+  "usa-header",
+  "usa-in-page-navigation",
+  "usa-input-mask",
+  "usa-language-selector",
+  "usa-modal",
+  "_usa-password",
+  "usa-range",
+  "usa-search",
+  "usa-skipnav",
+  "usa-table",
+  "usa-time-picker",
+  "usa-tooltip",
+  "usa-validation",
+];
+
+// The 22 barrel keys exported by packages/uswds-core/src/js/index.js.
+const BARREL_KEYS = [
+  "accordion",
+  "banner",
+  "button",
+  "characterCount",
+  "comboBox",
+  "datePicker",
+  "dateRangePicker",
+  "fileInput",
+  "footer",
+  "inPageNavigation",
+  "inputMask",
+  "languageSelector",
+  "modal",
+  "navigation",
+  "password",
+  "range",
+  "search",
+  "skipnav",
+  "table",
+  "timePicker",
+  "tooltip",
+  "validator",
+];
+
+describe("package exports surface", function () {
+  // ── 1. Per-component subpath resolution ────────────────────────────────────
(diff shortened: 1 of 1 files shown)


ryparker:
@coderabbitai full review

ryparker (approved):
Looks good against the updated `develop` branch. This adds one package-export test file without changing component code or dependencies. All 48 focused tests, ESLint, and formatting checks pass locally, and CI and signature checks are green.

PR #6903, opened 2026-09-07
test(build): characterize the compiled JavaScript bundle

### LINKED ISSUE #6847: USWDS - JS: Add a behavioral characterization test for the built JS bundle
## Summary

Add an automated test that loads the actual *built* JavaScript bundle
(`dist/js/uswds.js` / `dist/js/uswds.min.js`) and asserts on its runtime
behavior. This does not exist today, and it is the primary safety net for
an upcoming change that replaces the bundler (browserify) that produces
this file -- see the parent issue #6845 for full context on that migration.

## Why this is needed

- `gulpfile.js`'s `exports.test` task chain (`typeCheck, lintSass,
  sassTests, unitTests, tasksTests`) never runs `compileJS`. `npm run
  test:ci` is `lint && gulp test && test:a11y && build:html` -- also no
  `compileJS`. So `dist/js/uswds.js` is only ever built when running
  `npm pack` (via the `prepare` script) or via the manually-triggered
  `build-diff.yml` GitHub Action. **There is currently no CI signal at all
  on whether the built JS bundle actually works.**
- A future PR will replace the bundler that produces this file. A byte-diff
  of the old and new bundle output will be totally uninformative (a
  different bundler produces different bytes for equivalent code), so a
  *behavioral* test is the only way to verify the swap didn't change
  anything a consumer can observe.

## Scope

- [ ] New file: `packages/uswds-core/src/test/dist-bundle.spec.js` (Mocha
      spec, matches the existing `packages/uswds-*/**/*.spec.js` glob so no
      new glob entry is needed).
- [ ] New gulp task, e.g. `distTests`, in `tasks/test.js`, that runs this
      spec. Export it from `gulpfile.js` as `exports.distTests`, and add a
      composed task (e.g. `exports.testDist = series(compileJS,
      distTests)`) that builds the JS bundle first, then runs the spec
      against the built output.
- [ ] `package.json`: add a script, e.g. `"test:dist": "gulp testDist"`.
- [ ] `npm run test:dist` is added to the `test:ci` script, positioned
      *before* `test:a11y` (so a broken bundle fails fast, before the
      slower Storybook/axe suite runs).

## What the spec must assert

Load the built bundle into a real DOM (e.g. via `jsdom` and Node's `vm`
module -- read the built file's contents and evaluate them in a fresh jsdom
window context; do not `require()` it as a module, since it's a plain
browser script with no module exports). Then assert:

1. **`window.uswdsPresent === true` after the bundle runs.** This is set as
   the very first statement in `packages/uswds-core/src/js/start.js`
   (`window.uswdsPresent = true;`). `packages/uswds-core/src/js/uswds-init.js`
   (a separate, unbundled script meant to run in `<head>` before the main
   bundle) polls this flag to know when USWDS has finished loading. If a
   future build tool's dead-code elimination ever strips this line as an
   unused side effect, every USWDS site's JS silently stops initializing --
   with no error and no visible difference in a static screenshot, since
   the CSS still loads fine. This is the single most important assertion in
   this whole test.
2. **The component barrel exposes exactly these 22 keys and no others:**
   `accordion, banner, button, characterCount, comboBox, datePicker,
   dateRangePicker, fileInput, footer, inPageNavigation, inputMask,
   languageSelector, modal, navigation, password, range, search, skipnav,
   table, timePicker, tooltip, validator`. Note two of these do **not**
   match their package directory name 1:1: `navigation` corresponds to the
   `usa-header` package, and `password` corresponds to the `_usa-password`
   package. These are the two easiest keys to accidentally lose or rename
   in a refactor -- assert the full set, not just a spot check.
3. **Every one of those 22 values has a callable `.on`, `.off`, `.add`, and
   `.remove` method** (this is the shared "behavior" interface all
   components implement).
4. **No global leaks.** After the bundle runs, `window.uswds`, `window.USWDS`,
   `window.require`, `window.module`, `window.exports`, and `window.define`
   must all be `undefined`. `README.md` explicitly documents that USWDS
   components are *not* accessible in the global browser scope by default
   (a consumer has to opt in with `window.uswds = require("./components")`)
   -- assert this stays true.
5. **An end-to-end interaction test**: build a small HTML fixture containing
   a valid `usa-accordion` markup structure, mount it in the jsdom document,
   dispatch a real click event on the accordion button, and assert
   `aria-expanded` flips to `"true"` and the corresponding content panel's
   `hidden` attribute is removed. This exercises the full behavior-wiring
   chain through actual minification, not just "the file parses."
6. **Sourcemap sanity**: the accompanying `.map` file exists, has a
   non-empty `sourcesContent` array, and its `sources` array contains at
   least one path referencing a real package (e.g. assert
   `sources.some(s => s.includes("usa-accordion"))` -- use a substring
   check, not an exact-path match, since different bundlers format source
   paths differently).
7. **`dist/js/uswds-init.js` / `.min.js` stay plain ES5** (no arrow
   functions, `const`/`let`, or template literals) -- this file runs in
   `<head>` before any feature detection can happen, so it must work in the
   oldest browsers USWDS supports, independent of whatever the main bundle
   targets.

## Verification
```
npm ci
npm run lint
gulp typecheck
gulp test
npm run test:dist
```
plus `npm run test:dist` (new script added by this issue).

## Depends on
Nothing in `packages/` source. Independent of the other Phase 0 issues (#6846, #6848, #6849).

ravitejapioneerblaze-code:
## Summary

**Check the built JavaScript bundle's behavior in CI.** Load a freshly compiled bundle in jsdom, verify initialization and accordion interaction, and fail on unexpected globals or browser errors.

## Breaking change

This is not a breaking change. Component implementation and build output are unchanged; this adds tests and their task wiring.

## Related issue

Related to #6847. Keep that issue open for the deferred minified-initializer compatibility requirement.

## Problem and solution

A successful compilation does not establish that the emitted browser script initializes correctly. Add a dedicated `test:dist` task that compiles JavaScript before running the bundle characterization suite, and run it from `test:ci`. Exclude this artifact-dependent suite from ordinary unit-test discovery.

Use jsdom's actual browser VM context, complete the media-query mock, wait for normal DOM initialization, capture browser/event errors, and close test windows afterward. Compare window properties before and after execution so unexpected globals fail. Check the actual source barrel and component interfaces, the bundle presence flag, an accordion click, source-map contents, and initializer syntax.

The source-barrel/interface checks characterize package exports; the initialization, global, and accordion checks exercise the compiled browser bundle.

## Validation

- `npm run build` and `npm test`: pass locally.
- `npm run test:dist`: 37 passing, without browser initialization errors.
- Negative control: adding an unexpected window global to the compiled bundle fails the suite.
- Negative control: throwing during `DOMContentLoaded` fails the suite.
- Restored the unmodified generated bundle after both controls; final bundle suite passes.
- Changed JavaScript passes ESLint and Prettier.

## CodeRabbit follow-up

The suggestion to apply the ES5 assertions to `uswds-init.min.js` was tested. Current `compileJS` already emits an arrow-function wrapper in that artifact: the expanded suite reports 39 passing and one failure. Kept the existing passing characterization scope; changing the minifier's output target belongs in a separate build-compatibility fix. The unminified initializer's ES5 checks remain active.

The follow-up suggestion to register `jsdom-global` inside the spec is already satisfied by the root `.mocharc.js`, which loads `jsdom-global/register`. The Gulp Mocha invocation discovers that config; an additional suite-local global setup would duplicate it. Isolated `JSDOM` instances remain necessary to observe the bundle's own window globals.

## Review scope clarification

The browser bundle intentionally exposes no module exports or public component barrel. The source-export checks are labeled separately from the compiled-bundle behavior checks. This test-only PR changes no production ARIA or keyboard behavior and makes no new assistive-technology conformance claim.


ryparker:
@coderabbitai full review

ryparker (approved):
The updated tests look good. The full local suite and all 37 bundle tests pass, and the bundle tests correctly fail when I introduce an unexpected global or a `DOMContentLoaded` error. They check the actual barrel exports and clean up their isolated browser context. Component code is unchanged. Ready to merge once the refreshed CI run passes.

PR #6909, opened 2026-09-09
Update references to design systems in README

mgifford:
Removed outdated links and added GOV.UK Frontend link.

Even pointing to an Archive.org page for a design system reference isn't a good idea so I removed the link to:  https://web.archive.org/web/20250830060151/https://uspto.github.io/USPTO-Design-System


msbtterswrth (approved):
Thanks for the update Mike!

PR #6912, opened 2026-09-09
fix(nav): preserve hover icons and theme overrides

mgifford:
## Summary

Correct the collapsed navigation icon selector from `::after:hover` to `:where(:hover)::after`. Applying hover to the containing span produces a valid selector. Using `:where()` preserves the previous specificity, so application styles that customize the icon continue to win.

A plain `:hover::after` introduces an override regression. With this later application rule, the icon should remain red on hover:

```css
.usa-nav__primary button[aria-expanded=false] span::after {
  background-color: rgb(190, 0, 0);
}
```

The previous PR head changed the icon to the system button-text color. The repaired head retains red. Originally reported by @nektro.

## Breaking change

This is not a breaking change. Existing markup and customization precedence remain supported. The demonstrated override regression in the earlier PR head is repaired.

## Testing and review

Verified at `ec3e720edcbcd20f2dd69cf13da70ae6621d5e43`:

- Compiled the actual base and PR Sass and reproduced the consumer override failure in Chromium, Firefox, and WebKit. All three engines pass the repaired case.
- Compared idle, hover, keyboard-focus, and expanded states at 375px and 1200px with normal and emulated forced colors. All 16 base/repair observations per engine match for default styling.
- Retained before/after screenshots using the pinned Material icon assets. Formatting and whitespace checks pass.
- Current build, signature, CodeQL, and security checks pass. The title follows the current conventional-title policy.

These are browser probes, not a committed regression test in the repository suite. A jsdom hover assertion cannot verify rendered pseudo-element styling. Durable CSS-aware browser coverage remains a useful follow-up. Forced-colors emulation does not establish actual Windows high-contrast behavior, and WebKit is not released Safari.

Before merge, refresh the branch base and rerun final verification. The current `PR title` workflow checks an older trusted base that lacks `.github/scripts/check-pr-title.test.mjs`, so it fails before validating the title. This PR remains open.

### DIFF
--- a/COMMUNITY.md
+++ b/COMMUNITY.md
@@ -216,7 +216,7 @@ If anyone involved with USWDS in the past would like to be reflected on this pag
 
 None of us can do this alone. We never have. We never will. We [learn from each other](https://designsystem.digital.gov/design-principles/#listen), and [share with each other](https://designsystem.digital.gov/about/product-values/#share-what-we-do), and continuing _that_ will continue to make things better over time. Thanks for being here.
 
-Total number of contributors to this particular repository: <!--CONTRIBUTOR COUNT START--> 190 <!--CONTRIBUTOR COUNT END--> (!)
+Total number of contributors to this particular repository: <!--CONTRIBUTOR COUNT START--> 191 <!--CONTRIBUTOR COUNT END--> (!)
 
 <!-- readme: contributors -start -->
 <table role="presentation">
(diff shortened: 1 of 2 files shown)


mgifford:
I thought I addressed the author response, but perhaps not. This is a trivial patch.

chandracarney:
Hey @mgifford thanks for the contribution! The label is automatically added to issues without a verified signature. You'll have to follow the steps in the GitHub Actions comment above for commit verification, then push again.

mgifford:
I think it is sorted now.

ryparker:
@coderabbitai full review

ryparker:
Fixed the hover selector in `ec3e720e` using `:where(:hover)::after`. The earlier `:hover::after` correction made the selector valid but increased specificity, overriding custom icon colors on hover.

With `.usa-nav__primary button[aria-expanded=false] span::after { background-color: rgb(190, 0, 0); }` loaded after USWDS, the original code keeps the icon red. The previous PR version changed it to the system button-text color; this fix keeps it red again.

Chromium, Firefox, and WebKit reproduce the old failure and pass the fixed case. I also compared idle, hover, keyboard-focus, and expanded states at 375px and 1200px, with normal and emulated forced colors. All 16 observations per engine match the original defaults. These checks used the actual compiled Sass and pinned Material icon assets; before/after screenshots are saved, and formatting checks pass.

The demonstrated override regression is fixed. Checks in actual Windows high-contrast mode and released Safari would still be useful; emulation and WebKit don't cover those platforms. The PR remains open.

PR #6913, opened 2026-09-09
Just Dependabot Updates

mgifford:
These are all just dependabot updates that I applied locally that I think could allow us to close these https://github.com/uswds/uswds/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies

since I've squashed them here https://github.com/mgifford/uswds/pulls?q=is%3Apr+is%3Aopen


ethangardner:
@mgifford - I'm not sure why, but when I look at https://github.com/uswds/uswds/pull/6913/changes on this PR, it's not showing any changes at all.

Issue #6881, opened 2026-08-29
USWDS - Code Review: Add review digest tool


ethangardner:
Part of #6838.

Add a review digest tool on top of the review cache added in #6878: a summary table/markdown view of all cached reviews, grouped by recommendation (Approve / Approve with suggestions / Request changes / Hold / Other) and filterable by PR state (open / merged).

This was originally part of #6878 but was split out because it's a separate, decomposable concern — the digest only reads from the cache and nothing in the cache depends on it — and including it pushed that PR well past this repo's PR size norms.

**Depends on #6878 (review cache) landing first.**

## Scope

- `review-digest.mjs`: table/markdown/JSON renderers, grouped by recommendation, filterable by state.
- Rebuild the report link using the cache's exported `CACHE_DIR_NAME` (the old `reportPath` field was removed from cache entries in #6878).
- Fix `--json` leaking the entire unfiltered cache via a `raw` key regardless of `--state`.
- Share newline-collapsing between the markdown and terminal renderers so a multi-line summary can't corrupt either layout.
- Wrap the terminal summary at the existing column width instead of printing it unbounded.
- Cap markdown summary cells at a fixed length with word-boundary truncation.
- Stop echoing the group header (e.g. "Approve") into the summary cell when a review has no summary.
- Restore `review-digest.spec.mjs` with tests for each of the above, plus the digest sections of `SKILL.md`/`VERIFICATION.md`.

ryparker:
I'm closing this digest proposal as not planned for now. We need a recurring review or triage task that would benefit from it before adding the extra report formats.

The review cache and its reliability fixes remain in place; #6917 didn't implement the digest. If you need this view, please describe the task and the smallest report that would help.

PR #6917, opened 2026-09-14
USWDS - Code Review: Fix silent diff truncation on pr-metrics.mjs timeout

### LINKED ISSUE #6883: USWDS - Code Review: Fix silent diff truncation on pr-metrics.mjs timeout
Part of #6838.

`pr-metrics.mjs` has its own local `exec` helper (separate from `execCmd` in `review-cache.mjs`, added in #6878) that has a bug `execCmd` was specifically written to avoid: on a timed-out/SIGKILLed child process, `code` comes back `null`, and `pr-metrics.mjs`'s `exec` treats `code === null` as success — resolving with whatever partial stdout had arrived rather than rejecting.

`fetchPRDiff` calls this with a 90s timeout on `gh pr diff`. A PR slow enough to hit that timeout gets a truncated diff that still passes the "non-empty" check downstream, so it gets scored as a complete diff — wrong line counts, wrong budget verdict, no error surfaced anywhere.

Since `execCmd` in `review-cache.mjs` already exists and handles this correctly (rejects on `signal`, with a comment explaining exactly this failure mode), the fix is to import and reuse it in `pr-metrics.mjs` instead of maintaining a second, buggier copy. Add a regression test using a command that emits some output and then hangs past the timeout.

**Independent of the other #6838 follow-ups, but hold until the review cache (#6878) and review digest PRs land**, so this doesn't need to rebase onto a moving base in the meantime.

Raydman:
# Summary

**Fixed silent diff truncation when a PR diff fetch times out.** `pr-metrics.mjs` now reuses the shared `execCmd` helper, which rejects when a child process is killed by its timeout instead of treating the partial output as a complete diff.

## Breaking change

## Related issue

Closes #6883

## Related pull requests

None. This touches only `pr-metrics.mjs` and adds a new spec, so it is independent of the review digest work in #6881.

> The issue asked to hold this until the review cache (#6878) and the review digest PRs landed, to avoid rebasing onto a moving base. #6878 is merged; #6881 has no open PR yet but is scoped to `review-digest.mjs`, so there is no overlap with this change. Happy to hold the merge if you would still rather sequence it behind the digest.

## Problem statement

`pr-metrics.mjs` carried its own local `exec` helper, separate from `execCmd` in `review-cache.mjs` (added in #6878), and it had the exact bug `execCmd` was written to avoid:

```js
proc.on("close", (code) => {
  if (code !== 0 && code !== null)
    reject(new Error(`${cmd} exited ${code}: ${stderr}`));
  else resolve(stdout);   // code === null on a SIGKILLed child
});
```

When `spawn`'s `timeout` fires, the child is killed by a signal and `code` comes back `null`. The helper treated that as success and resolved with whatever stdout had already been buffered.

`fetchPRDiff` calls this with a 90s timeout on `gh pr diff`. A PR slow enough to hit that timeout produced a truncated diff that still passed the downstream non-empty check, so it was scored as a complete diff: wrong line counts, a wrong budget verdict, and no error surfaced anywhere. The failure is silent, and it gets worse the larger the PR, which is precisely when the metrics matter most.

## Solution

As the issue proposed, import and reuse `execCmd` from `review-cache.mjs` rather than maintaining a second, buggier copy. `execCmd` rejects on `err.signal`, so a timed-out `gh` now falls through to the existing REST API fallback instead of supplying a partial diff.

Two supporting changes were needed to make that swap safe and testable:

1. **Raised `maxBuffer` at both call sites.** `execCmd` is backed by `execFile`, whose default `maxBuffer` is 1 MB; past that it rejects with `ERR_CHILD_PROCESS_STDIO_MAXBUFFER`. The old `spawn`-based helper accumulated stdout with no limit, so a straight swap would have made every diff over 1 MB desert `gh` for the fallback — the same large PRs most likely to hit the timeout in the first place. Both call sites now pass a raised cap via a named `DIFF_MAX_BUFFER` constant.

2. **Made the module importable.** `fetchPRDiff` takes its exec helper as an argument (defaulting to `execCmd`), mirroring how `fetchPRInfo` in `review-cache.mjs` accepts `execFn`. The module also gets the same CLI entrypoint guard `review-cache.mjs` uses, so importing it in a spec no longer runs `main()`.

## Major changes

- `pr-metrics.mjs`: deleted the local `exec` helper; `fetchPRDiff` uses injectable `execCmd` with a raised `maxBuffer`; added the `import.meta.url` entrypoint guard.
- `pr-metrics.spec.mjs` (new): four tests covering the regression, both fallback paths, and the buffer ceiling.

## AI disclosure

What AI tools did you use to create this MR?

- [ ] No AI
- [ ] Chat
- [ ] Completions (e.g. Copilot)
- [x] Coding Agent

## Author's checklist

- [x] Removed redundant AI comments (if applicable)
- [x] The changes on this PR are strictly limited to the ticket requirements (i.e., no scope creep)
- [x] I personally witnessed this code working in a local environment
- [x] Automated tests are passing
- [ ] I have manually tested accessibility using physical devices and relevant assistive technology
  - Not applicable: this is Node-side agent tooling with no user-facing surface.
- [x] I understand all the code and could debug it without AI 6 months from now

## Testing and review

```sh
npm run test:agents
# 36 passing (was 32)

npm run test:agents -- --grep fetchPRDiff
```

The new spec drives the real `execCmd` against a local `bash` stand-in rather than the network, so nothing here reaches GitHub. The timeout cases run a command that prints one diff line and then hangs past the timeout, which is what a slow `gh pr diff` looks like from the caller's side: a child that is SIGKILLed with output already buffered. Each case keeps the `curl` branch resolving, so a regression shows up as the wrong diff rather than as the `process.exit(1)` that a total fetch failure would take the run down with.

**To confirm the regression test actually discriminates**, inject the deleted helper and watch the old behavior come back:

```js
import { fetchPRDiff } from "./pr-metrics.mjs";

// verbatim copy of the `exec` helper this PR deletes
const buggyExec = (cmd, args, opts = {}) => new Promise((res, rej) => {
  const proc = spawn(cmd, args, { timeout: opts.timeout });
  let stdout = "", stderr = "";
  proc.stdout?.on("data", (d) => { stdout += d; });
  proc.stderr?.on("data", (d) => { stderr += d; });
  proc.on("error", rej);
  proc.on("close", (code) => {
    if (code !== 0 && code !== null) rej(new Error(`${cmd} exited ${code}: ${stderr}`));
    else res(stdout);
  });
});

const execFn = (cmd, args, opts) =>
  cmd === "gh"
    ? buggyExec("bash", ["-c", "echo 'diff --git a/partial b/partial'; sleep 5"], { ...opts, timeout: 300 })
    : Promise.resolve("diff --git a/complete b/complete\n");

await fetchPRDiff("https://github.com/uswds/uswds/pull/1", execFn);
// old helper -> "diff --git a/partial b/partial\n", and `curl` is never reached
// execCmd     -> falls through to the fallback and returns the complete diff
```

The CLI was also exercised directly to confirm the entrypoint guard did not break it:

```sh
git diff develop > changes.patch
node .agents/skills/uswds-code-review/scripts/pr-metrics.mjs --diff changes.patch
```


ryparker:
@coderabbitai full review

ryparker (approved):
The timeout fix looks good on the updated branch. All 36 agent-tooling tests pass, as do CI and signature checks. This only changes the internal PR-diff tooling.

The REST fallback already has the same 64 MiB output cap as `gh`. A test specifically covering that fallback would be useful, but it's a follow-up suggestion and doesn't block this fix.

PR #6944, opened 2026-09-17
fix(combo-box): preserve button focus contrast

### LINKED ISSUE #6424: USWDS - Bug: Combo Box clear input button focus outline does not match assigned value
### Describe the bug

Currently, the combo box clear input button applies a lighter focus outline color than what is assigned. For example, when looking at the USWDS component demo and applying focus to the button, the _focus.scss looks like it should be applying an outline color of #2491ff, but in reality, #7CBDFF is being assigned.

When accepting the default outline colors, this fails WCAG 2.1 Non-text Contrast (AA) as the contrast ratio is 2:1.

1. Go to the [Combo Box preview](https://designsystem.digital.gov/components/combo-box/#component-preview)
2. Apply focus on the Clear Input button
3. Compare assigned focus outline color to displayed color

Image

### Expected Behavior

Expected behavior is that the assigned outline color for focus is applied to the Clear Input button.

### Related code

_No response_

### Screenshots

_No response_

### System setup

- This is experienced with both Chrome 133.0.6943.127 and Firefox 135.0.1

### Additional context

To get around the 508 issue, a darker color can be assigned, however the assigned color is still not the same as the displayed color.

Image

### Code of Conduct

ryparker:
## Summary

Preserve the configured focus-outline color on combo-box buttons. The clear button now has full opacity while focused, so its outline is no longer faded together with the icon. Unfocused icons retain their existing appearance.

## Related issue

Closes #6424

## Breaking change

The new focused-button opacity declaration overrides existing consumer styles because it uses `button.usa-combo-box__toggle-list:focus` and `button.usa-combo-box__clear-input:focus`. For example, a later `.usa-combo-box__toggle-list:focus { opacity: .85; }` rule previously produced opacity `0.85`; it now loses to the more specific USWDS rule and produces `1`. Later class-only opacity overrides are affected too. The rule also applies to the shared time-picker controls.

The [compatibility review](https://github.com/uswds/uswds/pull/6944#issuecomment-5725085898) requests moving the opacity fix into a separate rule with no more specificity than the existing base class, while retaining the outline-offset rule. That compatibility fix remains outstanding; the default focus-contrast improvement alone does not preserve consumer overrides.

## Solution

Set `opacity: 1` in the existing focused combo-box button rule. This also covers the shared time-picker control and a focused dropdown-toggle button. The outline width, offset, theme color, markup and interaction behavior are unchanged. Existing custom opacity overrides are affected as described above.

## Testing and review

- Matched Chromium keyboard-focus checks measure default-theme outline contrast against the white input at **1.99:1 before** and **3.20:1 after**. The retained contrast assertion fails on the baseline and passes with this change.
- Keyboard clearing empties the input and underlying select, hides the clear button, and returns focus to the input. The time-picker clear button receives the same full-opacity outline. Unfocused and disabled controls retain their appearance.
- `npm test` passes: 128 Sass, 924 component and 28 task tests, plus lint and typecheck.
- `npm run build` and all 37 distribution checks pass. Compiled CSS changes only the intended opacity declaration.

No real mobile, forced-colors or screen-reader test is claimed. This is a CSS-only focus-style change.

## Visual comparison

The same synthetic fixture in Chromium at 1280x720, with keyboard focus on the first clear button. The other controls show the unchanged unfocused and disabled appearance.

Before (baseline `eaef3953`):

Before: opacity fades the focused clear-button outline

After:

After: the focused clear button retains its full theme outline color

[Reproduction fixture, measurements, and contrast assertion](https://github.com/uswds/uswds/tree/a7289d0f974d2406f9b084aaeff71e467cb9ebef/focus-opacity). Fixed revision: `a26e8305ba3018c999481a7ef844c979935b274a`.

Regression rationale: the linked browser contrast assertion uses actual computed opacity, outline color, and input background, and fails on the baseline. A Sass selector-presence check or jsdom style assertion would not validate the rendered alpha-composited contrast. The evidence also records focused and unfocused controls, keyboard clearing, and the shared time picker.

## Focused review verification

Reviewed commit `a26e8305ba3018c999481a7ef844c979935b274a`: a fresh Chromium run reproduced outline contrast of 1.985983:1 on the baseline and measured 3.196749:1 with the fix. Keyboard Tab and Enter still clear the input and select, then return focus to the input. The shared time-picker clear control and focused toggle have opacity 1; unfocused controls retain opacity 0.6. All eight measured button rectangles are unchanged. Fresh compiled CSS matches the tested focus rule. Those checks do not resolve the custom-style regression documented in the breaking-change section. These measurements cover the default theme, not every consumer-defined color combination.


ryparker:
The focus-contrast fix in `a26e8305` overrides some existing custom styles. The [new opacity declaration](https://github.com/uswds/uswds/blob/a26e8305ba3018c999481a7ef844c979935b274a/packages/usa-combo-box/src/styles/_usa-combo-box.scss#L85) inherits the `button` selector, so a later `.usa-combo-box__toggle-list:focus { opacity: .85; }` rule now loses to `button.usa-combo-box__toggle-list:focus`. The compiled-CSS browser comparison gives `0.85` before and `1` after. That's why this has the breaking-change label.

Please put the opacity fix in a separate rule with no more specificity than the existing base class, for example with a zero-specificity focus condition. Removing only `button` would still override later class-only styles. The outline-offset rule can stay as it is.

We need to check default focus contrast and both class-only and class-with-focus overrides before lifting the hold. That would preserve custom styles while fixing the faded default outline.

PR #6945, opened 2026-09-18
fix(checkbox): preserve indeterminate indicators when checked

### LINKED ISSUE #6381: USWDS - Feature: Raise precedence of indeterminate checkbox styling when checked=true
## Summary

When the USWDS checkbox has `checked=true` and some sort of indeterminate state (either using JS or the `data-indeterminate` attribute), there are different behaviors depending on how the indeterminate state is set, and what browser is being used.

### Safari and Firefox
- `indeterminate=true` (JS) AND/OR `data-indeterminate` (HTML)
  - USWDS Checkbox: Shows the "checked" state
  - Default Checkbox: Shows the "indeterminate" state

### Chrome
- `indeterminate=true` (JS)
  - USWDS Checkbox: Shows the "indeterminate" state
  - Default Checkbox: Shows the "indeterminate" state
- `data-indeterminate` (HTML)
  - USWDS Checkbox: Shows the "checked" state
  - Default Checkbox: Shows the "indeterminate" state

## Indeterminate Precedence

My first question is about the precedence of checked vs indeterminate styling! It looks like [it was decided](https://github.com/uswds/uswds/pull/5713#discussion_r1504766405) to [set the `checked` styling with a higher precedence](https://github.com/uswds/uswds/pull/5713#pullrequestreview-1909783239) than `indeterminate` because "Indeterminate implies that it is not checked", but I'm curious about that interpretation of [the quoted CSS specification passage](https://drafts.csswg.org/selectors/#indeterminate) because it says that the checkbox is "neither checked **nor unchecked**."

Based on that description, I would expect the `indeterminate` attribute to _visually override_ the `checked` attribute, whether it's checked or unchecked (i.e. the `checked` attribute technically has to be in one of those two states, but it should essentially be ignored for the purposes of rendering.)

This seems to more closely match the interpretation of all three mentioned browsers.

## indeterminate=true vs. data-indeterminate

My second question is about the behavior in Chrome, which seems to behave differently when using the browser-native `element.indeterminate=true` setting vs. the USWDS-provided `data-indeterminate` attribute.

Regardless of the browser implementation, I would expect the USWDS checkbox to behave consistently when both options are available to the developer. It appears we are currently treating the combination of `checked` and `indeterminate` states as undefined behavior; if we continue to do so, I'm curious if we'd like to discourage this usage via usability guidance.

The alternative would be to keep the existing undefined behavior, under the assumption that developers are not using `checked` and `indeterminate` attributes together. (Is there usability guidance we can/should provide in the docs if this is the preferred thinking?)

### Additional context

Some images as a reference of behavior.

## USWDS vs. default checkbox in Chrome

|indeterminate=true with JS|data-indeterminate="true"|
|---|---|
|Image|Image|

## 🤔 USWDS vs. default checkbox in Safari and Firefox

|Safari|Firefox|
|---|---|
|Image|Image|

### Code of Conduct

ryparker:
# Summary

Fixed indeterminate checkboxes so the dash remains visible when the checkbox is also checked. Native `indeterminate` and `data-indeterminate` now use the same indicator across default, tile, disabled, and dark styles.

## Breaking change

The current implementation changes CSS precedence for ordinary checked, disabled checkboxes that are not indeterminate. Removing the explicit checked-and-disabled checkmark selectors reduces their specificity from `(0,4,1)` to `(0,3,1)`. A later custom disabled-state `background-image` rule can therefore replace a checkmark that previously remained visible. This affects both native `disabled` and `aria-disabled="true"` controls.

Separately, the intended change makes the indeterminate dash take precedence when a checkbox is also checked. No JavaScript API or markup migration is required, but the disabled-checkbox customization regression must still be resolved. The [compatibility review](https://github.com/uswds/uswds/pull/6945#issuecomment-5726494255) requests retaining the checked-disabled selectors and placing equally specific indeterminate-disabled rules after them.

## Related issue

Closes #6381.

## Problem and solution

With `checked=true`, `data-indeterminate` displayed a checkmark in Chromium, Firefox, and WebKit. Firefox and WebKit also displayed a checkmark for the native indeterminate property. This hides the mixed-selection state, including the issue's "select all" use case.

Apply the checked indicator before the indeterminate indicator and remove disabled checked selectors that otherwise override it. Removing those selectors also weakens ordinary checked-disabled customization compatibility, as described above. Checked values and interaction behavior are preserved. The data attribute remains a styling hook.

This revises the precedence chosen in #5713. The [HTML checkbox definition](https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox)) allows indeterminateness to obscure either checked value, and [existing USWDS guidance](https://designsystem.digital.gov/components/checkbox/) supports both ways of applying indeterminate styles.

## Related pull requests

No companion documentation PR is required. The existing checkbox documentation already describes native and data attribute indeterminate styling.

## Screenshots

Matched markup and 1280 by 1250 viewport. Native and data attribute states are shown with both checked values, disabled modes, tiles, and dark colors.

| Before | After |
| --- | --- |
| Before: checked data-indeterminate boxes show checkmarks | After: all indeterminate boxes show dashes |

## Testing and review

- Added six Storybook regression plays covering both state sources, both checked values, default and tile inputs, disabled and aria-disabled states, and indicator restoration. All six fail baseline; all nine checkbox stories and existing Axe checks pass with the fix.
- Chromium 147, Firefox 151, and WebKit 26.5 each pass all 48 normal visual states and 48 state-restoration checks. Chromium and WebKit also pass all 48 forced-color checks.
- `npm test`: 128 Sass, 924 component/core, and 28 task tests pass. Build, formatting, lint, and whitespace checks pass.
- Review by toggling checked and indeterminate independently in Checkbox stories. The dash should obscure either checked value; clearing indeterminate should restore the correct checkmark or empty box.

Limitations: Firefox forced-color emulation suppresses all indicator background images on both baseline and fix. Existing print-specific indicator behavior remains outside this change. WebKit verification is not physical Safari/iOS testing; physical screen reader testing was not performed. No ARIA, focus management, or keyboard behavior changed.

[Retained fixture and cross-browser measurements](https://github.com/uswds/uswds/tree/0dd9de113c7393e9089df2e1bbad2b8f3099c0a1/checkbox-indeterminate). Fixed revision: `33cd8c3147dc873cc966b7aefc2d2defe95ce57f`.

## Focused review verification

Reviewed commit `33cd8c3147dc873cc966b7aefc2d2defe95ce57f`: a fresh Chromium run passed all 144 assertions across normal rendering, forced colors, and restoration. Fresh Sass compilation matched the retained fixture CSS. The checked-and-indeterminate screen-rendering scope of #6381 is covered. Print-specific indicators, attribute-only ARIA semantics, and physical screen-reader testing are outside the claimed result.


ryparker:
There's a separate regression in `33cd8c3`: the checkmark can disappear from an ordinary checked, disabled checkbox. The [new checked rule](https://github.com/uswds/uswds/blob/33cd8c3147dc873cc966b7aefc2d2defe95ce57f/packages/uswds-core/src/styles/mixins/helpers/checkbox-and-radio-colors.scss#L143-L156) removes the more specific checked-and-disabled selectors, changing which background image wins. That's the reason for the breaking-change label.

For example, load this after USWDS:

```css
.usa-checkbox__input:disabled + .usa-checkbox__label::before {
  background-image: linear-gradient(red, red);
}
```

For a checked, disabled checkbox that isn't indeterminate, the previous `(0,4,1)` selector keeps the USWDS checkmark. The new `(0,3,1)` selector lets the custom rule replace it. Chromium checks reproduce this with `aria-disabled="true"` too; enabled-checked and disabled-unchecked controls are unchanged.

Please keep the existing checked-disabled selectors and place equally specific indeterminate-disabled rules after them. That should fix the mixed-state indicator without changing ordinary disabled checkboxes.

Issue #5913, opened 2024-05-06
USWDS - Bug: $theme-table-header-text-color throws error


ajanickiv:
### Describe the bug

When attempting to use the `$theme-table-header-text-color` setting for the table found under the [Component Settings -> Table area here](https://designsystem.digital.gov/documentation/settings/#component-settings-2) to set the color in a CSS class, I get the following error: 
```
Undefined variable.
   ╷
59 │     @if map.has-key($project-color-shortcodes, $value) {
   │                     ^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
  node_modules\@uswds\uswds\packages\uswds-core\src\styles\functions\utilities\color.scss 59:21  color()
```
I am also using the `$theme-table-header-background-color` in the same CSS class with no issues.

### Steps to reproduce the bug

1. Import `uswds-core` into a .scss file
2. Create a css class
3. Set the color attribute using the color function like  `color: color($theme-table-header-text-color);`
4. Run build process to generate styles
5. Receive error

### Expected Behavior

The stylesheet gets generated without error.

### Related code

```
@import 'uswds-core';
.my-class {
    color: color($theme-table-header-text-color);
}
```

### Screenshots

![image](https://github.com/uswds/uswds/assets/16595489/fb82a57d-e33e-4f1f-8e2c-72449f93f685)

### System setup

- USWDS Version: 3.8.0
- WIndows laptop
- Edge browser

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22) for duplicate bug reports.

ryparker:
I reproduced this at `05cc4e65` with both the original `@import "uswds-core"` example and the equivalent `@use` form. Sass reports `Undefined variable.` at `$project-color-shortcodes`.

By default, `$theme-table-header-text-color` is `default`, which tells the table mixin to choose a color based on the background. Passing it directly to `color()` produces an invalid-token error, but that error handling itself references an unavailable variable. An explicit token such as `"ink"` compiles successfully, as does the normal `usa-table` stylesheet.

This is a confirmed bug in the error handling. The fix should explain that the token is invalid while preserving the table mixin's automatic color choice, rather than assigning an arbitrary color to `default`.

PR #6948, opened 2026-09-18
fix(sass): restore invalid color token diagnostics

ryparker:
# Summary

Corrected the `color()` compiler diagnostic for invalid tokens. Calls such as `color($theme-table-header-text-color)` with the default setting now identify the invalid token instead of failing with an unrelated undefined-variable error.

## Breaking change

Direct legacy imports of `uswds-core/src/styles/functions/utilities/color` no longer resolve custom colors supplied through the caller's `$project-color-shortcodes` map. For example, defining `("brand-custom": #123456)` before that `@import` previously allowed `color("brand-custom")` to compile to `#123456`; the current implementation instead reports an invalid color token. The removed fallback could access that map through the legacy import's shared scope. Standard core `@use` and theme configuration do not exercise the same path.

The [compatibility review](https://github.com/uswds/uswds/pull/6948#issuecomment-5725085485) requests retaining the fallback when its variable exists while preserving the improved module-scope diagnostic. That compatibility fix remains outstanding in this PR.

## Related issue

Related to #5913. This fixes the diagnostic path. The component's `default` sentinel remains invalid as a standalone color token, so this PR does not automatically close the broader usage report.

## Problem and solution

In the standard core module path, `$all-color-shortcodes` contains the configured project color map. The subsequent fallback references a variable unavailable in the function's module scope, causing the unrelated undefined-variable error for invalid tokens. A direct legacy `@import` can instead expose the caller's custom map to that fallback.

This PR removes the fallback so invalid values reach the existing `error-not-token()` helper. Standard system, theme, state, and configured custom colors retain their behavior, but custom colors supplied through the direct legacy import regress as described above.

## Related pull requests

No companion documentation PR is linked. Function signatures and setting names are unchanged, but legacy direct-import compatibility still requires the fix described above.

## Screenshots

The matched browser fixture shows actual Sass compiler output and the unchanged valid color and default table controls.

| Before | After |
| --- | --- |
| Before: invalid color token produces an undefined-variable error | After: compiler identifies the invalid default color token |

## Testing and review

- Added 12 real compilation tests. Four diagnostic regressions fail baseline and pass with the fix; remaining checks preserve ordinary tokens, custom colors, override flags, and existing error policies.
- `npm run build` and `npm test` pass: 128 Sass, 936 component/core, and 28 task tests. Formatting and whitespace checks pass.
- Compiled default table CSS is byte-identical before and after. Chromium confirms the rendered compiler messages match actual exceptions and the valid color and table controls have unchanged computed styles.
- Reproduce with `@use "uswds-core" as *; .example { color: color($theme-table-header-text-color); }`. The compiler should identify `default` as an invalid USWDS color token. A supported token such as `"ink"` still compiles normally.

This change affects compilation diagnostics and custom-color resolution through legacy direct imports. No component interaction or accessibility behavior changes; physical assistive technology testing was not performed.

[Retained compiler fixture, measurements, and screenshots](https://github.com/uswds/uswds/tree/70fbb2b974e9f995f9c40208881d14dd32f015b4/color-diagnostic). Baseline: `eaef39533a0a0f78976d6888edf8ae56fe9334d6`. Fixed revision: `968bacb743ae8b6feb4e972e08246f01856c5ee8`.


ryparker:
At `968bacb7`, a direct import of the legacy `color()` function stops recognizing a custom color. That's the breaking change here; ordinary theme configuration isn't affected in the same way.

With `packages` on the Sass load path:

```scss
$project-color-shortcodes: ("brand-custom": #123456);
@import "uswds-core/src/styles/functions/utilities/color";
.example { color: color("brand-custom"); }
```

This previously compiled to `color: #123456`, but now reaches the [invalid-token error](https://github.com/uswds/uswds/blob/968bacb743ae8b6feb4e972e08246f01856c5ee8/packages/uswds-core/src/styles/functions/utilities/color.scss#L57). The deleted fallback could see the caller's custom map through the direct `@import`; the standard core `@use` and top-level import tests didn't exercise that case.

Please keep the fallback when its variable exists, while retaining the improved diagnostic for unknown tokens in module scope. Adding this direct-import case alongside the theme and system-token tests would let us check the fix before removing the label.

Issue #5484, opened 2023-08-31
Refactor Input Mask: Add JSDocs comments


amyleadem:
## Summary
Add JSDocs to all input mask functions.

The update should meet the standards outlined in the [Input mask requirements doc](https://docs.google.com/document/d/1kLHaCl5PceSByz0NvZqSc-ot7tSjcbe1EJPbBAZM7LY/edit#heading=h.fa3sv1ray42g) (:lock: Google docs).

We should be able to use the work for this originally done in https://github.com/uswds/uswds/pull/5227.

ryparker:
I'm closing this as not planned alongside #6951. The input-mask refactor in #5227 already includes function documentation, and both #5227 and #6203 are still under review. A separate documentation pass on the older implementation would leave us with two versions to reconcile.

We'll need to review the function documentation with the implementation we choose. The refactor requirements are still outstanding. If an undocumented behavior in the current release is blocking your work, please share an example here so we can consider a smaller fix.

Issue #5217, opened 2023-04-07
USWDS - Core: Update comment style for SASS functions


mejiaj:
### Is your feature request related to a problem? Please describe.

Use [documentation comments](https://sass-lang.com/documentation/syntax/comments#documentation-comments) instead of CSS style comments to prevent them from being included in compiled CSS. 

**Example of commented code seen on compile**
https://github.com/uswds/uswds/pull/5209

### Describe the solution you'd like

For example:

```diff
- /*
- ========================================
- General-purpose functions
- ----------------------------------------
- */


+ // ==========================================================================
+ // General-purpose functions
+ // ==========================================================================
```

```diff
- /* ln()
- ----------------------------------------
- */

+ //
+ // ln()
+ // ==========================================================================
+
+ /// Computes an exponent.
+ ///
+ /// @param {number} $value
+ ///   The exponent.
+ /// @return {number} `$value` to the power of 10.
```

### Additional context

Following this style will make it easier to integrate sassdoc.
https://sass-lang.com/documentation/syntax/comments#documentation-comments

Related issue:
https://github.com/uswds/uswds/issues/2642

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://designsystem.digital.gov/about/community/#community-conduct).
- [X] I checked the [current issues](https://github.com/uswds/uswds/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Feature+Request%22) for duplicate feature requests.

annepetersen:
May also add to/link to from the USWDS internal engineering guide

ryparker:
I'm closing this and #6951 as not planned for now. Converting the comments would remove them from expanded CSS, but it wouldn't change compressed CSS or behavior. We'll improve comments as we work on the affected files instead of doing a separate, broad cleanup.

If the current comments cause a problem in your expanded CSS or SassDoc workflow, please share an example here. We can revisit a smaller change around that need. The request and patch remain available if we revisit this.

PR #6951, opened 2026-09-18
docs: clarify internal function documentation

ryparker:
## Summary

Documented the current input-mask functions and kept Sass function implementation notes out of compiled CSS. Existing parameter documentation now uses SassDoc comments directly above its function, and the contributor guide explains the comment conventions.

## Related issues

Related to #5484 and #5217. This documents the current input-mask implementation without replacing the separate rewrite in #5227. The private acceptance-document link in #5484 was unavailable for independent review.

## Breaking change

This is not a breaking change. JavaScript behavior and CSS rules are unchanged.

## Validation

- JavaScript syntax trees are identical after excluding comments and source formatting metadata.
- Sass source differs only in comments and whitespace.
- Expanded CSS decreased from 749,438 to 730,024 bytes, removing 19,414 bytes of implementation comments.
- Compressed CSS is byte-identical at 557,899 bytes.
- Scoped ESLint, Prettier, and `git diff --check` pass.

The remaining plain function headers are silent comments. This change does not claim to provide complete SassDoc metadata for every function or generate a documentation site.


ryparker:
I'm closing this PR for now. It changes comments across 62 files without changing compressed CSS or runtime behavior, and the input-mask documentation overlaps work still under review in #5227 and #6203.

We'll update those comments alongside the relevant implementation changes. I'm also closing #5217 and #5484 as not planned; those fixes haven't been completed. The patch and commit history remain available.

If part of this addresses a specific maintenance or build problem for your team, please share that example. We can reconsider a smaller change around it.

Issue #6516, opened 2025-08-12
USWDS - ADR: Prefer intrinsic layouts and container queries to page level media queries for responsive design


heymatthenry:

ryparker:
Let's continue this in [the discussion on intrinsic layouts and container queries](https://github.com/uswds/uswds/discussions/6958). I'm closing this issue as not planned for now: we haven't chosen a component change or adopted this as an architectural preference. The draft in #6956 would also create a second place to record decisions alongside the existing proposals repository.

If this approach solves a problem for your team, please share the component, a small layout example, and the behavior you'd like in the discussion. That would help us define a change and record the decision in the existing repository. The original issue and draft remain available.

PR #6956, opened 2026-09-18
docs(architecture): propose intrinsic component layouts

ryparker:
A component can have little available space even when the viewport is wide. Add a proposed architecture decision describing when intrinsic layout or a size container query should guide component layout, and when viewport media queries remain appropriate.

The record covers container ownership, a usable base layout, compatibility, public markup changes, review evidence, alternatives, and incremental adoption. Its status is explicitly **Proposed** with a pending acceptance record. It does not change support requirements, public APIs, component CSS, or establish a migration deadline.

Related to #6516. Maintainers still need to accept, revise, or reject the architectural preference under existing project governance.

Validation: reviewed the proposal against current W3C and MDN references, rendered both Markdown files, verified repository links, and ran `git diff --check`. Documentation only; no application suite was needed.


ryparker:
I'm withdrawing this PR. Architectural decisions already live in [uswds-proposals](https://github.com/uswds/uswds-proposals/tree/main/decisions), and we don't yet have a component example that needs a new layout rule. A second place to maintain those decisions would add unnecessary work.

Let's continue with concrete examples in [the intrinsic-layout discussion](https://github.com/uswds/uswds/discussions/6958). The draft and branch remain available. I'm closing #6516 as not planned, not as an implemented decision. If an example supports a focused change, we can take it through the existing proposal process.

